Issue #1, opened 2021-02-19
Describe cryptographic dependencies


chris-wood:
Key encapsulation mechanisms, PRGs, and a sprinkle finite fields. (We can of course elaborate on these in the actual PR.)

PR #2, opened 2021-02-24
Describe the cryptographic dependencies for the candidate protocol

cjpatton:
Solves #1.

This adds a high level description of the current candidate for our version of Prio. (Note that this is subject to change as the threat 
model and system requirements evolve.) 

It then enumerates the cryptographic primitives with which the protocol could be instantiated, including the FF parameters, KEM, and PRG.

### DIFF
--- a/design-document.md
+++ b/design-document.md
@@ -1,5 +1,16 @@
 # Prio v3 Design Document
 
+## Terminology
+1. Aggregator: A server that runs the input-validation protocol and accumulates
+   input shares.
+1. Client: the endpoint from which the user sends data to be aggregated, e.g., a
+   web browser.
+1. Input-validation protocol: The protocol executed by the client and aggregators
+   in order to validate the clients's input without leaking its value to the
+   aggregators.
+1. Leader: A distinguished aggregator that coordinates input validation and data
+   aggregation.
+
 ## Sample user stories
 
 ## Threat model
@@ -14,4 +25,212 @@
 
 ## Open questions and system parameters
 
-## Cryptographic dependencies
+## Cryptographic components
+
+### The input-validation protocol
+
+**Notation.**
+Fix a finite field K. We let K^n denote the set of vectors over K of length n.
+We write the linear, s-way secret sharing of an element x of K^n as {x:1}, ...,
+{x:s}, where {x:i} is the share held by the i-th party. We write {x} as
+shorthand for the sequence {x:1}, ..., {x:s}. Each {x:i} is also an element of
+K^n.
+
+Prio combines standard [linear secret
+sharing](https://en.wikipedia.org/wiki/Secret_sharing#t_=_n) with a new type of
+probabilistically checkable proof (PCP) system, called a fully linear PCP. The
+input-input validation protocol can be described in terms of three main
+algorithms:
+
+1. pf := Prove(x) denotes generation of a proof pf of the validity of input x.
+   This algorithm is executed by the client.
+1. {vf:i} := Query({x:i}, {pf:i}, r) denotes computation of the verification
+   share {vf:i} for input share {x:i} and proof share {pf:i}. This algorithm is
+   executed by each of the aggregators; input r denotes the joint randomness
+   shared by all of the aggregators.
+1. b := Decide({vf}, r) denotes the execution of the decision procedure on input
+   shares {vf} and joint randomness r. The output b is a boolean indicating
+   whether the input is deemed valid. This algorithm is run by the leader.
+
+The values above have following types:
+
+1. Input x is an element of K^n for some integer n.
+1. Proof pf is an element of K^p(n) for some function p.
+1. The joint randomness r is an element of K^u(n) for some function u.
+1. Each verification share {vf:i} is an element of K^v(n) for some function v.
+
+Before the protocol begins, the aggregators agree on joint randomness r and
+designate one of the aggregators as the leader. The protocol proceeds as
+follows:
+
+1. The client runs pf := Prove(x). It splits x and pf into {x} and {pf}
+   respectively and sends ({x:i}, {pf:i}) to aggregator i.
+1. Each aggregator i runs {vf:i} := Query({x:i}, {pf:i}, r) ands sends {vf:i} to
+   the leader.
+1. The leader runs b := Decide({vf}, r) and sends b to each of the aggregators.
+
+If b=1, then each aggregator i adds its input share {x:i} into its share of the
+aggregate. Once a sufficient number of inputs have been validated and
+aggregated, the aggregators send their aggregate shares to the leader, who adds
+them together to obtain the final result.
+
+[[TODO: Sketch out the b=1 path.]]
+
+**Proof generation and verification.**
+[[TODO: Describe how to construct proof systems for languages recognized by
+validity circuits with G-gates, a la [BCC+19, Theorem 4.3].]]
+
+**Security parameters.**
+[[TODO: Define completeness, soundness, and honest-verifier zero-knowledge for
+fully linear PCPs and state bounds for [BBC+19, Theorem 4.3]. This bound will
+guide the selection of the field best suited for the data type and
+application.]]
+
+**Consensus protocol.**
+[[TODO: Describe how the aggregators pick the leader and the joint randomness.]]
+
+**Key distribution.**
+[[TODO: Decide how clients obtain aggregators' public keys.]]
+
+### Changes to the input-validation protocol
+
+**Coordinating state.**
+The state of the input-validation protocol is maintained by the leader; except
+for aggregation of the input shares, the other aggregators are completely
+stateless. This is accomplished by making the following changes to the core
+protocol.
+
+1. The client sends all of its shares to the leader. To maintain privacy, the
+   client encrypts each (input, proof) share under the public key of the share's
+   recipient.
+1. The leader forwards each encrypted share to its intended recipient. Each
+   aggregator decrypts its input and proof share, computes its verification
+   share, and sends its verification share to the aggregator as usual.
+1. If b=1 in the last step, then the leader also sends along the encrypted input
+   share to each aggregator so that they can decrypt and aggregate the share
+   without needing to cache the input share from the previous step.
+
+**Minimizing communication overhead.**
+In most linear secret sharing schemes, the length of each share is equal to the
+length of the input. Therefore, the communication overhead for the client is
+O(s\*(n+p(n))). This can be reduced to O(s+n+p(n)) with the following standard
+trick.
+
+Let x be an element of K^n for some n. Suppose we split x into {x} by choosing
+{x:1}, ..., {x:s-1} at random and letting {x:s} = x - ({x:1} + ... + {x:s-1}).
+We could instead choose s-1 random seeds k[s-1], ..., k[s-1] for a pseudorandom
+number generator PRG and let {x:i} = PRG(k[i], n) for each i. This effectively
+"compresses" s-1 of the shares to O(1) space.
+
+### Primitives
+
+This section describes the core cryptographic primitives of the system.
+
+#### Finite field arithmetic
+
+The algorithms that comprise the input-validation protocol --- Prove, Query, and
+Decide --- are constructed by generating and evaluating polynomials over a
+finite field. As such, the main ingredient of Prio is an implementation of
+arithmetic in a finite field suitable for the given application.
+
+We will use a prime field. The choice of prime is influenced by the following
+criteria:
+
+1. **Field size.** How big the field needs to be depends on the type of data
+   being aggregated and how many users there are. The field size also impacts
+   the security level: the longer the validity circuit, the larger the field
+   needs to be in order to affectively detect malicious clients.
+1. **Fast polynomial operations.** In order to make Prio practical, it's
+   important that implementations employ FFT to speed up polynomial operations.
+   In particular, the prime modulus p should be chosen so that (p-1) = 2^n * s
+   for large n and odd s. Then g^s is a principle, 2^n-th root of unity (i.e.,
+   g^(s\*2^n) = 1), where g is the generator of the multiplicative subgroup.
+   This fact allows us to quickly evaluate and interpolate polynomials at 2^d-th
+   roots of unity for 1 <= d <= n.
+1. **Highly composite subgroup.** Suppose that (p-1) = 2^n * s. It's best if s
+   is highly composite because this minimizes the number of multiplications
+   required to compute the inverse or apply Fermat's Little Theorem. (See
+   [BBC+19, Section 5.2].)
+1. **Code opttimziation.** [[TODO: What properties of the field make
+   it possible to write faster implementations?]]
+
+The table below lists parameters that meet these criteria at various levels of
+security. (Note that \#2 is the field used in "Prio v2".) The "bits" column
+indicates the size of the prime. For example, for p=3221225473,
+bits=log2(3221225473), which is approximately 31.53.
+
+| # | bits   | p                                      | g  | n   | s                |
+|---|--------|----------------------------------------|----|-----|------------------|
+| 1 | 31.58  | 3221225473                             | 5  | 30  | 3                |
+| 2 | 31.99  | 4293918721                             | 19 | 20  | 3^2 * 5 * 7 * 13 |
+| 3 | 63.75  | 15564440312192434177                   | 5  | 59  | 3^3              |
+| 4 | 66.17  | 83010348331692982273                   | 11 | 63  | 3^2              |
+| 5 | 122.81 | 9304595970494411110326649421962412033  | 3  | 120 | 7                |
+| 6 | 125.81 | 74769074762901517850839147140769382401 | 7  | 118 | 3^2 * 5^2        |
+
+**Finding suitable primes.**
+One way to find suitable primes is to first choose choose n, then "probe" to
+find a prime of the desired size. The following SageMath script prints the
+parameters of a number of primes larger than 2^n for a given n:
+
+```
+n = 116
+for s in range(0,1000,1):
+    N = 2^n
+    p = (N*s).next_prime()
+    if p-(N*s) == 1:
+        bits = round(math.log2(p), 2)
+        print(bits, p, GF(p).multiplicative_generator(), n, factor(s))
+```
+
+#### Key encapsulation
+
+Our instantiation of the input-validation protocol involves two additional
+operations: public key encryption and cryptographically secure pseudorandom
+number generation (CSPRNG). The combination of these primitives that we use here
+allows us to make an additional simplification. We assume that clients communicate
+with the leader over a confidential and authenticated channel, such as TLS. As a
+result, we only need to encrypt CSPRNG seeds, which requires only a
+key-encapsulation mechanism (KEM) rather than full-blown encryption.
+
+A KEM is comprised of two algorithms:
+
+1. (c, k) := Encaps(pk) denotes generation and encapsulation of symmetric key k
+   under the recipient's public key pk.
+1. k := Decaps(sk, c) denotes decapsulation of symmetric key k under the
+   recipient's secret key sk.
+
+To generate an aggregator's share, the client runs (c[i], k[i]) := Encaps(pk[i])
+and sends c[i] to the aggregator. To compute its share, the aggregator would run
+k[i] := Decaps(sk[i], c[i]) and compute its share as {x:i} = PRG(k[i], n).
+
+[HPKE](https://datatracker.ietf.org/doc/draft-irtf-cfrg-hpke/) is a natural
+candidate for instantiating the KEM. In "Export-Only" mode, HPKE provides an
+efficient scheme with all the cryptographic agility we would ever need. And
+although it's still an Internet-Draft, it has high quality implementations in a
+variety of languages.
+
+[[TODO: Specify how HPKE is used to implement Encaps() and Decaps().]]
+
+#### Pseudorandom number generation
+
+A suitable PRG will have the following syntax. Fix a finite field K:
+
+1. x := PRG(k, n) denotes generation of a pseudorandom element x of K^n, i.e., a
+   vector of n elements of K.
+
+This can be instantiated using a standard stream cipher, e.g.., ChaCha20 as
+follows. Interpret k as the cipher key, and using a fixed nonce, generate l\*n
+bytes of output, where l is the number of bytes needed to encode an element of
+K, then map each chunk of l bytes to an element of K by interpreting the chunk
+as an l-byte integer and reducing it modulo the prime modulus.
+
+[[OPEN ISSUE: Mapping the output of PRG(.,.) to a vector over K induces a
+small amount of bias on the output. How much bias is induced depends on the how
+close the prime is to a power of 2. Should this be a criterion for selecting the
+prime?]]
+
+## References
+
+* [BBC+19] Boneh et al. "Zero-Knowledge Proofs on Secret-Shared Data via Fully
+  Linear PCPs". Crypto 2019.
--- a/design-document.md
+++ b/design-document.md
@@ -15,6 +15,15 @@
 
 ## Threat model
 
+[[OPEN ISSUE: The threat model for Prio --- as it's described in the original
+paper and [BBC+19] --- considers **either** a malicious client (attacking
+soundness) **or** a malicious subset of aggregators (attacking privacy). In
+particular, soundness isn't guaranteed if any one of the aggregators is
+malicious; in theory it may be possible for a malicious client and aggregator to
+collude and break soundness. Is this a contingency we need to address? There are
+techniques in [BBC+19] that account for this; we need to figure out if they're
+practical.]]
+
 ## System requirements
 
 ### Data types


design-document.md
```diff
@@ -14,4 +24,169 @@
...
+
+The algorithms that comprise the input-validation protocol --- Prove, Query, and
+Decide --- are constructed by generating and evaluating polynomials over a
+finite field. As such, the main ingredient of Prio is an implementation of
+arithmetic in a finite field suitable for the given application.
+
+We will use a prime field. The choice of prime is influenced by the following
+criteria:
+
+1. **Field size.** How big the field needs to be depends on the type of data
```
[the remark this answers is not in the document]
acmiyaguchi: Does arbitrary sized finite fields imply the use of bignums? Will the field sizes be carefully chosen for performance (e.g. recommended EC primes of 256, 384, etc)?

design-document.md
```diff
@@ -14,4 +24,169 @@
 
 ## Open questions and system parameters
 
-## Cryptographic dependencies
+## Cryptographic components
+
+### Overview of the core protocol
+
+**Notation.**
+Fix a finite field K. We write the linear, s-way secret sharing of an element x
```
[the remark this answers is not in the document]
cjpatton: It's worth adding in this flexibility now in case we end up needing it later. Depending on our security considerations and how address resilience, it may end up being useful to have 3 or more aggregators.

design-document.md
```diff
@@ -14,4 +24,169 @@
...
+
+1. The user sends all of its shares to the leader. To maintain privacy, the user
+   encrypts each input/proof share under the public key of the share's
+   recipient.
+1. The leader forwards each encrypted share to its intended recipient. Each
+   aggregator decrypts its input and proof share, computes its verification
+   share, and sends its verification share to the aggregator as usual.
+1. If b=1 in the last step, then the leader also sends along the encrypted input
+   share to each aggregator so that they can decrypt and aggregate the share
+   without needing to maintain state from the previous step.
```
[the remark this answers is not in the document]
cjpatton: Changed "maintain state" to "keep around the input share".

design-document.md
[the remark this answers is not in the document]
cjpatton: I don't think we should try to support arbitrary fields; instead, we should pick a few fields of different sizes and develop optimized versions of each. I suspect that generic bignum libraries would be too slow. (That said, a coworker of mine developed some pretty fast code that works for any prime field of size < 2^126. We'll be open sourcing this very soon.) As far as using off-the-shelf code for ECC fields. My sense is that he design criteria for ECC are quite different from ours. Hence, I'm not sure that this would be suitable. Worth looking into, though.

design-document.md
```diff
@@ -14,4 +24,169 @@
...
+arithmetic in a finite field suitable for the given application.
+
+We will use a prime field. The choice of prime is influenced by the following
+criteria:
+
+1. **Field size.** How big the field needs to be depends on the type of data
+   being aggregated and how many users there are. The field size also impacts
+   the security level: the longer the validity circuit, the larger the field
+   needs to be in order to effectively root out malicious clients.
+1. **Fast polynomial operations.** In order to make Prio practical, it's
```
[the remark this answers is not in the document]
cjpatton: Most Markdown renderers turn these into an ascending sequence, i.e., "1.", "2.", "3.", and so on. This just indicates that I want an enumerated list.

design-document.md
```diff
@@ -14,4 +24,169 @@
...
+
+Our instantiation of the core protocol involves two additional operations:
+public key encryption and pseudorandom number generation. The combination of
+these primitives that we use here allows us to make an additional
+simplification. If we assume that a secure channel will be used between the user
+and the leader (e.g., a server-authenticated TLS connection), then there is no
+need to encrypt the leaders input/proof share. In fact, all that we need to
+encrypt are seeds for a PRG. As a result, in our application we can get away
+with using a simple key-encapsulation mechanism (KEM) rather than full-blown
+encryption.
```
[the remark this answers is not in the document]
cjpatton: No, since ECIES is full blown encryption.

Note that we may end up wanting full-blown encryption, depending on our security considerations. If we do, then HPKE would still provide it.

design-document.md
```diff
@@ -14,4 +24,169 @@
...
+   roots of unity for 1 <= d <= n.
+1. **Highly composite subgroup.** Suppose that (p-1) = 2^n * s. It's best if s
+   is highly composite because this minimizes the number of multiplications
+   required to compute the inverse or apply Fermat's Little Theorem. (See
+   [BBC+19, Section 5.2].)
+
+The table below lists parameters that meet these criteria at various levels of
+security. (Note that \#2 is the field used in "Prio v2".)
+
+| #  | bits   | p                                      | g  | n   | s                |
```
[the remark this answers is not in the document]
cjpatton: "bits" indicates the size of the prime, i.e., the prime is roughly 2^bits. E.g, log2(3221225473) \approx 31.58. I added a note to clarify.

design-document.md
```diff
@@ -14,4 +24,171 @@
...
+   whether the input is deemed valid. This algorithm is run by the leader.
+
+The values above have following types:
+
+1. Input x is an element of K^n for some integer n.
+1. Proof pf is an element of K^p(n) for some function p.
+1. The joint randomness r is an element of K^u(n) for some function u.
+1. Each verification share {vf:i} is an element of K^v(n) for some function v.
+
+Before the protocol begins, the aggregators agree on joint randomness r and
```
chris-wood, line 57: Should we mark this consensus protocol (for both leader and r selection) as a dependency?

design-document.md
```diff
@@ -14,4 +24,171 @@
...
+1. b := Decide({vf}, r) denotes the execution of the decision procedure on input
+   shares {vf} and joint randomness r. The output b is a boolean indicating
+   whether the input is deemed valid. This algorithm is run by the leader.
+
+The values above have following types:
+
+1. Input x is an element of K^n for some integer n.
+1. Proof pf is an element of K^p(n) for some function p.
+1. The joint randomness r is an element of K^u(n) for some function u.
+1. Each verification share {vf:i} is an element of K^v(n) for some function v.
```
chris-wood, line 55: Out of curiosity, why use separate functions to denote the extension of K for each type? Is it not the case that `n = p(n) = u(n) = v(n)`?

design-document.md
```diff
@@ -14,4 +24,171 @@
...
+1. The user runs pf := Prove(x). It splits x and pf into {x} and {pf}
+   respectively and sends ({x:i}, {pf:i}) to aggregator i.
+1. Each aggregator i runs {vf:i} := Query({x:i}, {pf:i}, r) ands sends {vf:i} to
+   the leader.
+1. The leader runs b := Decide({vf}, r) and sends b to each of the aggregators.
+
+If b=1, then each aggregator i adds its input share {x:i} into its share of the
+aggregate. Once all of the inputs have been validated and aggregated, the
+aggregators send their aggregate shares to the leader, who adds them together to
+obtain the final result.
```
chris-wood, line 70:
```suggestion
obtain the final result.

[[OPEN ISSUE: sketch out the b=1 path]]
```

design-document.md
```diff
@@ -14,4 +24,171 @@
...
+
+The values above have following types:
+
+1. Input x is an element of K^n for some integer n.
+1. Proof pf is an element of K^p(n) for some function p.
+1. The joint randomness r is an element of K^u(n) for some function u.
+1. Each verification share {vf:i} is an element of K^v(n) for some function v.
+
+Before the protocol begins, the aggregators agree on joint randomness r and
+designate one of the aggregators as the leader. The protocol proceeds as
```
chris-wood, line 58:
```suggestion
designate one of the aggregators as the leader.

[[OPEN ISSUE: specify consensus protocol.]]

The protocol proceeds as
```

design-document.md
```diff
@@ -14,4 +24,171 @@
...
+If b=1, then each aggregator i adds its input share {x:i} into its share of the
+aggregate. Once all of the inputs have been validated and aggregated, the
+aggregators send their aggregate shares to the leader, who adds them together to
+obtain the final result.
+
+**Proof generation and verification.**
+[[TODO: Describe how to construct proof systems for languages recognized by
+validity circuits with G-gates, a la [BCC+19, Theorem 4.3].]]
+
+### Changes to the core protocol
```
chris-wood, line 76: What is the core protocol? (Ideally this doc would be self-contained, so the core should be sketched here or somewhere nearby.)

design-document.md
```diff
@@ -14,4 +24,171 @@
...
+
+**Coordinating state.**
+The state of the input-validation protocol is maintained by the leader; except
+for aggregation of the input shares, the other aggregators are completely
+stateless. This is accomplished by making the following changes to the core
+protocol.
+
+1. The user sends all of its shares to the leader. To maintain privacy, the user
+   encrypts each input/proof share under the public key of the share's
```
chris-wood, line 85:
```suggestion
   encrypts each (input, proof) share under the public key of the share's
```

design-document.md
```diff
@@ -14,4 +24,171 @@
...
+   recipient.
+1. The leader forwards each encrypted share to its intended recipient. Each
+   aggregator decrypts its input and proof share, computes its verification
+   share, and sends its verification share to the aggregator as usual.
+1. If b=1 in the last step, then the leader also sends along the encrypted input
+   share to each aggregator so that they can decrypt and aggregate the share
+   without needing to keep around the input share from the previous step.
```
chris-wood, line 92:
```suggestion
   without needing to cache the input share from the previous step.
```

design-document.md
```diff
@@ -14,4 +24,171 @@
...
+finite field. As such, the main ingredient of Prio is an implementation of
+arithmetic in a finite field suitable for the given application.
+
+We will use a prime field. The choice of prime is influenced by the following
+criteria:
+
+1. **Field size.** How big the field needs to be depends on the type of data
+   being aggregated and how many users there are. The field size also impacts
+   the security level: the longer the validity circuit, the larger the field
+   needs to be in order to affectively detect malicious clients.
```
chris-wood, line 123: Can we mark as an open issue to parameterize this (as HCG provided)?

design-document.md
```diff
@@ -14,4 +24,171 @@
...
+   is highly composite because this minimizes the number of multiplications
+   required to compute the inverse or apply Fermat's Little Theorem. (See
+   [BBC+19, Section 5.2].)
+
+The table below lists parameters that meet these criteria at various levels of
+security. (Note that \#2 is the field used in "Prio v2".) The "bits" column
+indicates the size of the prime. For example, for p=3221225473,
+bits=log2(3221225473), which is approximately 31.53.
+
+| #  | bits   | p                                      | g  | n   | s                |
```
chris-wood, line 141: How did we generate these primes?

design-document.md
```diff
@@ -14,4 +24,171 @@
...
+| 2  | 31.99  | 4293918721                             | 19 | 20  | 3^2 * 5 * 7 * 13 |
+| 3  | 63.75  | 15564440312192434177                   | 5  | 59  | 3^3              |
+| 4  | 66.17  | 83010348331692982273                   | 11 | 63  | 3^2              |
+| 5  | 122.81 | 9304595970494411110326649421962412033  | 3  | 120 | 7                |
+| 6  | 125.40 | 56076806072176138388129360355577036801 | 13 | 116 | 3^3 * 5^2        |
+
+#### Key encapsulation
+
+Our instantiation of the core protocol involves two additional operations:
+public key encryption and pseudorandom number generation. The combination of
```
chris-wood, line 153:
```suggestion
public key encryption and cryptographically secure pseudorandom number generation (CSPRNG). The combination of
```

chris-wood: (common form in IETF standards)

design-document.md
```diff
@@ -14,4 +24,171 @@
...
+these primitives that we use here allows us to make an additional
+simplification. If we assume that a secure channel will be used between the user
+and the leader (e.g., a server-authenticated TLS connection), then there is no
+need to encrypt the leaders input/proof share. In fact, all that we need to
+encrypt are seeds for a PRG. As a result, in our application we can get away
+with using a simple key-encapsulation mechanism (KEM) rather than full-blown
+encryption.
```
chris-wood, line 160:
```suggestion
these primitives that we use here allows us to make an additional
simplification. However, we assume users and leaders communicate over a secure,
authenticated channel, such as TLS. As a result, we only need to encrypt
CSPRNG seeds, which requires only a key-encapsulation mechanism (KEM) 
rather than full-blown encryption.
```

design-document.md
```diff
@@ -14,4 +24,171 @@
...
+
+To generate an aggregator's share, the user runs (c[i], k[i]) := Encaps(pk[i])
+and sends c[i] to the aggregator. To compute its share, the aggregator would run
+k[i] := Decaps(sk[i], c[i]) and compute its share as {x:i} = PRG(k[i], n).
+
+[HPKE](https://datatracker.ietf.org/doc/draft-irtf-cfrg-hpke/) is a natural
+candidate for instantiating the KEM. In "Export-Only" mode, HPKE provides an
+efficient scheme with all the cryptographic agility we would ever need. And
+although it's still an Internet-Draft, it has high quality implementations in a
+variety of languages.
```
chris-wood, line 177:
```suggestion
variety of languages.

[[OPEN ISSUE: specify how HPKE can implement Encaps and Decaps above.]]
```

design-document.md
```diff
@@ -14,4 +24,171 @@
...
+
+A suitable PRG will have the following syntax. Fix a finite field K:
+
+1. x := PRG(k, n) denotes generation of a pseudorandom element x of K^n, i.e., a
+   vector of n elements of K.
+
+This can be instantiated using a standard stream cipher, e.g.., ChaCha20 as
+follows. Interpret k as the cipher key, and using a fixed nonce, generate l\*n
+bytes of output, where l is the number of bytes needed to encode an element of
+K, then map each chunk of l bytes to an element of K in the natural way.
```
chris-wood, line 189:
```suggestion
K, then map each chunk of l bytes to an element of K in the natural way.

[[OPEN ISSUE: determine if more than k*l byes are needed to deal with bias for each element.]]
```

chris-wood (approved):
🚢 great start!

design-document.md
cjpatton: This is something we'll have to figure out once we have picked a threat model. I'll mark it as an open issue for now.

design-document.md
cjpatton: Actually "K^n" is meant to denote the set of vectors over K of length n. :grimacing: I added this notation to the "Notation" paragraph above.

To your question: in general, the length of the proof, verification message, and joint randomness is a function of the length of length of the input. For example, for an input length of n, the corresponding proof might be p(n) = a*n + b for some constants a, b.

design-document.md
cjpatton: The 'core protocol" is the input-validation protocol described in the previous section.

design-document.md
cjpatton: Done, here and elsewhere.

design-document.md
cjpatton: Done.

design-document.md
cjpatton: Added an explanation below.

cjpatton: (Also updated one of the primes.)

design-document.md
cjpatton: Great point! Done.

design-document.md
```diff
@@ -14,4 +24,205 @@
...
+### Changes to the input-validation protocol
+
+**Coordinating state.**
+The state of the input-validation protocol is maintained by the leader; except
+for aggregation of the input shares, the other aggregators are completely
+stateless. This is accomplished by making the following changes to the core
+protocol.
+
+1. The user sends all of its shares to the leader. To maintain privacy, the user
+   encrypts each (input, proof) share under the public key of the share's
```
aaomidi, line 98: How does the public key of the other recipients get advertised?

cjpatton: That's TBD. I was thinking the user would simply make an HTTP request to each aggregator, but there might be a better way.

Added a TODO below.

design-document.md
```diff
@@ -14,4 +24,210 @@
...
+   without needing to cache the input share from the previous step.
+
+**Minimizing bandwidth overhead.**
+In most linear secret sharing schemes, the length of each share is equal to the
+length of the input. Therefore, the communication overhead for the user is
+O(s\*(n+p(n))). This can be reduced to O(s+n+p(n)) with the following standard
+trick.
+
+Let x be an element of K^n for some n. Suppose we split in into {x} by choosing
+{x:1}, ..., {x:s-1} at random and letting {x:s} = in - ({x:1} + ... + {x:s-1}).
```
tgeoghegan, line 117: I'm a bit confused by `in` vs. `x` in this paragraph. Is `in` meant to be the input before being split into the vector `{x}`? If so, should line 116 read "Let in be an element of..."?

design-document.md
```diff
@@ -14,4 +24,210 @@
...
+        print(bits, p, GF(p).multiplicative_generator(), n, factor(s))
+```
+
+#### Key encapsulation
+
+Our instantiation of the input-validation protocol involves two additional
+operations: public key encryption and cryptographically secure pseudorandom
+number generation (CSPRNG). The combination of these primitives that we use here
+allows us to make an additional simplification. We assume that users communicate
+with the leader over a confidential and authenticated channel, such as TLS. As a
```
tgeoghegan, line 189: Users (/clients) communicate with each aggregator, not just the leader, right?

design-document.md
```diff
@@ -14,4 +24,210 @@
...
+result, we only need to encrypt CSPRNG seeds, which requires only a
+key-encapsulation mechanism (KEM) rather than full-blown encryption.
+
+A KEM is comprised of two algorithms:
+
+1. (c, k) := Encaps(pk) denotes generation and encapsulation of symmetric key k
+   under the recipient's public key pk.
+1. k := Decaps(pk, c) denotes decapsulation of symmetric key k under the
+   recipient's secret key sk.
```
tgeoghegan, line 198: Should this be `Decaps(sk, c)`?

design-document.md
```diff
@@ -14,4 +24,210 @@
...
 
-## Cryptographic dependencies
+## Cryptographic components
+
+### The input-validation protocol
+
+**Notation.**
+Fix a finite field K. We let K^n denote the set of vectors over K of length n.
+We write the linear, s-way secret sharing of an element x of K^n as {x:1}, ...,
+{x:s}, where {x:i} is the share held by the i-th party. We write {x} as
```
tgeoghegan, line 34: I think it's worth clarifying that `{x:i}` is itself an element of `K^n` (right?)

design-document.md
```diff
@@ -14,4 +24,210 @@
...
+
+1. The user runs pf := Prove(x). It splits x and pf into {x} and {pf}
+   respectively and sends ({x:i}, {pf:i}) to aggregator i.
+1. Each aggregator i runs {vf:i} := Query({x:i}, {pf:i}, r) ands sends {vf:i} to
+   the leader.
+1. The leader runs b := Decide({vf}, r) and sends b to each of the aggregators.
+
+If b=1, then each aggregator i adds its input share {x:i} into its share of the
+aggregate. Once all of the inputs have been validated and aggregated, the
+aggregators send their aggregate shares to the leader, who adds them together to
```
tgeoghegan, line 71: I think this should say something like "Once all of the inputs have been validated and a sufficient number of input shares have been aggregated, the aggregators send their aggregate shares to the leader...", to make it clear that aggregators aren't disclosing individual input shares to the leader.

design-document.md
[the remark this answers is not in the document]
tgeoghegan: For what it's worth, I think ECIES as used in Prio v2 is overkill for v3 for two reasons:

1. Prio v2 needs to tunnel a confidential+authenticated channel from mobile devices to aggregators through Apple or Google servers, so we couldn't rely on TLS. In v3, if clients send shares directly to aggregators, we don't have that problem.
2. In v2, clients will generate an ephemeral ECDSA key and do ECDH with the aggregator's long term public key to make a fresh symmetric for every single datum moving through the system. We were constrained there by the fact that the mobile OS vendors had already built and shipped their clients and were reluctant to make changes to them, but I think we could reuse symmetric keys for >1 datum.

tgeoghegan (changes requested):
Thanks for getting the ball rolling!

design-document.md
```diff
@@ -1,5 +1,15 @@
...
 
+**Terminology.**
+1. Aggregator: A server that runs the input-validation protocol and accumulates
+   input shares.
+1. Input-validation protocol: The protocol executed by the user and aggregators
+   in order to validate the user's input without leaking its value to the
+   aggregators.
+1. Leader: A distinguished aggregator that coordinates input validation and data
+   aggregation.
+1. User: the party contributing data into an aggregate.
```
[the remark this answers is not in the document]
cjpatton: Here I mean "client". Will fix.

design-document.md
```diff
@@ -14,4 +24,210 @@
...
+
+### The input-validation protocol
+
+**Notation.**
+Fix a finite field K. We let K^n denote the set of vectors over K of length n.
+We write the linear, s-way secret sharing of an element x of K^n as {x:1}, ...,
+{x:s}, where {x:i} is the share held by the i-th party. We write {x} as
+shorthand for the sequence {x:1}, ..., {x:s}.
+
+Prio combines standard linear secret sharing with a new type of
```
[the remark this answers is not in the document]
cjpatton: Done.

design-document.md
```diff
@@ -14,4 +24,210 @@
...
+allows us to make an additional simplification. We assume that users communicate
+with the leader over a confidential and authenticated channel, such as TLS. As a
+result, we only need to encrypt CSPRNG seeds, which requires only a
+key-encapsulation mechanism (KEM) rather than full-blown encryption.
+
+A KEM is comprised of two algorithms:
+
+1. (c, k) := Encaps(pk) denotes generation and encapsulation of symmetric key k
+   under the recipient's public key pk.
+1. k := Decaps(pk, c) denotes decapsulation of symmetric key k under the
```
chris-wood, line 197:
```suggestion
1. k := Decaps(sk, c) denotes decapsulation of symmetric key k under the
```

design-document.md
chris-wood: I think, in this proposed design, clients communicate only with the leader, who mediates all things. @cjpatton?

design-document.md
chris-wood, line 117:
```suggestion
Let x be an element of K^n for some n. Suppose we split x into {x} by choosing
{x:1}, ..., {x:s-1} at random and letting {x:s} = x - ({x:1} + ... + {x:s-1}).
```

design-document.md
```diff
@@ -14,4 +24,210 @@
...
+
+1. pf := Prove(x) denotes generation of a proof pf of the validity of input x.
+   This algorithm is executed by the user.
+1. {vf:i} := Query({x:i}, {pf:i}, r) denotes computation of the verification
+   share {vf:i} for input share {x:i} and proof share {pf:i}. This algorithm is
+   executed by each of the aggregators; input r denotes the joint randomness
+   shared by all of the aggregators.
+1. b := Decide({vf}, r) denotes the execution of the decision procedure on input
+   shares {vf} and joint randomness r. The output b is a boolean indicating
+   whether the input is deemed valid. This algorithm is run by the leader.
```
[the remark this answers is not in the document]
cjpatton: The threat model for Prio --- as it's described in the original paper and [BBC+19] --- considers **either** a malicious client (attacking soundness) **or** a malicious subset of aggregators (attacking privacy). In particular, soundness isn't guaranteed if any one of the aggregators is malicious; in theory it may be possible for a malicious client and aggregator to collude and break soundness.

There are techniques described in [BBC+19] that address the stronger threat model in which a malicious client may collude with a malicious aggregator. I think it's worth exploring how practical these techniques are. Meanwhile, thinking about attacks is a great idea.

design-document.md
chris-wood: `{x:i}` is a share of `x`, so it's just an element of `K`.

chris-wood, line 34:
```suggestion
{x:s}, where {x:i} is the share held by the i-th party and an element of K. We write {x} as
```

design-document.md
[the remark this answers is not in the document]
chris-wood: Maybe just flag this as an open issue?
```suggestion
   whether the input is deemed valid. This algorithm is run by the leader.
   
[[OPEN ISSUE: what can go wrong when the leader is responsible for verifying everything on behalf of all aggregators?]]
```

design-document.md
```diff
@@ -14,4 +24,210 @@
...
+1. The joint randomness r is an element of K^u(n) for some function u.
+1. Each verification share {vf:i} is an element of K^v(n) for some function v.
+
+Before the protocol begins, the aggregators agree on joint randomness r and
+designate one of the aggregators as the leader. The protocol proceeds as
+follows:
+
+1. The user runs pf := Prove(x). It splits x and pf into {x} and {pf}
+   respectively and sends ({x:i}, {pf:i}) to aggregator i.
+1. Each aggregator i runs {vf:i} := Query({x:i}, {pf:i}, r) ands sends {vf:i} to
```
chris-wood, line 65:
```suggestion
1. Each aggregator i runs {vf:i} := Query({x:i}, {pf:i}, r) and sends {vf:i} to
```

design-document.md
cjpatton: Oops, typo: s/in/x/.

design-document.md
cjpatton: I envision the client sending its encrypted shares to the leader. However, it does need to speak to the aggregators prior to the input-validation protocol to get their public keys.

design-document.md
cjpatton: Yup, good catch!

design-document.md
cjpatton: Yes, done.

design-document.md
cjpatton: Changed to "Once a sufficient number of shares have been ..."

design-document.md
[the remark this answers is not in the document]
cjpatton: Added this to "Thread model".

design-document.md
chris-wood: Does it? Why can't the client get these keys from the leader?

cjpatton: The leader is incentivized to break privacy... it could just hand the client fake public keys for which it knows the secret keys and decrypt all of the shares.

tgeoghegan (approved):

PR #3, opened 2021-02-24
Start system overview and requirements.

chris-wood:
Imperfect, incomplete, maybe even inaccurate! Starting to get some stuff down.

### DIFF
--- a/design-document.md
+++ b/design-document.md
@@ -18,7 +18,7 @@ counting the number of times a sensitive or private event occurs and approximati
 that sensitive tokens or strings occur.
 
 Client applications hold private inputs to the aggregation function, server processors,
-or aggregators, invoke  multi-party computation to compute the output, and a final collector 
+or aggregators, run a protocol that validates each input x[1], x[2], ... and computes the final output y. The final collector  
 obtains the output of the aggregation function. At a high level, the flow of data through
 these entities works roughly as follows:
 
--- a/design-document.md
+++ b/design-document.md
@@ -3,8 +3,8 @@
 ## Architecture overview
 
 Prio is a system and protocol for privately computing aggregation functions over private 
-input. An aggregation function F is one that computes an output y = F(x1,x2,...) for inputs
-xi. In general, Prio supports any aggregation function whose inputs can be encoded in a 
+input. An aggregation function F is one that computes an output y = F(x[1],x[2],...) for inputs
+x[i]. In general, Prio supports any aggregation function whose inputs can be encoded in a 
 particular way. However, not all aggregation functions admit an efficient encoding, rendering
 them impractical to implement. Thus, Prio supports a limited set of aggregation functions, 
 some of which we highlight below:
--- a/design-document.md
+++ b/design-document.md
@@ -62,13 +62,13 @@ At a high level, the flow of data through these entities works roughly as follow
 ~~~ 
 
 1. Upload: Clients split inputs into s >= 2 shares, encrypt each share for a different 
-   Aggregator, and send these encrypted shares to the aggregators.
-2. Verify: Upon receipt of an encrypted share, the Leader sends it to the respective 
-   Aggregator. Each Aggregator decrypts the share, computes a proof from the respective
-   share, sends this proof to the leader. Once the Leader collects all proofs for 
-   the batch, it determines whether or not the data for each entry is correct. 
-   (Details about input validation and how it pertains to system security properties 
-   is in {{CITE}}.)
+   Aggregator, and send these encrypted shares to the Aggregators. (Details about Aggregator
+   discovery is in {{CITE}}.)
+2. Verify: Upon receipt of an encrypted share, an Aggregator decrypts the share, 
+   computes a proof from the respective share, and sends this proof to the Leader. 
+   Once the Leader collects all proofs for the batch, it determines whether or not the
+   data for each entry is correct. (Details about input validation and how it pertains 
+   to system security properties is in {{CITE}}.)
 3. Aggregate: Assuming the input share is valid, the Leader instructs each Aggregator 
    to combine aggregate their corresponding input share locally. When complete, each
    Aggregator sends their aggregated input shares to the Leader, who then combines all
--- a/design-document.md
+++ b/design-document.md
@@ -11,7 +11,7 @@ some of which we highlight below:
 
 - Simple statistics, including sum, mean, min, max, variance, and standard deviation;
 - Bit vector OR and AND operations; and
-- Count-min sketch (approximated frequency counts) over a closed universe of strings.
+- Data structures, like Bloom filters, counting Bloom filters, and count-min sketches, that approximately represent (multi-)sets of strings.
 
 The applications for such aggregations functions are large, including, though not limited to:
 counting the number of times a sensitive or private event occurs and approximating the frequency
--- a/design-document.md
+++ b/design-document.md
@@ -18,43 +18,43 @@ counting the number of times a sensitive or private event occurs and approximati
 that sensitive tokens or strings occur.
 
 Client applications hold private inputs to the aggregation function, server processors,
-or custodians, interact in a multi-party computation to compute the output, and a final
-collector obtains the output of the aggregator. At a high level, the flow of data through
+or aggregators, invoke  multi-party computation to compute the output, and a final collector 
+obtains the output of the aggregation function. At a high level, the flow of data through
 these entities works roughly as follows:
 
 ~~~
                             +------------+     
  (1) Batch submission       |            |        (3) Collection
-    +-----------------------> Processor  +------------------+
+    +-----------------------> Aggregator +------------------+
     |                       |            |                  |
     |                       +-^-------^--+                  |
     |                         |       |                     |
     |                         |       |                     |
     |                         |       |  (2) MPC            |
-+--------+           +--------v--+    |      eval      +----v------+
-|        |           |           |    |                |           |
-| Client +-----------> Processor |    |                | Collector |
-|        |           |           |    |                |           |
-+--------+           +--------^--+    |                +----^------+
++--------+           +--------v---+   |      eval      +----v------+
+|        |           |            |   |                |           |
+| Client +-----------> Aggregator |   |                | Collector |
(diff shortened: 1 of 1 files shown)


design-document.md
```diff
@@ -1,8 +1,89 @@
 # Prio v3 Design Document
 
-## Sample user stories
+## Architecture overview
 
-## Threat model
+Prio is a system and protocol for privately computing aggregation functions over private 
+input. An aggregation function F is one that computes an output y = F(x1,x2,...) for inputs
+xi. In general, Prio supports any aggregation function whose inputs can be encoded in a
```
cjpatton, line 7:
```suggestion
input. An aggregation function F is one that computes an output y = F(x[1],x[2],...) for inputs
x[i]. In general, Prio supports any aggregation function whose inputs can be encoded in a 
```

design-document.md
```diff
@@ -1,8 +1,89 @@
...
 
-## Threat model
+Prio is a system and protocol for privately computing aggregation functions over private 
+input. An aggregation function F is one that computes an output y = F(x1,x2,...) for inputs
+xi. In general, Prio supports any aggregation function whose inputs can be encoded in a 
+particular way. However, not all aggregation functions admit an efficient encoding, rendering
+them impractical to implement. Thus, Prio supports a limited set of aggregation functions, 
+some of which we highlight below:
+
+- Simple statistics, including sum, mean, min, max, variance, and standard deviation;
```
cjpatton, line 12: [[OPEN ISSUE: It's possible to estimate quantiles such as the median. How practical is this?]]

design-document.md
```diff
@@ -1,8 +1,89 @@
...
+- Bit vector OR and AND operations; and
+- Count-min sketch (approximated frequency counts) over a closed universe of strings.
```
cjpatton, line 14:
```suggestion
- Data structures, like Bloom filters, counting Bloom filters, and count-min sketches, that approximately represent (multi-)sets of strings.
```

design-document.md
```diff
@@ -1,8 +1,89 @@
...
+
+The applications for such aggregations functions are large, including, though not limited to:
+counting the number of times a sensitive or private event occurs and approximating the frequency
+that sensitive tokens or strings occur.
+
+Client applications hold private inputs to the aggregation function, server processors,
+or aggregators, invoke  multi-party computation to compute the output, and a final collector
```
cjpatton, line 21:
```suggestion
or aggregators, run a protocol that validates each input x[1], x[2], ... and computes the final output y. The final collector  
```

("invoke multi-party computation" sounds like we're doing general purpose MPC.)

design-document.md
```diff
@@ -1,8 +1,89 @@
...
+|        |           |            |   |                |           |
++--------+           +--------^---+   |                +----^------+
+    |                         |       |                     |
+    |                         |       |                     | 
+    |                         |       |                     |
+    |                       +-v-------v--+                  |
+    |                       |            |                  |
+    +-----------------------> Aggregator +------------------+
+                            |            |
+                            +------------+
```
cjpatton, line 46: Pretty! Though I think we need to align this with the candidate protocol #2. In particular, I would remove the wires going from the client to the top and bottom aggregators, and I would rename the middle aggregator to "Leader'.

design-document.md
```diff
@@ -1,8 +1,89 @@
...
+~~~ 
+
+1. Applications split inputs into multiple (at least two) anonymized and encrypted shares,
+   and upload each share to different aggregators that do not collude or otherwise share 
+   data with one another. Applications continue this process until a "batch" of data is 
+   collected. Upon receipt of a share, each aggregator verifies it for correctness. 
+   (Details about input validation and how it pertains to system security properties is 
+   in {{CITE}}.)
```
cjpatton, line 54: Needs alignment with the candidate protocol of #2.

design-document.md
```diff
@@ -1,8 +1,89 @@
...
+2. Each aggregator combines its shares into a partial sum. The aggregators then engage 
+   in a multi-party protocol to combine these sums into a final, aggregated output.
+3. The aggregated output is sent to the collector.
+
+The output of a single batch aggregation reveals little to nothing beyond the value itself.
+
+## Security overview
+
+Prio assumes a powerful adversary with the ability to compromise an unbounded number of 
+clients. In doing so, the adversary can input malicious (yet truthful) to the aggregation
```
cjpatton, line 64:
```suggestion
clients. In doing so, the adversary can provide malicious (yet truthful) inputs to the aggregation 
```
Also, what does "yet truthful" mean here?

design-document.md
```diff
@@ -12,6 +93,9 @@
 
 ## System design
 
-## Open questions and system parameters
+## Open questions and system parameters {#questions-and-params}
+
+[[OPEN ISSUE: discuss batch size parameter and thresholds]]
+[[OPEN ISSUE: discuss f^ leakage differences from HCG's paper]]
```
cjpatton, line 99: What does "f^" mean?

cjpatton (commented):
Good start. The main thing is that this needs to be aligned with the candidate protocol in #2. Let's merge that PR, then rebase this.

design-document.md
```diff
@@ -1,8 +1,89 @@
...
 
-## Threat model
+Prio is a system and protocol for privately computing aggregation functions over private 
+input. An aggregation function F is one that computes an output y = F(x[1],x[2],...) for inputs
+x[i]. In general, Prio supports any aggregation function whose inputs can be encoded in a 
+particular way. However, not all aggregation functions admit an efficient encoding, rendering
+them impractical to implement. Thus, Prio supports a limited set of aggregation functions, 
+some of which we highlight below:
+
+- Simple statistics, including sum, mean, min, max, variance, and standard deviation;
```
chris-wood, line 12:
```suggestion
- Simple statistics, including sum, mean, min, max, variance, and standard deviation; [[OPEN ISSUE: It's possible to estimate quantiles such as the median. How practical is this?]]


```

cjpatton (approved):

design-document.md
```diff
@@ -11,9 +11,98 @@
...
+that servers communicate over secure and mutually authenticated channels. In practice,
+this can be done by TLS or some other form of application-layer authentication.
+
+In the presence of this adversary, Prio provides two important properties for computing 
+an aggergation function F:
+
+1. Privacy. The adversary learns only the output of F computed over all client inputs, 
+   and nothing else. 
+1. Robustness. The adversary can influence the output of F only by reporting false 
+   (untruthful) data. The output cannot be influenced in any other way.
```
tgeoghegan, line 95: The 2017 paper's section 2 discusses _anonymity_ alongside privacy and robustness. However I'm not sure how to draw a clear distinction between anonymity and privacy in this context so maybe it's not worth discussing in this doc.

design-document.md
```diff
@@ -11,9 +11,98 @@
...
 
-## Sample user stories
-
-## Threat model
+## Architecture overview
+
+Prio is a system and protocol for privately computing aggregation functions over private 
+input. An aggregation function F is one that computes an output y = F(x[1],x[2],...) for inputs
+x[i]. In general, Prio supports any aggregation function whose inputs can be encoded in a 
+particular way. However, not all aggregation functions admit an efficient encoding, rendering
```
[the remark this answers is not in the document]
chris-wood: Indeed -- and the mechanism is described later on in this doc, so I figured best to keep it high level for now.

design-document.md
```diff
@@ -11,9 +11,98 @@
...
+                            +-v-------v--+     
+                            |            |     
+                            | Aggregator |
+                            |            |
+                            +------------+
+~~~ 
+
+1. Upload: Clients split inputs into s >= 2 shares, encrypt each share for a different 
+   Aggregator, and send these encrypted shares to the Leader. Clients continue 
+   this process until a "batch" of data is collected.
```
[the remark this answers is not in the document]
chris-wood: That's a good point! I'll move this batch cutoff to the leader step.

design-document.md
chris-wood: Indeed. I suggest we leave this as-is for now.

design-document.md
```diff
@@ -11,9 +11,98 @@
...
+2. Verify: Upon receipt of an encrypted share, the Leader sends it to the respective 
+   Aggregator. Each Aggregator decrypts the share, computes a proof from the respective
```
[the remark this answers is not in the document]
chris-wood: That's one possibility, sure, but I don't think that's necessary. The leader could package up signed keys that it collects from aggregators and send them to the client. Clients could also fetch them directly from aggregators (if they knew what aggregators to fetch from). From a client perspective, the former seems simpler, since it would allow us to punt on aggregator discovery.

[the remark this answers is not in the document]
tgeoghegan: The leader could vend signed keys on behalf of aggregators, but the client still needs to discover a trusted public key for each aggregator to then verify the signed keys. I think you have to bootstrap trust between the client and each participating aggregator independently from the leader, or the leader can impersonate aggregators and defeat the system's privacy guarantees.

design-document.md
cjpatton: To resolve this, please add a reference to the paper.

design-document.md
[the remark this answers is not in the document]
cjpatton: IMO it's better to avoid creating a new PKI and just have aggregators distribute their own encryption keys. I'm open to being wrong, however :)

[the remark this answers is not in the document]
chris-wood: I'm not suggesting a new PKI for this.

design-document.md
cjpatton: I would resolve this by adding a refereence to the paper.

design-document.md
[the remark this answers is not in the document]
chris-wood: >  but the client still needs to discover a trusted public key for each aggregator to then verify the signed keys.

Yes, it does, and I'm suggesting this can be done by asking the leader to provide the signed keys.

design-document.md
chris-wood: Can you please leave what you're looking for as a suggestion? (There is a reference -- "HCG's paper" -- no?)

tgeoghegan (approved):

PR #6, opened 2021-03-02
Add an initial section on system constraints

acmiyaguchi:
This PR adds a draft of system constraints. I had difficulty finding example documents to lend structure here, so apologies if the style or ideas are off the rails.

I've added three constraints that should be relevant to future discussions: data resolution, timeliness, and integrity. I'm likely missing others. Some of the text involves services or applications that implement/use Prio -- I'm not sure if it's appropriate for this document.

### DIFF
--- a/design-document.md
+++ b/design-document.md
@@ -37,34 +37,37 @@ applications may choose to utilize services implementing the specification.
 ### Data resolution limitations
 
 Privacy comes at the cost of computational complexity. While affine-aggregatable
-encodings can compute many useful statistics, they require more bandwidth and
-CPU cycles to account for finite-field arithmetic during input-validation. The
-throughput of the system bounds the volume of data processed, which is
-related to the verification circuit's complexity and the rate of verifications
-performed by each work unit in an aggregator.
+encodings (AFEs) can compute many useful statistics, they require more bandwidth
+and CPU cycles to account for finite-field arithmetic during input-validation.
+The increased work from verifying inputs decreases the throughput of the system
+or the inputs processed per unit time. Throughput is related to the verification
+circuit's complexity and the available compute-time to each aggregator.
 
-Applications that utilize proofs with a large number of multiplication gates or a high
-frequency of inputs may need to limit inputs into the system to meet bandwidth
-or compute constraints. Some methods of overcoming these limitations include
-choosing a better representation for the data or introducing sampling into the
-data collection methodology.
+Applications that utilize proofs with a large number of multiplication gates or
+a high frequency of inputs may need to limit inputs into the system to meet
+bandwidth or compute constraints. Some methods of overcoming these limitations
+include choosing a better representation for the data or introducing sampling
+into the data collection methodology.
+
+[[TODO: Discuss explicit key performance indicators, here or elsewhere.]]
 
 ### Aggregation utility and soft batch deadlines
 
-A soft real-time system should produce a response within a deadline in order to be useful. This
-constraint may be relevant when the value of an aggregate decreases over time.
+A soft real-time system should produce a response within a deadline to
+be useful. This constraint may be relevant when the value of an aggregate
+decreases over time. A missed deadline can reduce an aggregate's utility
+but not necessarily cause failure in the system.
 
-An example of a soft real-time constraint is the expectation that input data can be
-verified and aggregated in a period equal to data collection, given some
+An example of a soft real-time constraint is the expectation that input data can
+be verified and aggregated in a period equal to data collection, given some
 computational budget. Meeting these deadlines will require efficient
 implementations of the input-validation protocol. Applications might batch
 requests or utilize more efficient serialization to improve throughput.
 
 Some applications may be constrained by the time that it takes to reach a
 privacy threshold defined by a minimum number of input shares. One possible
-solution is to increase the reporting period so more samples can be collected
-before the period is over, although this may need to be balanced with the need
-to produce a response within a soft deadline.
+solution is to increase the reporting period so more samples can be collected,
+balanced against the urgency of responding to a soft deadline.
 
 ### Aggregation services may need to fulfill data integrity constraints
 
@@ -76,15 +79,17 @@ every share must be processed exactly once by all aggregators. Data integrity
 constraints may be at odds with the threat model if meeting the constraints
 requires replaying data.
 
-Invalid shares are expected when operating an aggregator and do not necessarily
-constitute failures in the protocol. However, a large number of invalid
-or identical inputs may be concerning participants in the protocol.
+Aggregator operators should expect to encounter invalid inputs during regular
+operation due to misconfigured or malicious clients. Low volumes of errors are
+tolerable; the input-verification protocol and AFEs are robust in the face of
+malformed data. Aggregators may need to detect and mitigate statistically
+significant floods of invalid or identical inputs that affect accuracy, e.g.,
+denial of service (DoS) events.
 
 Certain classes of errors do not exist in the input-validation protocol
-considered in this document. For example, errors from packet loss when clients
+considered in this document. For example, packet loss errors when clients
 make requests directly to aggregators are not relevant when the leader proxies
-requests and controls the schedule for signaling the end of the aggregation
-rounds.
+requests and controls the schedule for signaling aggregation rounds.
 
 ## System design
 
--- a/design-document.md
+++ b/design-document.md
@@ -69,7 +69,7 @@ privacy threshold defined by a minimum number of input shares. One possible
 solution is to increase the reporting period so more samples can be collected,
 balanced against the urgency of responding to a soft deadline.
 
-### Aggregation services may need to fulfill data integrity constraints
+### Data integrity constraints
 
 Data integrity concerns the accuracy and correctness of the outputs in the
 system. The integrity of the output can be influenced by an incomplete round of
--- a/design-document.md
+++ b/design-document.md
@@ -30,6 +30,62 @@ practical.]]
 
 ## System constraints
 
+Prio has inherent constraints derived from the tradeoff between privacy
+guarantees and computational complexity. These tradeoffs influence how
+applications may choose to utilize services implementing the specification.
+
+### Data types have limited resolution due to input-validation complexity
+
+Privacy comes at the cost of computational complexity. While affine-aggregatable
+encodings can compute many useful statistics, they require more bandwidth and
+CPU cycles to account for finite-field arithmetic during input-validation. The
+throughput of the system bounds the volume of data processed, which is
+proportional to the verification circuit's size and the rate of verifications
+performed by each work unit in an aggregator.
+
+Applications that utilize proofs with a large number of gates or a high
+frequency of inputs may need to limit inputs into the system to meet bandwidth
+or compute constraints. Some methods of overcoming these limitations include
+choosing a better representation for the data or introducing sampling into the
+data collection methodology.
+
+### Aggregation services may need to fulfill near-real-time constraints
+
+A near-real-time system must produce a response within a deadline. This
+constraint may be relevant when the value of an aggregate decreases over time.
+
+An example of a near-real-time constraint is the expectation that a can be
+verified and aggregated in a period equal to data collection, given some
+computational budget. Meeting these deadlines may require efficient
+implementations of the input-validation protocol. Applications might batch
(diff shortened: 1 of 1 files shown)


design-document.md
```diff
@@ -30,6 +30,62 @@ practical.]]
...
+applications may choose to utilize services implementing the specification.
+
+### Data types have limited resolution due to input-validation complexity
+
+Privacy comes at the cost of computational complexity. While affine-aggregatable
+encodings can compute many useful statistics, they require more bandwidth and
+CPU cycles to account for finite-field arithmetic during input-validation. The
+throughput of the system bounds the volume of data processed, which is
+proportional to the verification circuit's size and the rate of verifications
+performed by each work unit in an aggregator.
```
cjpatton, line 44: Instead of "proportional to the verification circuit's size" I would suggest "related to the circuit's complexity". The size of the proof, and the time complexity of generating and verifying it, can be reduced significantly if the circuit has a lot of redundancy.

design-document.md
```diff
@@ -30,6 +30,62 @@ practical.]]
...
+data collection methodology.
+
+### Aggregation services may need to fulfill near-real-time constraints
+
+A near-real-time system must produce a response within a deadline. This
+constraint may be relevant when the value of an aggregate decreases over time.
+
+An example of a near-real-time constraint is the expectation that a can be
+verified and aggregated in a period equal to data collection, given some
+computational budget. Meeting these deadlines may require efficient
```
cjpatton, line 59:
```suggestion
computational budget. Meeting these deadlines will require efficient
```

design-document.md
```diff
@@ -30,6 +30,62 @@ practical.]]
...
+implementations of the input-validation protocol. Applications might batch
+requests and utilize binary serialization to improve throughput.
```
cjpatton, line 61: What's meant by "binary serialization"?

design-document.md
```diff
@@ -30,6 +30,62 @@ practical.]]
...
+solution is to increase the reporting period so more samples can be collected
+before the period is over, although this may need to be balanced with the need
+to produce a response within a deadline.
+
+### Aggregation services may need to fulfill data integrity constraints
+
+Data integrity concerns the accuracy and correctness of the outputs in the
+system. The integrity of the output can be influenced by an incomplete round of
+aggregation caused by network partitions, or by bad actors attempting to cause
+inaccuracies in the aggregates. An example data integrity constraint may be that
```
cjpatton, line 74:
```suggestion
inaccuracies in the aggregates. An example data integrity constraint is that
```
Definitely an important constraint :)

design-document.md
```diff
@@ -30,6 +30,62 @@ practical.]]
...
+every share must be processed at most once by all aggregators. Data integrity
+constraints may be at odds with the threat model if meeting the constraints
+requires replaying data.
+
+Invalid shares are expected when operating an aggregator and do not necessarily
+constitute failures in the protocol. However, a large number of invalid
+or identical inputs may be concerning participants in the protocol.
```
cjpatton, line 81: Invalid shares are always a problem if they go undetected. It sounds like this paragraph is more about DoS mitigation? I.e., regardless of their validity, the aggregators need to be able to detect if they're being flooded by inputs?

design-document.md
```diff
@@ -30,6 +30,62 @@ practical.]]
...
+
+Certain classes of errors do not exist in the input-validation protocol
+considered in this document. For example, errors from packet loss when clients
+make requests directly to aggregators are not relevant when the leader proxies
+requests and controls the clock cycle for signaling the end of the aggregation
+rounds.
```
cjpatton, line 87: This is good stuff. I feel like we're definitely gonna expand this paragraph as we implement this :)

design-document.md
```diff
@@ -30,6 +30,62 @@ practical.]]
...
+before the period is over, although this may need to be balanced with the need
+to produce a response within a deadline.
+
+### Aggregation services may need to fulfill data integrity constraints
+
+Data integrity concerns the accuracy and correctness of the outputs in the
+system. The integrity of the output can be influenced by an incomplete round of
+aggregation caused by network partitions, or by bad actors attempting to cause
+inaccuracies in the aggregates. An example data integrity constraint may be that
+every share must be processed at most once by all aggregators. Data integrity
```
cjpatton, line 75:
```suggestion
every share must be processed exactly once by all aggregators. Data integrity
```

cjpatton (commented):
This looks good, minor comments only.

design-document.md
```diff
@@ -30,6 +30,62 @@ practical.]]
...
+Prio has inherent constraints derived from the tradeoff between privacy
+guarantees and computational complexity. These tradeoffs influence how
+applications may choose to utilize services implementing the specification.
+
+### Data types have limited resolution due to input-validation complexity
+
+Privacy comes at the cost of computational complexity. While affine-aggregatable
+encodings can compute many useful statistics, they require more bandwidth and
+CPU cycles to account for finite-field arithmetic during input-validation. The
+throughput of the system bounds the volume of data processed, which is
```
chris-wood, line 42: What is the throughput of the system? How many inputs/second can be processed?

Taking a step back: Should we have a section where we discuss this and other key performance indicators that we're trying to be mindful of? (Time between batch start and batch end, where the aggregate is available, comes to mind, for example.)

design-document.md
```diff
@@ -30,6 +30,62 @@ practical.]]
...
+proportional to the verification circuit's size and the rate of verifications
```
chris-wood, line 43:
```suggestion
related to the verification circuit's complexity and the rate of verifications
```

design-document.md
```diff
@@ -30,6 +30,62 @@ practical.]]
...
+performed by each work unit in an aggregator.
+
+Applications that utilize proofs with a large number of gates or a high
```
chris-wood, line 46:
```suggestion
Applications that utilize proofs with a large number of multiplication gates or a high
```

design-document.md
```diff
@@ -30,6 +30,62 @@ practical.]]
 
 ## System constraints
 
+Prio has inherent constraints derived from the tradeoff between privacy
+guarantees and computational complexity. These tradeoffs influence how
+applications may choose to utilize services implementing the specification.
+
+### Data types have limited resolution due to input-validation complexity
```
chris-wood, line 37:
```suggestion
### Data resolution limitations
```

design-document.md
```diff
@@ -30,6 +30,62 @@ practical.]]
...
+
+Applications that utilize proofs with a large number of gates or a high
+frequency of inputs may need to limit inputs into the system to meet bandwidth
+or compute constraints. Some methods of overcoming these limitations include
+choosing a better representation for the data or introducing sampling into the
+data collection methodology.
+
+### Aggregation services may need to fulfill near-real-time constraints
+
+A near-real-time system must produce a response within a deadline. This
```
chris-wood, line 54:
```suggestion
A soft real-time system should produce a response within a deadline in order to be useful. This
```

Any sort of time synchronization or deadline enforcement here seems challenging, given the nature of the distributed system. I'd suggest relaxing this text a bit to say that deadlines are soft, and missing a deadline just means that the output has lesser value.

design-document.md
```diff
@@ -30,6 +30,62 @@ practical.]]
...
+proportional to the verification circuit's size and the rate of verifications
+performed by each work unit in an aggregator.
+
+Applications that utilize proofs with a large number of gates or a high
+frequency of inputs may need to limit inputs into the system to meet bandwidth
+or compute constraints. Some methods of overcoming these limitations include
+choosing a better representation for the data or introducing sampling into the
+data collection methodology.
+
+### Aggregation services may need to fulfill near-real-time constraints
```
chris-wood, line 52:
```suggestion
### Aggregation utility and soft batch deadlines
```

design-document.md
```diff
@@ -30,6 +30,62 @@ practical.]]
...
+
+A near-real-time system must produce a response within a deadline. This
+constraint may be relevant when the value of an aggregate decreases over time.
+
+An example of a near-real-time constraint is the expectation that a can be
```
chris-wood, line 57:
```suggestion
An example of a soft real-time constraint is the expectation that input data can be
```

design-document.md
chris-wood, line 61:
```suggestion
requests or utilize more efficient serialization to improve throughput.
```

design-document.md
```diff
@@ -30,6 +30,62 @@ practical.]]
...
+verified and aggregated in a period equal to data collection, given some
+computational budget. Meeting these deadlines may require efficient
+implementations of the input-validation protocol. Applications might batch
+requests and utilize binary serialization to improve throughput.
+
+Some applications may be constrained by the time that it takes to reach a
+privacy threshold defined by a minimum number of input shares. One possible
+solution is to increase the reporting period so more samples can be collected
+before the period is over, although this may need to be balanced with the need
+to produce a response within a deadline.
```
chris-wood, line 67:
```suggestion
to produce a response within a soft deadline.
```

design-document.md
```diff
@@ -30,6 +30,62 @@ practical.]]
...
+requires replaying data.
+
+Invalid shares are expected when operating an aggregator and do not necessarily
+constitute failures in the protocol. However, a large number of invalid
+or identical inputs may be concerning participants in the protocol.
+
+Certain classes of errors do not exist in the input-validation protocol
+considered in this document. For example, errors from packet loss when clients
+make requests directly to aggregators are not relevant when the leader proxies
+requests and controls the clock cycle for signaling the end of the aggregation
```
chris-wood, line 86:
```suggestion
requests and controls the schedule for signaling the end of the aggregation
```

chris-wood (commented):
Nice work! Beyond @cjpatton's comments, my only blocking point here is on the notion of a "deadline", and what that means in a distributed system. I've suggested relaxing this to be soft. (Other comments and suggestions around additional content can be addressed in a followup PR.)

design-document.md
acmiyaguchi: The example I had in mind was using something like Avro over JSON+base64 in cases where string parsing takes up more time than it should.

design-document.md
acmiyaguchi: I mean for this paragraph to distinguish between normal and unusual. In my day-to-day work, I've trawled through garbage coming through public ingestion endpoint (a lot of fuzzers and malformed data). I think some error is tolerable and normal. A large number of invalid inputs could be attributed to misconfigured application, and not necessarily an explicit denial of service. I did have DoS in mind too; I'll be more specific.

design-document.md
acmiyaguchi: I'm interested in learning more about the optimizations at some point after overcoming my admittedly weak understanding of arithmetic circuits.

stpeter:
> Nice work! Beyond @cjpatton's comments, my only blocking point here is on the notion of a "deadline", and what that means in a distributed system. I've suggested relaxing this to be soft.

In certain use cases (e.g., click measurement), most of the data might need to be passed along within a certain timeframe (say, 1 hour). That's not quite a deadline, but a constraint that we'll want to keep in mind.

design-document.md
acmiyaguchi: I think inputs/second is a reasonable definition of throughput, which is dependent on input parameters and the resources allocated for work. I guess "rate of work" would do here too. 

Using a maybe not so great physics analogy to explain in a different frame:
* verifying input is work
* throughput is power (rate of work)
* compute-time is energy (capacity for work)
* the amount of data that can be processed is constrained by throughput and available compute-time

> Should we have a section where we discuss this and other key performance indicators that we're trying to be mindful of?

It's probably a good idea. Maybe something to consider in the requirements or a separate performance section. I imagine an implementation would also want to monitor some of these things (e.g time to completion, error counts).  This section could also be expanded for discussion.

chris-wood:
> That's not quite a deadline, but a constraint that we'll want to keep in mind.

Yeah, that's fine. I was just trying to stress that any tightly synchronized version of a deadline ("all nodes stop collecting the batch at time X") would be quite hard to implement.

design-document.md
chris-wood: > It's probably a good idea. Maybe something to consider in the requirements or a separate performance section. I imagine an implementation would also want to monitor some of these things (e.g time to completion, error counts). This section could also be expanded for discussion.

Both seem fine. Maybe just mark an OPEN ISSUE here for us to resolve later?

chris-wood (approved):

PR #13, opened 2021-03-04
Update field parameters

cjpatton:
@ekr and others have discussed using an 80 bit field in some application. This PR adds one. It also removes the smallest field (unlikely to ever be used) and renumbers the table.

### DIFF
--- a/design-document.md
+++ b/design-document.md
@@ -311,48 +311,49 @@ criteria:
 1. **Field size.** How big the field needs to be depends on the type of data
    being aggregated and how many users there are. The field size also impacts
    the security level: the longer the validity circuit, the larger the field
-   needs to be in order to affectively detect malicious clients.
+   needs to be in order to effectively detect malicious clients. Typically the
+   soundness error (i.e., the probability of an invalid input being deemed valid
+   by the aggregators) will be 2n/(p-n), where n is the size of the input and p
+   is the prime modulus.
 1. **Fast polynomial operations.** In order to make Prio practical, it's
    important that implementations employ FFT to speed up polynomial operations.
-   In particular, the prime modulus p should be chosen so that (p-1) = 2^n * s
-   for large n and odd s. Then g^s is a principle, 2^n-th root of unity (i.e.,
-   g^(s\*2^n) = 1), where g is the generator of the multiplicative subgroup.
-   This fact allows us to quickly evaluate and interpolate polynomials at 2^d-th
-   roots of unity for 1 <= d <= n.
-1. **Highly composite subgroup.** Suppose that (p-1) = 2^n * s. It's best if s
+   In particular, the prime modulus p should be chosen so that (p-1) = 2^b * s
+   for large b and odd s. Then g^s is a principle, 2^b-th root of unity (i.e.,
+   g^(s\*2^b) = 1), where g is the generator of the multiplicative subgroup.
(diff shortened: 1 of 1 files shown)


ekr:
@henrycg would love to get your input here.

ekr:
As far as I can tell, this accurately subsets the previous fields. I have no opinion on the substance.

ekr:
Oh, I see the 72 bit field. No idea if that one is right :)

henrycg:
Thanks for looping me in. Here are a couple of comments:

- It might be a good idea to explain somewhere that soundness error (probability that the servers accept a false proof) is `2n/(p-n)` when carrying out all arithmetic modulo `p` and with input size `n`. As written, the "bits" column in the parameters table might give the impression that using modulus 1 gives ~31 bits of security, when it could actually give only ~15 bits of security when used with inputs of size 2^16.

- Nit: In this table and in the preceding discussion on [Finite field arithmetic](https://github.com/abetterinternet/prio-documents/blob/main/design-document.md#finite-field-arithmetic), it might make sense to use a symbol other than `n`, since `n` is used to mean something else in the earlier section on [Cryptographic Components](https://github.com/abetterinternet/prio-documents/blob/main/design-document.md#cryptographic-components).

cjpatton:
> Good call! Changed `n` to `s`.

Oof, this is supposed to be `b` instead of `s`. Edited above.

tlepoint:
Out of curiosity, why do you need to hardcode prime numbers rather than making the whole system more flexible and accept any prime that would respect the constraints?

cjpatton:
> Out of curiosity, why do you need to hardcode prime numbers rather than making the whole system more flexible and accept any prime that would respect the constraints?

It seems sensible to settle on a handful of primes and develop optimized implementations of the arithmetic for each field. Initially, however, we plan to have generic arithmetic: see https://github.com/abetterinternet/libprio-rs/issues/10.

design-document.md
```diff
@@ -311,48 +311,49 @@ criteria:
 1. **Field size.** How big the field needs to be depends on the type of data
    being aggregated and how many users there are. The field size also impacts
    the security level: the longer the validity circuit, the larger the field
-   needs to be in order to affectively detect malicious clients.
+   needs to be in order to affectively detect malicious clients. Typically the
```
tgeoghegan, line 314:
```suggestion
   needs to be in order to effectively detect malicious clients. Typically the
```

tgeoghegan (approved):
I'm well out of my depth on the math here but one typo that this PR didn't even introduce aside, this LGTM.

design-document.md
```diff
@@ -311,48 +311,49 @@ criteria:
 1. **Field size.** How big the field needs to be depends on the type of data
    being aggregated and how many users there are. The field size also impacts
    the security level: the longer the validity circuit, the larger the field
-   needs to be in order to affectively detect malicious clients.
+   needs to be in order to effectively detect malicious clients. Typically the
```
chris-wood, line 314: Why "typically"?

chris-wood (approved):

design-document.md
cjpatton: It's going to depend on the circuit.

chris-wood: Will it be higher? Lower?

cjpatton:
Rebased and squahsed.

PR #16, opened 2021-03-09
design document: threat model

tgeoghegan:
Initial threat model, enumerating the system's actors and their
capabilities, as well as current or hypothetical mitigations.

### DIFF
--- a/design-document.md
+++ b/design-document.md
@@ -9,8 +9,8 @@
 1. Collector: The endpoint that receives the final aggregate. It also specifies
    the parameters of the protocol.
 1. False input: Input which is valid under the input validation protocol but is
-   not truthful. e.g., if the data being gathered is whether or not users have
-   clicked on a particular button, a client could report clicks when none
+   not truthful. For example, if the data being gathered is whether or not users
+   have clicked on a particular button, a client could report clicks when none
    occurred.
 1. Input: The original data emitted by a client, before any encryption or secret
    sharing scheme is applied.
@@ -106,10 +106,11 @@ this can be done by TLS or some other form of application-layer authentication.
 In the presence of this adversary, Prio provides two important properties for computing 
 an aggergation function F:
 
-1. Privacy. The adversary learns only the output of F computed over all client inputs, 
-   and nothing else.
-1. Robustness. The adversary can influence the output of F only by reporting false 
-   (untruthful) input. The output cannot be influenced in any other way.
+1. Privacy. The aggregators and collector learn only the output of F computed
+   over all client inputs, and nothing else.
+1. Robustness. As long as the aggregators execute the input-validation protocol
+   correctly, a malicious client can skew the output of F only by reporting
+   false (untruthful) input. The output cannot be influenced in any other way.
 
 There are several additional constraints that a Prio deployment must satisfy in order
 to achieve these goals:
@@ -215,7 +216,7 @@ mitigations available to aggregators also apply to the leader.
    by dropping messages.
      * This capability is no stronger than any aggregator's ability to refuse to
        emit output shares.
-1. Shinking the anonymity set. The leader instructs aggregators to construct
+1. Shrinking the anonymity set. The leader instructs aggregators to construct
    output parts and so could request aggregations over few inputs.
 
 ##### Mitigations
--- a/design-document.md
+++ b/design-document.md
@@ -131,7 +131,7 @@ confer some capability that enables further attack on the system), the
 capabilities that a malicious or compromised actor has, and potential
 mitigations for attacks enabled by those capabilities.
 
-#### Client
+#### Client/user
 
 ##### Assets
 
@@ -142,14 +142,19 @@ mitigations for attacks enabled by those capabilities.
 
 ##### Capabilities
 
-1. Individual clients can trivially and undetectably defeat the privacy property
-   by leaking data outside of the Prio system. Accordingly, such attacks are
-   outside of the threat model.
+1. Individual users can reveal their own data and compromise their own privacy.
+   This does not affect the privacy of others in the system.
+1. Clients (that is, software which might be used by many users of the system)
+   can defeat privacy by leaking data outside of the Prio system. Other
+   participants have no insight into what clients do besides uploading data
+   shares and accordingly, such attacks are outside of the current threat model.
 1. Clients may affect the quality of aggregations by reporting false data.
 
 [[OPEN ISSUE: Is there anything that could be done to either mitigate or detect
 a client compromising privacy? Do users have any choice but to trust that
-clients have faithfully implemented the protocol?]]
+clients have faithfully implemented the protocol? If we introduce the notion of
+a trusted OS or platform actor, distinct from the client, that could attest to
+the client's behavior, can we improve this?]]
 
 ##### Mitigations
 
@@ -164,9 +169,6 @@ clients have faithfully implemented the protocol?]]
    whitebox cryptography to reproducible builds and remote attestation of
    software measurements from a trusted computing base.
 
-[[OPEN ISSUE: The problem of clients sending bogus data exists in any metrics
-system, orthogonally to Prio, so perhaps it is out of scope.]]
-
 #### Aggregator
 
 ##### Assets
@@ -178,35 +180,44 @@ system, orthogonally to Prio, so perhaps it is out of scope.]]
 
 ##### Capabilities
 
-1. Aggregators may trivially and undetectably defeat the robustness of the
-   system by emitting bogus aggregation shares. Accordingly, such attacks are
-   outside of the threat model.
-1. If client identities are used, aggregators may weaken the anonymity of the
-   system by revealing that a particular client contributed data to the system.
+1. Aggregators may defeat the robustness of the system by emitting bogus
+   aggregation shares.
+1. If clients reveal identifying information to aggregators (such as a trusted
+   identity during client authentication), aggregators can learn which clients
+   are contributing data.
+     1. Aggregators may weaken anonymity by revealing that a particular client
+        contributed data to the system.
+     1. Aggregators may choose to selectively omit data from certain clients.
+          * For example, omitting submissions from a particular geographic
+            region to falsely suggest that a particular localization is not
+            being used.
 1. Individual aggregators may compromise availability of the system by refusing
    to emit aggregation parts.
 
 ##### Mitigations
 
-1. Infrastructure diversity. Prio deployments should ensure that aggregators'
-   security do not have a common point of failure. For instance, if all
-   participating aggregators stored unencrypted data shares on a single cloud
-   object storage service, then that cloud vendor would be able to decrypt and
-   reassemble all the data shares and defeat privacy.
-1. Implementation diversity. Similarly to infrastructure diversity, diversity in
-   implementations would help ensure that a single vulnerability in an
-   aggregator or one of its dependencies would not compromise _all_ the
-   participating aggregators.
+1. The linear secret sharing scheme employed by the client ensures that privacy
+   is preserved unless every participating aggregator reveals their unencrypted
+   data shares.
+1. Infrastructure diversity. Prio deployments should ensure that aggregators do
+   not have common dependencies that would enable a single vendor to reassemble
+   data.
+     * For example, if all participating aggregators stored unencrypted data
+       shares on the same cloud object storage service, then that cloud vendor
+       would be able to reassemble all the data shares and defeat privacy.
 1. Introducing a further actor to the system whose sole responsibility would be
    to verify the identity of participating clients before forwarding encrypted
    data shares to aggregators would mitigate the aggregator's ability to weaken
    client anonymity.
-1. Running the protocol over multiple subsets of the available aggregators in
-   parallel would prevent an individual aggregator from compromising
-   availability.
-
-[[OPEN ISSUE: can anything mitigate or detect the aggregator's capability to
-compromise robustness?]]
+1. Running the protocol over multiple subsets of the available aggregators
+   chosen so that no aggregator appears in all subsets and accepting any result
+   would prevent an individual aggregator from compromising availability.
+1. Running the protocol over multiple subsets of the available aggregators
+   chosen so that no aggregator appears in all subsets and requiring that the
+   results agree would prevent an individual aggregator from compromising
+   robustness
+     * This could also allow identifying of defective aggregators, by finding
+       the aggregator that appears in every subset whose results were wrong.
 
 #### Leader
 
@@ -240,7 +251,7 @@ mitigations available to aggregators also apply to the leader.
 
 ##### Mitigations
 
-1. Aggregators should refuse shared parameters that are obviously insecure
+1. Aggregators should refuse shared parameters that are trivially insecure
    (i.e., aggregation threshold of 1 contribution).
 
 #### Aggregator collusion
@@ -248,6 +259,42 @@ mitigations available to aggregators also apply to the leader.
 If all aggregators collude (e.g. by promiscuously sharing unencrypted data
 shares), then none of the properties of the system hold. Accordingly, such
 scenarios are outside of the threat model.
+
+#### Attacker on the network
+
+We assume the existence of attackers on the network links between participants.
+
+##### Capabilities
+
+1. Observation of network traffic. Attackers may observe messages exchanged
+   between participants at the IP layer.
+     1. The time of transmission of data shares by clients could reveal
+        information about user activity.
+          * For example, if a user opts into a new feature, and the client
+            immediately reports this to aggregators, then just by observing
+            network traffic, the attacker can infer what the user did.
+     1. Observation of message size could allow the attacker to learn how much
+        data is being submitted by a client.
+          * For example, if the attacker observes an encrypted message of some
+            size, they can infer the size of the plaintext, plus or minus the
+            cipher block size. From this they may be able to infer which
+            aggregations the user has opted into or out of.
+1. Tampering with network traffic. Attackers may drop messages or inject new
+   messages into communications between participants.
+
+##### Mitigations
+
+1. All messages exchanged between participants in the system should be
+   encrypted.
+1. All messages exchanged between aggregators, the collector and the leader
+   should be mutually authenticated so that network attackers cannot impersonate
+   participants.
+1. Messages sent from clients to aggregators can be authenticated using client
+   identities to prevent attackers from injecting false data into aggregations.
+1. Clients should be required to submit data at regular intervals so that the
+   timing of individual messages does not reveal anything.
+1. A fixed-length encoding should be used for data shares. Additionally, clients
+   should submit dummy data even for aggregations the user has not opted into.
 
 [[OPEN ISSUE: The threat model for Prio --- as it's described in the original
 paper and [BBC+19] --- considers **either** a malicious client (attacking
--- a/design-document.md
+++ b/design-document.md
@@ -8,27 +8,26 @@
    web browser.
 1. Collector: The endpoint that receives the final aggregate. It also specifies
    the parameters of the protocol.
-1. Input-validation protocol: The protocol executed by the client and aggregators
+1. False input: Input which is valid under the input validation protocol but is
+   not truthful. e.g., if the data being gathered is whether or not users have
+   clicked on a particular button, a client could report clicks when none
+   occurred.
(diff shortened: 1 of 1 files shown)


chris-wood (changes requested):
This is a great start! I think my only blocking comment is around the network adversary.

design-document.md
```diff
@@ -106,6 +123,132 @@ to achieve these goals:
...
+
+1. Unshared data. Clients are the only actor that can ever see the original
+   data.
+1. Unencrypted data shares.
+1. Client identity (optionally).
+
+##### Capabilities
+
+1. Individual clients can trivially and undetectably defeat the privacy property
+   by leaking data outside of the Prio system. Accordingly, such attacks are
```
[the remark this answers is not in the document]
tgeoghegan: I'm making a distinction between the user and the client here (I think my use of "individual" is confusing). I'm imagining a scenario where an application vendor might adopt Prio to elicit trust from users, but ship client software that covertly leaks user data over some non-Prio channel. 

I'm not sure we can do anything about it, unless the model is extended to include a trusted OS that a client runs in, which could attest to various properties of the client (e.g., measurements of reproducible builds, or attest to a log of the client's network activity so that external observers can verify that there are no unexpected connections to `secret-cleartext-metrics.appvendor.com`).

I'm going to reword this to explicitly separate the case of an individual user/client disclosing their own data (which as you say does not compromise anyone else's privacy) and the case of a client used by many users leaking data (which is out of scope in the current model).

design-document.md
```diff
@@ -106,6 +123,132 @@ to achieve these goals:
...
+   to emit aggregation parts.
+
+##### Mitigations
+
+1. Infrastructure diversity. Prio deployments should ensure that aggregators'
+   security do not have a common point of failure. For instance, if all
+   participating aggregators stored unencrypted data shares on a single cloud
+   object storage service, then that cloud vendor would be able to decrypt and
+   reassemble all the data shares and defeat privacy.
+1. Implementation diversity. Similarly to infrastructure diversity, diversity in
```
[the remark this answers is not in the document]
tgeoghegan: I agree that this one isn't specific to Prio but rather would apply to any distributed computing system, so I'll strike it.

design-document.md
```diff
@@ -106,6 +123,132 @@ to achieve these goals:
...
+1. Individual aggregators may compromise availability of the system by refusing
+   to emit aggregation parts.
+
+##### Mitigations
+
+1. Infrastructure diversity. Prio deployments should ensure that aggregators'
+   security do not have a common point of failure. For instance, if all
+   participating aggregators stored unencrypted data shares on a single cloud
+   object storage service, then that cloud vendor would be able to decrypt and
+   reassemble all the data shares and defeat privacy.
```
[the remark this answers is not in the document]
tgeoghegan: The point of Prio, as I see it, is that you need to subvert all of the aggregators in order to defeat privacy. But if all the aggregators are using the same cloud storage vendor, then the attacker no longer needs to subvert _n_ different aggregators -- they just need to subvert a single cloud vendor. So users are back to hoping that a single entity will not have any breaches, will not get any national security letters and will consistently comply with its own privacy policy, which I think undermines Prio.

Reading this back, I don't like the sentence "Prio deployments should ensure that aggregators' security do not have a common point of failure.". I'll try to find a better way to express this.

design-document.md
```diff
@@ -106,6 +123,132 @@ to achieve these goals:
...
+   minimize the impact of false data.
+1. Many deployments could allow anyone to contribute data shares to aggregators,
+   but others could require that clients establish a trusted identity before
+   they may contribute data. A variety of techniques providing different degrees
+   of assurance could be employed to establish trust in clients, ranging from
+   whitebox cryptography to reproducible builds and remote attestation of
+   software measurements from a trusted computing base.
+
+[[OPEN ISSUE: The problem of clients sending bogus data exists in any metrics
+system, orthogonally to Prio, so perhaps it is out of scope.]]
```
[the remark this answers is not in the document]
tgeoghegan: I'm not so sure, because at least one mitigation could be applied within the Prio protocol: requiring client identities. If clients had to authenticate to aggregators (or to a trusted authenticator who would then relay anonymized data shares to aggregators), then the system could detect clients sending excessive contributions and reject them.

design-document.md
```diff
@@ -106,6 +123,132 @@ to achieve these goals:
...
+1. Data share decryption keys.
+1. Aggregation parts.
+1. Aggregator identity.
+
+##### Capabilities
+
+1. Aggregators may trivially and undetectably defeat the robustness of the
+   system by emitting bogus aggregation shares. Accordingly, such attacks are
+   outside of the threat model.
+1. If client identities are used, aggregators may weaken the anonymity of the
```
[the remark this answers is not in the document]
tgeoghegan: I would phrase it as "if a client reveals identifying information to aggregators, such as via client auth, then aggregators have this capability".

design-document.md
```diff
@@ -106,6 +123,132 @@ to achieve these goals:
    function output itself. {{questions-and-params}} discusses the leakage profiles of 
    various aggregation functions in more detail.
 
+### Threat model
```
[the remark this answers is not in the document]
tgeoghegan: I think you're right that there should be a section for an attacker on the network. However I find it hard to reason about a threat like the one in your example because we haven't yet sketched out how aggregator discovery works. But certainly we can already call out the necessity of confidential and mutually authenticated links between participating servers.

design-document.md
```diff
@@ -106,6 +123,167 @@ to achieve these goals:
...
+       would be able to reassemble all the data shares and defeat privacy.
+1. Running the protocol over multiple subsets of the available aggregators
+   chosen so that no aggregator appears in all subsets and accepting any result
+   would prevent an individual aggregator from compromising availability.
+1. Running the protocol over multiple subsets of the available aggregators
+   chosen so that no aggregator appears in all subsets and requiring that the
+   results agree would prevent an individual aggregator from compromising
+   robustness
+     * This could also allow identifying of defective aggregators, by finding
+       the aggregator that appears in every subset whose results were wrong.
```
cjpatton, line 210: ^ Does this defeat the bogus aggregation share attack? I'm not sure anything does. There may be other, more subtle attacks it detects, but given how easy it is for an aggregator to submit a bogus aggregation share, I don't think it matters much. (See comment in "Future work and possible extensions".)

However, detecting bad aggregators is a really good idea. We'll need to flesh this out later.

design-document.md
```diff
@@ -106,6 +123,167 @@ to achieve these goals:
...
+
+#### Leader
+
+The leader is also an aggregator, and so all the assets, capabilities and
+mitigations available to aggregators also apply to the leader.
+
+##### Capabilities
+
+1. Data validity proof verification. The leader can forge proofs and collude
+   with a malicious client to trick aggregators into aggregating invalid data.
```
cjpatton, line 220: This capability is only marginally stronger than the bogus aggregation share attack, IMO.

design-document.md
```diff
@@ -106,6 +123,167 @@ to achieve these goals:
...
+     * This could also allow identifying of defective aggregators, by finding
+       the aggregator that appears in every subset whose results were wrong.
+
+#### Leader
+
+The leader is also an aggregator, and so all the assets, capabilities and
+mitigations available to aggregators also apply to the leader.
+
+##### Capabilities
+
```
cjpatton, line 218: Another capability to call out: The leader can try to shrink the anonymity set by requesting aggregation shares early (cf. https://github.com/abetterinternet/prio-documents/issues/15).

design-document.md
```diff
@@ -106,6 +123,167 @@ to achieve these goals:
...
+1. Data validity proof verification. The leader can forge proofs and collude
+   with a malicious client to trick aggregators into aggregating invalid data.
+1. Relaying messages between aggregators. The leader can compromise availability
+   by dropping messages.
+
+##### Mitigations
+
+1. Rather than having all aggregators trust the leader to validate proofs, each
+   aggregator could exchange proof shares with every other aggregator, enabling
+   each aggregator to independently verify the proofs.
```
cjpatton, line 228: I think what's meant by this is that the aggregators could exchange verification shares so that each aggregator can run the decision algorithm locally. Is that right?

I agree that this would help mitigate the leader's additional capabilities. However, doing so would require a secure channel between each pair of aggregators. I think this cost is too high.

design-document.md
```diff
@@ -106,6 +123,167 @@ to achieve these goals:
...
+1. Running the protocol over multiple subsets of the available aggregators and
+   choosing different leaders for each run would prevent any individual leader
+   from compromising availability.
+
+#### Collector
+
+##### Capabilities
+
+1. Advertising shared configuration parameters (e.g., minimum thresholds for
+   aggregations, polynomial identity test values, arithmetic circuits).
```
cjpatton, line 238: In [BBC+19] (and hence libprio-rs) the concept of "polynomial identity test" is generalized to distributed evaluation of a PCP over joint randomness. Hence, I would use the term "joint randomness" instead of "polynomial identity test values".

To your point: We haven't yet decided how the joint randomness is chosen. It could very well be the collector, but this is TBD. (See "Consensus protocol" below.) Also, I would add "designation the leader role" to this list.
But

design-document.md
```diff
@@ -106,6 +123,167 @@ to achieve these goals:
...
+1. Collectors may trivially defeat availability by discarding aggregation
+   parts submitted by aggregators.
```
cjpatton, line 240:
```suggestion
   shares submitted by aggregators.
```

design-document.md
```diff
@@ -106,6 +123,167 @@ to achieve these goals:
...
+
+##### Mitigations
+
+1. Aggregators should refuse shared parameters that are trivially insecure
+   (i.e., aggregation threshold of 1 contribution).
```
cjpatton, line 245: The parameters need to be authenticated and enforced by each aggregator, but I think it's up to the collector to choose how large the anonymity set should be.

design-document.md
```diff
@@ -106,6 +123,167 @@ to achieve these goals:
...
+        information about user activity.
+          * For example, if a user opts into a new feature, and the client
+            immediately reports this to aggregators, then just by observing
+            network traffic, the attacker can infer what the user did.
+     1. Observation of message size could allow the attacker to learn how much
+        data is being submitted by a client.
+          * For example, if the attacker observes an encrypted message of some
+            size, they can infer the size of the plaintext, plus or minus the
+            cipher block size. From this they may be able to infer which
+            aggregations the user has opted into or out of.
```
cjpatton, line 271: The size of the input is a function of the data type, not the data itself. In particular, all clients will submit the same amount of data for a given aggregation job.

design-document.md
```diff
@@ -106,6 +123,167 @@ to achieve these goals:
...
+
+1. All messages exchanged between participants in the system should be
+   encrypted.
+1. All messages exchanged between aggregators, the collector and the leader
+   should be mutually authenticated so that network attackers cannot impersonate
+   participants.
+1. Clients should be required to submit data at regular intervals so that the
+   timing of individual messages does not reveal anything.
+1. A fixed-length encoding should be used for data shares. Additionally, clients
+   should submit dummy data even for aggregations the user has not opted into.
```
cjpatton, line 285: Aggregation isn't well-defined if input lengths vary, so I think the first part of this mitigation is irrelevant.

design-document.md
```diff
@@ -115,6 +293,50 @@ collude and break soundness. Is this a contingency we need to address? There are
...
+
+In this section we discuss attacks that are not considered in the above threat
+model, and suggest mitigations that could be incorporated into implementations
+of this protocol or future revisions of this specfication.
+
+#### Client authentication
+
+Attackers can impersonate Prio clients and submit large amounts of valid but
+false data in order to spoil aggregations. Deployments could require that
+clients present a trusted identity before they may contribute data. For example,
```
cjpatton, line 306:
```suggestion
false data in order to spoil aggregations. Deployments could require client
authentication. For example,
```

design-document.md
```diff
@@ -115,6 +293,50 @@ collude and break soundness. Is this a contingency we need to address? There are
...
+by requiring data submissions to be signed with a key trusted by aggregators. A
+variety of techniques providing different degrees of assurance could be employed
+to establish trust in clients, ranging from whitebox cryptography to public key
+infrastructures. However some deployments may opt to accept the risk of false
+data to avoid having to figure out how to distribute trusted identities to
+clients.
```
cjpatton, line 312: Another authentication mechanism: The leader might just require the user to login beforehand.

I'm curious how whitebox crypto might be used here?

design-document.md
```diff
@@ -115,6 +293,50 @@ collude and break soundness. Is this a contingency we need to address? There are
...
+
+#### Client attestation
+
+In the current threat model, servers participating in the protocol have no
+insight into the activities of clients except that they have uploaded data into
+a Prio aggregation, meaning that clients could covertly leak a user's data into
+some other channel which compromises privacy and anonymity. If we introduce the
```
cjpatton, line 319:
```suggestion
some other channel which compromises privacy. If we introduce the
```

As I mentioned above, privacy implies anonymity as the terms are defined above. (You may disagree?)

design-document.md
```diff
@@ -115,6 +293,50 @@ collude and break soundness. Is this a contingency we need to address? There are
 techniques in [BBC+19] that account for this; we need to figure out if they're
 practical.]]
 
+### Future work and possible extensions
```
cjpatton, line 296: An attacker capability explicitly considered in [BBC+19], but not in [GB17], is that a coalition of verifiers may attempt to convince the honest verifiers that an invalid data is valid. This doesn't matter much in our setting, since the bogus aggregation share attack descrbed above is equally effective and much easier to pull off. It would matter, however, if the aggregator were divided into two entities: one which validates input shares (the verifier); and another that accumulates them and emits the aggregation share. I think it's worth noting this somewhere, probably here.

cjpatton (commented):
This is super useful, thank you for getting this started. My only high-level comment is that this analysis doesn't completely settle the question of whether we care if a malicious server colludes with a malicious client in order to corrupt the final output. As I mention below, this threat model is only briefly considered in [BBC+19] (and not at all in the original Prio paper, [GB17]); and because a malicious aggregator could always submit a bogus aggregate share, this model doesn't seem (to me anyway) relevant for our deployment.

I think this PR should explicitly remove this threat model from scope. We should assume that aggregators do their very best to validate and accumulate data shares correctly. That said, we should make it a goal to detect misbehaving aggregators. However, the target there needs to be detecting bogus shares.

design-document.md
```diff
@@ -13,6 +13,22 @@
    aggregators.
 1. Leader: A distinguished aggregator that coordinates input validation and data
    aggregation.
+1. Data: The original data emitted by a client, before any encryption or secret
+   sharing scheme is applied.
+1. Data share: one of the _n_ shares output by feeding a datum into a linear
+   secret sharing scheme. Each share is to be transmitted to one of the _n_
+   participating aggregators.
```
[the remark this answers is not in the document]
tgeoghegan: That's a good idea. I keep finding myself writing "original data" or "unshared data" to disambiguate, so I'll adopt your "input" instead.

design-document.md
```diff
@@ -13,6 +13,22 @@
...
+1. Invalid data: Data which cannot be represented in the affine-aggregatable
+   encoding chosen by participants. e.g., a UTF-8 string if the chosen encoding
+   is a vector of 16 bits.
```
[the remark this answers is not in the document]
tgeoghegan: My phrasing here was based on a misreading of the 2017 paper. I'll rewrite this to make it plain that invalid data is simply data for which the validity predicate does not hold.

design-document.md
```diff
@@ -95,6 +111,7 @@ an aggergation function F:
    and nothing else. 
 1. Robustness. The adversary can influence the output of F only by reporting false 
    (untruthful) data. The output cannot be influenced in any other way.
+1. Anonymity. The adversary cannot learn which client submitted which data value.
```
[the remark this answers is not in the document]
tgeoghegan: The [2017 paper](https://crypto.stanford.edu/prio/paper.pdf) calls out privacy and anonymity as distinct security properties (section 2). Anonymity is the adversary not knowing which client submitted which input, privacy is the adversary not knowing what the inputs are. The reason I added anonymity to this list is because I think that in the setting where aggregators talk directly to clients, privacy is preserved but client anonymity is weakened since the aggregator can identify clients e.g. by IP. Further we can sketch out a mitigation -- an anonymizing proxy -- so I think it's worth discussing the anonymity property.

design-document.md
tgeoghegan: Plus, HCG clarified for us yesterday that any aggregator, leader or not, can collude with malicious clients to achieve the same thing. I'll rewrite this.

design-document.md
```diff
@@ -13,6 +13,22 @@
...
+   participating aggregators.
+1. Invalid data: Data which cannot be represented in the affine-aggregatable
+   encoding chosen by participants. e.g., a UTF-8 string if the chosen encoding
+   is a vector of 16 bits.
+1. False data: Data which is valid under the chosen affine-aggregatable encoding
+   but is not truthful. e.g., if the data being gathered is whether or not users
+   have clicked on a particular button, a client could report clicks when none
+   occurred.
+1. Aggregation: a statistical aggregation over a body of data which is of
+   interest to a collector.
```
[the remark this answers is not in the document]
tgeoghegan: Good point! I chose "output" for symmetry with your other good suggestion of "input".

design-document.md
```diff
@@ -106,6 +123,167 @@ to achieve these goals:
...
+   data shares.
+1. Infrastructure diversity. Prio deployments should ensure that aggregators do
+   not have common dependencies that would enable a single vendor to reassemble
+   data.
+     * For example, if all participating aggregators stored unencrypted data
+       shares on the same cloud object storage service, then that cloud vendor
+       would be able to reassemble all the data shares and defeat privacy.
+1. Running the protocol over multiple subsets of the available aggregators
+   chosen so that no aggregator appears in all subsets and accepting any result
+   would prevent an individual aggregator from compromising availability.
```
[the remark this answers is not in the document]
tgeoghegan: I think this isn't something we'd tackle in an initial deployment of Prio anyway, so I wrote up #22 on this topic, and will move this mitigation into future work and possible extensions.

design-document.md
tgeoghegan: The notion with whitebox is that you could use it to distribute a shared secret to all "legitimate" instances of your client, and so having a message HMACed with a whiteboxed key gives you some kind of assurance on the provenance of the message. But thinking about it further, I've never been a big fan of whitebox, and mentioning it here isn't particularly helpful to explaining Prio, so I'll just strike it.

design-document.md
tgeoghegan: I think the leader in our design is the verifier you describe, no? What HCG argued to us on 3/24 was that a verifier colluding with a malicious client doesn't have any capability that isn't already available to any aggregator colluding with a malicious client. I think what you're sketching out is a further variation where the leader is _only_ a verifier and not an aggregator?

design-document.md
tgeoghegan: For any individual aggregation, yes, but imagine if a client supports dozens or hundreds of individual aggregations, and users can opt out of any of them. If the client only transmits input shares for the aggregations the client has opted into, then the volume of traffic observable by the adversary would change. Hence I argue that a client should always transmit the same amount of data, regardless of which aggregations a user has consented to.

design-document.md
tgeoghegan: I agree, but the intent here is that aggregators should implement basic checks independent of the configuration of any aggregation to prevent abuse.

design-document.md
tgeoghegan: HCG convinced us (or at least me!) that it's OK to have a designated proof verifier because the verifier is no more powerful than any aggregator that colludes with a malicious client, so I'm going to remove this item.

design-document.md
cjpatton: Works for me.

design-document.md
cjpatton: > I think the leader in our design is the verifier you describe, no? 

No. Here I'm imagining if any aggregator were split into two services: one which validates input shares and another that accumulates them.

In any case, the point here is moot I think. We will assume that the aggregators execute the input-validation correctly (this is setting I of [BBC+19].)

cjpatton (approved):
LGTM modulo editorial things.

chris-wood (approved):

PR #17, opened 2021-03-09
Add batch size constraint and aggregator count constraint to the design document

aaomidi:
Introduces two new sections under `System constraints`

### DIFF
--- a/design-document.md
+++ b/design-document.md
@@ -182,6 +182,29 @@ considered in this document. For example, packet loss errors when clients
 make requests directly to aggregators are not relevant when the leader proxies
 requests and controls the schedule for signaling aggregation rounds.
 
+### Batch size constraints
+
+In this batch processing system, the __n__ aggregators will share processed
+batches between each other. The size of the batches have a direct impact on the
+efficiency and throughput of the aggregator. 
+
+The aggregators may introduce a maximum content size measured in bytes that the
+clients can send to the leader. This requires the client to ensure batches do
+not exceed the maximum size of any aggregator. The content size is applied to
+the entire batch, rather than specific data points.
+
+
+### Aggregator count constraints
+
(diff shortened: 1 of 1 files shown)


design-document.md
```diff
@@ -182,6 +182,29 @@ considered in this document. For example, packet loss errors when clients
...
+the entire batch, rather than specific data points.
+
+
+### Aggregator count constraints
+
+The user of the Prio privacy preserving aggregation system may opt into using
+more than the minimum number of aggregators required for preserving privacy of
+clients. The likelihood of failures due to external factors increases
+exponentially with each additional aggregator. An example of such external
+factors are packet loss, infrastructure failure, and latency spikes.
```
acmiyaguchi, line 203: Might be useful to include an approach of how to resolve one of the issues that arises in this context (e.g., ensuring redundancy of some kind).

design-document.md
```diff
@@ -182,6 +182,29 @@ considered in this document. For example, packet loss errors when clients
...
+
+The aggregators may introduce a maximum content size measured in bytes that the
+clients can send to the leader. This requires the client to ensure batches do
+not exceed the maximum size of any aggregator. The content size is applied to
+the entire batch, rather than specific data points.
+
+
+### Aggregator count constraints
+
+The user of the Prio privacy preserving aggregation system may opt into using
```
cjpatton, line 199:
```suggestion
The collector may opt into using
```
(Not sure if this is what you mean ... see "Terminology" above.)

cjpatton (commented):
I don't think the aggregators should be able to enforce minimum batch sizes. IMO this should be a parameter chosen by the collector.

cjpatton:
I believe the intention of this change is now subsumed by #59. Can we close this?

PR #23, opened 2021-03-26
EKR pass 1.

ekr:
- Try to keep things concrete. We always send things to the leader
  so just say so.

- Move the "System Constraints" section to the end and rename it.
  This will remove forward refs when the rest of the doc is
  written.

Other editorial.

### DIFF
--- a/design-document.md
+++ b/design-document.md
@@ -18,8 +18,9 @@
 
 Prio is a system and protocol for privately computing aggregation functions over private 
 input. An aggregation function F is one that computes an output y = F(x[1],x[2],...) for inputs
-x[i]. In general, Prio supports any aggregation function whose inputs can be encoded in a 
-particular way. However, not all aggregation functions admit an efficient encoding, rendering
+x[i]. Internally, Prio computes the aggregation function "sum" and can compute
+other aggregation functions by specially encoding the inputs.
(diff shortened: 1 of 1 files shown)


design-document.md
```diff
@@ -210,8 +153,12 @@ K^n.
 
 Prio combines standard [linear secret
 sharing](https://en.wikipedia.org/wiki/Secret_sharing#t_=_n) with a new type of
-probabilistically checkable proof (PCP) system, called a fully linear PCP. The
-input-input validation protocol can be described in terms of three main
+probabilistically checkable proof (PCP) system, called a fully linear PCP. 
+The aggregrators jointly validate the proof of correctness of the input.
```
tgeoghegan, line 157: The most recent consensus is that aggregators will decrypt proof+input shares received from clients, forward the proof shares to the leader (while keeping input shares secret), and the leader is responsible for validating the proof and instructing aggregators to include an input in an aggregation.

A further nit: the proof proves that the input is _valid_ (#16 adds a definition of invalid input to the glossary). Prio can't say anything useful about the correctness or truthfulness of the input.

chris-wood (approved):
🚢

cjpatton (approved):

PR #26, opened 2021-03-31
Try to make the on-ramp to the crypto easier and clean up a bit of no…

ekr:
…tation to make it more friendly to comsec types


design-document.md
```diff
@@ -477,12 +477,17 @@ uploading, open issues, and system parameters?]]
...
-K^n.
+As described in {{overview}} each client wishes to report a set of
+inputs. For simple measurements, such as having each client report a
+count and computing the total, each input is just mapped onto a single
+integer. For more complicated measurements, it may be necessary to map
+a given input onto multiple integers. In any case, the client takes
+all of its inputs and maps them onto a vector of integers each 
+of which is a member of a finite field K. We denote this vector as x.
+
+In order to share x between s servers, we break it up into s shares
```
cjpatton, line 488: s/break/split/

design-document.md
```diff
@@ -506,10 +511,12 @@ algorithms:
...
-1. Input x is an element of K^n for some integer n.
-1. Proof pf is an element of K^p(n) for some function p.
-1. The joint randomness r is an element of K^u(n) for some function u.
-1. Each verification share {vf:i} is an element of K^v(n) for some function v.
+1. Input x is vector of length n elements of K.
+1. Proof pf is a vector of length p(n) of elements of K.
+1. The joint randomness r is a vector of length u(n) of elements of K.
+1. Each verification share {vf:i} is a vector of length v(n) of elements of K.
+
+Where p(n), u(n), and v(n) are functions that we specify later.
```
cjpatton, line 519: s/Where/Above,/

design-document.md
```diff
@@ -477,12 +477,17 @@ uploading, open issues, and system parameters?]]
...
-{x:s}, where {x:i} is the share held by the i-th party. We write {x} as
-shorthand for the sequence {x:1}, ..., {x:s}. Each {x:i} is also an element of
-K^n.
+As described in {{overview}} each client wishes to report a set of
+inputs. For simple measurements, such as having each client report a
+count and computing the total, each input is just mapped onto a single
+integer. For more complicated measurements, it may be necessary to map
+a given input onto multiple integers. In any case, the client takes
+all of its inputs and maps them onto a vector of integers each 
+of which is a member of a finite field K. We denote this vector as x.
```
cjpatton, line 486: I understand the intention of this paragraph, however I think it's a bit confusing as it is. For instance, you say that a client wishes to report a "set of inputs", whereas above we think of each clients sending exactly one input, and we refer to the inputs of the aggregation function collectively as the "set of inputs"

WDYTA:
> The Prio computation is parameterized by a finite field, which we will call K, and an integer n. Each client encodes its input as a length-n vector of element of K. The length of the vector depends on the type of data being collected. A single field element may be sufficient for some applications, whereas more sophisticated measurements will require larger encodings.

cjpatton (changes requested):

design-document.md
ekr: Perhaps this is just me being confused. When you and I talked the other day, what I thought I heard was that if I wanted to report, say, "number of urls" and "number of clicks", I would turn that into a single 2-length vector of integers rather then sending two entirely different 1-length vectors and proofs.' Is that not correct?

cjpatton: That's correct. My point is that, in the language of this doc, the "input" in this case is the pair (number_of_urls, number_of_clicks). I'm trying to be precise about the language.

design-document.md
chris-wood, line 488:
```suggestion
In order to share x between s servers, we split it up into s shares
```

design-document.md
chris-wood, line 519:
```suggestion
Above, p(n), u(n), and v(n) are functions that we specify later.
```

chris-wood (approved):

design-document.md
cjpatton: Taking my suggestion.

cjpatton (approved):

PR #28, opened 2021-04-13
More background

ekr:


design-document.md
```diff
@@ -135,6 +142,37 @@ validity. The aggregators use these proof shares in a protocol designed to
...
+the mapping from measurements to reports. The set of measurements
+that go into a report is referred to as the "input" to the report.
+Because each report is independent, for the remainder of this document
+we focus on a single report and its inputs.
+
+The client uses the statistic to be computed in order to know how to
+encode the measurement. For instance, if the statistic is mean, then
+the measurement can be encoded directly. However, if the statistic is
+standard deviation, then the client must send both x and x^2.  Section
+[TODO] describes how to encode measurements for each statistic.
```
cjpatton, line 171:
```suggestion
{{CITE}} describes how to encode measurements for each statistic.
```

cjpatton (approved):
:boat:

design-document.md
ekr: I think this is actually going to be a cross-reference to some to-be-written section of this doc.

design-document.md
```diff
@@ -57,36 +57,42 @@ document.
...
-
-Now the first aggregator is in possession of input shares x[1,1], x[2,1], and
-x[3,1], and the second aggregator is in possession of input shares x[1,2],
-x[2,2], and x[3,2]. Each aggregator computes the sum of its input shares. Let
+Consider an illustrative example. Suppose there are three clients and
+two aggregators. Each client i holds a single measurement in the form of
+a positive integer x[i], and our goal is to compute the sum of the
+measurements of all clients. Using an additive secret-sharing scheme, the first client splits its
+measurement x[1] into a pair of integers x[1,1] and x[1,2] for which x[1] =
+x[1,1] + x[1,2] modulo a prime p. (For convenience, we will omit the
```
chris-wood, line 88:
```suggestion
measurements of all clients. In this case, the protocol input is a single measurement consisting of 
a single positive integer; no additional encoding is done. Given this input, the first client splits its
measurement x[1] with additive secret-sharing into a pair of integers x[1,1] and x[1,2] for which 
x[1] = x[1,1] + x[1,2] modulo a prime p. (For convenience, we will omit the
```

design-document.md
cjpatton: Indeed. "{{CITE}}" the way we've been writing this.

ekr: That's not going to be ideal if/when we want to turn this into an IETF draft, b/c it will cause the formatter to choke.

Issue #30, opened 2021-04-21
Message encoding


chris-wood:
We should specify structure and syntax for messages exchange in the protocol, e.g., data/proof shares. We'll pass these around with high-level HTTP APIs, maybe carried in JSON messages or whatever.

cjpatton:
+1 to JSON blobs, as long as input shares can be encoded in binary. Seems to be the simplest to implement in JS.

cjpatton:
I guess I don't much care if we need to encode input shares in base64 or whatever.

cjpatton:
With #43 we are now using a binary format, specifically TLS syntax. Does anyone have any objections to continuing to use it?

chris-wood:
None -- let's just stick with TLS. :-)

PR #31, opened 2021-04-22
Specify the ingestion flow

### LINKED ISSUE #4: Cross-aggregator data consistency
We assume that all aggregators process the share of a client's input, but what happens if one aggregator loses its share of the client input? How do we accommodate one aggregator losing its view of a share, either maliciously or accidentally?

### LINKED ISSUE #8: Consensus protocol
We currently have an open issue for tracking the consensus protocol needed for this system to work. In working through this text, we should make note of what sort of synchronization we assume between different participants of the system.

### LINKED ISSUE #9: Aggregator discovery
Clients need to obtain aggregator public keys to encrypt shares. While we may not want to specify the exact mechanism by which deployments do this, we should minimally state the requirements for discovery, perhaps with a RECOMMENDED version.

cc @ekr

### LINKED ISSUE #18: Heavy Hitters is in scope
See [the paper](https://eprint.iacr.org/2021/017.pdf).

### LINKED ISSUE #21: Client configuration information
We should probably assume that configuration information, including the desired aggregation function, encoding details, choice of aggregators, etc., are all known to the client before it creates any data for submission. Online parameters such as batch identifiers should also be assumed known for a given client when producing data for the system. In practice, clients might be given batch identifiers to use for a given aggregation from some trusted party, or they might query it from the leader, or something else. The details are probably out of scope. 

Essentially, clients can submit data given configuration parameters and a batch identifier. And that's probably(?) the minimum dependencies here.

### LINKED ISSUE #22: Explore Shamir secret sharing (was "Multiple protocol runs and/or threshold secret sharing")
Prio v2 employs a linear secret sharing scheme in which all input shares are required in order to produce coherent outputs. This gives us robustness against malicious clients, but not malicious servers ([robustness per section 2 of the 2017 paper](https://crypto.stanford.edu/prio/paper.pdf)).

We could provide robustness against a minority of malicious servers by running the protocol over multiple subsets of aggregators. If the outputs across all aggregator subsets do not agree, then participants know that there is at least one malicious server.

If linear secret sharing is employed, then clients would have to generate input shares once for each subset. We could also use a threshold secret sharing scheme (e.g., [Shamir](https://en.wikipedia.org/wiki/Shamir%27s_Secret_Sharing)), which would reduce work for clients.

cjpatton:
Specifies the ingestion flow, including parameter discovery, uploading of shares, and validating of the shares. The same flow applies to both Prio and heavy hitters ("Hits"). The protocol-specific messages are left unspecified. (We will specify them in future PRs.)

Note that I decided to go with TLS syntax for this PR. This was done only because I'm familiar with this style and it was easy for me to write quickly. I'm happy to completely re-do message encoding, but let's do so in a future PR.

Issues addressed by this change:
* Closes #4: In this design, the leader specifies any number of helpers. For each helper, the client initiates a run of the protocol with the leader and the helper. This allows the a helper to drop out without impacting data processing later on. However, the leader cannot drop out. (This seems OK, since this is the same up-time requirement as usual.)
* Addresses #8. In this design, the joint randomness needed for a run of the protocol is picked by the leader and sent to the helpers in the `PAVerifyStartReq` message.
* Closes #9. In this design, the leader tells the client the URL of each helper. The client gets each helper's public key by making an HTTP request.
* Addresses #18. This PR specifies a high-level flow that should fit the input-validation protocol for heavy hitters.
* Closes #21. Any protocol-specific parameters are carried by the `PAClientParam` message.
* Addresses #22. In this design, multiple protocol runs are used to add resilience to aggregator drop-out. We still don't know whether we can use threshold secret sharing (i.e., Shamir) for the same purpose. (In any case, this design disallows it.) We should think about whether this works before closing that issue.

### DIFF
--- a/design-document.md
+++ b/design-document.md
@@ -295,7 +295,7 @@ task IDs to be sparse.]]
 ### Discovery
 
 To begin, the client first needs to discover the parameters and the URL of each
-helper. The client sends a GET request to `[leader]/parameters`. The leader
+helper. The client sends a POST request to `[leader]/upload`. The leader
 responds with status 200 and the following message body:
 
 ```
@@ -399,13 +399,13 @@ opaque PALeaderShare<0..2^24-1>;
 ```
 
 **Response.**
-The leader handles POST requests to `[leader]/upload` as follows. It first
-checks the request for a well-formed `PAUpload` message. If not found, then the
-leader aborts and alerts the client with "malformed upload".  It then looks up
-the PA parameters `PAParam` for which `PAUpload.task == PAParam.task` and
-`PAUpload.helper_url` is in `PAParam.helper_urls`. If none are found, then the
-leader aborts and alerts the client with "unrecognized task". Otherwise, if the
-upload is well-formed and contains a recognized PA task, then the leader
+The leader handles POST requests to `[leader]/upload_finish` as follows. It
+first checks the request for a well-formed `PAUpload` message. If not found,
+then the leader aborts and alerts the client with "malformed upload".  It then
+looks up the PA parameters `PAParam` for which `PAUpload.task == PAParam.task`
+and `PAUpload.helper_url` is in `PAParam.helper_urls`. If none are found, then
+the leader aborts and alerts the client with "unrecognized task". Otherwise, if
+the upload is well-formed and contains a recognized PA task, then the leader
 responds to the GET request with status 200.
 
 ### Verify Start
--- a/design-document.md
+++ b/design-document.md
@@ -51,6 +51,8 @@ document.
    invalid.
 1. Leader: A distinguished aggregator that coordinates input validation and data
    collection.
+1. Measurement: A single value (e.g., a count) being reported by a client.
+   Multiple measurements may be grouped into a single protocol input.
 1. Output: A reduction over the inputs, for instance a statistical aggregation,
    which is of interest to a collector. This is the output of the aggregation
    function.
@@ -62,8 +64,6 @@ document.
    client's input.
 1. Proof share: A share of a proof, used by an aggregator during the
    input-validation protocol.
-1. Measurement: A single value (e.g., a count) being reported by a client.
-   Multiple measurements may be grouped into a single protocol input.
 
 ## Overview {#overview}
 
@@ -231,6 +231,352 @@ is as follows.
 Note: Two non-colluding aggregators --- one leader and one standard aggregator
 --- are required to provide protect the inputs' privacy. Additional aggregators
 may be used to make the system more resilient; see {{CITE}}.
+
+## Ingestion {#proto-ingestion}
+
+[[TODO: Decide how to authenticate the leader in leader-to-helper connections.
+One option is to use client certificates for TLS; another is to have the leader
+sign its messages directly, as in Prio v2.]]
+
+Data ingestion involves the client uploading its secret shared inputs and the
+aggregators collectively validating the corresponding input. Before this process
+begins, the following must be true (the acronym "PA" stands for "Private
+Aggregation"):
+1. The client knows the URL of the leader endpoint, e.g., `example.com/metrics`.
+   We write this URL as `[leader]` below. (We write `[helper]` for the helper's
+   URL and `[aggregator]` for the URL of some aggregator.)
+1. The client has a root certificate it can use to authenticate the leader.
+1. The leader's root certificates are sufficient to authenticate each helper.
+1. The helper has chosen an HPKE key pair.
+1. The aggregators agree on the PA parameters associated with a given PA task.
+
+The PA parameters are structured as follows:
+
+```
+struct {
+  PATask task;
+  select (PAClientParam.task.proto) {
+    case prio: PrioParam;
+    case hits: HitsParam;
+  }
+} PAParam;
+
+struct {
+  PAProto proto;
+  uint16 version;
+  uint64 id;
+} PATask;
+
+Url opaque<0..255>;
+
+enum { prio(0), hits(1) } PAProto;
+```
+
+[[TODO: Add batch parameters, including min/max batch size, batch time window,
+etc.]]
+
+[[NOTE: `PAParamters.task.id` is a 64-bit integer because it may be useful for
+task IDs to be sparse.]]
+
+`PrioParam` is defined in {{prio-messages}} and `HitsParam` is defined in
+{{hits-messages}}.
+
+### Discovery
+
+To begin, the client first needs to discover the parameters and the URL of
+each helper. The client sends an HTTP request to `[leader]/parameters`. The
+leader responds with following message:
+
+```
+struct {
+  PATask task; // Same as PAParam.task
+  select (PAClientParam.task.proto) {
+    case prio: PrioClientParam;
+    case hits: HitsClientParam;
+  }
+  Url helper_urls<0..2^16-1>; // The URL of each helper
+} PAClientParam;
+
+Url opaque<0..255>;
+```
+
+`PrioClientParam` is defined in {{prio-messages}}; `HitsClientParam` is defined
+in {{hits-messages}}. These contain protocol-specific values, including any
+*public* randomness the client will use for proof generation. If the client
+doesn't recognize the protocol or parameters, then it aborts and alerts the
+leader with "unrecognized protocol", as described in {{ingest-alerts}}.
+
+For each URL `[helper]` in `PAClientParam.helper_urls`, the client sends an HTTP
+request to `[helper]/key_config`. The helper's response contains an
+[HPKE](https://datatracker.ietf.org/doc/draft-irtf-cfrg-hpke/) public key and
+encryption parameters:
+
+```
+struct {
+  uint8 id; // Used by helper to decide if it recognizes this config.
+  HpkeKemId kem_id;
+  HpkeKdfId kdf_id;
+  HpkePublicKey public_key;
+} HPKEKeyConfig;
+
+opaque HpkePublicKey<0..2^16-1>;
+uint16 HpkeAeadId; // Defined in I-D.irtf-cfrg-hpke.
+uint16 HpkeKemId;  // Defined in I-D.irtf-cfrg-hpke.
+uint16 HpkeKdfId;  // Defined in I-D.irtf-cfrg-hpke.
+```
+
+We call this the helper's *key configation*. The key configuration is used to
+set up a base-mode HPKE context to use to protect the shares sent to the helper
+via the leader.
+
+Finally, the client collects the set of helpers it will upload shares to. It
+ignores a helper if:
+* the client and helper failed to establish a secure, helper-authenticated
+  channel;
+* the request to the helper URL failed or didn't return a valid key config; or
+* the key config specifies a KEM or KDF algorithm the client doesn't recognize.
+
+If the set of helpers is empty, then the client aborts and alerts the leader
+with "no supported helpers".
+
+### Upload
+
+For each helper, the client issues an HTTP request to the leader.
+
+**Request.**
+The client begins by setting up an HPKE context for the helper by running
+
+```
+helper_enc, context = SetupBaseS(pk, PAClientParam.task)
+```
+
+where `pk` is the KEM public key corresponding to `HPKEKeyConfig.public_key`.
+The outputs are the encapsulated context `helper_enc` and the context `context`.
+
+Next, the client encodes its measurements as an input to the given PA protocol.
+It then generates a validity proof for its input and uses `context` to split the
+input and proof into a leader share and a helper share, where the latter is
+encrypted by the HPKE context. (Note that the secret-sharing and encryption
+procedures are specific to the PA protocol.)
+
+[[OPEN ISSUE: Is it safe to generate the proof once, then secret-share between
+each (leader, helper) pair? Probably not in general, but maybe for Prio?]]
+
+Finally, the clients sends an HTTP request to `[leader]/upload` with the
+following payload:
+
+```
+struct {
+  PATask task;                // PAParameter.task
+  uint8 helper_key_config_id; // HPKEKeyConfig.id
+  Url helper_url;             // URL of helper endpoint
+  PAHelperShare helper_share; // The helper's encrypted share
+  PALeaderShare leader_share; // The leader's share
+} PAUplaod;
+
+struct {
+  opaque enc<0..2^16-1>;     // Encapsulated HPKE context
+  opaque payload<0..2^24-1>; // The encrypted share
+} PAHelperShare;
+
+opaque PALeaderShare<0..2^24-1>;
+```
+
+### Input validation
+
+The leader handles HTTP requests to `[leader]/upload` as follows. It first
+checks the request for a well-formed `PAUpload` message. If not found, then the
+leader aborts and alerts the client with "malformed upload". Otherwise, if a
+well-formed `PAUpload` message is found, the leader responds to the request with
+"200 OK".
+
+#### Verify Start
+
+The leader begins the input-validation protocol by looking up the PA parameters
+`PAParam` for which `PAUpload.task == PAParam.task` and `PAUpload.helper_url` is
+in `PAParam.helper_urls`. If none are found, then the leader aborts without
+alerting the client.
+
+**Request.**
+Let `[helper]` denote the URL corresponding to `PAUpload.helper_url`. The leader
+sends an HTTP request to `[helper]/verify_start` with the following payload:
+
+```
+struct {
+  PATask task;                           // PAUpload.task
+  uint8 key_config_id;                   // PAUpload.helper_key_config_id
+  PAHelperShare shares<0..2^24-1>;       // Any number of helper shares
+  select (PAVerifyStartReq.task.proto) { // Payload for each helper share
+    case prio: PrioVerifyStartReq payloads<0..2^16-1>;
+    case hits: HitsVerifyStartReq payloads<0..2^16-1>;
+  }
+} PAVerifyStartReq;
+```
+
+This message contains any number of encrypted helper shares: it might contain
+the shares uploaded by a single client, or it might contain the shares uploaded
+by multiple clients; we only require that all of the shares correspond to the
+same PA task and HPKE key configuration. This allows the aggregator to
+batch-process a number of inputs at once.
+
+Each helper share has a corresponding message payload contained in
+`PAVerifyStartReq.shares`. This includes any protocol-specific information the
+helper needs for input validation, e.g., the joint randomness used by the leader
+and helper for the protocol run.
+
+Note that a well-formed message contains as many payloads as shares, i.e., the
+sequence `PAVerifyStartReq.payloads` has the same number of elements as
+`PAVerifyStartReq.shares`. Moreover, the i-th payload should correspond to the
+i-th share for each i.
+
+**Response.**
+The helper handles HTTP requests to `[helper]/verify_start` as follows. It first
+checks for a well-formed `PAVerifyStartReq`. If not found, then the helper
+aborts and alerts the leader with "malformed verify start request".
+
+Next, it looks for PA parameters `PAParam` for which `PAVerifyStartReq.task ==
+PAParam.task`. If not found, then it aborts and alerts the leader with
+"unrecognized task".
+
+Next, it looks up the HPKE config and corresponding secret key associated with
+`PAVerifyStartReq.key_config_id`. If not found, then it aborts and alerts the
+leader with "unrecognized key config".
+
+Finally, for each pair of shares and message payloads, the helper does as
+follows. Let `share` denote the share. It computes the HPKE context as
+
+```
+context = SetupBaseR(share.enc, sk, PAParam.task)
+```
+
+where `sk` is the secret key corresponding to the HPKE key config. It then
+derives its proof share from `context` and `share.payload` as specified by the
+PA protocol. Finally, the helper uses the proof share and corresponding message
+to derive its response for the protocol run.
+
+The helper responds to the HTTP request from the leader with the following
+message:
+
+```
+struct {
+  PAProto proto;
+  bool resend_shares; // Whether the leader should resend shares
+  select (PAVerifyStartResp.proto) {
+    case prio: PrioVerifyStartResp payloads<0..2^16-1>;
+    case hits: HitsVerifyStartResp payloads<0..2^16-1>;
+  }
+} PAVerifyStartResp;
+```
+
+The i-th element of `PAVerifyStartResp.payloads` corresponds to the i-th element
+of `PAVerifyStartReq.payloads`. The boolean `resend_shares` is set to `true` if
+the helper wants the leader to retransmit its shares in the last step of the
+input-validation protocol. This allows the helper to implement the protocol
+statelessly, if desired.
+
+#### Verify Finish
+
+The leader handles the helper's response to its HTTP request by first checking
+for a well-formed `PAVerifyStartResp` message. If not found, then it aborts and
+alerts the helper with "malformed verify start response".
+
+**Request.**
+The leader sends an HTTP request to `[helper]/verify_finish` with the following
+message:
+
+```
+struct {
+  PATask task;                     // PAUpload.task
+  uint8 key_config_id;             // PAUpload.helper_key_config_id
+  PAHelperShare shares<0..2^24-1>; // Shares (may be empty)
+  select (PAVerifyFinishReq.task.proto) {
+    case prio: PrioVerifyFinishReq payloads<0..2^16-1>;
+    case hits: HitsVerifyFinishReq payloads<0..2^16-1>;
+  }
+} PAVerifyFinishReq;
+```
+
+The i-th payload corresponds to the i-th payload of the `PAVerifyStartResp` sent
+by the helper in response to the previous request. If
+`PAVerifyStartResp.resend_shares == true`, then `PAverifyFinishReq.shares` is
+equal to `PAVerifyStartReq.shares`. Otherwise, `PAVerifyFinishedReq.shares` is
+empty.
+
+**Response.**
+The helper handles HTTP requests to `[helper]/verify_finish` as follows. It
+first checks for a well-formed `PAVerifyFinishReq` message. If not found, then
+it aborts and alerts the leader with "malformed verify finish request".
+
+Next, it looks for PA parameters `PAParam` for which `PAVerifyFinishReq.task ==
+PAParam.task`. If not found, then it aborts and alerts the leader with
+"unrecognized task".
+
+Next, it looks up the HPKE config and corresponding secret key associated with
+`PAVerifyStartReq.key_config_id`. If not found, then it aborts and alerts the
+leader with "unrecognized key config".
+
+Next, the helper decides whether the input corresponding to each of its shares
+is valid. For each valid input, the helper derives the input share and stores it
+for processing later on, as described in {{CITE}}.
+
+Finally, the helper responds with the following message:
+
+```
+struct {
+  PAProto proto;
+  select (PAVerifyFinishResp.proto) {
+    case prio: PrioVerifyFinishResp payloads<0..2^16-1>;
+    case hits: HitsVerifyFinishResp payloads<0..2^16-1>;
+  }
+} PAVerifyFinishResp;
+```
+
+As usual, the i-th payload corresponds to the i-th payload of the
+`PAVerifyFinishReq`.
+
+The leader handles the response to its HTTP request by checking for a
+well-formed `PAVerifyFinishResp` message. If not found, it aborts and alerts the
+helper with "malformed verify finish response".
+
+Finally, the helper decides whether the input corresponding to each of its
+shares is valid. For each valid input, it derives the input shares and stores it
+for processing later on.
+
+### Prio messages {#prio-messages}
+
+[[TODO: Define `PrioParam`.]]
+[[TODO: Define `PrioClientParam`.]]
+[[TODO: Define `PrioVerifyStartReq`.]]
+[[TODO: Define `PrioVerifyStartResp`.]]
+[[TODO: Define `PrioVerifyFinishReq`.]]
+[[TODO: Define `PrioVerifyFinishResp`.]]
+
+### Heavy-Hitters messages {#hits-messages}
+
+[[TODO: Define `HitsParam`.]]
+[[TODO: Define `HitsClientParam`.]]
+[[TODO: Define `HitsVerifyStartReq`.]]
+[[TODO: Define `HitsVerifyStartResp`.]]
+[[TODO: Define `HitsVerifyFinishReq`.]]
+[[TODO: Define `HitsVerifyFinishResp`.]]
+
+### Alerts {#ingest-alerts}
+
+An *alert* is a message sent either in an HTTP request or response that signals
+to the receiver that the peer has aborted the protocol. The payload is
+
+```
+struct {
+  PATask task;
+  opaque payload<0..255>;
+} PAAlert;
+```
+
+where `task` is the associated PA task (this value is always known) and
+`payload` is the message. When sent by an aggregator in response to an HTTP
+request, the response status is [[TODO]]. When sent in a request to an
+aggregator, the URL is always `[aggregator]/error`, where `[aggregator]` is the
+URL of the aggregator endpoint.
 
 ## Security overview {#security-requirements}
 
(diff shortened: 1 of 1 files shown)


design-document.md
```diff
@@ -232,6 +232,352 @@ Note: Two non-colluding aggregators --- one leader and one standard aggregator
...
+sign its messages directly, as in Prio v2.]]
+
+Data ingestion involves the client uploading its secret shared inputs and the
+aggregators collectively validating the corresponding input. Before this process
+begins, the following must be true (the acronym "PA" stands for "Private
+Aggregation"):
+1. The client knows the URL of the leader endpoint, e.g., `example.com/metrics`.
+   We write this URL as `[leader]` below. (We write `[helper]` for the helper's
+   URL and `[aggregator]` for the URL of some aggregator.)
+1. The client has a root certificate it can use to authenticate the leader.
```
ekr, line 248: This just uses TLS, right? If so, we usually just assume the WebPKI and wave at 6125.

design-document.md
```diff
@@ -232,6 +232,352 @@ Note: Two non-colluding aggregators --- one leader and one standard aggregator
...
+
+struct {
+  PAProto proto;
+  uint16 version;
+  uint64 id;
+} PATask;
+
+Url opaque<0..255>;
+
+enum { prio(0), hits(1) } PAProto;
```
ekr, line 272: Can you reorder these so we don't have a forward reference?

design-document.md
```diff
@@ -232,6 +232,352 @@ Note: Two non-colluding aggregators --- one leader and one standard aggregator
...
+  }
+} PAParam;
+
+struct {
+  PAProto proto;
+  uint16 version;
+  uint64 id;
+} PATask;
+
+Url opaque<0..255>;
```
ekr, line 270: This is kind of a confused thing in TLS, but based on 8446, I believe you want:

```
opaque<0..255> URl;
```

With that said, you also want 0..2^16-1;

design-document.md
```diff
@@ -232,6 +232,352 @@ Note: Two non-colluding aggregators --- one leader and one standard aggregator
...
+    case prio: PrioParam;
+    case hits: HitsParam;
+  }
+} PAParam;
+
+struct {
+  PAProto proto;
+  uint16 version;
+  uint64 id;
+} PATask;
```
ekr, line 268: You need to explain these structs.

design-document.md
```diff
@@ -232,6 +232,352 @@ Note: Two non-colluding aggregators --- one leader and one standard aggregator
...
+1. The leader's root certificates are sufficient to authenticate each helper.
+1. The helper has chosen an HPKE key pair.
+1. The aggregators agree on the PA parameters associated with a given PA task.
+
+The PA parameters are structured as follows:
+
+```
+struct {
+  PATask task;
+  select (PAClientParam.task.proto) {
```
ekr, line 258: You usually need a length field here so that you can extent his to new protos

design-document.md
```diff
@@ -232,6 +232,352 @@ Note: Two non-colluding aggregators --- one leader and one standard aggregator
...
+[[NOTE: `PAParamters.task.id` is a 64-bit integer because it may be useful for
+task IDs to be sparse.]]
+
+`PrioParam` is defined in {{prio-messages}} and `HitsParam` is defined in
+{{hits-messages}}.
+
+### Discovery
+
+To begin, the client first needs to discover the parameters and the URL of
+each helper. The client sends an HTTP request to `[leader]/parameters`. The
```
ekr, line 287: A GET? A POST? [ACME](https://www.rfc-editor.org/rfc/rfc8555.html) does a pretty good job of describing how to talk to this kind of HTTP services server.

design-document.md
```diff
@@ -232,6 +232,352 @@ Note: Two non-colluding aggregators --- one leader and one standard aggregator
...
+struct {
+  PATask task; // Same as PAParam.task
+  select (PAClientParam.task.proto) {
+    case prio: PrioClientParam;
+    case hits: HitsClientParam;
+  }
+  Url helper_urls<0..2^16-1>; // The URL of each helper
+} PAClientParam;
+
+Url opaque<0..255>;
```
ekr, line 300: This seems redundant.

design-document.md
```diff
@@ -232,6 +232,352 @@ Note: Two non-colluding aggregators --- one leader and one standard aggregator
...
+
+```
+struct {
+  PATask task; // Same as PAParam.task
+  select (PAClientParam.task.proto) {
+    case prio: PrioClientParam;
+    case hits: HitsClientParam;
+  }
+  Url helper_urls<0..2^16-1>; // The URL of each helper
+} PAClientParam;
```
ekr, line 298: With an OK or...? What's the content type.

design-document.md
```diff
@@ -232,6 +232,352 @@ Note: Two non-colluding aggregators --- one leader and one standard aggregator
...
+[HPKE](https://datatracker.ietf.org/doc/draft-irtf-cfrg-hpke/) public key and
+encryption parameters:
+
+```
+struct {
+  uint8 id; // Used by helper to decide if it recognizes this config.
+  HpkeKemId kem_id;
+  HpkeKdfId kdf_id;
+  HpkePublicKey public_key;
+} HPKEKeyConfig;
```
ekr, line 320: How does this interact with the related structs in ECH?

ekr (changes requested):
I got about half-way through this and left some comments, but I think it probably needs a fair amount of rework before it's ready for detailed review. In particular, it's pretty informal about the HTTP mechanics. There's a lot of prior art in how to describe this. I would suggest imitating ACME (RFC 8555) which will also help with the client auth if you don't want to do TLS client auth.

design-document.md
```diff
@@ -232,6 +232,352 @@ Note: Two non-colluding aggregators --- one leader and one standard aggregator
...
+
+opaque HpkePublicKey<0..2^16-1>;
+uint16 HpkeAeadId; // Defined in I-D.irtf-cfrg-hpke.
```
chris-wood, line 323: This is unused -- did you mean to include in the key config?

design-document.md
```diff
@@ -232,6 +232,352 @@ Note: Two non-colluding aggregators --- one leader and one standard aggregator
...
+opaque PALeaderShare<0..2^24-1>;
+```
+
+### Input validation
+
+The leader handles HTTP requests to `[leader]/upload` as follows. It first
+checks the request for a well-formed `PAUpload` message. If not found, then the
+leader aborts and alerts the client with "malformed upload". Otherwise, if a
+well-formed `PAUpload` message is found, the leader responds to the request with
+"200 OK".
```
chris-wood, line 391: Should it wait to send "200 OK" until the helper confirms that the upload is well formed? (That is, it can decrypt the share?)

design-document.md
```diff
@@ -232,6 +232,352 @@ Note: Two non-colluding aggregators --- one leader and one standard aggregator
...
+
+#### Verify Start
+
+The leader begins the input-validation protocol by looking up the PA parameters
+`PAParam` for which `PAUpload.task == PAParam.task` and `PAUpload.helper_url` is
+in `PAParam.helper_urls`. If none are found, then the leader aborts without
+alerting the client.
```
chris-wood, line 398: Why would we not relay this signal to clients? It could mean that their configuration information is out of date, or something else. Would that be important for monitoring purposes?

design-document.md
```diff
@@ -232,6 +232,352 @@ Note: Two non-colluding aggregators --- one leader and one standard aggregator
...
+
+**Request.**
+Let `[helper]` denote the URL corresponding to `PAUpload.helper_url`. The leader
+sends an HTTP request to `[helper]/verify_start` with the following payload:
+
+```
+struct {
+  PATask task;                           // PAUpload.task
+  uint8 key_config_id;                   // PAUpload.helper_key_config_id
```
chris-wood, line 407: We should probably say that the leader buckets up uploads based on matching task IDs and helper config IDs above. (It's listed below.)

design-document.md
```diff
@@ -232,6 +232,352 @@ Note: Two non-colluding aggregators --- one leader and one standard aggregator
...
+This message contains any number of encrypted helper shares: it might contain
+the shares uploaded by a single client, or it might contain the shares uploaded
+by multiple clients; we only require that all of the shares correspond to the
+same PA task and HPKE key configuration. This allows the aggregator to
+batch-process a number of inputs at once.
+
+Each helper share has a corresponding message payload contained in
+`PAVerifyStartReq.shares`. This includes any protocol-specific information the
+helper needs for input validation, e.g., the joint randomness used by the leader
+and helper for the protocol run.
```
chris-wood, line 425: How can `PAVerifyStartReq.shares` carry protocol-specific information if it only includes `enc` and `payload`? Is `payload` the share encryption and then other goop? (Reading on, that does seem to be the case. I'm not sure how we'd  clarify that now.)

design-document.md
```diff
@@ -232,6 +232,352 @@ Note: Two non-colluding aggregators --- one leader and one standard aggregator
...
+
+Note that a well-formed message contains as many payloads as shares, i.e., the
+sequence `PAVerifyStartReq.payloads` has the same number of elements as
+`PAVerifyStartReq.shares`. Moreover, the i-th payload should correspond to the
+i-th share for each i.
```
chris-wood, line 430: What happens if this is not the case? Would it be better to just have a list of (share, payload) tuple in `PAVerifyStartReq`?

design-document.md
```diff
@@ -232,6 +232,352 @@ Note: Two non-colluding aggregators --- one leader and one standard aggregator
...
+    case hits: HitsVerifyStartResp payloads<0..2^16-1>;
+  }
+} PAVerifyStartResp;
+```
+
+The i-th element of `PAVerifyStartResp.payloads` corresponds to the i-th element
+of `PAVerifyStartReq.payloads`. The boolean `resend_shares` is set to `true` if
+the helper wants the leader to retransmit its shares in the last step of the
+input-validation protocol. This allows the helper to implement the protocol
+statelessly, if desired.
```
chris-wood, line 475: It also requires the leader to maintain a mapping between shares and messages it sent to the helper. That sort of state seems error prone to manage. Hmm... could this message indicate specifically which shares should be re-sent, if any?

Also, I wonder if we should even make this an option. State management is hard enough as-is, so perhaps requiring just one entity to do it is better than possibly allowing others to do it?

chris-wood (commented):
Good start! Lots of ways we can do this, and I'm happy to just iterate going forward.

design-document.md
cjpatton: Yup. Changed to "The client and leader can establish a leader-authenticated TLS channel."

design-document.md
cjpatton: What's a "forward reference"?

design-document.md
cjpatton: It is, removed.

design-document.md
cjpatton: It's not the same struct, so they're not related. We (Cloudflare) wouldn't use our ECHConfig for this purpose, if that's what you're asking.

design-document.md
cjpatton: I think it should be a GET here. Specified this here and elsewhere.

design-document.md
cjpatton: 200. Specified here and elsewhere.

design-document.md
cjpatton: Didn't seem necessary here, but I don't mind adding it.

design-document.md
cjpatton: Good catch, added AEAD code point to key config

design-document.md
```diff
@@ -232,6 +232,352 @@ Note: Two non-colluding aggregators --- one leader and one standard aggregator
...
+struct {
+  PATask task;                // PAParameter.task
+  uint8 helper_key_config_id; // HPKEKeyConfig.id
+  Url helper_url;             // URL of helper endpoint
+  PAHelperShare helper_share; // The helper's encrypted share
+  PALeaderShare leader_share; // The leader's share
+} PAUplaod;
+
+struct {
+  opaque enc<0..2^16-1>;     // Encapsulated HPKE context
```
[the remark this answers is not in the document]
cjpatton: Ah, good call.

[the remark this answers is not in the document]
cjpatton: Actually it's better if the config id is not in the `PAHelperShare`. This struct is reused in the `PAVerifyStartReq` message, and there a single config id is used for several helper shares.

design-document.md
```diff
@@ -232,6 +232,352 @@ Note: Two non-colluding aggregators --- one leader and one standard aggregator
...
+helper_enc, context = SetupBaseS(pk, PAClientParam.task)
+```
+
+where `pk` is the KEM public key corresponding to `HPKEKeyConfig.public_key`.
+The outputs are the encapsulated context `helper_enc` and the context `context`.
+
+Next, the client encodes its measurements as an input to the given PA protocol.
+It then generates a validity proof for its input and uses `context` to split the
+input and proof into a leader share and a helper share, where the latter is
+encrypted by the HPKE context. (Note that the secret-sharing and encryption
```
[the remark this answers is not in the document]
cjpatton: As we discussed today, some PA protocols (like Prio) might derive an HPKE context, export a PRG seed, and use that seed for secret sharing. Others (like HH) might secret share first, then encrypt.

design-document.md
```diff
@@ -232,6 +232,352 @@ Note: Two non-colluding aggregators --- one leader and one standard aggregator
...
+  channel;
+* the request to the helper URL failed or didn't return a valid key config; or
+* the key config specifies a KEM or KDF algorithm the client doesn't recognize.
+
+If the set of helpers is empty, then the client aborts and alerts the leader
+with "no supported helpers".
+
+### Upload
+
+For each helper, the client issues an HTTP request to the leader.
```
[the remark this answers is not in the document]
cjpatton: I suppose the PAUploads could be sent in the same POST request. This seemed easier to implement. Besides, the leader will never share state between the individual runs.

design-document.md
```diff
@@ -232,6 +232,352 @@ Note: Two non-colluding aggregators --- one leader and one standard aggregator
...
+following payload:
+
+```
+struct {
+  PATask task;                // PAParameter.task
+  uint8 helper_key_config_id; // HPKEKeyConfig.id
+  Url helper_url;             // URL of helper endpoint
+  PAHelperShare helper_share; // The helper's encrypted share
+  PALeaderShare leader_share; // The leader's share
+} PAUplaod;
```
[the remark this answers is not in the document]
cjpatton: It doesn't end up being a simple copy-and paste. IF you look at PAVerifyStartReq, there is a U16 between key_config_id and the first helper share that encodes the length of the list of helper shares.

design-document.md
cjpatton: Hmm, that's an interesting suggestion. I'm leaning towards "no" because the leader is permitted to batch a bunch of verify-start requests. This might cause latency that the client would notice. Besides, there's no reason to wait, since the client doesn't care if its shares were deemed valid.

design-document.md
cjpatton: In fact, this is spelled out above on the section on verify start request. Would you like me to repeat it here?

design-document.md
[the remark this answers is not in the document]
cjpatton: Done, but I called the inner struct `PAHelperShare`.

design-document.md
```diff
@@ -232,6 +232,352 @@ Note: Two non-colluding aggregators --- one leader and one standard aggregator
...
+sends an HTTP request to `[helper]/verify_start` with the following payload:
+
+```
+struct {
+  PATask task;                           // PAUpload.task
+  uint8 key_config_id;                   // PAUpload.helper_key_config_id
+  PAHelperShare shares<0..2^24-1>;       // Any number of helper shares
+  select (PAVerifyStartReq.task.proto) { // Payload for each helper share
+    case prio: PrioVerifyStartReq payloads<0..2^16-1>;
+    case hits: HitsVerifyStartReq payloads<0..2^16-1>;
```
tgeoghegan, line 422: The use of "payload" here is confusing. Aren't the shares the payload? I'm also not clear on why there is a unique `{Prio|Hits}VerifyStartReq` for each `PAHelperShare`. Would `PrioVerifyStartReq` contain the random value used for the polynomial identity test?

design-document.md
```diff
@@ -232,6 +232,352 @@ Note: Two non-colluding aggregators --- one leader and one standard aggregator
...
+```
+
+This message contains any number of encrypted helper shares: it might contain
+the shares uploaded by a single client, or it might contain the shares uploaded
+by multiple clients; we only require that all of the shares correspond to the
+same PA task and HPKE key configuration. This allows the aggregator to
+batch-process a number of inputs at once.
+
+Each helper share has a corresponding message payload contained in
+`PAVerifyStartReq.shares`. This includes any protocol-specific information the
```
tgeoghegan, line 428: "share" vs. "payload" is confusing in these lines.

design-document.md
```diff
@@ -232,6 +232,352 @@ Note: Two non-colluding aggregators --- one leader and one standard aggregator
...
+sequence `PAVerifyStartReq.payloads` has the same number of elements as
+`PAVerifyStartReq.shares`. Moreover, the i-th payload should correspond to the
+i-th share for each i.
+
+**Response.**
+The helper handles HTTP requests to `[helper]/verify_start` as follows. It first
+checks for a well-formed `PAVerifyStartReq`. If not found, then the helper
+aborts and alerts the leader with "malformed verify start request".
+
+Next, it looks for PA parameters `PAParam` for which `PAVerifyStartReq.task ==
```
tgeoghegan, line 440: Looks where? I think the idea is that each helper maintains a list of `PAParam`s it is aware of and participating in?

design-document.md
```diff
@@ -232,6 +232,352 @@ Note: Two non-colluding aggregators --- one leader and one standard aggregator
...
+1. The client has a root certificate it can use to authenticate the leader.
+1. The leader's root certificates are sufficient to authenticate each helper.
+1. The helper has chosen an HPKE key pair.
+1. The aggregators agree on the PA parameters associated with a given PA task.
+
+The PA parameters are structured as follows:
+
+```
+struct {
+  PATask task;
```
tgeoghegan, line 259: Is there a 1:1 relationship between `PAParam` and `PATask`? I think yes, since participants need to be able to use a task ID to look up a `PAParam` value.

design-document.md
tgeoghegan: One way or the other, I don't think `resend_shares` is an attribute of an individual `PAVerifyStartResp`. A helper server either has storage or not and so would presumably always send the same value of `resend_shares`. Should this be part of the `PATask` or `PAParam` that the leader and all helpers have to agree upon before the protocol starts?

design-document.md
```diff
@@ -232,6 +232,352 @@ Note: Two non-colluding aggregators --- one leader and one standard aggregator
...
+The helper handles HTTP requests to `[helper]/verify_finish` as follows. It
+first checks for a well-formed `PAVerifyFinishReq` message. If not found, then
+it aborts and alerts the leader with "malformed verify finish request".
+
+Next, it looks for PA parameters `PAParam` for which `PAVerifyFinishReq.task ==
+PAParam.task`. If not found, then it aborts and alerts the leader with
+"unrecognized task".
+
+Next, it looks up the HPKE config and corresponding secret key associated with
+`PAVerifyStartReq.key_config_id`. If not found, then it aborts and alerts the
```
tgeoghegan, line 518:
```suggestion
`PAVerifyFinishReq.key_config_id`. If not found, then it aborts and alerts the
```

design-document.md
```diff
@@ -232,6 +232,352 @@ Note: Two non-colluding aggregators --- one leader and one standard aggregator
...
+
+#### Verify Finish
+
+The leader handles the helper's response to its HTTP request by first checking
+for a well-formed `PAVerifyStartResp` message. If not found, then it aborts and
+alerts the helper with "malformed verify start response".
+
+**Request.**
+The leader sends an HTTP request to `[helper]/verify_finish` with the following
+message:
```
tgeoghegan, line 488: IIUC, the leader will run the FLPCP `decide` algorithm for the input using the `PAVerifyStartResp` messages it gathers from helpers, and the verify finish step means informing the helper(s) of the results of `decide`?

design-document.md
cjpatton: Done.

design-document.md
```diff
@@ -232,6 +232,352 @@ Note: Two non-colluding aggregators --- one leader and one standard aggregator
...
+
+```
+struct {
+  PATask task;                     // PAUpload.task
+  uint8 key_config_id;             // PAUpload.helper_key_config_id
+  PAHelperShare shares<0..2^24-1>; // Shares (may be empty)
```
tgeoghegan, line 494: If this is empty then I think you still need a list of share unique identifiers so that helper knows how to resolve the list of `{Prio|Hits}VerifyFinishReq` messages it gets against the shares it has stored.

design-document.md
cjpatton: Done.

design-document.md
cjpatton: Yes that's the case. `PAVerifyStartREq.shares` should be `PAVerifyStartReq.payloads` here.

design-document.md
cjpatton: That seemed tricky to do in TLS syntax, particularly because we need to select the type of `payload` on `task.proto`. I'm open to suggestions for how to spell this differently. In the meantime, I added a TODO below explaining that this would be better as a sequence of (share, payload) tuples.

design-document.md
cjpatton: To clarify why this bit exists: the helper shares may be so big that the leader and helper agree to use state rather than have to retransmit them. If set, then the leader is supposed to retransmit *all* of the shares.

> It also requires the leader to maintain a mapping between shares and messages it sent to the helper. That sort of state seems error prone to manage. Hmm... could this message indicate specifically which shares should be re-sent, if any?

The intention is that the leader needs to re-send *all* of the shares if the bit is true, or *none* if the bit is false. I don't think we should bother with supporting anything in the middle.


> Also, I wonder if we should even make this an option. State management is hard enough as-is, so perhaps requiring just one entity to do it is better than possibly allowing others to do it?

Shares are likely to be fairly large, especially for heavy hitters. I think this is going to need to be an option.

cjpatton: > One way or the other, I don't think `resend_shares` is an attribute of an individual `PAVerifyStartResp`. A helper server either has storage or not and so would presumably always send the same value of `resend_shares`. Should this be part of the `PATask` or `PAParam` that the leader and all helpers have to agree upon before the protocol starts?

That is an option, though the bit is more deployment-specific rather than protocol-specific, which is why I put it in the high level message.

design-document.md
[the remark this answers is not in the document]
cjpatton: That's what the next line says. (See updated text.)

design-document.md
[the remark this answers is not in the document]
cjpatton: The task id is presumably encoded by the string `[leader]`.

design-document.md
```diff
@@ -232,6 +232,352 @@ Note: Two non-colluding aggregators --- one leader and one standard aggregator
...
+```
+
+We call this the helper's *key configation*. The key configuration is used to
+set up a base-mode HPKE context to use to protect the shares sent to the helper
+via the leader.
+
+Finally, the client collects the set of helpers it will upload shares to. It
+ignores a helper if:
+* the client and helper failed to establish a secure, helper-authenticated
+  channel;
```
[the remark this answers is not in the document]
cjpatton: This refers to the TLS channel between client and helper.

design-document.md
```diff
@@ -232,6 +232,352 @@ Note: Two non-colluding aggregators --- one leader and one standard aggregator
...
+Data ingestion involves the client uploading its secret shared inputs and the
+aggregators collectively validating the corresponding input. Before this process
+begins, the following must be true (the acronym "PA" stands for "Private
+Aggregation"):
+1. The client knows the URL of the leader endpoint, e.g., `example.com/metrics`.
+   We write this URL as `[leader]` below. (We write `[helper]` for the helper's
+   URL and `[aggregator]` for the URL of some aggregator.)
+1. The client has a root certificate it can use to authenticate the leader.
+1. The leader's root certificates are sufficient to authenticate each helper.
+1. The helper has chosen an HPKE key pair.
```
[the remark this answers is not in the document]
cjpatton: > What is the scope of an HPKE key pair? Does the helper use a single HPKE key for all deployments it participates in, or one per `PATask`?

That's up to the deployment I think.

> If a helper has a single HPKE key pair across all `PATask`s (which makes sense to me if we think of the HPKE pub as analogous to a TLS server's certificate), then I think we should specify that `PAParam.task.id` gets used as a label in the HPKE KDF to ensure keys aren't re-used across `PATask`s.

It does! The context is computed as `SetupBaseR(share.enc, sk, PAParam.task)`

design-document.md
```diff
@@ -232,6 +232,352 @@ Note: Two non-colluding aggregators --- one leader and one standard aggregator
...
+1. The aggregators agree on the PA parameters associated with a given PA task.
+
+The PA parameters are structured as follows:
```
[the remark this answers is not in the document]
cjpatton: That's premature. We may end up re-doing the encoding for #30.

design-document.md
cjpatton: Not necessarily. In any case, `PATask` is supposed to encode what's common between `PAParam` and `PAClientParam`.

design-document.md
```diff
@@ -232,6 +232,352 @@ Note: Two non-colluding aggregators --- one leader and one standard aggregator
...
+
+```
+struct {
+  PATask task;
+  select (PAClientParam.task.proto) {
+    case prio: PrioParam;
+    case hits: HitsParam;
+  }
+} PAParam;
```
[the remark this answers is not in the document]
cjpatton: In fact, the symmetric name would be `PAAggregatorParam`. I have a weak preference for `PAParam` because it's less wordy.

design-document.md
```diff
@@ -232,6 +232,352 @@ Note: Two non-colluding aggregators --- one leader and one standard aggregator
...
+leader with "unrecognized protocol", as described in {{ingest-alerts}}.
+
+For each URL `[helper]` in `PAClientParam.helper_urls`, the client sends an HTTP
+request to `[helper]/key_config`. The helper's response contains an
+[HPKE](https://datatracker.ietf.org/doc/draft-irtf-cfrg-hpke/) public key and
+encryption parameters:
+
+```
+struct {
+  uint8 id; // Used by helper to decide if it recognizes this config.
```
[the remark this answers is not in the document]
cjpatton: I removed the comment and added some clarification in the paragraph below.

design-document.md
```diff
@@ -232,6 +232,352 @@ Note: Two non-colluding aggregators --- one leader and one standard aggregator
...
+  HpkeKemId kem_id;
+  HpkeKdfId kdf_id;
```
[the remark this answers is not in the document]
cjpatton: Yup.

design-document.md
```diff
@@ -232,6 +232,352 @@ Note: Two non-colluding aggregators --- one leader and one standard aggregator
...
+  uint8 helper_key_config_id; // HPKEKeyConfig.id
+  Url helper_url;             // URL of helper endpoint
+  PAHelperShare helper_share; // The helper's encrypted share
+  PALeaderShare leader_share; // The leader's share
+} PAUplaod;
+
+struct {
+  opaque enc<0..2^16-1>;     // Encapsulated HPKE context
+  opaque payload<0..2^24-1>; // The encrypted share
+} PAHelperShare;
```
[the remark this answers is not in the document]
cjpatton: For Prio, `PAHelperShare.payload` will be the empty string.

design-document.md
```diff
@@ -232,6 +232,352 @@ Note: Two non-colluding aggregators --- one leader and one standard aggregator
...
+in `PAParam.helper_urls`. If none are found, then the leader aborts without
+alerting the client.
+
+**Request.**
+Let `[helper]` denote the URL corresponding to `PAUpload.helper_url`. The leader
+sends an HTTP request to `[helper]/verify_start` with the following payload:
+
+```
+struct {
+  PATask task;                           // PAUpload.task
```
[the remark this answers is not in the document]
cjpatton: `task` also specifies the version (of this document) and the PA protocol ("prio" or "hits"). This info is needed in order to parse the rest of the message.

design-document.md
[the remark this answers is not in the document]
cjpatton: Shares might be big, but I don't think the corresponding protocol message will be big. Hence U24 for the former and U16 for the latter. For consistency, we could do U24 for both. I'm agnostic.

cjpatton: Right, `PrioVerifyStartReq` and `HItsVerifyStartReq` carries the protocol-specific stuff, e.g., the random value input to `query` in the context of Prio.

design-document.md
cjpatton: "share" is the thing that is encrypted and contains the input and proof share; "payload" is whatever else is needed by the helper in the next step. The reason these are separate is because the "share" is copied verbatim from the client's upload request, and the "payload" is generated by the leader.

design-document.md
cjpatton: Yup, that's the idea.

design-document.md
cjpatton: Yes.

design-document.md
cjpatton: That seems like a lot of overhead that's not really necessary. It should be clear that the first element of `payloads` corresponds to the first share, the second element of `payloads` to the second share, and so on. This should be fairly straightforward if the helper is already maintaining state.

cjpatton (commented):
Thanks, all, for your comments! I've tried to address them all... several remain unresolved.

One area of confusion and which we don't quite agree is the `resend_shares` flag in the verify start response. This allows the leader and helper to avoid retransmitting the shares, as long as the helper is able to maintain state between requests. This is helpful if the shares are really big. I don't think we have a use case with really big shares yet, so I'd be fine with just dropping this flag and having the leader always retransmit the shares.

Also, @ekr, at this stage I'm not trying to be overly formal with the spec. My goal is to describe the protocol in enough detail that we all basically know what's going on and we can start working on an implementation. There are also some open questions, which preclude a more formal description. That said, I did make an effort to be more clear about what's a POST and what's a GET, as well as the response status for each HTTP request.

design-document.md
cjpatton: OH wait, I misunderstood. Yes, there's 1:1 map.

cjpatton:
Be advised: I changed the first client->leader request to a POST. That way the leader can "challenge" the client (e.g., give it a random nonce), which may be needed for some protocols

design-document.md
```diff
@@ -38,6 +38,7 @@ document.
 1. False input: An input that is valid, but incorrect. For example, if the data
    being gathered is whether or not users have clicked on a particular button, a
    client could report clicks when none occurred.
+1. Helper: An aggreegator that is not a leader.
```
tgeoghegan, line 41:
```suggestion
1. Helper: An aggregator that is not a leader.
```

design-document.md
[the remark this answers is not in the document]
tgeoghegan: Oh, I think I got confused by the notation. I thought `PAHelperShare shares<0..2^24-1>` meant "between 0 and 2^24-1 `PAHelperShare` structures, but it means `0..2^24-1` bytes.

design-document.md
tgeoghegan: I think it'd help to have language explaining that all participants are configured with a list of known `PAParam`s before the start of the discovery or ingestion protocols.

tgeoghegan (approved):
With the understanding that we still need to specify lots of details about message encoding and HTTP semantics, this looks good to me. Thanks Chris!

design-document.md
cjpatton: Done. (See the changes to the beginning of {{proto-ingestion}}.

design-document.md
cjpatton: Quick follow-up: I'm fine with removing the `resend_shares` bit for simplicity and just require the leader to always resend. We can always add it later if it's needed.

tgeoghegan: I think we should err on the side of making the protocol simpler, and hence remove the option. My intuition is that shares should be small for Prio (since there aren't actually shares but rather just a PRNG seed from which shares are derived).

cjpatton:
@chris-wood has some major revisions (restructuring the doc, mostly) so I'll be closing this PR and opening a new one. I'm also making the following changes:
- Remove "resend_shares" flag
- Add an "upload ID" to PAClientParam.

design-document.md
cjpatton: I plan to remove the bit.

PR #33, opened 2021-04-27
Use mdbook to render doc

cjpatton:
This allows us to use `mdbook` to render the design doc into HTML, as suggested by @acmiyaguchi. To run, do
```
cargo install mdbook
mdbook server
```
This starts an HTTP server that updates the page each time the source is edited. This makes editing the document much a more pleasant experience, I think.


acmiyaguchi (commented):
I suggest adding a README with the instructions to build/serve the docs. Otherwise it's a good starting point 👍.

acmiyaguchi (approved):

REAMDE.md
```diff
@@ -0,0 +1,14 @@
+This document is built using Rust's
```
acmiyaguchi, line 1: The document is title REAMDE instead of README

cjpatton: Ugh, sorry.

book.toml
```diff
@@ -0,0 +1,6 @@
+[book]
+authors = ["Christopher Patton"]
```
tgeoghegan, line 2: Not that I don't appreciate your contributions but could we change this to "The authors" or "The contributors" or remove it altogether if it's optional?

tgeoghegan (changes requested):

book.toml
cjpatton: Oops, that was added auto-magically

cjpatton: Done.

tgeoghegan (approved):

PR #34, opened 2021-04-30
Doc rework and specification of upload and verify flows

### LINKED ISSUE #4: Cross-aggregator data consistency

### LINKED ISSUE #9: Aggregator discovery

### LINKED ISSUE #21: Client configuration information

cjpatton:
Specifies the upload and verify phases of a PA protocol. In the upload phase, the client uploads a report; in the verify phase, the aggregators verify that their shares correspond to a valid input. These flows apply to both Prio and heavy hitters ("Hits"), though the protocol-specific messages in these flows are left unspecified. (We will specify them in future PRs.)

I've also broken the various sections of the document into separate files so that `mdbook` can create a table of contents.

**Non-goal for this PR:** This PR is not meant to provide a definitive, formalized "standard document". It's meant only to get us one step closer to the final shape of the protocol.

Issues addressed by this change:

* Closes #4: In this design, the leader specifies any number of helpers. For each helper, the client initiates a run of the protocol with the leader and the helper. This allows the a helper to drop out without impacting data processing later on. However, the leader cannot drop out. (This seems OK, since this is the same up-time requirement as usual.)
*Addresses #8. In this design, the joint randomness needed for a run of the protocol is picked by the leader and sent to the helpers in the PAVerifyStartReq message.
* Closes #9. In this design, the leader tells the client the URL of each helper. The client gets each helper's public key by making an HTTP request.
* Addresses #18. This PR specifies a high-level flow that should fit the input-validation protocol for heavy hitters.
* Closes #21. Any protocol-specific parameters are carried by the PAClientParam message.
* Addresses #22. In this design, multiple protocol runs are used to add resilience to aggregator drop-out. We still don't know whether we can use threshold secret sharing (i.e., Shamir) for the same purpose. (In any case, this design disallows it.) We should think about whether this works before closing that issue.


src/02-pa-upload.md
```diff
@@ -0,0 +1,112 @@
...
+
+The message includes the URL of each helper and any protocol specific parameters
+the client needs to generate its report. The leader's response to malformed
+requests is specified in [the "Error handling"
+section](pa-error.md#common-abort-conditions).
+
+## Upload Finish
+
+For each URL `[helper]` in `PAClientParam.helper_urls`, the client sends a GET
+request to `[helper]/key_config`. The helper responds with status 200 and an
```
acmiyaguchi, line 51: I think the `key_config` request should be broken out into it's own sub-section.

src/02-pa-upload.md
```diff
@@ -0,0 +1,112 @@
...
+
+The payload of the POST request is structured as follows:
+
+```
+struct {
+  PATask task;
+  opaque id[16];              // Equal to PAUploadStartReq.id
+  uint8 helper_key_config_id; // Euqal to HPKEKeyConfig.id
+  Url helper_url;             // URL of helper endpoint
+  PAHelperShare helper_share; // The helper's encrypted share
```
acmiyaguchi, line 95: Should this be an array of helper shares, if there's more than a single helper?

src/02-pa-upload.md
```diff
@@ -0,0 +1,112 @@
...
+  }
+} PAUploadStartResp;
+```
+
+The message includes the URL of each helper and any protocol specific parameters
+the client needs to generate its report. The leader's response to malformed
+requests is specified in [the "Error handling"
+section](pa-error.md#common-abort-conditions).
+
+## Upload Finish
```
acmiyaguchi, line 48: Suppose we're dealing with an anonymizing proxy: Is it possible for the parameters to be known ahead of time (e.g. hpke parameters shipped directly into client code) and to omit the first 2 stages of `[leader]/upload_start` and `[helper(s)]/key_config`?

src/02-pa-config.md
```diff
@@ -0,0 +1,82 @@
+# Configuration
+
+**Tasks.**
+Each PA protocol is associated with a *PA task* that specifies the measurements
+that are to be collected and the protocol that will be used to collect them:
+
+```
+struct {
+  uint16 version;
+  uint16 id;
```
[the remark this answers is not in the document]
cjpatton: - Unspecified at this time, though I imagine it'll be the collector that assigns this.
- I'm not sure if 65k is enough (see OPEN ISSUE below).

src/02-pa-config.md
```diff
@@ -0,0 +1,82 @@
...
+    case hits: HitsParam; // Defined in Section 4
+  }
+} PAParam;
+```
+
+[TODO: Add batch parameters, including min/max batch size, batch time window,
+etc.]
+
+[OPEN ISSUE: Would it be useful if `PAoTask.id` were bigger so that task ids can
+be more sparse?]
```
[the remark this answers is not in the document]
cjpatton: +1 to UUID.

src/02-pa-upload.md
cjpatton: Ack, will do.

src/02-pa-upload.md
cjpatton: Actually, there will only be one helper for this message.

src/02-pa-upload.md
cjpatton: Yeah, that should be possible for most PA protocols. What's the added value of the anonymizing proxy? Unlinking any metadata, like UA, from the client IP?

acmiyaguchi: Yeah, unlinking metadata like ip addresses, timestamps, and frequency of requests.

I was thinking about ingestion services (e.g. the prio-server architecture) and whether it was possible to support that use-case with the http api proposed here.

PR #36, opened 2021-05-03
Port document and toolchain to IETF I-D format.

chris-wood:
cc @ekr, @cjpatton


chris-wood:
FYI, I'll fix the build after this is merged to main. (It might be due to me creating this on a branch that's not main, though I'm not entirely sure.)

cjpatton (approved):
This dumps some text related to choosing parameters for Prio. I need to rework this text anyway, so I'm happy to add it back later.

PR #38, opened 2021-05-03
Rework overview and add skeleton for other stuff

cjpatton:


draft-pda-core.md
```diff
@@ -26,6 +26,27 @@ informative:
       - ins: H. Corrigan-Gibbs
       - ins: D. Boneh
 
+  BBG+19:
```
chris-wood, line 29: This is {{BBCp19}} in the document, referenced by DOI, so maybe we can drop this (or the other) so there aren't duplicates?

draft-pda-core.md
```diff
@@ -102,41 +128,44 @@ document.
...
-x[1] = x[1,1] + x[1,2] modulo a prime p. (For convenience, we will omit the
-the "mod p" in the rest of this section.) It then uploads x[1,1] to
-one sever x[1,2] to the other. The second client splits its measurement x[2]
-into x[2,1] and x[2,2], uploads them to the servers, and so on.
-
-Now the first aggregator is in possession of shares x[1,1], x[2,1], and
-x[3,1], and the second aggregator is in possession of shares x[1,2],
-x[2,2], and x[3,2]. Each aggregator computes the sum of its  shares. Let
-A[1] denote the first aggregator's share of the sum and let A[2] denote the
+The protocol is executed by a large set of clients and a small set of servers.
```
chris-wood, line 131: Small = 2?

draft-pda-core.md
```diff
@@ -102,41 +128,44 @@ document.
...
+We call the servers the *aggregators*. Each client's input to the protocol is a
+set of measurements (e.g., counts of some user behavior). Given the input set
+of measurements `x_1, ..., x_n` held by `n` users, the goal of a
+*private aggregation (PA) protocol* is to compute `y = F(x_1, ..., x_n)` for
+some aggregation function `F` while revealing nothing else about the
+measurements.
+
+## Private aggregation via secret sharing
+
+The main cryptographic tool we'll use for achieving this privacy goal is
```
chris-wood, line 141:
```suggestion
The main cryptographic tool used for achieving this privacy goal is
```

draft-pda-core.md
```diff
@@ -166,24 +199,44 @@ limited set of aggregation functions, some of which we highlight below:
...
-share, each client sends to each aggregator a share of a "proof" of the input's
-validity. The aggregators use these proof shares in a protocol designed to
-establish the input's validity, without leaking the input itself. We describe
-this input-validation protocol in detail in [[TODO:citeme]].
-
-## Assembling Reports
+**Hits.**
+A common PA task that can't be solved efficiently with Prio is the
+`t`-*heavy-hitters* problem {{HITS}}. In this setting, each user is in
+possession of an `n`-bit string, and the goal is to compute the compute the set
```
chris-wood, line 205:
```suggestion
possession of a single `n`-bit string, and the goal is to compute the compute the set
```

draft-pda-core.md
```diff
@@ -166,24 +199,44 @@ limited set of aggregation functions, some of which we highlight below:
...
+inputs are secret shared. In particular, a malicious client can corrupt the
+computation by submitting random integers instead of a proper secret sharing of
+a valid input.
+
+To solve this problem, each PA protocol in this document specifies a
+*zero-knowledge proof (ZKP) system* that allows the aggregators to verify that
+their shares correspond to as valid input. The system needs to have the
+following security properties (stated informally here; see {{BBG+19}} for a
+formal definition):
+1. *Completeness:* The verification procedure always succeeds on valid inputs.
```
chris-wood, line 227: Can we move these to security considerations? I don't think they're important here.

draft-pda-core.md
```diff
@@ -166,24 +199,44 @@ limited set of aggregation functions, some of which we highlight below:
...
+An essential task of any data collection pipeline is ensuring that the input
+data is "valid". Going back to the example above, it's often useful to assert
+that each measurement is in a certain range, e.g., `[0, 2^k)` for some `k`.
+This straight-forward task is complicated in our setting by the fact that the
+inputs are secret shared. In particular, a malicious client can corrupt the
+computation by submitting random integers instead of a proper secret sharing of
+a valid input.
+
+To solve this problem, each PA protocol in this document specifies a
+*zero-knowledge proof (ZKP) system* that allows the aggregators to verify that
```
chris-wood, line 223:
```suggestion
protocol that allows the aggregators to verify that
```

In general, I think we ought to try and keep these technical terms in the security considerations where possible.

draft-pda-core.md
cjpatton: Small = O(1). Remember that the plan is to execute the protocol among multiple sets of two servers.

draft-pda-core.md
cjpatton: I disagree.

draft-pda-core.md
cjpatton: Again, I disagree. "ZKP" is no less technical of a term than "encryption", which we use freely in this section. It's just that "ZKP" is less common.

draft-pda-core.md
ekr: FWIW, I agree with Wood.  The point here is that there is a proof system that demonstrates validity. The details should go below.

draft-pda-core.md
ekr: I would split the difference here and say "includes a zero-knowledge proof". That is somewhat less formal than "zero-knowledge proof system" and I think clearer in this context.

ekr: Or, for that matter, "proof"

draft-pda-core.md
chris-wood: Yes, I'm aware. My point was that the core protocol runs with s=2 servers.

draft-pda-core.md
```diff
@@ -26,6 +26,27 @@ informative:
...
+    target: "https://eprint.iacr.org/2019/188"
+    author:
+      -ins: D. Boneh
+      -ins: E. Boyle
+      -ins: H. Corrigan-Gibbs
+      -ins: N. Gilboa
+      -ins: Y. Ishai
+
+  HITS:
+    title: "ightweight Techniques for Private Heavy Hitters"
```
ekr: Lightweight,

draft-pda-core.md
```diff
@@ -35,21 +26,26 @@ TODO
...
-Prio. In the next section ({{security-requirements}}) we enumerate the security
-goals and non-goals of our protocol. In section {{operational-requirements}} we
-list the use cases our system needs to support, how it is configured, and any
-other operational constraints for the protocol. In {{protocol}} we specify the
-protocol itself.
+This document describes a framework for specifying protocols for
+privacy-preserving data-aggregation. Each protocol is executed by a large set of
+clients and a small set of servers. The servers' goal is to compute some
+aggregate statistic over the clients' inputs without learning the inputs
+themselves. This is made possible by "distributing trust" among the servers in
```
ekr, line 33:
```suggestion
themselves. This is made possible by distributing the computation among the servers in
```

draft-pda-core.md
```diff
@@ -35,21 +26,26 @@ TODO
...
+introduction to the underlying cryptographic techniques. In the next section
+({{pa}}) we specify a generic framework for PA protocols. in {{prio}} we
+describe the framework's instantiation with Prio {{BC17}}, and in {{hits}} we
+describe its instantiation with the "heavy hitters" protocol of Boneh et al.
+{{BBCp21}}.
+
+In section {{security-requirements}} we enumerate the security goals and
+non-goals of our protocol. In section {{operational-requirements}} we list the
+use cases our system needs to support, how it is configured, and any other
+operational constraints for the protocol.
```
ekr, line 48: I would rework this and the previous graf. It's not conventional in specifications to provide this kind of "in this section, we do X, then in the next section we do Y". Rather,  we just try to make the specification evolve clearly.

draft-pda-core.md
```diff
@@ -102,41 +98,44 @@ document.
...
+set of measurements (e.g., counts of some user behavior). Given the input set
+of measurements `x_1, ..., x_n` held by `n` users, the goal of a
+*private aggregation (PA) protocol* is to compute `y = F(x_1, ..., x_n)` for
+some aggregation function `F` while revealing nothing else about the
+measurements.
+
+## Private aggregation via secret sharing
+
+The main cryptographic tool used for achieving this privacy goal is *additive
+secret sharing*. Rather than send its input in the clear, each client "splits"
```
ekr, line 112:
```suggestion
secret sharing*. Rather than send its input in the clear, each client splits
```

draft-pda-core.md
```diff
@@ -102,41 +98,44 @@ document.
...
+its measurements into a sequence of *shares* and sends a share to each of the
+aggregators. Additive secret sharing has two important properties:
+- First, it's impossible to deduce the measurement without knowing *all* of the
```
ekr, line 115:
```suggestion
- It's impossible to deduce the measurement without knowing *all* of the
```

draft-pda-core.md
```diff
@@ -102,41 +98,44 @@ document.
...
+  shares.
+- Second, it allows the aggregators to compute the final output by first adding
```
ekr, line 117:
```suggestion
- It allows the aggregators to compute the final output by first adding
```

Bulleted lists do not need verbal indices.

draft-pda-core.md
```diff
@@ -147,17 +146,21 @@ because modular addition is commutative. I.e.,
...
-- Simple statistics, like sum, mean, min, max, variance, and standard
-  deviation; [[OPEN ISSUE: It's possible to estimate quantiles such as the
-  median. How practical is this?]]
+**Prio.**
+This approach can be used to privately compute any function `F` that can be
+expressed as a function of the sum of the users' inputs. In Prio {{BC17}}, each
+user encodes its measurement as a sequence of elements over some prime field. It
+then splits its input into shares and sends each share to the aggregators. The
+aggregators sum up their input shares. Once all the shares have been aggregated,
+they combine their shares of the aggregate to get the final output.
```
ekr, line 155: I think we could do without the "prime field" bit here. It's not necessary to explain what Prio is for.

draft-pda-core.md
```diff
@@ -166,24 +169,44 @@ limited set of aggregation functions, some of which we highlight below:
...
-validity. The aggregators use these proof shares in a protocol designed to
-establish the input's validity, without leaking the input itself. We describe
-this input-validation protocol in detail in [[TODO:citeme]].
-
-## Assembling Reports
+**Hits.**
+A common PA task that can't be solved efficiently with Prio is the
+`t`-*heavy-hitters* problem {{BBCp21}}. In this setting, each user is in
+possession of a single `n`-bit string, and the goal is to compute the compute
+the set of strings that occur at least `t` times.
```
ekr, line 176: ```With Prio, it would be necessary to encode 2^n values, where all but one is zero. This is clearly inefficient.```

draft-pda-core.md
```diff
@@ -166,24 +169,44 @@ limited set of aggregation functions, some of which we highlight below:
...
+
+[TODO: Provide an overview of the protocol of [BBG+21] and provide some
+intuition about how additive secret sharing is used. Be sure to introduce the
+notion of distributed point functions.]
```
ekr, line 180: I would not actually talk about DPFs here. These sections should treat the protocols as black boxes.

ekr (changes requested):

draft-pda-core.md
cjpatton: Ah, gotcha. I still think "small set of servers" is more accurate than "two servers" here, since this section is meant to describe the general of case of one leader and multiple helpers.

draft-pda-core.md
cjpatton: Good catch! Thanks.

draft-pda-core.md
cjpatton: Is that true? I believe there's more to it than this.

draft-pda-core.md
cjpatton: +1 to not mentioning DPFs.

draft-pda-core.md
cjpatton: What would you put instead?

draft-pda-core.md
cjpatton: Ack, I'll remove the paragraphs for now.

draft-pda-core.md
ekr: Why?

draft-pda-core.md
ekr: I would remove the entire sentence.

draft-pda-core.md
cjpatton: I would push back as follows: several weeks ago I had to explain to @ekr that input validation fails to detect an invalid input with non-zero probability, but that this probability could be bounded. In general, this is a property of ZKP systems: the ZKP system is *sound* if verification correctly detects invalid inputs except with small probability. I don't think this is overly technical. Moreover, it's necessary to explain to the reader that input validation may fail.

ekr: Nobody is saying that you don't say it. Rather, we're saying that you should say it *later*, because it's a detail and is not needed to understand the big picture.

The more detail you have in this section the harder it is for the reader to get the overall idea.

draft-pda-core.md
cjpatton: I reworked it to avoid using the word "system".

draft-pda-core.md
ekr: As with the other text, I think the purpose of this section is to give people the right intuition.

What's misleading about what I said?

draft-pda-core.md
cjpatton: In general, I think there's a danger here of contorting the language so much in order to avoid using "advanced" crypto terminology that we confuse the reader.

draft-pda-core.md
cjpatton: There's two things: 
1. Instead of "encode 2^n values" it would be more accurate to say "generate a proof of length `O(some_function_of(n))`". where `some_function_of(.)` depends on the proof strategy. I suppose we could just say "generate a huge proof".
2. The "collect" step doesn't work the same way. It's not a simple matter of adding up shares; the collector has to interact with the aggregators over several rounds.

draft-pda-core.md
chris-wood: That may be true, but I think Ekr's point is that we want to try and decrease cognitive load on the reader (implementer). They should exactly what they need to know to implement the protocol. Everything else can be described elsewhere!

draft-pda-core.md
cjpatton: That works.

draft-pda-core.md
chris-wood: I think the intuition ("should I use Prio or HH?") is good to add here. (@ekr's suggestion seems to clarify the "this can't be solved efficiently claim" with a simple example, so I'd support adding it)

draft-pda-core.md
cjpatton: I think implementers are going to have to understand the term "ZKP" to the same depth they understand "encrypt" or "secret share". I'm defining "ZKP" here because w'er also defining "secret sharing" here. I don't think it makes sense to define one but not the other. They are both equally important at this stage.

In any case, this is a purely editorial point. I would suggest one of two paths forward: (1) we punt on this conversation and leave an "OPEN ISSUE" for cleaning up the presentation. (We'll have to do this at some point anyway.) (2) remove the text completely and leave a "TODO" for describing input validation.

draft-pda-core.md
cjpatton: How about: `One reason Prio isn't applicable to the use case is that the proof that is generated would be huge.`

ekr:
That seems fine.

draft-pda-core.md
cjpatton: Done.

draft-pda-core.md
cjpatton: Or (3) does someone want to propose alternate text? The things I think are important to convey are (a) that the verification protocol does not leak the secret inputs to the aggregators and (b) the verification protocol may fail to detect invalid inputs with small probability.

ekr: @cjpatton I think we're talking past each other: I agree that this is useful text, I just don't think it belongs *here*.

The generic structure of this kind of specification is:

- Problem statement
- Overview of the solution that just is enough to give people the idea
- Detailed protocol specification
- Security considerations and the like

The key idea with point (2) is to be high level. That means avoiding detail which isn't absolutely necessary. In particular, why do we need to tell users at this point that it might fail to detect invalid inputs? They don't need to know that to understand the protocol specification. See https://tools.ietf.org/rfcmarkup?doc=4101 for more on this.

cjpatton: > The key idea with point (2) is to be high level. That means avoiding detail which isn't absolutely necessary. In particular, why do we need to tell users at this point that it might fail to detect invalid inputs?

I guess I don't quite understand what "absolutely necessary" means. It's important to know that the input validation protocol can be attacked, since mitigating such attacks is one of the core design criteria for this class of protocols. IMO we have to mention it up front.

Towards finding common ground, I'm wondering if it's strictly about using the term "ZKP"? Wood suggested that I might just be trying to use a fancy crypto term here. I assure you that I'm not. I'm just trying to provide the context one needs to understand the system. Using the term itself is useful because it connects what we're doing to a fairly well-known body of work. I don't want to give people the impression that the crypto herein is one-off.

> They don't need to know that to understand the protocol specification. See https://tools.ietf.org/rfcmarkup?doc=4101 for more on this.

So, up until this point, I haven't been thinking of this document as a formal standard. Based on the feedback I'm getting from you and Wood, it seems like I might need to recalibrate. I thought the goal of this change was merely to get us all on the same page about overall shape of the protocol, not (yet!) to provide a reference for others to implement. It's probably worth working this out in the next design call.

By the way, I would love to participate in the task of shaping this into a formal standard. I just didn't think that's what we were doing yet.

ekr: I don't care much one way or the other about the term "zero-knowledge proof". I think you could say "proof" just as well. What I am saying is that your list of properties is unnecessary here. You can just say:

"To solve this problem, in each PA protocol, the client generates a
*zero-knowledge proof (ZKP)* of its input's validity, which the aggregators use
to verify that their shares correspond to as valid input."

And stop.  Or maybe say "This proof can be verified without learning the input".

You then move the properties to the Security Considerations. If you must, you can put a pointer to that here. Anyone who knows enough about ZKPs to care about the details can check there, and anyone who doesn't will read the text I have proposed and get substantively the right impression.


> Using the term itself is useful because it connects what we're doing to a fairly well-known body of work. I don't want to give people the impression that the crypto herein is one-off.

That can be done later in the specification

> So, up until this point, I haven't been thinking of this document as a formal standard. Based on the feedback I'm getting from you and Wood, it seems like I might need to recalibrate. I thought the goal of this change was merely to get us all on the same page about overall shape of the protocol, not (yet!) to provide a reference for others to implement. It's probably worth working this out in the next design call.

Well, I don't understand the point of this text at all if it's not targeted at an eventual specification. But yes, I do think this should be an attempt to write down pieces of a specification to use as input into the standards process.

cjpatton: How's this:
```
To solve this problem, in each PA protocol, the client generates a
zero-knowledge proof of its input's validity that the aggregators use
to verify that their shares correspond to as valid input. The verification
procedure is designed to ensure that the aggregators learn nothing about the
input beyond its validity.
```
To me it's useful to use the term "zero-knowledge" because it signals what's under the hood. If you want I can remove it.

> Well, I don't understand the point of this text at all if it's not targeted at an eventual specification. But yes, I do think this should be an attempt to write down pieces of a specification to use as input into the standards process.

Of course that's what we're doing :) I meant more that I don't see why we need to settle on the presentation right now. I'd prefer to focus on the protocol itself and less on how it's presented. I figure that we will need to rework this doc quite a lot before we consider trying to get buy in from the wider community.

draft-pda-core.md
chris-wood: Sounds good!

chris-wood (approved):

PR #40, opened 2021-05-04
Fixup gh-pages

martinthomson:
No idea if this is going to fix your problems here, but it might.  Something went pretty badly wrong somewhere along the line and this might help fix it.

The fallback is to completely remove the branch and start over.  There's a script that you can use for that, but it's pretty scary.


chris-wood:
Yeah, I tried blowing away gh-pages and trying from scratch. Maybe your fix will stick?

PR #43, opened 2021-05-05
Specify the upload and verify phases of the PA protocol

cjpatton:
Based on #38.

@chris-wood, @ekr

### DIFF
--- a/draft-pda-core.md
+++ b/draft-pda-core.md
@@ -296,16 +296,15 @@ Each PA task in this document is divided into three sub-protocols as follows.
 This section specifies a protocol for executing generic PA tasks. Concrete
 PA protocols are described in {{prio}} and {{hits}}.
 
-Each round of the protocol corresponds to an HTTP request and response. We
-assume that some transport layer security protocol (e.g., TLS or QUIC) is used
-between each pair of parties and that the server is always authenticated.
+Each round of the protocol corresponds to an HTTP request and response. The
+content type of each request is "application/octet-stream". We assume that some
+transport layer security protocol (e.g., TLS or QUIC) is used between each pair
+of parties and that the server is always authenticated.
 
 [TODO: Decide how to authenticate the leader in leader-to-helper and
 aggregator-to-collector connections. One option is to use client certificates
 for TLS; another is to have the leader sign its messages directly, as in Prio
 v2.]
-
-[TODO: Decide the "content type" for HTTP requests.]
 
 [TODO: @chris-wood suggested we specify APIs for producing and consuming each of
 the messages in the protocol. Specific PA protocols would implement this API.]
@@ -341,7 +340,7 @@ field, `id` is an opaque identifier used by the clients, aggregators, and
 collector to uniquely identify the PA task at hand. We will call it the *task
 id* in the remainder.
 
-[TODO: Decide how the `PATask` is negotiated. Eventually this will be
+[TODO: Decide how the `PATask` is configured. Eventually this will be
 distributed, in an authenticated manner, from the collector the other parties.
 For now, we just assume this value is negotiated out-of-band.]
 
@@ -399,7 +398,8 @@ decrypt a share it receives.
 
 ## Pre-conditions
 
-We assume the following conditions hold before the protocol begins:
+We assume the following conditions hold before the client begins uploading its
+data:
 1. The client, aggregators, and collector are configured with a specific PA task.
 1. The client knows the URL of the leader endpoint, e.g., `example.com/metrics`.
    We write this URL as `[leader]` below. (We write `[helper]` for a helper's
--- a/draft-pda-core.md
+++ b/draft-pda-core.md
@@ -331,9 +331,6 @@ struct {
 } PATask;
 
 ~~~
-[TODO: Decide how to serialize protocol messages. We're using TLS syntax for
-now, but there's no reason to stick with it other than most folks are familiar
-with it.]
 
 The first field, `version` specifies the version of this document. The second
 field, `id` is an opaque identifier used by the clients, aggregators, and
--- a/draft-pda-core.md
+++ b/draft-pda-core.md
@@ -650,7 +650,7 @@ struct {
 
 For each sub-request `PAVerifyReq`, the helper computes the corresponding
 sub-response as follows. It first checks that checks that `PAVerifyReq.proto ==
-PAPram.proto`. If not, it aborts and alerts the leader with "incorrect protocol
+PAParam.proto`. If not, it aborts and alerts the leader with "incorrect protocol
 for sub-request". Otherwise, It computes the HPKE context as
 
 ~~~
--- a/draft-pda-core.md
+++ b/draft-pda-core.md
@@ -398,6 +398,7 @@ decrypt a share it receives.
 
 We assume the following conditions hold before the client begins uploading its
 data:
+
 1. The client, aggregators, and collector are configured with a specific PA task.
 1. The client knows the URL of the leader endpoint, e.g., `example.com/metrics`.
    We write this URL as `[leader]` below. (We write `[helper]` for a helper's
--- a/draft-pda-core.md
+++ b/draft-pda-core.md
@@ -672,9 +672,6 @@ the body of the `PAVerifyResp` according to the PA protocol.
 
 ## Error handling {#pa-error}
 
-A protocol participant *aborts* the protocol by tearing down the TLS connection
-with each of its peers.
-
 An *alert* is a message sent either in an HTTP request or response that signals
 to the receiver that the peer has aborted the protocol. The payload is
 
--- a/draft-pda-core.md
+++ b/draft-pda-core.md
@@ -104,6 +104,8 @@ document.
 
 # Overview {#overview}
 
+[OPEN ISSUE: Rework this section in light of issue #44.]
+
 The protocol is executed by a large set of clients and a small set of servers.
 We call the servers the *aggregators*. Each client's input to the protocol is a
 set of measurements (e.g., counts of some user behavior). Given the input set
@@ -143,13 +145,13 @@ second aggregator's share of the sum. In the last step, aggregators combine
 their sum shares to obtain the final output `y = A[1] + A[2]`. This is correct
 because modular addition is commutative. I.e.,
 
-```
+~~~
     y = A[1] + A[2]
       = (x[1,1] + x[2,1] + x[3,1]) + (x[1,2] + x[2,2] + x[3,2])
       = (x[1,1] + x[1,2]) + (x[2,1] + x[2,2]) + (x[3,1] + x[3,2])
       = x[1] + x[2] + x[3]
       = F(x[1], x[2], x[3])
-```
+~~~
 
 **Prio.**
 This approach can be used to privately compute any function `F` that can be
@@ -252,7 +254,7 @@ multiple pairs of aggregators.]
 
 Each PA task in this document is divided into three sub-protocols as follows.
 
-```
+~~~
                     +------------+
                     |            |
                     |   Helper   <---------------+
@@ -274,7 +276,7 @@ Each PA task in this document is divided into three sub-protocols as follows.
                     |   Helper   <---------------+
                     |            |
                     +------------+
-```
+~~~
 
 1. **Upload:** Each client assembles the measurements into an input for the given
    PA protocol. It generates a proof of its input's validity and splits the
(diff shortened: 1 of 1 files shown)


csharrison:
I have a high level concern with the architecture here, following up from our meeting on Wednesday related to heavy hitters.

**TL;DR evaluating a hierarchy at only a subset of levels (which are specified at query time) is not optimal for performance if the verification step does not have access to those specified levels.**

For heavy hitters, we are considering a model where records form a binary tree, where aggregates can be reported at different levels of the tree. There are a few possible ways this could go:
1. For every record, verify and evaluate the entire binary tree (i.e. bit-by-bit evaluation)
2. For every record, verify and evaluate only a subset of the tree (at the limit, this can look more like Prio which only evaluates the last level of the tree)
  2a. The subset of the tree to evaluate is configured within a particular record at record-creation time
  2b. The subset of the tree to evaluate is configured dynamically at _collection / aggregation_ time

For our use-case (discussed in https://github.com/abetterinternet/prio-documents/issues/18#issuecomment-801248636), we are interested in (2b), for a few reasons:
- (1) is potentially inefficient from both a performance and accuracy POV in that it requires more rounds, and additionally when used with differential privacy requires splitting a DP budget across all levels. Some prefixes might just not be interesting to the caller so we shouldn't waste privacy budget / compute on them.
- (2a) is problematic in two dimensions.
  - Embedding configuration in records may compromise user privacy if the configurations can be set by an adversary colluding with one of the aggregators (e.g. you can leak log2(n choose k) bits of information if you allow any subset of k prefixes for an n-bit domain).
  - Levels must be pre-specified at record-creation time, which is non-ideal if new information comes up at collection time which informs how aggregation should work (e.g. realizing you have fewer records than expected so it is better to focus on querying only up to a given prefix).

However, I think (2b) runs into a few issues with this architecture that separates verification from collection, in that without knowing the specific levels to evaluate, we run the risk of spending unnecessary compute verifying levels that will never end up being used. The protocol is more efficient (less computation, fewer rounds) if we can verify multiple levels at once (i.e. the levels the collector cares about).

Possible solutions:
1. Add a step where the collector and leader communicate prior to verification (but keep verification and collection separate)
2. More tightly couple "verify" and "collect" stages, so that verification can be done based on communication with the collector.

In our setting, the "leader" and "collector" roles are somewhat merged which aligns with option (1), but it seems like a good idea to make the general architecture robust to this use-case.

cc @schoppmp who is working on our C++ implementation of IDPFs.

draft-pda-core.md
```diff
@@ -291,11 +291,437 @@ Each PA task in this document is divided into three sub-protocols as follows.
...
+## Configuration {#pa-config}
+
+**Tasks.**
+Each PA protocol is associated with a *PA task* that specifies the measurements
+that are to be collected and the protocol that will be used to collect them:
+
+~~~
+struct {
+  uint16 version;
+  uint16 id;
```
chris-wood, line 323: This probably needs to be larger, say, `uint8 id[32]`?

draft-pda-core.md
```diff
@@ -291,11 +291,437 @@ Each PA task in this document is divided into three sub-protocols as follows.
...
+[TODO: Decide the "content type" for HTTP requests.]
+
+[TODO: chris-wood suggested we specify APIs for producing and consuming each of
+the messages in the protocol. Specific PA protocols would implement this API.]
+
+**Error handling.**
+In this section, we will use the verbs "abort" and "alert with `[some error
+message]`" to describe how protocol participants react to various error
+conditions. The behavior is specified in {{pa-error}}. For common errors, we may
+elide the verbs altogether and refer to to {{pa-error-common-aborts}}.
```
chris-wood, line 312: This seems out of place -- was it supposed to be move to some later point in the document?

draft-pda-core.md
```diff
@@ -291,11 +291,437 @@ Each PA task in this document is divided into three sub-protocols as follows.
...
+[TODO: Add an illustration of this sub-protocol.]
+
+Uploading a report involves two requests to the leader. In the  *upload start
+request*, the client discovers the protocol-specific parameters and the endpoint
+URL of each helper. In the *upload finish request* uploads its report to the
+leader.
+
+[OPEN ISSUE: @acmiyaguchi pointed out that the use of an anonymizing proxy for
+uploading shares might be easier to implement if the "upload" phase involved a
+single HTTP request. Is this true? Does OHTTP
```
chris-wood, line 412: 👍 agree with this, though it seems deployment specific? I might include this as an option for some deployments, maybe in an appendix.

draft-pda-core.md
```diff
@@ -291,11 +291,437 @@ Each PA task in this document is divided into three sub-protocols as follows.
...
+
+~~~
+struct {
+  PATask task;
+  opaque id[16];
+} PAUploadStartReq;
+~~~
+
+The first field, `task` corresponds to the PA task for which a report will be
+generated. The second, `id` is a random, 16-byte string generated by the
```
chris-wood, line 431: The `id` seems to tie the upload start and finish together. A couple questions:
1. Can we drop it if we turn this into a single HTTP request?
2. if we keep both requests, why do they need to be tied together?

draft-pda-core.md
```diff
@@ -291,11 +291,437 @@ Each PA task in this document is divided into three sub-protocols as follows.
...
+[OPEN ISSUE: Maybe use a UUID for `PAUploadSTartReq.id`, as well as for
+`PATask.id`?]
+
+**Response.**
+The leader responds to well-formed requests to `[leader]/upload_start` with
+status 200 and the following message:
+
+~~~
+struct {
+  Url helper_urls<0..2^16-1>;
```
chris-wood, line 443:
```suggestion
  Url helper_urls<1..2^16-1>;
```

Also, has `Url` been defined yet?

draft-pda-core.md
```diff
@@ -291,11 +291,437 @@ Each PA task in this document is divided into three sub-protocols as follows.
...
+upload shares to. It ignores a helper if:
+* the client and helper failed to establish a secure, helper-authenticated
+  channel;
+* the GET request to the helper URL failed or didn't return a valid key config;
+  or
+* the key config specifies a KEM, KDF, or AEAD algorithm the client doesn't
+  recognize.
+
+If the set of supported helpers is empty, then the client aborts and alerts the
+leader with "no supported helpers". Otherwise, for each supported helper the
```
chris-wood, line 470: Why is this alert important? What is the leader to do with this information?

draft-pda-core.md
```diff
@@ -291,11 +291,437 @@ Each PA task in this document is divided into three sub-protocols as follows.
...
+  PATask task;
+  opaque id[16];              // Equal to PAUploadStartReq.id
+  uint8 helper_key_config_id; // Euqal to HPKEKeyConfig.id
+  Url helper_url;             // URL of helper endpoint
+  PAHelperShare helper_share; // The helper's encrypted share
+  PALeaderShare leader_share; // The leader's share
+} PAUploadFinishReq;
+
+struct {
+  opaque enc<0..2^16-1>;     // Encapsulated HPKE context
```
chris-wood, line 508:
```suggestion
  opaque enc<1..2^16-1>;     // Encapsulated HPKE context
```

draft-pda-core.md
```diff
@@ -291,11 +291,437 @@ Each PA task in this document is divided into three sub-protocols as follows.
...
+  opaque payload<0..2^16-1>; // The encrypted share (may be empty)
```
chris-wood, line 509: Why is this allowed to be empty?

draft-pda-core.md
```diff
@@ -291,11 +291,437 @@ Each PA task in this document is divided into three sub-protocols as follows.
...
+
+opaque PALeaderShare<0..2^16-1>;
+~~~
+
+This message is called a *report*.
+
+**Response.**
+The leader responds to well-formed requests to `[leader]/upload_finish` with
+status 200 and an empty payload. Malformed requests are handled as described in
+{{pa-error-common-aborts}}.
```
chris-wood, line 520: What does the client do if one helper returns a 200 and another returns an error?

draft-pda-core.md
```diff
@@ -291,11 +291,437 @@ Each PA task in this document is divided into three sub-protocols as follows.
...
+
+## Verify {#pa-verify}
+
+[TODO: Add an illustration of this sub-protocol.]
+
+After the client uploads a report to the leader, the leader and helper verify
+the validity proof as specified by the PA protocol. This phase involves two
+requests sent from by the leader to the helper: the *verify start request*, and
+the *verify finish request*. The contents of each request depend largely on the
```
chris-wood, line 529:
```suggestion
the *verify finish request*. The contents of each request depend on the
```

draft-pda-core.md
```diff
@@ -291,11 +291,437 @@ Each PA task in this document is divided into three sub-protocols as follows.
...
+specific PA protocol. However, both requests contain the helper's share of the
+report, allowing the helper to run the verification protocol statelessly.
+
+Another important feature of the verification protocol is that the leader may
+"batch" multiple runs of the protocol into the same message flow. This allows
+the leader to throttle the load on the helper endpoint.
+
+At the end of this phase, the leader and helper will have decided whether a set
+of client inputs are valid. For each valid input, they proceed as described in
```
chris-wood, line 538: Is the set here the batch in total, or the set of individual reports?

draft-pda-core.md
```diff
@@ -291,11 +291,437 @@ Each PA task in this document is divided into three sub-protocols as follows.
...
+    case prio: PrioVerifyStartReq payloads<0..2^24-1>;
+    case hits: HitsVerifyStartReq payloads<0..2^24-1>;
+  }
+} PAVerifyStartReq;
+~~~
+
+Each helper share has a corresponding message payload contained in
+`PAVerifyStartReq.payloads`. This includes any protocol-specific information the
+helper needs for the first step of input validation, e.g., the joint randomness
+used by the leader and helper for the protocol run. Note that a well-formed
```
chris-wood, line 567: We've not defined joint randomness, so I might drop this example text entirely. The preceding text seems clear enough as-is.

draft-pda-core.md
```diff
@@ -291,11 +291,437 @@ Each PA task in this document is divided into three sub-protocols as follows.
...
+below.) This is difficult to express in TLS syntax because we have to select on
+the protocol type.]
+
+**Response.**
+The helper handles well-formed requests as follows. (As usual, malformed
+requests are handled as described in {{pa-error-common-aborts}}.) It first looks
+for the PA parameters `PAParam` for which `PAVerifyStartReq.task.id ==
+PAParam.task.id`. Next, it looks up the HPKE config and corresponding secret key
+associated with `PAVerifyStartReq.key_config_id`. If not found, then it aborts
+and alerts the leader with "unrecognized key config".
```
chris-wood, line 585: What does the leader do for a batch in this situation?

draft-pda-core.md
```diff
@@ -291,11 +291,437 @@ Each PA task in this document is divided into three sub-protocols as follows.
...
+As usual, the `i`-th payload corresponds to the `i`-th payload of the
+`PAVerifyStartResp` sent by the helper in response to the previous request.
+
+**Response.**
+The helper handles POST requests to `[helper]/verify_finish` as follows. It
+begins just as before by looking up the PA parameters `PAParam` for which
+`PAVerifyStartReq.task.id == PAParam.task.id`. Next, it looks up the HPKE config
+and corresponding secret key associated with `PAVerifyStartReq.key_config_id`.
+If not found, then it aborts and alerts the leader with "unrecognized key
+config".
```
chris-wood, line 644: Is there missing text here? What does the helper actually do with the FinishReq message?

draft-pda-core.md
```diff
@@ -291,11 +291,437 @@ Each PA task in this document is divided into three sub-protocols as follows.
...
+} PAVerifyFinishResp;
+~~~
+
+As usual, the `i`-th payload corresponds to the `i`-th payload of the
+`PAVerifyFinishReq`.
+
+### Decision
+
+The helper decides whether each of its input shares is valid after it responds
+to the leader's verify finish request. Likewise, the leader decides whether each
```
chris-wood, line 665: To clarify, this means the helper knows the valid bit after sending `PAVerifyFinishResp`, right?

draft-pda-core.md
```diff
@@ -291,11 +291,437 @@ Each PA task in this document is divided into three sub-protocols as follows.
...
+use in [the "Collect" phase](pa-collect.md) of the PA protocol.
+
+# Collect {#pa-collect}
+
+[TODO]
+
+# Error handling {#pa-error}
+
+A protocol participant *aborts* the protocol by tearing down the TLS connection
+with each of its peers.
```
chris-wood, line 679: Is this what we want to do if the helper fails to process one share of a batch?

draft-pda-core.md
```diff
@@ -291,11 +291,437 @@ Each PA task in this document is divided into three sub-protocols as follows.
...
+~~~
+struct {
+  PATask task;
+  opaque payload<0..255>;
+} PAAlert;
+~~~
+
+where `task` is the associated PA task (this value is always known) and
+`payload` is the message. When sent by an aggregator in response to an HTTP
+request, the response status is [TODO]. When sent in a request to an
```
chris-wood, line 693:
```suggestion
request, the response status is 400. When sent in a request to an
```

draft-pda-core.md
```diff
@@ -291,11 +291,437 @@ Each PA task in this document is divided into three sub-protocols as follows.
...
+## Verify {#pa-verify}
+
+[TODO: Add an illustration of this sub-protocol.]
+
+After the client uploads a report to the leader, the leader and helper verify
+the validity proof as specified by the PA protocol. This phase involves two
+requests sent from by the leader to the helper: the *verify start request*, and
+the *verify finish request*. The contents of each request depend largely on the
+specific PA protocol. However, both requests contain the helper's share of the
+report, allowing the helper to run the verification protocol statelessly.
```
chris-wood, line 531:
```suggestion
report submitted during the upload protocol, allowing the helper to run the verification protocol statelessly.
```

chris-wood:
@csharrison can you please file an issue with your comment? It doesn't seem to apply to this PR, which punts on the aggregate protocol.

csharrison:
Ah yes sorry, I included it here because I thought it may impact design for upload / verify. Let me delete and re-upload as an issue.

draft-pda-core.md
cjpatton: No, it was meant to go here.

draft-pda-core.md
```diff
@@ -291,11 +291,437 @@ Each PA task in this document is divided into three sub-protocols as follows.
...
+[TODO: Add batch parameters, including min/max batch size, batch time window,
+etc.]
+
+
+### Helper key configuration
+
+Each helper specifies the
+[HPKE](https://datatracker.ietf.org/doc/draft-irtf-cfrg-hpke/) public key the
+client will use to encrypt the helper's share of the report. The public key and
+associated parameters are structured as follows:
```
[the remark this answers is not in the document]
cjpatton: The delta would be adding support for multiple HPKE ciphersuites. Although I see the value of aligning the structures across specs, I'd also like to push back on adding agility for now. I've added a TODO for resolving this later.

draft-pda-core.md
cjpatton: >  The id seems to tie the upload start and finish together. A couple questions:
>
>     1. Can we drop it if we turn this into a single HTTP request?

For some Prio proof systems, it's useful for the leader to send the client a "challenge" that it'll use to generate the proof. One way to use the id is as a way of generating the challenge in a way that's unique for each upload.

>     2. if we keep both requests, why do they need to be tied together?

Retransmitting the id allows the two HTTP requests to be handled without carrying state between them.

draft-pda-core.md
cjpatton: It hasn't! Added a definition.

I'm not taking the suggestion because I don't want to get nit-picky about size ranges yet. This is something that's likely to change a lot.

draft-pda-core.md
cjpatton: Ah, good point. This might happen, say, if the helper advertises the wrong key config during the upload phase. The shares are useless if the helper can't decrypt them, so I don't think the leader has a choice other than to abort. This falls into the bucket of situations that re-running the protocol with multiple helpers is meant to address.

cjpatton:
TODO(@cjpatton) Allow an aribtrary number of verify requests

draft-pda-core.md
chris-wood: Hmm, okay. It just feels out of place. We can always move later.

draft-pda-core.md
```diff
@@ -291,11 +291,437 @@ Each PA task in this document is divided into three sub-protocols as follows.
...
+~~~
+
+We call this the helper's *key configation*. The key configuration is used to
+set up a base-mode HPKE context to use to derive symmetric keys for protecting
+the shares sent to the helper. The *config id*, `HPKEKeyConfig.id`, is forwarded
+by the client to the helper, who uses this value to decide if it knows how to
+decrypt a share it receives.
+
+**Pre-conditions.**
+We assume the following conditions hold before the protocol begins:
```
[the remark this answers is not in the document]
chris-wood: 👍

draft-pda-core.md
```diff
@@ -291,11 +291,437 @@ Each PA task in this document is divided into three sub-protocols as follows.
...
+1. The leader and each helper can establish a leader-authenticated TLS channel.
+1. Each helper has chosen an HPKE key pair.
+1. The aggregators agree on a set of PA tasks, as well as the PA protocol and
+   parameters used for each task.
+
+## Upload {#pa-upload}
+
+[TODO: Add an illustration of this sub-protocol.]
+
+Uploading a report involves two requests to the leader. In the  *upload start
```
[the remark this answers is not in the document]
chris-wood: 👍

draft-pda-core.md
chris-wood: Well, can the helper_urls be empty? If not, then we should take the suggestion.

draft-pda-core.md
chris-wood: Can't the leader determine if helpers are "online"? This seems to reveal information that's specific to clients. Imagine, for example, that clients are prohibited from talking to helpers but not the leader. Is it OK that leaders learn that about a client? I'm not sure, so I'd be inclined to remove this unless we have a concrete use case.

draft-pda-core.md
chris-wood: The protocol (as specified here) allows n>1 helpers to be used, right? (We may just only ever use one in practice.) I'm asking what we do if there are n>1 helpers and they all don't return 200. Does that clarify?

draft-pda-core.md
chris-wood: Yeah, we might want to note this (or flag a TODO to spell it out later).

draft-pda-core.md
cjpatton: Well, a Url is itself a u16-prefixed string, so I guess the minimum is actually 2. Again, I don't think it's worth being picky about this right now.

cjpatton (commented):
Post-prototype clean-up.

draft-pda-core.md
```diff
@@ -291,11 +293,439 @@ Each PA task in this document is divided into three sub-protocols as follows.
...
+After the client uploads a report to the leader, the leader and helper verify in
+zero knowledge that the proof is well-formed. The exact procedure for doing so
+is protocol specific, but all protocols have the same basic structure. In
+particular, the protocol is comprised of a sequence of HTTPS requests from the
+leader to the helper. At the end of this phase, the leader and helper will have
+decided whether a set of client inputs are valid. For each valid input, they
+proceed as described in {{pa-collect}}.
+
+**Request.**
+The leader begins by collecting a sequence of reports that are all associated
```
chris-wood, line 596:
```suggestion
The leader begins with a sequence of reports that are all associated
```

Slightly reworded so as to not require servers to do some sort of batching if they don't have any reports to batch.

draft-pda-core.md
```diff
@@ -291,11 +293,439 @@ Each PA task in this document is divided into three sub-protocols as follows.
...
+struct {
+  PAProto;
+  select (PAVerifyResp.proto) {
+    case prio: PrioVerifyResp;
+    case hits: HitsVerifyResp;
+  }
+} PAVerifyReq;
+~~~
+
+Fopr each sub-request `PAVerifyReq`, the helper computes the corresponding
```
chris-wood, line 662:
```suggestion
For each sub-request `PAVerifyReq`, the helper computes the corresponding
```

draft-pda-core.md
chris-wood: Sorry, I was zeroing in on the "tear down the HTTPS connection" piece. I don't think that's really required. All that seems required is for the protocol (be it upload, verify, etc) to be aborted, which might manifest in tearing down HTTPS, or it might not. For upload, aborting is likely akin to tearing down the connection, but it might not be. (Clients could, e.g., keep open persistent connections to the leader if they're uploading many things.)

draft-pda-core.md
```diff
@@ -291,11 +293,439 @@ Each PA task in this document is divided into three sub-protocols as follows.
...
+Each round of the protocol corresponds to an HTTP request and response. We
+assume that some transport layer security protocol (e.g., TLS or QUIC) is used
+between each pair of parties and that the server is always authenticated.
+
+[TODO: Decide how to authenticate the leader in leader-to-helper and
+aggregator-to-collector connections. One option is to use client certificates
+for TLS; another is to have the leader sign its messages directly, as in Prio
+v2.]
+
+[TODO: Decide the "content type" for HTTP requests.]
```
chris-wood, line 308: Let's use ["application/octet-stream"](https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types/Common_types).

draft-pda-core.md
```diff
@@ -291,11 +293,439 @@ Each PA task in this document is divided into three sub-protocols as follows.
...
+} PALeaderShare;
+~~~
+
+Note that the leader share is sent not encrypted.
+
+**Response.**
+The leader responds to well-formed requests to `[leader]/upload_finish` with
+status 200 and an empty payload. Malformed requests are handled as described in
+{{pa-error-common-aborts}}.
+
```
chris-wood, line 582:
```suggestion
[[OPEN ISSUE: consider client->leader upload replays, and what leaders should do to mitigate against them. Prohibit early data? Strike register? Something else?]]
```

draft-pda-core.md
chris-wood: I don't see why that's valuable or desired, since it just pushes state onto the server. (Left a separate comment for that, so will resolve this.)

draft-pda-core.md
chris-wood: It's not really being picky. It's just saying that the URLs must be non-empty.

draft-pda-core.md
```diff
@@ -291,11 +293,439 @@ Each PA task in this document is divided into three sub-protocols as follows.
...
+~~~
+
+The structure contains the PA task, the HPKE config id, and a sequence of
+*sub-requests*, each corresponding to a unique client report. Sub-requests are
+structured as follows:
+
+~~~
+struct {
+  opaque upload_id[16];
+  opaque enc<0..2^16-1>;
```
chris-wood, line 616:
```suggestion
  opaque enc<1..2^16-1>;
```

draft-pda-core.md
```diff
@@ -291,11 +293,439 @@ Each PA task in this document is divided into three sub-protocols as follows.
...
+struct {
+  PATask task;
+  opaque id[16];
+} PAUploadStartReq;
+~~~
+
+The first field, `task` corresponds to the PA task for which a report will be
+generated. The second, `id` is the client's *upload id*, which it chooses at
+random.
+
```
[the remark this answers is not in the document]
cjpatton: I'm fine with saying whatever here. (I don't actually think we need language about this yet.) I'll take the suggestion and we can overwrite it later if we choose.

draft-pda-core.md
```diff
@@ -291,11 +293,439 @@ Each PA task in this document is divided into three sub-protocols as follows.
...
+   We write this URL as `[leader]` below. (We write `[helper]` for a helper's
+   URL.)
+1. The client and leader can establish a leader-authenticated secure channel.
+1. The leader and each helper can establish a leader-authenticated secure
+   channel.
+1. Each helper has chosen an HPKE key pair.
+1. The aggregators agree on a set of PA tasks, as well as the PA protocol and
+   parameters used for each task.
+
+[TODO: It would be clearer to include a "pre-conditions" section prior to each
```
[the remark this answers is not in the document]
cjpatton: Let's do so, but in a follow-up PR.

draft-pda-core.md
```diff
@@ -291,11 +293,439 @@ Each PA task in this document is divided into three sub-protocols as follows.
...
+### Upload Start
+
+**Request.**
+The client sends a POST request to `[leader]/upload_start` with the following
+message:
+
+~~~
+struct {
+  PATask task;
+  opaque id[16];
```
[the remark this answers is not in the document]
cjpatton: (Getting rid of upload id, as it's motivation is tied to a particular instantiation of Prio.)

draft-pda-core.md
cjpatton: Good point! "Aborting" just means forgetting any state associated with the protocol ... it needn't result in dropping the HTTPS state altogether.

I guess we don't need to be explicit about what "abort" means right now. I just deleted this paragraph.

draft-pda-core.md
```diff
@@ -291,11 +293,439 @@ Each PA task in this document is divided into three sub-protocols as follows.
...
+
+We call this the helper's *key configation*. The key configuration is used to
+set up a base-mode HPKE context to use to derive symmetric keys for protecting
+the shares sent to the helper. The *config id*, `HpkeConfig.id`, is forwarded
+by the client to the helper, who uses this value to decide if it knows how to
+decrypt a share it receives.
+
+## Pre-conditions
+
+We assume the following conditions hold before the protocol begins:
```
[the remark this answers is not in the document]
cjpatton: That's right. Added.

draft-pda-core.md
```diff
@@ -291,11 +293,439 @@ Each PA task in this document is divided into three sub-protocols as follows.
...
+[TODO: Decide how to serialize protocol messages. We're using TLS syntax for
+now, but there's no reason to stick with it other than most folks are familiar
+with it.]
+
+The first field, `version` specifies the version of this document. The second
+field, `id` is an opaque identifier used by the clients, aggregators, and
+collector to uniquely identify the PA task at hand. We will call it the *task
+id* in the remainder.
+
+[TODO: Decide how the `PATask` is negotiated. Eventually this will be
```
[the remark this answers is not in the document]
cjpatton: s/negotiated/configured

draft-pda-core.md
cjpatton: Done.

draft-pda-core.md
```diff
@@ -291,11 +293,427 @@ Each PA task in this document is divided into three sub-protocols as follows.
...
+~~~
+struct {
+  uint16 version;
+  opaque id[16];
+} PATask;
+
+~~~
+[TODO: Decide how to serialize protocol messages. We're using TLS syntax for
+now, but there's no reason to stick with it other than most folks are familiar
+with it.]
```
ekr, line 336:
```suggestion
```

A reasonable point, but we are accumulating a lot of TODOs.

draft-pda-core.md
```diff
@@ -291,11 +293,427 @@ Each PA task in this document is divided into three sub-protocols as follows.
...
+
+The first field, `version` specifies the version of this document. The second
```
ekr, line 338: Do you mean like draft-blah-blah-01?

draft-pda-core.md
```diff
@@ -291,11 +293,427 @@ Each PA task in this document is divided into three sub-protocols as follows.
...
+field, `id` is an opaque identifier used by the clients, aggregators, and
+collector to uniquely identify the PA task at hand. We will call it the *task
+id* in the remainder.
+
+[TODO: Decide how the `PATask` is configured. Eventually this will be
+distributed, in an authenticated manner, from the collector the other parties.
+For now, we just assume this value is negotiated out-of-band.]
```
ekr, line 345: I'm not quite following what's going on here. Why not just give each task version its own UID? Or alternatively have version somewhere else.

draft-pda-core.md
```diff
@@ -291,11 +293,427 @@ Each PA task in this document is divided into three sub-protocols as follows.
...
+
+~~~
+struct {
+  PAParam param;
+  Url helper_urls<1..2^16-1>;
+  select (PAUploadStartResp.param.proto) {
+    case prio: PrioUploadStartResp;
+    case hits: HitsUploadStartResp;
+  }
+} PAUploadStartResp;
```
ekr, line 466: I don't quite understand the underlying assumptions here. If the client wants to upload some data, it's not going to be confused about whether the task involves heavy hitters or prio. What happens if it thinks it's doing prio and the leader responds with heavy hitterS?

draft-pda-core.md
```diff
@@ -291,11 +293,427 @@ Each PA task in this document is divided into three sub-protocols as follows.
...
+opaque Url<1..2^16-1>;
+~~~
+
+The message includes the URL of each helper and any protocol-specific parameters
+the client needs to generate its report. The leader's response to malformed
+requests is specified in {{pa-error-common-aborts}}.
+
+### Upload Finish
+
+For each URL `[helper]` in `PAUploadStartResp.helper_urls`, the client sends a
```
ekr, line 477: This seems like it has some obvious security problems. What stops the leader from just telling the client it runs all the helpers?

draft-pda-core.md
```diff
@@ -291,11 +293,427 @@ Each PA task in this document is divided into three sub-protocols as follows.
...
+follows:
+
+~~~
+struct {
+  PATask task;                 // Equal to PAUploadStartReq.task
+  uint8 helper_hpke_config_id; // Equal to HpkeConfig.id
+  Url helper_url;
+  PAHelperShare helper_share;
+  PALeaderShare leader_share;
+} PAUploadFinishReq;
```
ekr, line 539: Why are we replicating the leader_share for each helper?

ekr (changes requested):

draft-pda-core.md
chris-wood: Yeah, I assume so, similar to what we did for ECH.

draft-pda-core.md
chris-wood: Versioning the ID of a task seems fine, though not much different from each task ID being generated randomly. We can address this in a followup change. I'll clarify in the TODO.

chris-wood: The version here is "version of the protocol." But maybe I'm misunderstanding you?

draft-pda-core.md
chris-wood: This is to accommodate different parameters for the actual upload that are protocol specific. For example, if Hits or some future protocol requires per-upload data, that'd be in the HitsUploadStartResp.

ekr: Sure, but why is PAParam here? Almost nothing there can change.

draft-pda-core.md
chris-wood: Presumably clients will just check against their set of helpers or some local policy?

draft-pda-core.md
chris-wood: There's only one helper here, so no duplication.

ekr: "Otherwise, for each supported helper the client issues a POST request to `[leader]/upload_finish` with a payload
constructed as described below.

draft-pda-core.md
chris-wood: I guess it depends on how things are configured? The assumption now is that clients start with a Task and then learn the Params for that task during Upload, but if we assume that Params are also distributed out-of-band, then we can drop it here.

draft-pda-core.md
ekr: But then why do I need the server to tell me.

I think there's a philosophical issue here: We should avoid having the leader tell the client anything that could be inconsistent with things it already needs to know. If we are trying to detect mismatch, we should instead have a config structure where the server sends the hash (or it's included in an AAD or something). That way we detect mismatch but don't rely on the client checking each parameter. Otherwise, we get into situations where the client might inappropriately trust the leader.

draft-pda-core.md
chris-wood: Oh I see. Will flag as an open issue.

draft-pda-core.md
chris-wood: That's a good point -- if the baseline assumption is that the leader is not trusted to not collude, and clients can't reasonably check for this, then this needs to change. (Let's do that in the next PR.)

draft-pda-core.md
```diff
@@ -291,11 +293,424 @@ Each PA task in this document is divided into three sub-protocols as follows.
...
+supporting Prio-like proof systems in which the leader sends the client a
+"challenge" it uses to generate the proof. (See the SIMD construction of
+{{BBCp19}}, Section 5.2. Here, the "challenge" is a randomly generated field
+element.) Because the client/leader are using TLS, another way to implement this
+would be to export a secret from the key schedule.]
+
+### Upload Start
+
+The client sends a POST request to `[leader]/upload_start` with the following
+message:
```
tgeoghegan, line 441: The client obviously has to perform the `upload_finish` part of the protocol each time they want to submit a report, but it doesn't make sense to require them to run `upload_start` each time they submit metrics. We should specify client caching behavior for helper URL lists obtained from leaders and key configs obtained from helpers (see also my comment just below about using `HTTP GET` for cacheable, idempotent requests).

tgeoghegan (changes requested):

draft-pda-core.md
```diff
@@ -291,11 +293,427 @@ Each PA task in this document is divided into three sub-protocols as follows.
...
+(https//www.ietf.org/archive/id/draft-thomson-http-oblivious-01.html) allows
+clients to make multiple requests through a proxy, so these kinds of use cases
+should work.]
+
+[NOTE: @cjpatton: Breaking the upload phase into two requests is useful for
+supporting Prio-like proof systems in which the leader sends the client a
+"challenge" it uses to generate the proof. (See the SIMD construction of
+{{BBCp19}}, Section 5.2. Here, the "challenge" is a randomly generated field
+element.) Because the client/leader are using TLS, another way to implement this
+would be to export a secret from the key schedule.]
```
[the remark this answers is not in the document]
chris-wood: Yeah, we're unlikely to make this change, so we can probably just delete the note.

draft-pda-core.md
```diff
@@ -291,11 +293,425 @@ Each PA task in this document is divided into three sub-protocols as follows.
...
```
chris-wood, line 437:
```suggestion
element.)]
```

draft-pda-core.md
```diff
@@ -291,11 +293,427 @@ Each PA task in this document is divided into three sub-protocols as follows.
...
+### Upload Start
+
+The client sends a POST request to `[leader]/upload_start` with the following
+message:
+
+~~~
+struct {
+  PATask task;
+} PAUploadStartReq;
+~~~
```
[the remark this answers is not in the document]
chris-wood: > IIUC the leader doesn't actually do anything in response to the upload_start request except serve up some static configuration parameters.

In some cases, the server may supply per-upload data to be used in the finish message, so I think we ought to keep this to accommodate those protocols.

draft-pda-core.md
```diff
@@ -291,11 +293,427 @@ Each PA task in this document is divided into three sub-protocols as follows.
...
+
+The message includes the URL of each helper and any protocol-specific parameters
+the client needs to generate its report. The leader's response to malformed
+requests is specified in {{pa-error-common-aborts}}.
+
+### Upload Finish
+
+For each URL `[helper]` in `PAUploadStartResp.helper_urls`, the client sends a
+GET request to `[helper]/key_config`. The helper responds with status 200 and an
```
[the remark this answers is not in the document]
chris-wood: @ekr pointed out that client's should probably dictate the helpers (on behalf of the data collector), so this is likely to change.

draft-pda-core.md
```diff
@@ -291,11 +293,427 @@ Each PA task in this document is divided into three sub-protocols as follows.
...
+`HpkeConfig` message. Next, the client collects the set of helpers it will
```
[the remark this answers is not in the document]
chris-wood: I think a Cache-Control header seems appropriate, but that does seem like an implementation detail. If the key has expired then the client's share will fail to be aggregated. (Though we would need to make sure this happens during Verify.)

draft-pda-core.md
```diff
@@ -291,11 +293,427 @@ Each PA task in this document is divided into three sub-protocols as follows.
...
+* the client and helper failed to establish a secure, helper-authenticated
+  channel;
+* the GET request to the helper URL failed or didn't return a valid key config;
+  or
+* the key config specifies a KEM, KDF, or AEAD algorithm the client doesn't
+  recognize.
+
+If the set of supported helpers is empty, then the client aborts and alerts the
+leader with "no supported helpers". Otherwise, for each supported helper the
+client issues a POST request to `[leader]/upload_finish` with a payload
```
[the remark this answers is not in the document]
chris-wood: > Does this mean we only support 2-way secret sharing? I think that's OK, but we should make that constraint of the system more obvious.

Indeed -- this is limited to the two-server case. 

> Since all the reports are going to the same leader server, could they be combined into a single message, and hence a single request with a single state to be managed by the client and leader?

I don't see why not, though this seems like an optimization more than a functional change. 

> Is the client expected to secret share the input differently for each PAUploadFinishReq it constructs? If not, then PAUploadFinishReq.leader_share will be identical in each message, and since the leader share will be much larger than the helper share (at least in Prio where the helper's share is a PRNG seed), we should work out a way to absolve the client (which might be on an expensive and slow cellular network) of having to upload the leader share k times. 

Yes, I think that's expected. (That is, I assume the proof is fresh for each upload.)

draft-pda-core.md
```diff
@@ -291,11 +293,427 @@ Each PA task in this document is divided into three sub-protocols as follows.
...
+
+~~~
+enc, context = SetupBaseS(pk, [TODO])
+~~~
+
+where `pk` is the KEM public key encoded by `HpkeConfig.public_key`. The outputs
+are the helper's encapsulated context `enc` and the context `context`.
+
+[TODO: Decide what the info string should be. At a minimum it should include
+`PAParam` so that the PA parameters are implicitly authenticated by the helper.
```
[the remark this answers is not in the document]
chris-wood: Yeah, though we might want to fold in the entire PATask here anyway.

draft-pda-core.md
```diff
@@ -291,11 +293,427 @@ Each PA task in this document is divided into three sub-protocols as follows.
...
+when it decapsulates `enc`. We might consider using the hash of the transcript
+between the client and leader so far, i.e., `PAUploadStartReq`, ...,
+`PAUploadStartResp`, as this would be the most conservative thing. We just need
+to ensure that the helper can reproduce this transcript. This should be possible
+given the information it has.]
+
+Next, the client encodes its measurements as an input for the PA protocol. It
+then generates a validity proof for its input and uses `context` to split the
+input and proof into a *leader share* and a *helper share*, where the latter is
+protected by the HPKE context. Note that the details of each of these processing
```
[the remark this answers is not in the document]
chris-wood: That's right. Are you looking for a specific text change?

draft-pda-core.md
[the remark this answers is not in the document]
tgeoghegan: I suspect we can make this a GET without precluding those protocols, but we can pursue that question after merging this change: #48

draft-pda-core.md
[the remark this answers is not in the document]
tgeoghegan: Filed #49 to consider key rotation and how it fits into the protocol.

draft-pda-core.md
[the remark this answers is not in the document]
tgeoghegan: How about:
```suggestion
then generates a validity proof for its input, splits the input and proof into
a *leader share* and a *helper share*, using `context` to protect the latter.
Note that the details of each of these processing steps --- encode, prove,
split, and encrypt --- are specific to the PA protocol.
```

draft-pda-core.md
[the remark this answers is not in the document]
tgeoghegan: I think this is fine, so long as each of the `k` protocol runs the client initiates for a given report are totally independent from one another.

tgeoghegan (approved):
I wrote up my outstanding questions and comments in distinct issues in the name of unblocking progress here (#48, #49, #50). Otherwise this looks good to me.

PR #52, opened 2021-06-09
Add a TODO for resolving aggregate consistency

### LINKED ISSUE #51: Requirements for Collect (or, Revisiting the Pipeline Model)
At the moment, the spec thinks of a PA task as a one-way data-processing pipeline with three phases: upload, in which clients send inputs to the leader; verify, in which the leader and helper verify each input; and collect, in which the leader and helper aggregate inputs and push the output to the collector. The upload and verify phases are currently specified; our next task is to specify the collect phase.

This pipeline model is a perfectly fine way of thinking about Prio. (Although there are ways in which one might use Prio metrics that don't quite fit into this model; see #27.) However, HH doesn't fit into this model at all (see #44). It's apparently necessary to re-think the communication model for PA protocols. @chris-wood and I were kicking this around and wanted to share some thoughts.

First off, let's try to enumerate the relevant requirements for the PA framework.

1. The helper is stateless.
2. Aggregators (i.e., the leader and helper) enforce the batch size (#15): As long as one of the aggregators is honest, the collector should be unable to see outputs aggregated over a small amount of inputs.
3. Aggregators should never learn more information about outputs than the collector. Moreover, it should be possible to design a PA protocol in which neither the aggregator sees the output in the clear. (This is possible for Prio. In HH, the aggregators inevitably learn the set of candidate prefixes at each round, which leaks some information about the output.)
4. Implementing HH requires O(n) space for the leader, where n is the number of inputs. However, it should be possible for the leader to implement Prio with O(1) space.
5. The aggregators verify and aggregate inputs over multiple rounds. In each round, the collector specifies the parameters that will be used. (In HH, the parameters are the set of candidate prefixes.)
6. [Needs discussion] Not every PA protocol will verify inputs (#45).

We can't satisfy all of these requirements in the upload->verify->collect model. Instead, we might think of a PA task as two concurrent processes running asynchronously:
* **Upload process:** Clients *push* inputs to the leader via the upload start and upload finish requests.
* **Collect process:** Collector *pulls* outputs from the leader via a yet-to-be-specified collect request. The request carries the parameters the aggregators will use for verification and aggregation.

In this communication model, the collector may attempt to pull the output for a PA task whenever it likes. A pull will only be successful if the aggregators have verified and aggregated enough inputs.

If/when inputs are verified is up to the PA protocol, but must occur before the leader can respond to a collect request. (In Prio, reports can verified as soon as they're uploaded; in HH, verification can only occur once all the reports have been uploaded and the collector has issued a collect request.)

The helper will keep a long-term symmetric key for encrypting the state that's managed by the collector. This ensures privacy while allowing the helper to be stateless.

At the end of aggregation, each of the aggregators holds a share of the output. Before responding to a collect request, the leader requests the helper's output share. To ensure that the leader doesn't see the output in the clear, the helper encrypts its share under the collector's HPKE public key.

Open questions
1. Should collect requests be idempotent?
2. A collect request may take a long time. (For HH, the latency may be several minutes!) Is HTTP the right application?
3. How does the helper learn the collector's HPKE public key? The collector needs to prove ownership of the corresponding secret key in some way.

cjpatton:
Addresses #4 by adding a TODO. With #43 this issue has been solved in part: To ensure consistency in case of helper failure, the leader specifies a set of helpers, and the client runs the protocol with the leader and each helper. Each cohort produces an output, and the collector will need a way of deciding which output is "correct". Let's wait to answer this question until we close #51.

### DIFF
--- a/draft-pda-protocol.md
+++ b/draft-pda-protocol.md
@@ -573,6 +573,11 @@ The leader responds to well-formed requests to `[leader]/upload_finish` with
 status 200 and an empty body. Malformed requests are handled as described in
 {{pa-error-common-aborts}}.
 
+[TODO: Since we're running the protocol with multiple cohorts of aggregators, the
+collector needs to decide how to pick which cohort has the "correct" output.
+This might be the cohort with the largest batch of inputs. Figure this out once
+the collection is specified.]
+
 ## Verify {#pa-verify}
 
 [TODO: Add an illustration of this sub-protocol.]


tgeoghegan (approved):
LGTM but I suggest modifying the PR description so that GitHub doesn't close #51 when this gets merged.

PR #54, opened 2021-06-09
Define "Report"

### LINKED ISSUE #53: Terminology section should define "report"
The design document discusses "reports" which are a group of measurements being submitted to a DPA instantiation. There should be an entry in the "Terminology" section that explicitly defines a report and how it relates to "measurements" and "inputs".

cjpatton:
Closes #53.


tgeoghegan (approved):
Thanks for putting the list back in alphabetical order, too!

PR #59, opened 2021-06-21
Specify data collection

### LINKED ISSUE #44: Possible performance impact of decoupled verify / collect stages with heavy hitters
(moving a comment made in https://github.com/abetterinternet/prio-documents/pull/43 to an issue)
I have a high level concern with the architecture here, following up from our meeting on Wednesday related to heavy hitters.

**TL;DR evaluating a hierarchy at only a subset of levels (which are specified at query time) is not optimal for performance if the verification step does not have access to those specified levels.**

For heavy hitters, we are considering a model where records form a binary tree, where aggregates can be reported at different levels of the tree. There are a few possible ways this could go:

 1. For every record, verify and evaluate the entire binary tree (i.e. bit-by-bit evaluation)
 2. For every record, verify and evaluate only a subset of the tree (at the limit, this can look more like Prio which only evaluates the last level of the tree)
    2a. The subset of the tree to evaluate is configured within a particular record at record-creation time
    2b. The subset of the tree to evaluate is configured dynamically at collection / aggregation time

For our use-case (discussed in #18 (comment)), we are interested in (2b), for a few reasons:

(1) is potentially inefficient from both a performance and accuracy POV in that it requires more rounds, and additionally when used with differential privacy requires splitting a DP budget across all levels. Some prefixes might just not be interesting to the caller so we shouldn't waste privacy budget / compute on them.

(2a) is problematic in two dimensions.
-  Embedding configuration in records may compromise user privacy if the configurations can be set by an adversary colluding with one of the aggregators (e.g. you can leak log2(n choose k) bits of information if you allow any subset of k prefixes for an n-bit domain).
- Levels must be pre-specified at record-creation time, which is non-ideal if new information comes up at collection time which informs how aggregation should work (e.g. realizing you have fewer records than expected so it is better to focus on querying only up to a given prefix).

However, I think (2b) runs into a few issues with this architecture that separates verification from collection, in that without knowing the specific levels to evaluate, we run the risk of spending unnecessary compute verifying levels that will never end up being used. The protocol is more efficient (less computation, fewer rounds) if we can verify multiple levels at once (i.e. the levels the collector cares about).

Possible solutions:

1. Add a step where the collector and leader communicate prior to verification (but keep verification and collection separate)
2. More tightly couple "verify" and "collect" stages, so that verification can be done based on communication with the collector.
3. 
In our setting, the "leader" and "collector" roles are somewhat merged which aligns with option (1), but it seems like a good idea to make the general architecture robust to this use-case.

cc @schoppmp who is working on our C++ implementation of IDPFs.

cjpatton:
Closes #44 by having the collector pass parameters to the aggregators for each round of verification and aggregation.

Reframes the execution of a PA protocol in terms of clients *pushing* input shares to the aggregators and the collector *pulling* output shares from the aggregators. This is meant to satisfy the requirements discussed in #51.

### DIFF
--- a/draft-pda-protocol.md
+++ b/draft-pda-protocol.md
@@ -77,8 +77,10 @@ document.
 1. Aggregation function: The function computed over the users' inputs.
 1. Aggregator: An endpoint that runs the input-validation protocol and
    accumulates input shares.
-1. Batch size: The number of valid input shares accumulated by each aggregator
-   before computing the final output.
+1. Batch: A set of reports that are aggregated into an output.
+1. Batch size: The minimum size of a batch.
+1. Batch window: The minimum time difference between the oldest and newest
+   report in a batch (in seconds).
 1. Client: The endpoint from which a user sends data to be aggregated, e.g., a
    web browser.
 1. Collector: The endpoint that receives the output of the aggreagtion function.
@@ -114,6 +116,7 @@ document.
    input-validation protocol.
 1. Report: Uploaded to the leader from the client. A report contains the
    secret-shared and encrypted input and proof.
+1. Server: An aggregator or collector.
 
 # Overview {#overview}
 
@@ -257,52 +260,48 @@ proof showing that the encoded measurement is valid.
 
 ## Data flow
 
-[TODO: Rework this subsection so that all terms needed in the rest of the
-document are defined.]
+Each PA task consists of two sub-protocols, *upload* and *collect*, which are
+executed concurrently. Each sub-protocol consists of a sequence of HTTP requests
+made from one entity to another.
 
-[TODO: Explain that the downside of using secret sharing is that the protocol
-requires at least two servers to be online during the entire data aggregation
-process. To ameliorate this problem, we run the protocol in parallel with
-multiple pairs of aggregators.]
+~~~~
++-------------+ 1.      +-------------+
+|             +--------->             |
+|   Client    |         |   Leader    |
+|             |    +----+             |
++------+------+    | 2. +------^------+
+       | 1.        |           |
+       |           |           |
+       |           |           | 2.
++------v------+    |    +------+------+
+|             <----+    |             |
+|   Helper    |         |  Collector  |
+|             |         |             |
++-------------+         +-------------+
+~~~~
+{: #pa-topology title="Who makes requests to whom while executing a PA task."}
 
-Each PA task in this document is divided into three sub-protocols as follows.
+1. **Upload:** Each client assembles its measurements into an input for the
+   given PA protocol. It generates a proof of its input's validity and splits
+   the input and proof into two shares, one for the leader and another for a
+   helper. Rather than send each share to each aggregator directly, the client
+   sends a request to the helper to learn its public key, encrypts its share
+   under this public key, and sends the ciphertext to the leader. The leader's
+   share and the helper's encrypted share are referred to collectively as the
+   client's *report*.
+2. **Collect:** The collector makes one or more requests to the leader in order
+   to obtain the final output of the protocol. Before the output can be
+   computed, the aggregators (i.e, the leader and helpers) need to have verified
+   and aggregated a sufficient number of inputs. Depending on the PA protocol,
+   it may be possible for the aggregators to do so immediately when reports are
+   uploaded. (See {{prio}}.) However, in general it is necessary for them to
+   wait until (a) enough reports have been uploaded and (b) the collector has
+   made a request. (See {{hits}}.)
 
-~~~
-                    +------------+
-                    |            |
-                    |   Helper   <---------------+
-                    |            |               |
-                    +-----^------+               |
-                          |                      |
-                       2. |                   3. |
-                          |                      |
-+--------+  1.      +-----v------+         +-----v-----+
-|        +---------->            |      3. |           |
-| Client +---------->   Leader   +---------> Collector |
-|        +---------->            |         |           |
-+--------+          +-----^------+         +-----^-----+
-                          |                      |
-                       2. |                   3. |
-                          |                      |
-                    +-----V------+               |
-                    |            |               |
-                    |   Helper   <---------------+
-                    |            |
-                    +------------+
-~~~
-
-1. **Upload:** Each client assembles the measurements into an input for the given
-   PA protocol. It generates a proof of its input's validity and splits the
-   input and proof into two shares, one for the leader and another for a helper.
-   Rather than send each share to each aggregator directly, the client encrypts
-   each share under the helper's public key and sends the ciphertext to the
-   leader. The client repeats this procedure for each helper specified by the
-   leader.
-1. **Verify:** The leader initializes the input-validation protocol by sending
-   the encrypted shares to the aggregators. If the input is deemed valid, then
-   each aggregator stores its input share for processing later on.
-1. **Collect:** Finally, the collector interacts with the aggregators in order
-   to obtain the final output of the protocol.
+[TODO: Say that the protocol involves multiple helpers. The downside of using
+secret sharing is that the protocol requires at least two servers to be online
+during the entire data aggregation process. To ameliorate this problem, we run
+the protocol in parallel with multiple helpers.]
 
 # PA protocols {#pa}
 
@@ -314,16 +313,13 @@ content type of each request is "application/octet-stream". We assume that some
 transport layer security protocol (e.g., TLS or QUIC) is used between each pair
 of parties and that the server is always authenticated.
 
-[TODO: Decide how to authenticate the leader in leader-to-helper and
+[TODO: Decide how to provide mutual authentication in leader-to-helper and
 aggregator-to-collector connections. One option is to use client certificates
 for TLS; another is to have the leader sign its messages directly, as in Prio
 v2.]
 
 [TODO: @chris-wood suggested we specify APIs for producing and consuming each of
 the messages in the protocol. Specific PA protocols would implement this API.]
-
-[OPEN ISSUE: This needs to be reworked in order to account for protocols, like
-Hits, that involve multiple rounds of verify/collect. See issue #44.]
 
 **Error handling.**
 In this section, we will use the verbs "abort" and "alert with `[some error
@@ -334,35 +330,38 @@ elide the verbs altogether and refer to {{pa-error-common-aborts}}.
 ## Configuration {#pa-config}
 
 ### Tasks
+
 Each PA protocol is associated with a *PA task* that specifies the measurements
-that are to be collected and the protocol that will be used to collect them:
+that are to be collected:
 
 ~~~
 struct {
   uint16 version;
   opaque id[16];
 } PATask;
-
 ~~~
 
 The first field, `version` specifies the version of this document. The second
-field, `id` is an opaque identifier used by the clients, aggregators, and
+field, `id`, is an opaque identifier used by the clients, aggregators, and
 collector to uniquely identify the PA task at hand. We will call it the *task
 id* in the remainder.
 
 [TODO: Decide how the `PATask` is configured. Eventually this will be
-distributed, in an authenticated manner, from the collector the other parties.
-For now, we just assume this value is negotiated out-of-band.]
+distributed, in an authenticated manner, from the collector to the other
+parties. For now, we just assume this value is negotiated out-of-band.]
 
 ### Parameters
 
-Associated to each task is the set of PA protocol parameters. These are encoded
-by the `PAParam` structure, which also includes the task:
+Associated to each task are the *PA Parameters* encoded by the `PAParam`
+structure:
 
 ~~~
 struct {
   PATask task;
+  Url helper_urls<1..2^16-1>;
+  HpkeConfig collector_config;
   uint64 batch_size;
+  uint64 batch_window;
   PAProto proto;
   select (PAClientParam.proto) {
     case prio: PrioParam;
@@ -371,18 +370,35 @@ struct {
 } PAParam;
 
 enum { prio(0), hits(1) } PAProto;
+
+opaque Url<1..2^16-1>;
 ~~~
 
-The `batch_size` field encodes the *batch size*, the number of input shares
-accumulated by each aggregator before emitting its output share.  The `proto`
-field identifies the specific PA protocol in use. The rest of the structure
-contains any protocol-specific parameters that are required.
+* `task`: The PA task.
+* `helper_urls`: The helpers' endpoint URLs.
+* `collector_config`: The HPKE configuration of the collector (described in
+  {{hpke-config}}). [OPEN ISSUE: Maybe the collector's HPKE config should be
+  carried by the collect request?]
+* `batch_size`: The batch size, i.e., the minimum number of reports that are
+  aggregated into an output.
+* `batch_window`: The batch window, i.e., the minimum time difference between
+  the oldest and newest report in a batch.
+* `proto`: The PA protocol, e.g., Prio or Hits. The rest of the structure
+  contains the protocol specific parameters.
 
-### Helper key configuration
+[TODO: Decide how the `PAParam` is configured. Like the task, this will need to
+be authenticated by the collector.]
 
-Our protocol uses HPKE for public-key encryption {{!I-D.irtf-cfrg-hpke}}.  Each
+[OPEN ISSUE: Is there just one PAParam for each PATask, or do we want to allow
+PAParam to change overtime?]
+
+### HPKE key configuration {#hpke-config}
+
+Our protocol uses HPKE for public-key encryption {{!I-D.irtf-cfrg-hpke}}. Each
 helper specifies the HPKE public key that clients use to encrypt the helper's
-share. The public key and associated parameters are structured as follows:
+input share, and the collector specifies the HPKE public key that helpers use to
+encrypt output shares during collection. The public key and associated
+parameters are structured as follows:
 
 ~~~
 struct {
@@ -398,42 +414,59 @@ uint16 HpkeAeadId; // Defined in I-D.irtf-cfrg-hpke
 uint16 HpkeKemId;  // Defined in I-D.irtf-cfrg-hpke
 uint16 HpkeKdfId;  // Defined in I-D.irtf-cfrg-hpke
 ~~~
-[TODO: Decide whether to re-use the config from OHTTP/ECH. This would add
-support for multiple cipher suites.]
+[TODO: Decide whether to use the same config structure as OHTTP/ECH. This would
+add support for multiple cipher suites.]
 
-We call this the helper's *key configation*. The key configuration is used to
-set up a base-mode HPKE context to use to derive symmetric keys for protecting
-the shares sent to the helper. The *config id*, `HpkeConfig.id`, is forwarded
-by the client to the helper, who uses this value to decide if it knows how to
-decrypt a share it receives.
+We call this a *key configation*. The key configuration is used to set up a
+base-mode HPKE context to use to derive symmetric keys for protecting: (1) input
+shares sent from the client to the helper; or (2) output shares sent from the
+helper to the collector. The *config id*, `HpkeConfig.id`, is forwarded by the
+sender to the receiver to help the receiver decide if it knows the decryption
+key.
 
 ## Pre-conditions
 
-We assume the following conditions hold before the client begins uploading its
-data:
+We assume the following conditions hold before execution of any PA task begins:
 
-1. The client, aggregators, and collector are configured with a specific PA task.
-1. The client knows the URL of the leader endpoint, e.g., `example.com/metrics`.
-   We write this URL as `[leader]` below. (We write `[helper]` for a helper's
-   URL.)
-1. The client and leader can establish a leader-authenticated secure channel.
-1. The leader and each helper can establish a leader-authenticated secure
+1. The aggregators agree on a set of PA tasks, as well as the PA parameters
+   associated to each task.
+1. Each aggregator has a clock that is roughly in sync with true time, i.e.,
+   within the batch window specified by the PA parameters. (This is necessary to
+   prevent the same report from appearing in multiple batches.)
+1. Each client has selected a PA task for which it will upload a report.
+1. Each client knows the URL of the leader endpoint, e.g.,
+   `example.com/metrics`. We write this URL as `[leader]` below. (We write
+   `[helper]` for a helper's URL.)
+1. Each client and the leader can establish a leader-authenticated secure
    channel.
+1. The leader and each helper can establish a helper-authenticated secure
+   channel.
+1. The collector and leader can establish a leader-authenticated secure channel.
+1. The collector has chosen an HPKE key pair.
 1. Each helper has chosen an HPKE key pair.
-1. The aggregators agree on a set of PA tasks, as well as the PA protocol and
-   parameters used for each task.
+1. Each helper has chosen a key for an AEAD scheme.
 
 [TODO: It would be clearer to include a "pre-conditions" section prior to each
 "phase" of the protocol.]
 
 ## Upload {#pa-upload}
 
-[TODO: Add an illustration of this sub-protocol.]
+~~~~
+Client            Leader         Helper
+  |  upload start   |              |
+  <----------------->              |
+  |                 |  key config  |
+  <-------------------------------->
+  |  upload finish  |              |
+  <----------------->              |
+  v                 v              v
+~~~~
+{: #pa-upload-flow title="Flow of the upload process"}
 
-Uploading a report involves two requests to the leader. In the  *upload start
+Uploading a report involves two requests to the leader. In the *upload start
 request*, the client discovers the protocol-specific parameters it needs to
-generate the report, as well as the endpoint URL of each helper. In the *upload
-finish request*, it uploads its report to the leader.
+generate the report. In the *upload finish request*, it uploads its report to
+the leader.
 
 [NOTE: @acmiyaguchi pointed out that the use of an anonymizing proxy for
 uploading shares might be easier to implement if the "upload" phase involved a
@@ -448,7 +481,7 @@ supporting Prio-like proof systems in which the leader sends the client a
 {{BBCp19}}, Section 5.2. Here, the "challenge" is a randomly generated field
 element.)]
 
-### Upload Start
+### Upload Start Request
 
 The client sends a POST request to `[leader]/upload_start` with the following
 message:
@@ -468,21 +501,24 @@ status 200 and the following message:
 ~~~
 struct {
   PAParam param;
-  Url helper_urls<1..2^16-1>;
   select (PAUploadStartResp.param.proto) {
     case prio: PrioUploadStartResp;
     case hits: HitsUploadStartResp;
   }
 } PAUploadStartResp;
-
-opaque Url<1..2^16-1>;
 ~~~
 
-The message includes the URL of each helper and any protocol-specific parameters
-the client needs to generate its report. The leader's response to malformed
-requests is specified in {{pa-error-common-aborts}}.
+The message includes the PA parameters the client will use to generate its
+report. It also includes a protocol-specific message, e.g., PrioUploadStartResp,
+the semantics of which is up to the PA protocol.
 
-### Upload Finish
+The leader's response to malformed requests is specified in
+{{pa-error-common-aborts}}.
+
+[TODO: Maybe drop the protocol-specific message and make this an idempotent GET
+(see issue#48).]
+
+### Upload Finish Request
 
 For each URL `[helper]` in `PAUploadStartResp.helper_urls`, the client sends a
 GET request to `[helper]/key_config`. The helper responds with status 200 and an
@@ -501,6 +537,9 @@ leader with "no supported helpers". Otherwise, for each supported helper the
 client issues a POST request to `[leader]/upload_finish` with a payload
 constructed as described below.
 
+[OPEN ISSUE: Should the request URL encode the PA task? This would be necessary
+if we make `upload_start` an idempotent GET per issue#48.]
+
 [OPEN ISSUE: @chris-wood: Can't the leader determine if helpers are "online"?
 This seems to reveal information that's specific to clients. Imagine, for
 example, that clients are prohibited from talking to helpers but not the leader.
@@ -517,25 +556,23 @@ enc, context = SetupBaseS(pk, [TODO])
 where `pk` is the KEM public key encoded by `HpkeConfig.public_key`. The outputs
 are the helper's encapsulated context `enc` and the context `context`.
 
-[TODO: Decide what the info string should be. At a minimum it should include
-`PAParam` so that the PA parameters are implicitly authenticated by the helper.
-when it decapsulates `enc`. We might consider using the hash of the transcript
-between the client and leader so far, i.e., `PAUploadStartReq`, ...,
-`PAUploadStartResp`, as this would be the most conservative thing. We just need
-to ensure that the helper can reproduce this transcript. This should be possible
-given the information it has.]
-
 Next, the client encodes its measurements as an input for the PA protocol. It
 then generates a validity proof for its input and uses `context` to split the
 input and proof into a *leader share* and a *helper share*, where the latter is
 protected by the HPKE context. Note that the details of each of these processing
 steps --- encode, prove, split, and encrypt --- are specific to the PA protocol.
 
+[TODO: Fully specify encryption of the helper's share. We need to make sure we
+authenticate things like the PA parameters and the report timestamp. We need to
+decide what the info string for SetupBaseS() will be, as well as the aad for
+context.Seal(). The aad might be the entire "transcript" between the client and
+helper.]
+
 [OPEN ISSUE: Is it safe to generate the proof once, then secret-share between
 each (leader, helper) pair? Probably not in general, but maybe for Prio?]
 
 [OPEN ISSUE: allow server to send joint randomness in UploadStartResp, and then
-enforce uniqueness via double-spend state or something else]
+enforce uniqueness via double-spend state or something else (see issue#48).]
 
 The payload of the POST request to `[leader]/upload_finish` is structured as
 follows:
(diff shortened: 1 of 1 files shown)


tgeoghegan (commented):
Thanks for all this great work! Besides the various inline comments, I have some higher-level comments/questions, which I think we should discuss in person during tomorrow (Weds 6/23)'s design call.

### [leader]/collect calls block on multiple [helper]/aggregate calls

The diagram in 3.4. illustrates that each collect request can cause the leader to make _L_ aggregate calls. The PACollectReq contains `Url helper`, so that can't be an aggregate call to each of the _L_ helpers enumerated in the `PAParams.helper_urls`.

* Why the _L_ calls then? Is the idea to e.g. allow a leader computing output over 10,000 inputs to feed the inputs to helpers 1,000 at a time to limit how big individual leader->helper HTTP requests can be?
* What happens if an insufficient number of inputs (i.e., fewer than `PAParam.batch_size`) have been gathered when the leader receives the collect request? I assume the collector gets an error like "not yet; try again later".

In this draft, the collector's `[leader]/collect` request will block until the _L_ `[helper]/aggregate` requests return. This seems like a problem to me: that would require the leader and collector to keep a connection open for what could be a long time while all the helpers finish aggregating. Additionally, it'd be hard for leaders to maintain any kind of response time SLOs if they are beholden to the response times of all the helpers (though of course the overall system's response times and throughput will be at the mercy of any defective server regardless). If we have to define a means for the leader to tell the collector to back off and try again later anyway, then I think we could use that to avoid blocking for a long time.

Collectors would be configured to hit `[leader]/collect` at some fixed interval (say every 12 hours) (I suspect this is already implicitly the case). If an insufficient number of inputs have been received since the last time a `PACollectResp` was delivered, the collector gets `EAGAIN` and tries again after the fixed interval elapses. If the leader decides it has enough inputs to generate output shares, it synchronously returns a collect request ID to the collector, and then asynchronously makes all the necessary `[helper]/aggregate` calls. Subsequently, the collector can poll on `[leader]/collect/<collect request id>` at a more aggressive interval. If it polls on the collect request ID before the output shares are ready, it gets `EAGAIN`. Eventually, leader completes its transactions with the helper(s) and the next time collector polls `[leader]/collect/<collect request id>`, it provides a `PACollectResp` populated with the output shares.

### What state does the leader or collector have to maintain, and how does `state.Finished()` work?

Obviously it's protocol specific (prio vs. hits) but some concrete examples would help me understand.

In the Prio case, it's finished when the collector gets a `PACollectResp` that contains the leader and helper output shares, and then it can reassemble them into the output.

* Is the leader then responsible for keeping track of which reports/inputs were included in the most recent `PACollectResp` delivered to the collector, to avoid double-counting reports in the next `PACollectResp`? If so, how does the leader track this state -- does a date suffice (i.e. inputs that arrived after the recorded date should be included in the next `PACollectResp`)?

In the Hits case, then IIUC the successive collect requests sent by the collector represent traversing to the next level of a tree, so only the collector can know whether it feels it's gone deep enough.

* How does the leader know what set of inputs to include in the tree that the collector's query/collect request is applied to? Is it all the inputs gathered in the last _n_ hours, where _n_ is a Hits-specific portion of the PAParams? If so, is the leader responsible for keeping track of whether a given query has been applied to inputs from a given time window? Or is the query applied to all inputs gathered so far in the PATask? How does that affect data retention?

draft-pda-protocol.md
```diff
@@ -246,52 +246,52 @@ proof showing that the encoded measurement is valid.
...
+
+~~~~
+                   +------------+
+                   |            |
+    +-------------->   Helper   |
+    |              |            |
+    |              +-----^------+
+    |                    |
+    |                    |
+    | 1.                 |
```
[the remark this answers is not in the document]
cjpatton: The leader doesn't interact with the helper during upload, only during collect.

draft-pda-protocol.md
```diff
@@ -323,34 +320,36 @@ elide the verbs altogether and refer to {{pa-error-common-aborts}}.
...
-Associated to each task is the set of PA protocol parameters. These are encoded
-by the `PAParam` structure, which also includes the task:
+Associated to each task are the *PA Parameters* encoded by the `PAParam`
+structure:
 
 ~~~
 struct {
   PATask task;
+  Url helper_urls<1..2^16-1>;
+  HpkeConfig collector_config;
```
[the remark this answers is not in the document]
cjpatton: Yeah, some spelling of this is a good idea I think. I've left this as an OPEN ISSUE for now, since it's somewhat orthogonal. See below.

draft-pda-protocol.md
```diff
@@ -387,42 +402,54 @@ uint16 HpkeAeadId; // Defined in I-D.irtf-cfrg-hpke
...
 
-1. The client, aggregators, and collector are configured with a specific PA task.
-1. The client knows the URL of the leader endpoint, e.g., `example.com/metrics`.
-   We write this URL as `[leader]` below. (We write `[helper]` for a helper's
-   URL.)
-1. The client and leader can establish a leader-authenticated secure channel.
-1. The leader and each helper can establish a leader-authenticated secure
+1. The aggregators agree on a set of PA tasks, as well as the PA parameters
+   associated to each task.
+1. Each client is configured with a specific PA task.
```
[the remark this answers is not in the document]
cjpatton: Yes indeed. I don't this criterion is very clear. What we mean is "each client has selected a PA task for which it will upload a report". Fixed.

draft-pda-protocol.md
```diff
@@ -387,42 +402,54 @@ uint16 HpkeAeadId; // Defined in I-D.irtf-cfrg-hpke
...
-[TODO: Add an illustration of this sub-protocol.]
-
-Uploading a report involves two requests to the leader. In the  *upload start
+~~~~
+Client            Leader
+  |  upload start   |
+  <----------------->
+  |  upload finish  |
+  <----------------->
+  v                 v
```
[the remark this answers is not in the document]
cjpatton: Good call. Done.

draft-pda-protocol.md
```diff
@@ -562,116 +594,264 @@ PA protocol. The structure of the leader share is similarly protocol specific:
...
+    case prio: PrioLeaderInputShare;
+    case hits: HitsLeaderInputShare;
   };
-} PALeaderShare;
+} PALeaderInputShare;
 ~~~
 
-Note that the leader share is sent not encrypted.
+Note that only the helper's share is encrypted, and that the leader share is
+sent in plaintext.
```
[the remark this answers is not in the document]
cjpatton: Noted.

All we say about transport is that it should be a "server-authenticated secure channel". (I don't think we should be more specific than this.)

draft-pda-protocol.md
```diff
@@ -562,116 +594,264 @@ PA protocol. The structure of the leader share is similarly protocol specific:
...
+collector initializes local per-protocol state using the corresponding
+PAParam. This state object has the following member functions:
+
+- CreateRequest() -> PACollectReq: Creates a PACollectReq object, defined below,
+  that is sent to the leader to complete one iteration of the protocol.
+
+~~~
+struct {
+  PATask task;
+  Url helper;
```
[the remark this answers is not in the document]
cjpatton: > A `PAParam` can contain _n_ `helper_urls`, but a client may only select _k_ < _n_ of them to send reports to. So how does the collector know which `helper_urls` to use? I'm guessing that collectors are meant to send a `PACollectReq` for all _n_ helpers, but if _k_ << _n_ this seems wasteful. Are we assuming that _k_ is exactly or close to _n_?

The collector knows `helper_urls` because it's now specified by `PAParam`. (Note that this is change is new to this PR.) Nominally, _k_ would be close to _n_ (and _n_ would be be around 2 or 3).

> Supposing that some helper did not participate, what does the `PACollectResp` look like? Does it return an error (which should be described in 3.x. Common abort conditions) or does it provide a `PACollectResp` with empty `leader_share` and `encrypted_helper_share`?

Good question. (This is subsumed by the OPEN ISSUE below; I added your particular question as an example.) I think the right answer is that the leader would "abort" the request as described in {{pa-error}}. An important thing to note about this case is that it's (partially) recoverable, since the collector can try again with another helper.

> What about the case where `PACollectReq.helper` does not occur in `PAParam.helper_urls`? Does that get rejected by leader as an invalid message, or treated the same as a valid helper that did not participate?

This would be another "abort" I think.

draft-pda-protocol.md
```diff
@@ -562,116 +594,264 @@ PA protocol. The structure of the leader share is similarly protocol specific:
...
+  PAProto proto;
```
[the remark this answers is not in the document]
cjpatton: But there is no `PAParam` in this struct. I guess you could argue that the leader knows the `PAParam` since it has the task id. However, I tend to think that protocol messages should always be deserializaeable without extra context.

draft-pda-protocol.md
```diff
@@ -562,116 +594,264 @@ PA protocol. The structure of the leader share is similarly protocol specific:
...
+} PACollectReq;
+~~~
+
+- Update(resp: PACollectResp): Consumes an opaque PACollectResp object, defined
+  below, that is received from the leader in response to a PACollectReq.
+
+~~~
+struct {
+  PATask task;
+  PAProto proto;
```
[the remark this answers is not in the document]
cjpatton: I see what you mean. As I mentioned above, I think protocol messages should always be deserializeable without needing context. Another way of phrasing this is that protocol messages should be "self describing". Maybe this property isn't needed here.

draft-pda-protocol.md
```diff
@@ -562,116 +594,264 @@ PA protocol. The structure of the leader share is similarly protocol specific:
...
+  PAOutputShare leader_share;
+  opaque encrypted_helper_share;
```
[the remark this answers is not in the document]
cjpatton: The purpose of encryption here is to ensure that the aggregators don't learn the final output. (This is one of the requirements enumerated in #51.) This is a useful property to have if the collector "outsources" aggregation to a third-party leader. (See @ekr's comment on #51.)

draft-pda-protocol.md
```diff
@@ -562,116 +594,264 @@ PA protocol. The structure of the leader share is similarly protocol specific:
...
+HPKE config id. Let `[helper]` denote `PACollectReq.helper_url`. The leader
+sends a POST request to `[helper]/aggregate` with the following message:
 
 ~~~
 struct {
   PATask task;
-  uint8 hpke_config_id;
-  PAVerifyReq seq<1..2^24-1>;
-} PAVerifyReqSeq;
+  uint8 helper_hpke_config_id;
```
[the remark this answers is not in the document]
cjpatton: The config_id is meant to uniquely identify the HPKE config. We're using the same type for the config_id in ODoH and ECH, and it hasn't been a problem there. Usually the server will only have a small number of keys at any one time, e.g., 3 or 4.

cjpatton:
> ### [leader]/collect calls block on multiple [helper]/aggregate calls
> 
> The diagram in 3.4. illustrates that each collect request can cause the leader to make _L_ aggregate calls. The PACollectReq contains `Url helper`, so that can't be an aggregate call to each of the _L_ helpers enumerated in the `PAParams.helper_urls`.
> 
>     * Why the _L_ calls then? Is the idea to e.g. allow a leader computing output over 10,000 inputs to feed the inputs to helpers 1,000 at a time to limit how big individual leader->helper HTTP requests can be?

Each aggregate request corresponds to a round of the input validation protocol. For Prio, `L=1`; for Hits, `L = 2` I think. In general, I'd expect `L` to be small and constant.

>     * What happens if an insufficient number of inputs (i.e., fewer than `PAParam.batch_size`) have been gathered when the leader receives the collect request? I assume the collector gets an error like "not yet; try again later".

Yes.

> In this draft, the collector's `[leader]/collect` request will block until the _L_ `[helper]/aggregate` requests return. This seems like a problem to me: that would require the leader and collector to keep a connection open for what could be a long time while all the helpers finish aggregating. Additionally, it'd be hard for leaders to maintain any kind of response time SLOs if they are beholden to the response times of all the helpers (though of course the overall system's response times and throughput will be at the mercy of any defective server regardless). If we have to define a means for the leader to tell the collector to back off and try again later anyway, then I think we could use that to avoid blocking for a long time.

This is indeed a problem in general. For Hits the latency is unavoidable; for Prio we can avoid it because the aggregate requests can actually be made *as reports are uploade*. (Note that this is one of the requirements outline in #51. @acmiyaguchi and @ekr mention in the comments some ways we might deal with this. )


> Collectors would be configured to hit `[leader]/collect` at some fixed interval (say every 12 hours) (I suspect this is already implicitly the case). If an insufficient number of inputs have been received since the last time a `PACollectResp` was delivered, the collector gets `EAGAIN` and tries again after the fixed interval elapses. If the leader decides it has enough inputs to generate output shares, it synchronously returns a collect request ID to the collector, and then asynchronously makes all the necessary `[helper]/aggregate` calls. Subsequently, the collector can poll on `[leader]/collect/<collect request id>` at a more aggressive interval. If it polls on the collect request ID before the output shares are ready, it gets `EAGAIN`. Eventually, leader completes its transactions with the helper(s) and the next time collector polls `[leader]/collect/<collect request id>`, it provides a `PACollectResp` populated with the output shares.

That sounds right to me.

> ### What state does the leader or collector have to maintain, and how does `state.Finished()` work?
> 
> Obviously it's protocol specific (prio vs. hits) but some concrete examples would help me understand.

The collector's state encodes whatever information it needs from previous request to figure out what the next request will be. I envision this being pretty application specific... I'm not sure it has intrinsic value to either Prio or Hits.

> In the Prio case, it's finished when the collector gets a `PACollectResp` that contains the leader and helper output shares, and then it can reassemble them into the output.
> 
>     * Is the leader then responsible for keeping track of which reports/inputs were included in the most recent `PACollectResp` delivered to the collector, to avoid double-counting reports in the next `PACollectResp`? If so, how does the leader track this state -- does a date suffice (i.e. inputs that arrived after the recorded date should be included in the next `PACollectResp`)?

This question reflects a requirement that we worked out on the design call: that right now the "batch" of reports that are used to respond to a collect request is not well-defined. As a first stab at this, we will timestamp reports and have the collect request specify a time range that defines the batch.

> In the Hits case, then IIUC the successive collect requests sent by the collector represent traversing to the next level of a tree, so only the collector can know whether it feels it's gone deep enough.
> 
>     * How does the leader know what set of inputs to include in the tree that the collector's query/collect request is applied to? Is it all the inputs gathered in the last _n_ hours, where _n_ is a Hits-specific portion of the PAParams? If so, is the leader responsible for keeping track of whether a given query has been applied to inputs from a given time window? Or is the query applied to all inputs gathered so far in the PATask? How does that affect data retention?

See above.

cjpatton:
Rebased.

draft-pda-protocol.md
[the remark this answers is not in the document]
cjpatton: It also occurs to me that encrypting the output share is important to prevent the leader from doing Sybil attacks.

draft-pda-protocol.md
```diff
@@ -114,6 +116,7 @@ document.
    input-validation protocol.
 1. Report: Uploaded to the leader from the client. A report contains the
    secret-shared and encrypted input and proof.
```
tgeoghegan, line 118: Since we're getting into the -arity of things: does a report contain shares of a _single_ input or could several inputs be encoded into a single report?

draft-pda-protocol.md
```diff
@@ -375,23 +378,24 @@ enum { prio(0), hits(1) } PAProto;
...
-(described in {{hpke-config}}). The `batch_size` field encodes the *batch size*,
-the number of input shares accumulated by each aggregator before emitting its
-output share. The `proto` field identifies the specific PA protocol in use. The
-rest of the structure contains any protocol-specific parameters that are
-required.
+* `task`: The PA task.
+* `helper_urls`: The helpers' endpoint URLs.
+* `collector_config`: The HPKE configuration of the collector (described in
+  {{hpke-config}}). [OPEN ISSUE: Maybe the collector's HPKE config should be
+  carried by the collect request?]
```
tgeoghegan, line 385: This would allow the leader to impersonate the collector when it relays the HPKE config to the helper -- but as you say, it's an open issue so we don't have to solve it right now.

draft-pda-protocol.md
```diff
@@ -670,13 +675,19 @@ parameterized by a PAParam. At a high level, it proceeds as follows. First, the
...
 PAParam. This state object has the following member functions:
 
-- CreateRequest() -> PACollectReq: Creates a PACollectReq object, defined below,
-  that is sent to the leader to complete one iteration of the protocol.
+- CreateRequest(): Creates a PACollectReq object, defined below, that is sent to
+  the leader to complete one iteration of the protocol. Along with any
+  protocol-specific parameters, the message specifies a time interval
+  `[batch_start, batch_end)` that determines the batch of reports to be
+  aggregated. It must be that `batch_end - batch_start >= PAParam.batch_window`
+  and `batch_start` and `batch_end` are multiples of `PAParam.batch_window`.
```
tgeoghegan, line 683: I like making the collector specify the batch window, as I think it absolves us of specifying some tricky edge cases w.r.t. insufficient # of reports in the window (we can just say that it's up to the collector to try again with wider batch windows until they find one that works).

draft-pda-protocol.md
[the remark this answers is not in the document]
tgeoghegan: I had assumed that a helper wouldn't re-use HPKE configs across PATasks, but thinking about it further there's no reason to assume that.

tgeoghegan (approved):

draft-pda-protocol.md
cjpatton: I think one input per report. As has been said, it might be that the "input" is actually comprised of multiple measurements. The "input" is meant to capture the thing that's validated in a single run of the input validation protocol.

draft-pda-protocol.md
cjpatton: Actually, at the moment the helper is "pre-configured" with the PAParam (and thus the collectors HPKE config). But yeah, this something we have to work out.

draft-pda-protocol.md
```diff
@@ -257,52 +260,48 @@ proof showing that the encoded measurement is valid.
...
+
+1. **Upload:** Each client assembles its measurements into an input for the
+   given PA protocol. It generates a proof of its input's validity and splits
+   the input and proof into two shares, one for the leader and another for a
+   helper. Rather than send each share to each aggregator directly, the client
+   sends a request to the helper to learn its public key, encrypts its share
+   under this public key, and sends the ciphertext to the leader. The leader's
+   share and the helper's encrypted share are referred to collectively as the
+   client's *report*.
+2. **Collect:** The collector makes a request to the leader in order to obtain the
```
chris-wood, line 292:
```suggestion
2. **Collect:** The collector makes one or more requests to the leader in order to obtain the
```

draft-pda-protocol.md
```diff
@@ -257,52 +260,48 @@ proof showing that the encoded measurement is valid.
...
+   aggregated a sufficient number of inputs. Depending on the PA protocol, it
+   may be possible for the aggregators to do so immediately when reports are
+   uploaded. (See {{prio}}.) However, in general it is necessary for them to
+   wait until (a) enough reports have been uploaded and (b) the collector has
+   made a request. (See {{hits}}.)
+
+[TODO: Say that the protocol involves multiple helpers. The downside of using
+secret sharing is that the protocol requires at least two servers to be online
+during the entire data aggregation process. To ameliorate this problem, we run
+the protocol in parallel with multiple helpers.]
```
chris-wood, line 304: Can we move this to an issue and discuss separately? What does "we run the protocol in parallel with multiple helpers" actually mean? (Who is "we"?)

draft-pda-protocol.md
```diff
@@ -573,116 +609,281 @@ PA protocol. The structure of the leader share is similarly protocol specific:
...
+  |             <---------------->
+  |             |  output share  |
+  |             <---------------->
+  <-------------+                |
+  v             v                v
+~~~~
+{: #pa-collect-flow title="Flow of the collect process with N collect requests
+and L aggregate requests per collect request."}
+
+[TODO: Decide how the collector's request is authenticated.]
```
chris-wood, line 657:
```suggestion
[TODO: Decide if and how the collector's request is authenticated.]
```

draft-pda-protocol.md
chris-wood: Or maybe the collector can try again with a smaller window if it wants a different slice?

draft-pda-protocol.md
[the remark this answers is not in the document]
chris-wood: It seems a lot of complexity right now is coming from the fact that the collector allows for multiple helpers to be used by the client, and the client to make a selection. Can we just remove that option, and have the collector force which helpers to use?

draft-pda-protocol.md
```diff
@@ -371,18 +370,35 @@ struct {
...
-field identifies the specific PA protocol in use. The rest of the structure
-contains any protocol-specific parameters that are required.
+* `task`: The PA task.
+* `helper_urls`: The helpers' endpoint URLs.
+* `collector_config`: The HPKE configuration of the collector (described in
+  {{hpke-config}}). [OPEN ISSUE: Maybe the collector's HPKE config should be
+  carried by the collect request?]
+* `batch_size`: The batch size, i.e., the minimum number of reports that are
+  aggregated into an output.
+* `batch_window`: The batch window, i.e., the minimum time difference between
```
chris-wood, line 384: Are we concerned about skew between collectors and aggregators?

draft-pda-protocol.md
```diff
@@ -573,116 +609,281 @@ PA protocol. The structure of the leader share is similarly protocol specific:
...
-structured as follows:
+The structure contains the PA task, the helper's HPKE config id, an opaque
+*helper state* string, and a sequence of *sub-requests*, each corresponding to a
+unique client report.  Sub-requests are structured as follows:
 
 ~~~
 struct {
-  opaque enc<1..2^16-1>;
+  opaque enc<0..2^16-1>;
+  uint64 time; // UNIX timestamp (seconds) of the report.
```
chris-wood, line 764: This is generated by the leader, presumably? Is it set when the report is received from the client, or when aggregation is actually done during collection? Could the client set the timestamp in its encrypted share for the helper?

draft-pda-protocol.md
```diff
@@ -573,116 +609,281 @@ PA protocol. The structure of the leader share is similarly protocol specific:
...
 
-[OPEN ISSUE: encrypt and store Helper state at the Leader]
+##### Helper State
 
-## Collect {#pa-collect}
+The helper state is an optional parameter of an aggregate request that the can
+helper use to carry state accross requests. At least part of the state will
+usually need to be encrypted in order to protect user privacy. However, the
+details of precisely how the state is encrypted and the information that it
+carries is up to the implementation of the helper.
```
chris-wood, line 833:
```suggestion
carries is up to the helper implementation.
```

draft-pda-protocol.md
```diff
@@ -573,116 +609,281 @@ PA protocol. The structure of the leader share is similarly protocol specific:
...
+
+~~~
+struct {
+  PATask task;
+  opaque helper_state<0..2^16>;
+} PAOutputShareReq;
+~~~
+
+To respond to valid output share requests, the helper begins by decrypting its
+state. It then constructs its output share, which has the following structure:
```
chris-wood, line 853: Should we require that the helper check that the OutputShareReq is not sent before the batch limit is hit? (That is, should the helper track how many shares aggregated so far and error if the number is less than the size?)

draft-pda-protocol.md
```diff
@@ -573,116 +609,281 @@ PA protocol. The structure of the leader share is similarly protocol specific:
...
+encrypting it under the collector's HPKE public key.
+
+The leader sends a POST request to `[helper]/output_share` with the following
+message:
+
+~~~
+struct {
+  PATask task;
+  opaque helper_state<0..2^16>;
+} PAOutputShareReq;
```
chris-wood, line 849: Hmm, where does the batch window come into play here? Shouldn't it be the case that aggregate outputs are only produced for a given batch window once the size is hit? Do the collector's request parameters need to fold into this OutputShareReq?

chris-wood (commented):
First pass complete -- looking great!

draft-pda-protocol.md
cjpatton: Ah, good catch. I added the `batch_start` and `batch_end` parameters so that the helper can enforce the `batch_window` and `batch_size` parameters.

draft-pda-protocol.md
cjpatton: Yes indeed. I updated the text to make this check explicit.

draft-pda-protocol.md
cjpatton: We decided previously that multiple helpers would be used for resiliency. This has already been baked into the spec. This TODO is merely about motivating and describing this design choice. There are two issues that are impacted by this: #4 and #22.

draft-pda-protocol.md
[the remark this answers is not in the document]
cjpatton: This discussion seems regressive given that we've already decided that multiple helpers would be used to add resiliency. If we want to discuss changing course, then let's do so in an separate issue.

draft-pda-protocol.md
chris-wood: Why does a collector's request to get the aggregate output need to be authenticated? That seems like a deployment specific property.

draft-pda-protocol.md
chris-wood: I'll raise as a separate issue as I no longer thing this is necessary to bake into the protocol this way.

draft-pda-protocol.md
chris-wood: We should then (a) note that as a requirement and (b) note what happens if that does not happen.

draft-pda-protocol.md
chris-wood: What stops the leader from altering the timestamp, then? Should it be folded into the helper AAD, perhaps?

draft-pda-protocol.md
cjpatton: More precisely, I think all that's required for privacy is that each aggregator needs to have a clock that's roughly in sync with true time (within batch_window seconds). Added this to "Pre-conditions".

draft-pda-protocol.md
cjpatton: I don't think this is going to end up being deployment specific, but I'll concede the point for now.

draft-pda-protocol.md
cjpatton: Done -> https://github.com/abetterinternet/prio-documents/issues/68

draft-pda-protocol.md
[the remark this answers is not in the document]
cjpatton: Filed: https://github.com/abetterinternet/prio-documents/issues/68

draft-pda-protocol.md
cjpatton: Yup, that's the idea. There's a TODO for this in "Upload Finish Request".

draft-pda-protocol.md
```diff
@@ -573,116 +612,288 @@ PA protocol. The structure of the leader share is similarly protocol specific:
...
+~~~
+struct {
+  PATask task;
+  Url helper;
+  uint64 batch_start; // The beginning of the batch in UNIX time.
+  Uint64 batch_end;   // The end of the batch in UNIX time (exclusive).
+  PAProto proto;
+  select (PACollectReq.proto) {
+    case prio: PrioCollectReq;
+    case hits: HitsCollectReq;
```
ekr, line 693: Do we understand what the contents of this might be? Are they really going to be per-request? In particular, why do we need ```proto```. Each task should just have one proto

draft-pda-protocol.md
```diff
@@ -573,116 +612,288 @@ PA protocol. The structure of the leader share is similarly protocol specific:
...
+share (see {{pa-aggregate}}). In general, the procedure by which the aggregators
+verify and aggregate reports depends on parameters carried by the collect
+request. This procedure is described below in {{pa-aggregate}}.
+
+### Collect Request
+
+The collect protocol is an iterative procedure driven by the collector and
+parameterized by a PAParam. At a high level, it proceeds as follows. First, the
+collector initializes local per-protocol state using the corresponding
+PAParam. This state object has the following member functions:
```
ekr, line 675: I'd prefer we don't describe this in this object oriented fashion. It's incredibly hard to read. Just talk about the protocol messages, not member functions.

draft-pda-protocol.md
```diff
@@ -573,116 +612,288 @@ PA protocol. The structure of the leader share is similarly protocol specific:
...
+~~~
 
-[TODO: Add an illustration of this sub-protocol.]
+[OPEN ISSUE: Describe how intra-protocol errors yield collect errors (see
+issue#57). For example, how does a leader respond to a collect request if the
+helper drops out?]
+
+Each collect request involves one of the helpers specified by the PA parameters.
+If more than one helper is specified, the collector may issue the requests in
+any order.
```
ekr, line 733: As noted in in issue #68 this just seems like extra complexity. The helpers for any given task should be fixed.

draft-pda-protocol.md
```diff
@@ -573,116 +612,288 @@ PA protocol. The structure of the leader share is similarly protocol specific:
...
+  uint8 helper_hpke_config_id;
+  opaque helper_state<0..2^16>;
+  PAAggregateSubReq seq<1..2^24-1>;
+} PAAggregateReq;
 ~~~
 
-The structure contains the PA task, the HPKE config id, and a sequence of
-*sub-requests*, each corresponding to a unique client report. Sub-requests are
-structured as follows:
+The structure contains the PA task, the helper's HPKE config id, an opaque
```
ekr, line 760: What is the hpke_config_id used for? And why is it only one octet long?

draft-pda-protocol.md
```diff
@@ -573,116 +612,288 @@ PA protocol. The structure of the leader share is similarly protocol specific:
...
+*helper state* string, and a sequence of *sub-requests*, each corresponding to a
+unique client report.  Sub-requests are structured as follows:
 
 ~~~
 struct {
-  opaque enc<1..2^16-1>;
+  opaque enc<0..2^16-1>;
+  uint64 time; // UNIX timestamp (seconds) of the report.
   PAProto proto;
```
ekr, line 768: Again, why is ```proto``` being carried here? You can't change it on the fly.

draft-pda-protocol.md
```diff
@@ -573,116 +612,288 @@ PA protocol. The structure of the leader share is similarly protocol specific:
...
+for the PA parameters `PAParam` for which `PAAggregateReq.task.id ==
 PAParam.task.id`. Next, it looks up the HPKE config and corresponding secret key
-associated with `PAVerifyReq.key_config_id`. If not found, then it aborts and
-alerts the leader with "unrecognized key config". [NOTE: In this situation, the
-leader has no choice but to abort. This falls into the class of error scenarios
-that are addressable by running with multiple helpers.]
+associated with `PAAggregateReq.helper_hpke_config_id`. If not found, then it
+aborts and alerts the leader with "unrecognized key config". [NOTE: In this
+situation, the leader has no choice but to abort. This falls into the class of
+error scenarios that are addressable by running with multiple helpers.]
```
ekr, line 795: Why are you assuming that each report uses the same key_id. Suppose you had a situation in which you had a batch of a day's reports and the helper changed keys halfway through?

draft-pda-protocol.md
```diff
@@ -573,116 +612,288 @@ PA protocol. The structure of the leader share is similarly protocol specific:
...
+} PAAggregateSubResp;
 ~~~
 
-For each sub-request `PAVerifyReq`, the helper computes the corresponding
-sub-response as follows. It first checks that checks that `PAVerifyReq.proto ==
-PAParam.proto`. If not, it aborts and alerts the leader with "incorrect protocol
-for sub-request". Otherwise, It computes the HPKE context as
+The helper handles each sub-request `PAAggregateSubReq` as follows.  It first
+checks that checks that `PAAggregateSubReq.proto == PAParam.proto`. If not, it
+aborts and alerts the leader with "incorrect protocol for sub-request".
```
ekr, line 818: This could be avoided by just not having these fields.

draft-pda-protocol.md
```diff
@@ -573,116 +612,288 @@ PA protocol. The structure of the leader share is similarly protocol specific:
...
+~~~
+
+where `pk` is the HPKE public key encoded by the collector's HPKE key
+configuration and `output_share` is its serialized output share.
+
+It responds with the following message:
+
+~~~
+struct {
+  uint8 collector_hpke_config_id;
```
ekr, line 889: I don't see how this works and why is it just one byte?

draft-pda-protocol.md
cjpatton: - PrioCollectReq would minimally have some identifier for the validity circuit and, potentially, the field.
- HitsCollectReq might might specify one or two fields.

draft-pda-protocol.md
cjpatton: This was @chris-wood's idea. I think it's nice, but not strictly necessary. I added a TODO to decide whether to change this.

draft-pda-protocol.md
cjpatton: The structure of the rest of this message depends on the PA protocol. Maybe there's a better way to express this?

draft-pda-protocol.md
cjpatton: In that case, the leader would need to split the reports into two requests.

> Why are you assuming that each report uses the same key_id.

It seemed to me like it would be simpler if there's just one HPKE key used per request. 

> Suppose you had a situation in which you had a batch of a day's reports and the helper changed keys halfway through?

The leader would split the reports into two separate requests, one for the old key and another for the new.

draft-pda-protocol.md
cjpatton: This is basically the same thing we do in ECH: there is a 1-byte config identifier that's used to inform the decrypting party which HPKE key it needs to decrypt. It's a byte only because other HPKE applications have only a byte for the config id.

draft-pda-protocol.md
ekr: OK, I think it's a regression.

draft-pda-protocol.md
ekr: I don't see why any of this needs to be per-request. They should just be fixed per-task.

draft-pda-protocol.md
ekr: You don't need to carry that in the message. You just do like ```select (proto_in_use)``` or something and then note in the text that it's pointing to something external.

draft-pda-protocol.md
ekr: But that might end up with a tail that could never be aggregated because it was smaller than the minimal batch size.

draft-pda-protocol.md
cjpatton: For Hits this is also going to convey the set of candidate prefixes. It's therefore not fixed per request.

draft-pda-protocol.md
cjpatton: Ack, I'm sending a PR today to rip this out.

draft-pda-protocol.md
cjpatton: Added a TODO to address this later.

draft-pda-protocol.md
cjpatton: Not true, since the number of "sub-requests" isn't necessarily the same as the batch size. The leader makes several aggregate requests and not just one.

In any case, I'm good with relaxing this constraint. Added a TODO.

draft-pda-protocol.md
cjpatton: Ack, added a TODO.

draft-pda-protocol.md
ekr: Well, we should definitely factor out the things that are per-task and things that are per-collector-request.

draft-pda-protocol.md
cjpatton: [Oops, I thought I commented on this, but somehow the comment didn't make into the queue.] We're following the precedent set by ECH here: the config id is just a byte. It's used by the decrypting party to figure out which key to use.

Issue #32, opened 2021-04-23
Update field section criteria for Prio


cjpatton:
The current criteria:

> 1. **Field size.** How big the field needs to be depends on the type of data
>    being aggregated and how many users there are. The field size also impacts
>    the security level: the longer the validity circuit, the larger the field
>    needs to be in order to effectively detect malicious clients. Typically the 
>    soundness error (i.e., the probability of an invalid input being deemed valid
>    by the aggregators) will be 2n/(p-n), where n is the size of the input and p
>    is the prime modulus.
> 2. **Fast polynomial operations.** In order to make Prio practical, it's
>    important that implementations employ FFT to speed up polynomial operations.
>    In particular, the prime modulus p should be chosen so that (p-1) = 2^b * s 
>    for large b and odd s. Then g^s is a principle, 2^b-th root of unity (i.e.,
>    g^(s\*2^b) = 1), where g is the generator of the multiplicative subgroup.
>    This fact allows us to quickly evaluate and interpolate polynomials at 2^a-th
>    roots of unity for 1 <= a <= b.
> 3. **Highly composite subgroup.** Suppose that (p-1) = 2^b * s. It's best if s
>    is highly composite because this minimizes the number of multiplications
>    required to compute the inverse or apply Fermat's Little Theorem. (See
>    [BBG+19, Section 5.2].)
> 4. **Code optimization.** [[TODO: What properties of the field make
>    it possible to write faster implementations?]]

Criterion (3) doesn't seem that important to me any more. I've yet to think of a use case where we want to do inversion in a validity circuit, or for which we don't have an alternative to FLT. In any case, we already can get pretty good performance with a more standard, generic algorithm: https://github.com/abetterinternet/libprio-rs/blob/main/src/fp.rs#L156-L167

In addition to dropping (3), we should modify (2) by saying that `2^b` needs to be large enough to accommodate the largest proofs we would ever generate. `b=20` seems like a reasonable minimum.

cjpatton:
Closing this since it's no longer relevant to this document. (It will be relevant to https://github.com/cjpatton/vdaf, however.)

PR #60, opened 2021-06-21
Clean up Prio section and update field selection criteria

cjpatton:
- Removes the old Prio text, which by now is out-of-date.
- Addresses #32 by changing the criteria.

### DIFF
--- a/draft-pda-protocol.md
+++ b/draft-pda-protocol.md
@@ -26,6 +26,17 @@ informative:
       - ins: H. Corrigan-Gibbs
       - ins: D. Boneh
 
+  BBCp19:
+    title: "Zero-Knowledge Proofs on Secret-Shared Data via Fully Linear PCPs"
+    date: 2021-01-05
+    target: "https://eprint.iacr.org/2019/188"
+    author:
+      -ins: D. Boneh
+      -ins: E. Boyle
+      -ins: H. Corrigan-Gibbs
+      -ins: N. Gilboa
+      -ins: Y. Ishai
+
   BBCp21:
     title: "Lightweight Techniques for Private Heavy Hitters"
     date: 2021-01-05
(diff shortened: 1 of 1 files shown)


draft-pda-protocol.md
```diff
@@ -852,11 +754,13 @@ criteria:
    for large b and odd s. Then g^s is a principle, 2^b-th root of unity (i.e.,
    g^(s\*2^b) = 1), where g is the generator of the multiplicative subgroup.
    This fact allows us to quickly evaluate and interpolate polynomials at 2^a-th
-   roots of unity for 1 <= a <= b.
-1. **Highly composite subgroup.** Suppose that (p-1) = 2^b * s. It's best if s
-   is highly composite because this minimizes the number of multiplications
-   required to compute the inverse or apply Fermat's Little Theorem. (See
-   [BBG+19, Section 5.2].)
+   roots of unity for any 1 <= a <= b. Note that b imposes n upper bound on the
```
tgeoghegan, line 757:
```suggestion
   roots of unity for any 1 <= a <= b. Note that b imposes an upper bound on the
```

draft-pda-protocol.md
```diff
@@ -887,36 +795,7 @@ for s in range(0,1000,1):
         print(bits, p, GF(p).multiplicative_generator(), b, factor(s))
 ~~~
 
-### Key encapsulation
```
tgeoghegan, line 890: The idea here is that we use HPKE and thus don't need to specify key encapsulation beyond a reference to https://datatracker.ietf.org/doc/draft-irtf-cfrg-hpke/?

tgeoghegan (changes requested):

draft-pda-protocol.md
cjpatton: No, we'll need to specify this.

tgeoghegan (approved):

Issue #48, opened 2021-06-03
Consider making Upload Start an idempotent GET


tgeoghegan:
Per #43, the first step in the upload protocol is a POST to the leader server which returns information about helpers. In some yet to be specified cases, the response could include a challenge to be incorporated into reports. However I suspect that in most cases, the response will be static and so an idempotent, cacheable GET would be a better fit, especially since it would save clients a roundtrip to the helper on the hot path of submitting reports, and would make it possible for helpers to implement upload_start by putting some static content in a CDN.

HTTP GET requests don't have bodies, so we would have to figure out how to encode all the data in the current `PAUploadStartReq` into the `upload_start` URI, which could be done as either path fragments in the URI (i.e., `[leader]/upload_start/[task.version]/[task.id]`) or as query parameters (i.e., `[leader]/upload_start?version=[task.version];id=[task.id]`).

We would of course need to make sure that [the semantics of HTTP GET](https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/GET) don't preclude proof systems like the one alleged to exist in section 5.2 of "BBCp19".

tgeoghegan:
Also, because the current protocol uses a POST for upload_start, the implication is that for every report they submit, clients need to hit `upload_start` and then `upload_finish`. If we make `upload_start` idempotent and cacheable, then that request can be skipped in the majority of cases. I suppose the significance of this depends on how we expect clients to use PDA: are they going to be continuously streaming hundreds or thousands of reports into the leader as events of interest occur (in which case I think eliminating `upload_start` requests is interesting) or are they expected to accumulate several reports locally before submitting them all in one or a few requests?

cjpatton:
I agree this would be a good change, if we can swing it. It's notable that Prio v2 and HH don't require a random challenge from the leader. But as discussed, this is useful for other proof systems one might want to implement for Prio. For example, see the data type implemented here: https://github.com/abetterinternet/libprio-rs/pull/48. See below for an explanation.

There might be other ways to implement the challenge.
* One possibility is to derive the challenge from the TLS key schedule (or from whatever state the client and leader share for the secure channel). @ekr and @tgeoghegan  have both pointed out that adding this dependency on the underlying transport is problematic.
[...]

Issue #61, opened 2021-06-22
Align spec with ACME (was "Version all API endpoints")


tgeoghegan:
The protocol doc specifies HTTP endpoints like `[leader]/upload_start` or `[helper]/key_config`. We should make sure to insert a protocol version in the endpoints so that we can iterate on this in the future. i.e., `[leader]/v1/upload_start`.

cjpatton:
I was thinking the same thing :)

In fact, we might always include the version and task id in the request URI. For example:

example.com/[version]/[task_id]/upload_start

ekr:
I don't believe that this is considered current best practice. See, rather,
the structure used by ACME:

https://tools.ietf.org/rfcmarkup?doc=8555#section-7.1

martinthomson:
I don't know if this is recorded anywhere, but that is a bad way to version APIs.  Mostly because it is unnecessary.  Just change the `[leader]` part to something different.  Our services team has a simple policy: if you are going to make an incompatible change,  spin up an entirely new endpoint on a new domain name.  Compatible changes (new request or response formats for example) can use content negotiation.  And generally we use extensible formats (like JSON) that allow you to add protocol elements in a consistent fashion that result in predictable outcomes.

cjpatton:
The reason to do this is not just about versioning, since we also want to specify a "task id" for the request.

ekr:
On Wed, Jun 23, 2021 at 2:59 PM Christopher Patton ***@***.***>
wrote:

> The reason to do this is not just about versioning, since we also want to
> specify a "task id" for the request.
>

I'm not following how that changes the situation. Can you expand?

cjpatton:
I'm responding to @martinthomson's comment that the URI should not encode the version. I'm fine with that, I'm just not sure how to structure the request in a way that conveys the task id as well as the version. (I haven't yet looked at your link to the ACME spec, but I will)

martinthomson:
If the goal is to ensure that a client is able to communicate specific information to a server, in HTTP that's something we use headers for sometimes, but mostly it is what the content of a request is for.  I imagine here that we're talking about posting a bunch of stuff to a server, so there will be a payload into which you can put lots of goodies.

The message content might include a version, but versioning for content can and should use media types instead.  That too doesn't directly need an explicit version (not `application/my-protocol-v1` or far worse `application/my-protocol;v=1`), you just use a media type to ensure that if the client wants to send something else, the server won't get confused.

[...]

PR #63, opened 2021-06-23
Make upload start request a GET

cjpatton:
In #48 we did away with the requirement for interaction between the client and leader. This PR resolves the issue by making the upload start request an idempotent GET.

### DIFF
--- a/draft-pda-protocol.md
+++ b/draft-pda-protocol.md
@@ -483,20 +483,11 @@ element.)]
 
 ### Upload Start Request
 
-The client sends a POST request to `[leader]/upload_start` with the following
-message:
-
-~~~
-struct {
-  PATask task;
-} PAUploadStartReq;
-~~~
-
-The `task` field corresponds to the PA task for which a report will be
-generated.
-
-The leader responds to well-formed requests to `[leader]/upload_start` with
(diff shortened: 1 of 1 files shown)


draft-pda-protocol.md
```diff
@@ -450,29 +450,16 @@ element.)]
...
-
-The `task` field corresponds to the PA task for which a report will be
-generated.
-
-The leader responds to well-formed requests to `[leader]/upload_start` with
-status 200 and the following message:
+The client sends a GET request to `[leader]/[version]/[task_id]/upload_start`,
+where `[version] == PATask.version` and `[task_id] == PATask.id`. [TODO: Decide
+if adding the protocol version and task id to the URL is the right way to go.
+Instead, we may want to make this a POST with a body that specifies the PATask.
```
tgeoghegan, line 456: Our experience with caching OCSP for Let's Encrypt has been that caching POST responses is tricky because CDNs don't always support this as well as caching GETs. However we can resolve that question in #61.

tgeoghegan (approved):

cjpatton:
Based on this week's design call (2021/7/7), "upload start" is likely going away altogether.

Issue #68, opened 2021-07-01
Multiple helpers (was "Should multipler helpers be baked into the core protocol?")


cjpatton:
The reason for having multiple helpers is to (partially) recover the protocol output in case a helper goes offline. (See #4.) We currently bake this into the protocol by associating a set of helper URLs with the PPM parameters. This complicates the protocol in two ways:
1. When uploading reports, the client has to pick the set of helpers it will upload reports for (see #50).
2. When collecting outputs (see #59), the collector has to specify which helper the leader will engage with.

When reviewing #59, @chris-wood pointed out that we can simplify the protocol by having one helper per task. If deployments want to add redundancy, they can do so by spinning up more tasks with identical parameter except with a different helper for each task.

ekr:
I don't think that high availability is a good reason to allow for multiple helpers. If you want HA, stand up multiple instances in different data centers, etc.

The reason to allow for multiple helpers is to provide a higher level of assurance, namely to require >2 entities to collude in order to violate user privacy. It might be that we could defer that, though it's not clear to me why a fixed set of 2 helpers is more complicated than a fixed set of 1 helper.

WRT #50 I don't think that the leader should e telling the client about helpers at all. See #65

cjpatton:
> I don't think that high availability is a good reason to allow for multiple helpers. If you want HA, stand up multiple instances in different data centers, etc.

When we discussed this last we agreed that multiple helpers would be used for resiliency and not for weakening the trust model. Making this change at this point is fine, but I just want to note that it's a course correction and want to make sure we're all on the same page about.

> The reason to allow for multiple helpers is to provide a higher level of assurance, namely to require >2 entities to collude in order to violate user privacy. It might be that we could defer that, though it's not clear to me why a fixed set of 2 helpers is more complicated than a fixed set of 1 helper.

There are two ways in which this complicates things:

1. Not every PA protocol is well-defined for more than one helper. In particular, heavy hitters, as specified in [1], is defined for exactly two servers (one leader and one helper).
2. Input validation gets more complicated. For Prio, the leader will have to make a request to each helper, combine the results and make the validity decision, then make a second request to each helper in order to disseminate the decision. Right now Prio can be implemented with just one request to the helper.

With these wrinkles in mind, I'd be in favor of allowing multiple helpers to be used this way.

[1] https://eprint.iacr.org/2021/017

ekr:
[...]

PR #70, opened 2021-07-07
Don't have the PA params specify multiple helpers

cjpatton:
Partially addresses #68. (We also need to decide if/how to allow a particular PA protocol to use multiple helpers for privacy.)

Right now, multiple helpers are used for redundancy in case one of the
helpers dropped out. This removes this mechanism from the core protocol.

### DIFF
--- a/draft-pda-protocol.md
+++ b/draft-pda-protocol.md
@@ -291,17 +291,12 @@ made from one entity to another.
    client's *report*.
 2. **Collect:** The collector makes one or more requests to the leader in order
    to obtain the final output of the protocol. Before the output can be
-   computed, the aggregators (i.e, the leader and helpers) need to have verified
+   computed, the aggregators (i.e, the leader and helper) need to have verified
    and aggregated a sufficient number of inputs. Depending on the PA protocol,
    it may be possible for the aggregators to do so immediately when reports are
    uploaded. (See {{prio}}.) However, in general it is necessary for them to
    wait until (a) enough reports have been uploaded and (b) the collector has
    made a request. (See {{hits}}.)
-
-[TODO: Say that the protocol involves multiple helpers. The downside of using
-secret sharing is that the protocol requires at least two servers to be online
-during the entire data aggregation process. To ameliorate this problem, we run
-the protocol in parallel with multiple helpers.]
 
 # PA protocols {#pa}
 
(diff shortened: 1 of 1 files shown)


cjpatton:
Squashed and rebased.

draft-pda-protocol.md
```diff
@@ -504,10 +499,11 @@ The leader's response to malformed requests is specified in
...
 
-For each URL `[helper]` in `PAUploadStartResp.helper_urls`, the client sends a
-GET request to `[helper]/key_config`. The helper responds with status 200 and an
-`HpkeConfig` message. Next, the client collects the set of helpers it will
-upload shares to. It ignores a helper if:
+Let `[helper]` denote the helper URL encoded by
+`PAUploadStartResp.param.helper_url`.  When a client sends a GET request to
+`[helper]/key_config`, the helper responds with status 200 and an `HpkeConfig`
+message. The client aborts and alerts the leader with "unsupported helper" if
+any of the following happen:
```
chris-wood, line 506:
```suggestion
message. The client aborts if any of the following happen:
```

chris-wood (approved):
LGTM. Multiple helpers "for privacy" seems like it's inherently a per-protocol thing (Prio supports it, Hits does not). If and when we decide to address that, should we move this to the protocol-specific slot in PAParam?

cjpatton:
> LGTM. Multiple helpers "for privacy" seems like it's inherently a per-protocol thing (Prio supports it, Hits does not). If and when we decide to address that, should we move this to the protocol-specific slot in PAParam?

That's a good idea,

tgeoghegan (approved):

ekr:
> LGTM. Multiple helpers "for privacy" seems like it's inherently a per-protocol thing (Prio supports it, Hits does not). If and when we decide to address that, should we move this to the protocol-specific slot in PAParam?

Hmm... Do we need to decide. Per our discussion on Wednesday, if the clients learn this out of band then we won't need this field at all because clients will get it separately and helpers don't need to know what the other helpers are, right?

chris-wood:
> Hmm... Do we need to decide. Per our discussion on Wednesday, if the clients learn this out of band then we won't need this field at all because clients will get it separately and helpers don't need to know what the other helpers are, right?

My takeaway is that the parameters must contain all input needed to run the protocol, including the helpers. And all of this is configured out of band. You could punt everything except for the cryptographic properties out of the parameters structure, forcing clients to get public keys out of band as well, which would be fine. It's just different. I think both work just fine, but I lean towards having params specify everything. It just seems simpler to me.

PR #73, opened 2021-07-07
Encrypt the leader's share as well

### LINKED ISSUE #69: Encrypt to the leader as well
ACME originally had this "just trust TLS" structure and then ended up with end-to-end signatures to support designs in which you might have some kind of TLS decryption device on-path (either a forward proxy on the client or a reverse proxy on the server). ISTM that similar concerns apply here, and we should just HPKE to the leader. Note that this would also simplify the protocol because you wouldn't need to have parallel structures of leader/helper and could just have:

```
PAEncryptedInputShare input_share<..2^24-1>;
```

With the convention that leader was 0.

cjpatton:
Based on #70 (review that first).
Closes #69.

### DIFF
--- a/draft-pda-protocol.md
+++ b/draft-pda-protocol.md
@@ -284,11 +284,10 @@ made from one entity to another.
 1. **Upload:** Each client assembles its measurements into an input for the
    given PA protocol. It generates a proof of its input's validity and splits
    the input and proof into two shares, one for the leader and another for a
-   helper. Rather than send each share to each aggregator directly, the client
-   sends a request to the helper to learn its public key, encrypts its share
-   under this public key, and sends the ciphertext to the leader. The leader's
-   share and the helper's encrypted share are referred to collectively as the
-   client's *report*.
+   helper. The client then encrypts the leader's share and helper's share under,
+   respectively, the leader's public key and the helper's public key. (The keys
+   are obtained by making requests to the leader and helper.) Finally, the
+   client uploads the encrypted shares to the leader.
 2. **Collect:** The collector makes one or more requests to the leader in order
    to obtain the final output of the protocol. Before the output can be
    computed, the aggregators (i.e, the leader and helper) need to have verified
@@ -323,6 +322,9 @@ In this section, we will use the verbs "abort" and "alert with `[some error
 message]`" to describe how protocol participants react to various error
 conditions. The behavior is specified in {{pa-error}}. For common errors, we may
 elide the verbs altogether and refer to {{pa-error-common-aborts}}.
+
+[TODO: Fix the bounds for length-prefixed parameters in protocol messages.
+(E.g., `<23..479>` instead of `<1..2^16-1>`.)]
 
 ## Configuration {#pa-config}
 
@@ -393,8 +395,8 @@ PAParam to change overtime?]
 ### HPKE key configuration {#hpke-config}
 
 Our protocol uses HPKE for public-key encryption {{!I-D.irtf-cfrg-hpke}}. Each
-helper specifies the HPKE public key that clients use to encrypt the helper's
-input share, and the collector specifies the HPKE public key that helpers use to
+aggregator specifies the HPKE public key that clients use to encrypt its input
+share, and the collector specifies the HPKE public key that helpers use to
 encrypt output shares during collection. The public key and associated
 parameters are structured as follows:
 
@@ -417,7 +419,7 @@ add support for multiple cipher suites.]
 
 We call this a *key configation*. The key configuration is used to set up a
 base-mode HPKE context to use to derive symmetric keys for protecting: (1) input
-shares sent from the client to the helper; or (2) output shares sent from the
+shares sent from the client to an aggregator; or (2) output shares sent from the
 helper to the collector. The *config id*, `HpkeConfig.id`, is forwarded by the
 sender to the receiver to help the receiver decide if it knows the decryption
 key.
@@ -441,7 +443,7 @@ We assume the following conditions hold before execution of any PA task begins:
    channel.
 1. The collector and leader can establish a leader-authenticated secure channel.
 1. The collector has chosen an HPKE configuration and corresponding secret key.
-1. The helper has chosen an HPKE configuration and corresponding secret key.
+1. Each aggregator has chosen an HPKE configuration and corresponding secret key.
 
 [TODO: It would be clearer to include a "pre-conditions" section prior to each
 "phase" of the protocol.]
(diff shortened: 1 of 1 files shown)


draft-pda-protocol.md
```diff
@@ -509,17 +508,16 @@ The leader's response to malformed requests is specified in
...
-any of the following happen:
-
-* the client and helper failed to establish a secure, helper-authenticated
-  channel;
-* the GET request to the helper URL failed or didn't return a valid key config;
-  or
-* the key config specifies a KEM, KDF, or AEAD algorithm the client doesn't
+message. Similarly, when a client sends a GET request to `[leader]/key_config`,
+the leader responds with status 200 and an `HpkeConfig`. The client aborts and
+alerts the leader with "unsupported helper" if any of the following happen:
```
ekr, line 513: Hmm.... So how does this work if the client can't connect to the leader.

draft-pda-protocol.md
```diff
@@ -529,81 +527,79 @@ example, that clients are prohibited from talking to helpers but not the leader.
...
+is structured as follows:
 
 ~~~
-enc, context = SetupBaseS(pk, [TODO])
+struct {
+  PATask task; // Equal to PAUploadStartReq.task
+  uint64 time; // UNIX time (in seconds).
+  PAEncryptedInputShare leader_input_share;
+  PAEncryptedInputShare helper_input_share;
+} PAUploadFinishReq;
```
ekr, line 539: This seems like a good opportunity to just be:

```
PAEncryptedInputShare shares<...>;
```

That is like the least we can do to lay the groundwork for >1 helper.

draft-pda-protocol.md
```diff
@@ -529,81 +527,79 @@ example, that clients are prohibited from talking to helpers but not the leader.
...
+leader_enc, leader_context = SetupBaseS(leader_pk, [TODO])
+helper_enc, helper_context = SetupBaseS(helper_pk, [TODO])
 ~~~
 
-Field `enc` encodes the helper's encapsulated HPKE context. The remainder of the
-structure contains the share itself, the structure of which is specific to the
-PA protocol. The structure of the leader share is similarly protocol specific:
+`leader_pk` and `helper_pk` are, respectively, the leader and helper public key.
+`leader_enc` and `helper_enc` are the encapsulated HPKE context of each
+aggregator. Finally, the payloads of the encrypted input shares are computed as
```
ekr, line 585: This is another opportunity to avoid duplication, but just talking about aggregators.

ekr (changes requested):

draft-pda-protocol.md
cjpatton: Done.

draft-pda-protocol.md
cjpatton: Done.

draft-pda-protocol.md
```diff
@@ -559,36 +559,39 @@ struct {
...
+    case prio, helper: PrioHelperInputShare;
+    case hits, leader: HitsLeaderInputShare;
+    case hits, helper: HitsHelperInputShare;
   }
 } PAInputShare;
 ~~~
 
-Next, the client generates an HPKE {{!I-D.irtf-cfrg-hpke}} context for the
-leader and helper by running
+[TODO: Not sure if the use of `aggregator_type` above is syntactically correct.]
```
ekr, line 576: Yeah, you can't do this in the TLS syntax. I would just say this stuff verbally
"Of the appropriate share type for the protocol and role" down on line 593 below.

ekr (approved):

draft-pda-protocol.md
cjpatton: Removed the PAInputShare structure altogether and clarified above that the the format of the input share depends on the protocol and aggreagtor role.

draft-pda-protocol.md
```diff
@@ -529,81 +527,69 @@ example, that clients are prohibited from talking to helpers but not the leader.
...
 ~~~
 struct {
-  PATask task;                 // Equal to PAUploadStartReq.task
-  uint64 time;                 // UNIX time (in seconds).
-  uint8 helper_hpke_config_id; // Equal to HpkeConfig.id
-  PAHelperInputShare helper_input_share;
-  PALeaderInputShare leader_input_share;
+  PATask task; // Equal to PAUploadStartReq.task
+  uint64 time; // UNIX time (in seconds).
+  PAEncryptedInputShare encrypted_input_shares<1..2^16-1>;
```
ekr, line 537: nit: isn't this 7..2^16-1:

1 for config id
2 each for the lengths of enc and payload
1 each for the minimum size of length and payload.

ekr (approved):

draft-pda-protocol.md
cjpatton: Yeah, well, we'll need to fix up the bounds elsewhere as well. I'll add a TODO to handle this later, once the protocol gets closer to settled.

draft-pda-protocol.md
cjpatton: What do you mean by "feedback channel"?

chris-wood: I assume he means the alert from client to leader. What is the leader actually supposed to do with that information?

cjpatton: Ah yes, good call. I removed the alert.

cjpatton:
Squashed and rebased.

tgeoghegan (approved):

draft-pda-protocol.md
```diff
@@ -522,78 +523,69 @@ example, that clients are prohibited from talking to helpers but not the leader.
...
+is structured as follows:
 
 ~~~
 struct {
-  PATask task;                 // Equal to PAUploadStartReq.task
-  uint64 time;                 // UNIX time (in seconds).
-  uint8 helper_hpke_config_id; // Equal to HpkeConfig.id
-  PAHelperInputShare helper_input_share;
-  PALeaderInputShare leader_input_share;
+  PATask task; // Equal to PAUploadStartReq.task
```
chris-wood, line 531:
```suggestion
  PATask task;
```
(There is no more UploadStartReq, right?)

draft-pda-protocol.md
```diff
@@ -522,78 +523,69 @@ example, that clients are prohibited from talking to helpers but not the leader.
...
+payload = context.Seal(input_share, [TODO])
 ~~~
 
-Note that only the helper's share is encrypted, and that the leader share is
-sent in plaintext. (Confidentiality of the leader's share is provided by the
-server-authenticated secure channel over which the request is made.)
+where `input_share` is the aggregator's input share.
+
+[TODO: Fully specify encryption of the shares. We need to make sure we
+authenticate things like the PA parameters and the report timestamp. We need to
```
chris-wood, line 579: I filed #75 to follow up on this, but: what good is authenticating the timestamp if the client can just lie about it?

chris-wood (approved):
LGTM with #75 spun out!

draft-pda-protocol.md
cjpatton: Authenticating the timestamp is for the client's sake. It's meant to ensure its report is only used for a single batch.

draft-pda-protocol.md
cjpatton: There is, as of this PR.

draft-pda-protocol.md
chris-wood: Hmm, I don't really see how it helps the client at all. Can you elaborate?

cjpatton: The idea was discussed a few weeks ago, but hasn't been fully fleshed out. Basically the PA parameters specify a batch window (batch_start, batch_end). To check that a report should be added to a batch, an aggregator decrypts its input share. Decryption will fail if the reported timestamp doesn't match the timestamp generated by the client. This ensures that an honest aggregator only includes an input share in a batch if the timestamp is in the batch window.

chris-wood: I get that, but how does this benefit the _client_? (We should just take this to #75.)

draft-pda-protocol.md
chris-wood: I may be missing something then. I don't see PAUploadStartReq defined in this PR's diff?

draft-pda-protocol.md
cjpatton: Done.

cjpatton:
Rebased.

draft-pda-protocol.md
cjpatton: Ohhh right! The message got removed, but the "upload start" flow still exists. Good call :)

cjpatton:
Squashed.

PR #74, opened 2021-07-07
Add a length field to PAParam

### LINKED ISSUE #66: PAParam needs a length field
PAParam now has:

```
struct {
  PATask task;
  uint64 batch_size;
  PAProto proto;
  select (PAClientParam.proto) {
    case prio: PrioParam;
    case hits: HitsParam;
  }
} PAParam;
```
If you introduce a new protocol, this structure will not be parseable by anyone who doesn't know about it. The fix is to put a length field here, typically before the ```select```

cjpatton:
Closes #66.

This ensures that this structure can be parsed once we add more PAProto variants.

### DIFF
--- a/draft-pda-protocol.md
+++ b/draft-pda-protocol.md
@@ -363,6 +363,7 @@ struct {
   uint64 batch_size;
   uint64 batch_window;
   PAProto proto;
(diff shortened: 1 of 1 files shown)


ekr (approved):
LGTM

PR #78, opened 2021-07-12
Allow multiple reports in the same upload request

### LINKED ISSUE #64: Supporting batch uploads from the client (and routing reports through the collector)
In today's design call we discussed the collector receiving encrypted reports from clients and forwarding them to the leader. This aligns with the design we have in the WICG with some of the reasoning documented [here](https://github.com/WICG/conversion-measurement-api/blob/main/SERVICE.md#sending-data-directly-to-the-helper-servers).

I also brought this up for discussion on our regular calls in the WICG ([minutes](https://github.com/WICG/conversion-measurement-api/blob/main/meetings/2021-06-28-minutes.md)). Where there was some agreement that this was a good idea.

### Pros / Cons of routing reports through the collector
These are probably non-exhaustive.
Pros:

- Doesn't require aggregation servers to be highly online / available
- Supports graceful failure ("If something goes wrong, we could re-query")
- Gives some indication that "the API is working" on the server without needing to wait until query time, or via some other side-channel.
- Distributes state out of the aggregation servers (% protection from replay attacks). Arguably this aligns more in our API with who "owns" the data at some fundamental level.
- Adds query flexibility "for free" without explicitly adding support in the aggregation servers by allowing querying subsets of reports (for instance)
- Allows support for some level of report authentication by the collector, who (in our model) is the entity that is best in the position to validate reports. This could be done completely outside the protocol.

Cons:

- Adds query flexibility, which could be detrimental to privacy
- Leaks some metadata about each request that may otherwise be only visible to the leader (e.g. ip address), unless using some anonymizing proxy
- Introduces a new vector for replay attacks

### Protocol solution
It seems there is a fairly simple solution to this problem, and that is to simply:

- Instantiate the protocol where the collector is also the client, where the interactions between the "real clients" and the "client/collector" is unspecified by the protocol.
- Allow the "client" to optionally batch upload reports in the protocol rather than sending them one by one.

In the existing protocol there is no client authentication so it is technically possible to have a collector that just collects encrypted reports from clients and forwards them on to the leader. Of course the actual clients would need to be set up to do this but it is permitted by the protocol. By allowing batch uploads we just optimize this already-permitted configuration.

Alternatively, if we deem collector-clients to be bad for the protocol, we ought to have a mechanism which actually forbids them (e.g. by authenticating clients). However, I think that it is pretty reasonable to have this allowed by the protocol and leave it up to specific instantiations how the "client" is configured/trusted.

cjpatton:
Based on #77 (review that first).
Closes #64.

cc/ @csharrison

### DIFF
--- a/draft-pda-protocol.md
+++ b/draft-pda-protocol.md
@@ -483,15 +483,23 @@ the leader's endpoint URL. The payload is structured as follows:
 ~~~
 struct {
   PATaskID task_id;
-  uint64 time; // UNIX time (in seconds).
-  PAEncryptedInputShare encrypted_input_shares<1..2^16-1>;
+  PAReport reports<1..2^24-1>;
 } PAUploadReq;
 ~~~
 
-We sometimes refer to this message as the *report*. The message contains the
-task id derived from the PA parameters. It also includes the time (in seconds
-since the beginning of UNIX time) at which the report was generated. The rest of
-the message consists of the encrypted input shares, each of which has the
+This message consists of a task id and any number of *reports* for the task.
+Reports are structured as:
+
+~~~
+struct {
+  uint64 time; // UNIX time (in seconds).
+  PAEncryptedInputShare encrypted_input_shares<1..2^16-1>;
+} PAReport;
+~~~
+
+where `time` is the time (in seconds since the beginning of UNIX time) at which
+the report was generated and `encrypted_input_shares` is the sequence of
+encrypted input shares, each intended for one of the aggregators. These have the
 following structure:
 
 ~~~
@@ -508,11 +516,11 @@ struct {
   input share.
 * `payload` is the encrypted input share.
 
-To generate the report, the client begins by encoding its measurements as an
-input for the PA protocol and splitting it into input shares. (Note that the
-structure of each input share depends on the PA protocol in use, its parameters,
-and the role of aggregator, i.e., whether the aggregator is a leader or helper.)
-To encrypt an input share, the client first generates an HPKE
+To generate a report, the client begins by encoding its measurements as an input
+for the PA protocol and splitting it into input shares. (Note that the structure
+of each input share depends on the PA protocol in use, its parameters, and the
+role of aggregator, i.e., whether the aggregator is a leader or helper.) To
+encrypt an input share, the client first generates an HPKE
 {{!I-D.irtf-cfrg-hpke}} context for the aggregator by running
 
 ~~~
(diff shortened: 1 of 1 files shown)


chris-wood (approved):

ekr:
I actually don't understand what the benefit of this is. I see why you would want to do multiple submissions in a batch, but modern networking protocols such as H2 and H3 can do multiple concurrent HTTP requests in parallel, so why not just do one report at a time?

chris-wood:
> I actually don't understand what the benefit of this is.

I see it as purely an optimization to avoid sending the task ID over and over for large batches.

ekr:
> I see it as purely an optimization to avoid sending the task ID over and over for large batches.

Hmm... The task ID is in general going to be quite small compared to almost any report, right? I mean, just the crypto overhead is going to be order 48 bytes.

chris-wood:
> Hmm... The task ID is in general going to be quite small compared to almost any report, right? I mean, just the crypto overhead is going to be order 48 bytes.

It is -- I didn't say it was the most _useful_ optimization ;)

csharrison:
cc @hostirosti FYI. I like that this matches the mechanism we use to implement batch uploads from leader -> helper. Whatever future optimizations we implement for that could be useful here too (e.g. if we move away from HTTP POST and pass some file handle thing).

ekr:
I don't understand why moving away from POST will make things better. The vast amount of HTTP overhead is just moving the actual data around.

In any case, this seems like a case of adding protocol complexity for very little additional value, so I think we should not merge it.

hostirosti:
We should also look at this from the server side, yes with HTTP/2 you can have non-blocking parallel connections yet the server side needs to be able to handle the amount of connections as well, every connection carries overhead for establishing the connection (TCP, TLS handshake, ...). So reducing the number of connections from the client side will have a benefit for cost and operation on the server side, after all we talk 10^5s to millions of clients here. I'd definitely recommend batching on the client side if possible.

chris-wood:
@hostirosti I think @ekr's point is that you can do batching _within_ the protocol, as is proposed in this PR, or outside the protocol using independent HTTP streams. In both cases, the number of client<>server connections remains the same, right?

cjpatton:
In the 2021/7/14 call we decided to "park" this change for now. We all agree that the issue needs to be addressed, i.e., that an entity (client or collector) should be able to upload multiple reports. We haven't decided on whether these should be in the same HTTP request.

Issue #58, opened 2021-06-16
Consider using OHTTP 


tgeoghegan:
Prio v2 incorporates ingestion servers which sit between clients and aggregators. The ingestion servers:

- verify attestations from a trusted computing base in the client (i.e., secure enclave or TrustZone), making it harder for malicious clients to submit invalid inputs;
- remove any potentially identifying metadata from client reports before forwarding them to aggregators.

[Oblivious HTTP](https://datatracker.ietf.org/doc/draft-thomson-http-oblivious/) "allows a client to make multiple requests of a server without the server being able to link those requests to the client or to identify the requests as having come from the same client." We should see if we can integrate OHTTP into the DPA protocol as an optional extension that achieves the goals of the Prio v2 ingestion servers.

cjpatton:
I think this issue is orthogonal to the PA protocol spec, but is certainly a useful enhancement for deployments.

Issue #15, opened 2021-03-08
Enforce batch size (was "Trust in the leader / Aggregation size enforcement at the aggregator")


tlepoint:
If I am not mistaken, the current design would actually allow a malicious leader to learn single client contributions, by only sending that specific data to the aggregator, reporting that the proof is valid, ask the aggregator to "sum" the single element and return the answer. While the current design is motivated by different types of constraints (soft deadlines / privacy threshold defined by a minimum number of input shares), it must be observed that such a responsibility should not only be that of the leader.

Is there a plan to have aggregation size enforcement by the aggregators, and avoid aggregation of a small amount of (even a single) records?

tgeoghegan:
We did discuss that problem, but it's good to have an open issue on it so we don't forget to address it; thank you for filing this! 

One tricky dimension to this question is what the minimum threshold should be. I'm not sure if a single threshold value is appropriate for all use cases, and it might be that it's something the client should select while onboarding into the system.

tlepoint:
Having that be a choice of the client is interesting, but such value would need to be visible to the leader (since the leader needs to ask the aggregators to aggregate at least that number of shares together, but also need to be authenticated by the client for the aggregators (so that the leader cannot arbitrarily modify the value). This brings quite a few complications, especially with the optimization of using a KEM.

Another option (beyond enforcement by the aggregators, which I still find preferable) could be for the aggregators to keep a public and transparent log of how many shares a specific leader asked to aggregate. Such a public log may act as a deterrent for leader to ask for small aggregations, and they would need a strong justification on why a small aggregation was performed. On the downside, this would reveal publicly how much data is flowing throughout the system.

cjpatton:
My $0.02: aggregators should enforce the number of input shares aggregated before releasing their share of the output. One way to do this is, as you suggested, might be to replace the KEM with IND-CCA-secure encryption --- say, full HPKE, which takes associated data --- and authenticate the threshold by passing it in as associated data.

PR #84, opened 2021-07-14
Tidy and expand security considerations

tgeoghegan:
Cleans up the "Security Considerations" section which had some excessive nesting and adds content to address #15 and #58

### DIFF
--- a/draft-pda-protocol.md
+++ b/draft-pda-protocol.md
@@ -1012,8 +1012,6 @@ requests and controls the schedule for signaling aggregation rounds.
 
 # Security Considerations {#sec-considerations}
 
-## Security overview {#security-requirements}
-
 Prio assumes a powerful adversary with the ability to compromise an unbounded
 number of clients. In doing so, the adversary can provide malicious (yet
 truthful) inputs to the aggregation function. Prio also assumes that all but one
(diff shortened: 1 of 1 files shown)


cjpatton (changes requested):

draft-pda-protocol.md
```diff
@@ -1205,21 +1205,30 @@ collude and break soundness. Is this a contingency we need to address? There are
...
-Such a proxy could also remove auxiliary information from inputs (such as source
-IP or HTTP user agent), which could otherwise identify participating clients or
-permit some attacks on robustness.
+could be achieved by having clients submit reports to the leader over a client
+authenticated channel. Note that having clients authenticate to aggregators
+compromises privacy, which can be remedied by introducing an anonymizing proxy
+(see {{anon-proxy}}).
+
+[TODO: update this to discuss clients sending reports directly to helper as well
+as leader if we end up choosing to go that direction. See issue #89]
```
cjpatton, line 1220: On today's call (2021/07/28) we decided to address issue#89 by sticking in an opaque blob that deployment can use for attestation purposes. I think we should just remove this section and replace it with:
```
## Client Attestation
[TODO: Solve issue#89]
```
Then we can fully address it in a follow-on PR.

cjpatton (approved):
I suggest we remove the client auth stuff together and punt to the PR we discussed today, then merge this bad boy.

PR #87, opened 2021-07-15
clarify PATaskID derivation

### LINKED ISSUE #72: Derive the PA task from the parameters
On the 2021/7/7 design call we decided that the ID should be derived deterministically from the serialized PAParam by, for example, hashing it. This solves a few problems simultaneously. This settles the question of whether there should be 1:1 correspondence between a task ID and the parameters used for that task.

@tgeoghegan brought up an interesting question, which is how to make the task ID unique when two different data collection tasks use the same parameters. Thoughts?

tgeoghegan:
Resolves #72

### DIFF
--- a/draft-pda-protocol.md
+++ b/draft-pda-protocol.md
@@ -48,6 +48,16 @@ informative:
       -ins: N. Gilboa
       -ins: Y. Ishai
 
+normative:
+
+  FIPS180-4:
+    title: NIST FIPS 180-4, Secure Hash Standard
+    author:
+      name: NIST
+      ins: National Institute of Standards and Technology, U.S. Department of Commerce
+    date: 2012-03
+    target: http://csrc.nist.gov/publications/fips/fips180-4/fips-180-4.pdf
+
 --- abstract
 
 TODO: writeme
@@ -337,6 +347,7 @@ verify and aggregate the clients' measurements:
 
 ~~~
 struct {
+  opaque nonce[16];
   Url leader_url;
   Url helper_url;
   HpkeConfig collector_config; // [TODO: Remove this?]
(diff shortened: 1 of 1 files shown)


cjpatton (approved):
Approved, with the following suggestion: Let's leave a [TODO: ...] comment under the definition of `nonce` for clarifying how it should be chosen.

tgeoghegan:
> Approved, with the following suggestion: Let's leave a [TODO: ...] comment under the definition of `nonce` for clarifying how it should be chosen.

I added a clarification that the nonce is random. IMO the question of *who* randomly generates the nonce goes with the question of how participants agree upon PDAParams, which is to say it's outside the protocol.

draft-pda-protocol.md
```diff
@@ -367,14 +370,17 @@ opaque Url<1..2^16-1>;
...
 opaque PDATaskID[32];
 ~~~
 
-The task id is derived using the following procedure. [TODO: Specify derivation
-of the task ID.]
+The task ID is derived using the following procedure:
+
+~~~
+task_id = SHA-256(PDAParam)
+~~~
```
chris-wood, line 383:
```suggestion
The task ID is of a `PDAParam` instance `param` is derived using the following procedure:

~~~
task_id = SHA-256(param)
~~~
```

Where SHA-256 is as specified in {{FIPS180-4}}.

draft-pda-protocol.md
```diff
@@ -355,6 +356,8 @@ enum { prio(0), hits(1) } PDAProto;
 opaque Url<1..2^16-1>;
 ~~~
 
+* `nonce`: A sequence of random bytes used  to ensure that two otherwise
```
chris-wood, line 359:
```suggestion
* `nonce`: A unique sequence of bytes used to ensure that two otherwise
```

draft-pda-protocol.md
```diff
@@ -355,6 +356,8 @@ enum { prio(0), hits(1) } PDAProto;
...
+  identical `PDAParam` instances will have distinct `PDATaskID`s.
```
chris-wood, line 360:
```suggestion
  identical `PDAParam` instances will have distinct `PDATaskID`s. It is RECOMMENDED
  that this be set to a random 16-byte string derived from a cryptographically secure 
  pseudorandom number generator.
```

chris-wood (approved):

PR #88, opened 2021-07-15
introduce definitions for time units

tgeoghegan:
Introduces `Time` and `Duration` definitions, principally to make it
more obvious that `batch_window` is a number of seconds.

### DIFF
--- a/draft-pda-protocol.md
+++ b/draft-pda-protocol.md
@@ -341,7 +341,7 @@ struct {
   Url helper_url;
   HpkeConfig collector_config; // [TODO: Remove this?]
   uint64 batch_size;
-  uint64 batch_window;
+  Duration batch_window;
   PDAProto proto;
   uint16 length; // Length of the remainder.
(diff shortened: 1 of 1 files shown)


cjpatton (approved):
Good idea!

draft-pda-protocol.md
```diff
@@ -353,6 +353,8 @@ struct {
 enum { prio(0), hits(1) } PDAProto;
 
 opaque Url<1..2^16-1>;
+
+Duration uint64; /* Number of seconds elapsed between two instants */
```
chris-wood, line 357:
```suggestion
Time uint64; /* seconds elapsed since start of UNIX epoch */
Duration uint64; /* Number of seconds elapsed between two Time instants */
```

chris-wood (approved):
Yeah, this is a nice improvement. I'd keep Time and Duration colocated so we can more clearly define in terms of the other.

PR #90, opened 2021-07-26
Add additional anti-replay protection

### LINKED ISSUE #81: Intra-batch replay detection
For privacy reasons, leaders should not be allowed to replay a given report more than once within the context of a single batch. The current verify+collect model runs in a "streaming mode," wherein the leader sends slices of batch shares to helpers for verification and aggregation. 

If we changed this to require the leader to send _all_ batch shares in a single verify+collect flow, then replay prevention would be simpler: just scan the list and check for dupes. 

If we stick with the current model, then the helper must necessarily store some state (on the leader, most likely) representing the shares processed. This state could then be consulted when asked to verify+collect any new share. (We can probably spell out a number of ways to do store this state, but ideally it should not be linear in the number of shares collected so far.) We'd likely also need a way for the leader to request that the helper produce the final aggregate output once the batch is "done" being verified and collected.

cjpatton:
Closes #81.

### DIFF
--- a/draft-pda-protocol.md
+++ b/draft-pda-protocol.md
@@ -392,7 +392,7 @@ Each task has a unique *task ID* derived from the PDA parameters:
 opaque PDATaskID[32];
 ~~~
 
-The task ID of a `PDAParam` `is derived using the following procedure:
+The task ID of a `PDAParam` is derived using the following procedure:
 
 ~~~
 task_id = SHA-256(param)
@@ -508,15 +508,24 @@ the leader's endpoint URL. The payload is structured as follows:
 struct {
   PDATaskID task_id;
   Time time;
+  uint64 jitter;
   PDAEncryptedInputShare encrypted_input_shares<1..2^16-1>;
-} PDAUploadReq;
+} PDAReport;
 ~~~
 
-We sometimes refer to this message as the *report*. The message contains the
-`task_id` of the previous request. It also includes the time at which the report
-was generated. This field is present to ensure that each report is included in
-at most one batch. The rest of the message consists of the encrypted input
-shares, each of which has the following structure:
+This message is called the client's *report*. It contains the following fields:
+
+* `task_id` is the task ID of the task for which the report is intended.
+* `time` is the time at which the report was generated. This field is used by
+  the aggregators to ensure the report appears in at most one batch. (See
+  {{anti-replay}}.)
+* `jitter` is a random number chosen by the client generating the report. This
+  and the timestamp field are used by the aggregators to ensure that each report
+  appears at most once in a batch. (See {{anti-replay}}.)
+* `encrypted_input_shares` contains the encrypted input shares of each of the
+  aggregators.
+
+Encrypted input shares are structured as follows:
 
 ~~~
 struct {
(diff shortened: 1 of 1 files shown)


draft-pda-protocol.md
```diff
@@ -1297,6 +1319,28 @@ participating aggregators stored unencrypted input shares on the same cloud
 object storage service, then that cloud vendor would be able to reassemble all
 the input shares and defeat privacy.
 
+#### Anti-replay {#anti-replay}
+
+Using a report in multiple batches, or multiple times within a single batch, is
+considered a privacy violation, since it may leak more information about that
+report than intended by the PDA protocol. (For example, this may violate
+differential privacy.) To mitigate this issue, the core specification imposes an
```
tgeoghegan, line 1327:
```suggestion
report than intended by the PDA protocol. For example, this may violate
differential privacy. To mitigate this issue, the core specification imposes an
```

draft-pda-protocol.md
```diff
@@ -1297,6 +1319,28 @@ participating aggregators stored unencrypted input shares on the same cloud
...
+R1.jitter`, then we say that `R2` *does not follow* `R1`.
+
+To prevent replay attacks, it suffices for each aggregator to ensure that each
+report it consumes follows the previous one it consumed. To prevent the
+adversary from tampering with the ordering of reports, honest clients
+incorporate the ordering-sensitive parameters `(time, jitter)` into the AAD for
+HPKE encryption.
+
+Note that this strategy may result in dropping reports that happen to have the
+same timestamp and jitter value.
```
tgeoghegan, line 1342: With 32 bit jitter, by the birthday paradox, we'd expect jitter collisions if a single task generates more than 2.1 billion reports in a second across all clients. That's definitely a lot, but why not make the jitter 64 bits so that it's astronomically unlikely?

draft-pda-protocol.md
```diff
@@ -540,24 +549,20 @@ To encrypt an input share, the client first generates an HPKE
 {{!I-D.irtf-cfrg-hpke}} context for the aggregator by running
 
 ~~~
-enc, context = SetupBaseS(pk, [TODO])
+enc, context = SetupBaseS(pk, "pda input share" || server_role)
```
tgeoghegan, line 552: Do we have to define "||" as concatenation somewhere?

draft-pda-protocol.md
```diff
@@ -711,17 +717,27 @@ struct {
...
-helper share and request parameters used for the current round.
+The `time` and `jitter` fields have the same value as those in the report
+uploaded by the client. Similarly, the `helper_share` field is the helper's
+encrypted input share as it appeared in the report. [OPEN ISSUE: We usually only
+need to send this in the first aggregate request. Shall we exclude it in
+subsequent requests somehow?] The remainder of the structure is dedicated to the
+protocol-specific request parameters.
+
+In order to provide replay protection, the leader is required to send aggregate
+sub-requests in ascending order, where the ordering on sub-requests is
```
tgeoghegan, line 728: If the leader is responsible for reordering reports, then it should allow a grace period after an aggregation window starts for late reports to arrive from clients. Unlike Prio v2, the two aggregators don't need to agree on the grace period, since the leader drives aggregations (at the request of a collector, but since those are going to have to be asynchronous anyway I think it's OK for a leader to acknowledge a collect request but not service it until the leader is satisfied the aggregation window's grace period has elapsed). So I think we don't need to put grace period into PDAParams or specify much about what leaders should do, but perhaps we should have a SHOULD statement urging leader implementations to implement a grace period proportional to the aggregation window or other deployment specific details (e.g. special knowledge of the schedule on which clients upload reports)?

tgeoghegan (approved):
I have a few questions but nothing blocking.

draft-pda-protocol.md
cjpatton: I don't think so.

draft-pda-protocol.md
cjpatton: Actually, by the birthday paradox, the probability of a collision among just 2^16 reports is about 1.  For 64 bits, this same is true for 2^32 reports. If we want this probability to be negligible (i.e., about as unlikely as cracking AES-GCM, say) we would need 128 bits.

Personally, I'd be happy with 64 bits.

draft-pda-protocol.md
cjpatton: Good idea, done.

draft-pda-protocol.md
cjpatton: Done.

draft-pda-protocol.md
```diff
@@ -540,24 +549,20 @@ To encrypt an input share, the client first generates an HPKE
...
 
-where `input_share` is the aggregator's input share.
-
-[TODO: Fully specify encryption of the shares. We need to make sure we
-authenticate things like the PDA parameters and the report timestamp. We need to
-decide what the info string for SetupBaseS() will be, as well as the aad for
-context.Seal(). The aad might be the entire "transcript" between the client and
-aggregator.]
+where `input_share` is the aggregator's input share and `task_id`, `time`, and
+`jitter` are the fields of `PDAReport`.
```
chris-wood, line 565: Should we say that leaders should drop reports whose timestamps fall outside of the batch window? (Happy for that to be an open issue, if desired.)

chris-wood (approved):

draft-pda-protocol.md
cjpatton: I think this would be a bit confusing here. Ideally this is clear from {{pa-aggregate}}.

Issue #19, opened 2021-03-10
Accommodating randomized inputs (a la DP)


chris-wood:
@tlepoint pointed out that PrioV2 requires all clients to submit randomized values for the purposes of ensuring that identical inputs don't yield the same output. Basically, with some very small probability, clients flip some bits in their data to maximize aggregate utility and privacy (in terms of differential privacy bounds). [Apple's paper](https://arxiv.org/abs/2012.12803) on this topic has details. 

We should probably design the system such that either clients or aggregators can inject this sort of noise.

cjpatton:
This seems orthogonal to the core protocol logic, since its applicability depends on the data type.

tgeoghegan:
There's an open question that fell out of the review of #91 that feels appropriate for this issue. Aggregators are encouraged to apply differential privacy noise to outputs. Collectors and aggregators may need to agree on DP parameters (e.g., the probability with which values are flipped) so that collectors can de-noise outputs. I think we might need to have a field in `PDAParams` for differential privacy parameters.

csharrison:
It's worth noting that in many situations (especially with server-added noise), the collectors don't need to do anything special to "de-noise" since it's often going to have a mean of 0. However, collectors may want to know things like the variance of the noise, or what distribution the noise was sampled from so they can understand the effective error.

csharrison:
I was originally going to post this as a new issue but I think this one should suffice. Originally titled "Optional DAP configuration that achieves differential privacy". Apologies in advance for the wall of text, I hope it is useful to the discussion.

-----

Differential privacy (DP) is a useful property of an aggregate system. Among other things, it is a mitigation against Sybil attacks that aim to violate user privacy. This is because differential privacy holds under worst-case assumptions, including when the adversary knows the contents of the entire database except for the targeted user. In other words, malicious clients submitting data to the system does not adversely affect the privacy (in this worst-case sense) of any honest user.

We should consider whether we want to specify an optional configuration in DAP which deployments can use to achieve DP output. @cjpatton mentions [here](https://github.com/ietf-wg-ppm/ppm-specification/pull/235#discussion_r865016936) it would be useful, so here are my thoughts:

DP typically involves noise addition, we ought to consider two flavors, since I think both are relevant and useful to this work:

## 1. Clients add noise to their input

[...]

PR #91, opened 2021-07-28
Add differential privacy as a threat mitigation

csharrison:
Differential privacy can protect against attacks where a portion of the input data is known a priori, or when the size of a batch is small. This PR adds DP as an optional mitigation for PDA deployments and lists where it helps mitigate various threats.

Additionally:
- Removes a false statement that revealing user input does not compromise other users' privacy. This is not necessarily the case (e.g. imagine if the batch size is 2 or if many users reveal their input).
- Removes a statement that clients revealing their inputs is outside the threat model (is this still relevant?)

### DIFF
--- a/draft-pda-protocol.md
+++ b/draft-pda-protocol.md
@@ -48,6 +48,20 @@ informative:
       -ins: N. Gilboa
       -ins: Y. Ishai
 
+  JD02:
+    title: "The Sybil Attack"
+    date: 2022-10-10
+    target: "https://link.springer.com/chapter/10.1007/3-540-45748-8_24"
+    author:
+      -ins: J. Douceur
+
+  SV16:
+    title: "The Complexity of Differential Privacy"
+    date: 2016-08-09
+    target: "https://privacytools.seas.harvard.edu/files/privacytools/files/complexityprivacy_1.pdf"
+    author:
+      -ins: S. Vadhan
+
 normative:
 
   FIPS180-4:
@@ -1175,6 +1189,7 @@ mitigations available to aggregators also apply to the leader.
 1. Known input injection. Collectors may collude with clients to send known
    input to the aggregators, allowing collectors to shrink the effective
    anonymity set by subtracting the known inputs from the final output.
+   Sybil attacks {{JD02}} could be used to amplify this capability.
 
 #### Mitigations
 
@@ -1260,14 +1275,13 @@ choose minimum batch sizes.
 ## Differential privacy {#dp}
 
 Optionally, PDA deployments can choose to ensure their output F achieves
-[differential privacy](https://en.wikipedia.org/wiki/Differential_privacy).
-A simple approach would require the aggregators to add two-sided
-noise (e.g. sampled from a two-sided geometric distribution) to outputs.
-Since each aggregator is adding noise independently, privacy can be guaranteed
-even if all but one of the aggregators is malicious. Differential privacy is a strong
-privacy definition, and protects users in extreme circumstances: Even if an
-adversary has prior knowledge of every input in a batch except for one, that
-one record is still protected.
+differential privacy {{SV16}}. A simple approach would require the aggregators
+to add two-sided noise (e.g. sampled from a two-sided geometric distribution)
+to outputs. Since each aggregator is adding noise independently, privacy can be
+guaranteed even if all but one of the aggregators is malicious. Differential
+privacy is a strong privacy definition, and protects users in extreme
+circumstances: Even if an adversary has prior knowledge of every input in a
+batch except for one, that one record is still protected.
 
 ## Multiple protocol runs
 
--- a/draft-pda-protocol.md
+++ b/draft-pda-protocol.md
@@ -1099,8 +1099,8 @@ can defeat privacy by leaking input outside of the Prio system.
 1. The input validation protocol executed by the aggregators prevents either
 individual clients or coalitions of clients from compromising the robustness
 property.
-1. If aggregator output satisifes [differential privacy](#differential-privacy)
-then all records not leaked by malicious clients are still protected.
+1. If aggregator output satisifes differential privacy {{dp}}, then all records
+not leaked by malicious clients are still protected.
 
 ### Aggregator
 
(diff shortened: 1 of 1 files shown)


draft-pda-protocol.md
```diff
@@ -1173,11 +1172,16 @@ mitigations available to aggregators also apply to the leader.
    aggregations, joint randomness, arithmetic circuits).
 1. Collectors may trivially defeat availability by discarding output shares
    submitted by aggregators.
+1. Known input injection. Collectors may collude with clients to send known
+   input to the aggregators, allowing collectors to shrink the effective
+   anonymity set by subtracting the known inputs from the final output.
```
cjpatton, line 1177: I would work in the term "Sybil attack" here. While we're at it, let's add a reference for Sybil attacks (i.e., `{{sybil}}`). I think the canonical reference is https://link.springer.com/chapter/10.1007/3-540-45748-8_24.

draft-pda-protocol.md
```diff
@@ -1104,6 +1099,8 @@ can defeat privacy by leaking input outside of the Prio system.
 1. The input validation protocol executed by the aggregators prevents either
 individual clients or coalitions of clients from compromising the robustness
 property.
+1. If aggregator output satisifes differential privacy {{dp}}, then all records
+not leaked by malicious clients are still protected.
```
cjpatton, line 1103: Would you mind adding a reference for `{{dp}}` and verify the document builds by running `make`? See the top of the document for guidance. If you don't have time, then please at least provide a link to an article you think would be a suitable reference. The most useful reference would be something that describes the distinction between client-side and server-side noise addition.

draft-pda-protocol.md
```diff
@@ -1253,6 +1257,18 @@ but server implementations may also opt out of participating in a PDA task if
 the minimum batch size is too small. This document does not specify how to
 choose minimum batch sizes.
 
+## Differential privacy {#dp}
+
+Optionally, PDA deployments can choose to ensure their output F achieves
+[differential privacy](https://en.wikipedia.org/wiki/Differential_privacy).
```
cjpatton, line 1263: Given that the likely trajectory of this document is an IETF WG draft, it would be better to replace this markdown with a `{{dp}}` reference.

draft-pda-protocol.md
```diff
@@ -1253,6 +1257,18 @@ but server implementations may also opt out of participating in a PDA task if
...
+A simple approach would require both the helper and leader to add two-sided
```
cjpatton, line 1264:
```suggestion
A simple approach would require the aggregators to add two-sided
```
Depending on what we do for #68 there may be more than one helper.

draft-pda-protocol.md
```diff
@@ -1253,6 +1257,18 @@ but server implementations may also opt out of participating in a PDA task if
...
+noise (e.g. sampled from a two-sided geometric distribution) to outputs.
+Since each aggregator is adding noise independently, privacy can be guaranteed
+even if one of the aggregators is malicious. Differential privacy is a strong
```
cjpatton, line 1267:
```suggestion
even if all but one of the aggregators is malicious. Differential privacy is a strong
```

cjpatton (changes requested):
Excellent! Requested changes are editorial.

draft-pda-protocol.md
```diff
@@ -1253,6 +1257,18 @@ but server implementations may also opt out of participating in a PDA task if
 the minimum batch size is too small. This document does not specify how to
 choose minimum batch sizes.
 
+## Differential privacy {#dp}
```
tgeoghegan, line 1260: Besides this informal recommendation, do we need explicit protocol support for differential privacy so that collectors can de-noise outputs? We can leave it up to aggregators to decide how they're going to implement DP but I wonder if `PDAOutputShare` should have a field for the epsilon value that was used by the aggregator. Forgive me if I'm talking nonsense about DP, I am speaking in the terms that we used in Prio v2.

tgeoghegan (approved):
I have one question about whether we need more explicit protocol support for DP but if so, that can land in a subsequent PR.

draft-pda-protocol.md
cjpatton: Oops, I missed that {{dp}} refers to a section and not a paper. Disregard.

draft-pda-protocol.md
cjpatton: Edited this comment.

draft-pda-protocol.md
csharrison: Done

draft-pda-protocol.md
csharrison: Hm. I am nervous about being to prescriptive here. In the simplest protocol design nothing is needed since the epsilon is hardcoded into the specific protocol instantiation and won't change.

In practice, some specific instantiations may want to reveal even more information about how noise was applied, e.g.
- The distribution noise is sampled from (Laplace, Gaussian, etc)
- Parameters of the noise distribution
- Any kind of threshold used (for example, if you are using approximate DP)

I think we should make this as opaque to the protocol as possible vs. prescribing some single "epsilon" field which might be too constraining. What do you think?

draft-pda-protocol.md
csharrison: Done, I added a general DP reference in place of the wikipedia link. It discusses central, local, and multi-party DP.

draft-pda-protocol.md
csharrison: However, I still think this current PR is land-able given that a basic instantiation can hardcode everything without requiring any communication.

cjpatton: The way to go here, I think, is to document the open question by adding an `[OPEN ISSUE: blah blah blah]`.

cjpatton (approved):

draft-pda-protocol.md
csharrison: Done

tgeoghegan: I put a note about this in #19, which I think is an appropriate issue to track this discussion.

PR #93, opened 2021-07-29
Document minimal operational capabilities.

chris-wood:
I also deleted a bunch of the (outdated) system design text. I'm sure this can be simplified, but I think this captures the basic assumptions (at least in my mental model).

### DIFF
--- a/draft-pda-protocol.md
+++ b/draft-pda-protocol.md
@@ -1064,7 +1064,7 @@ In addition, for each PDAParam instance, leaders are required to:
 ### Collector capabilities
 
 Collectors statefully interact with aggregators to produce an aggregate output. Their
-input to the protocol is the PAParam structure, configured out of band, which contains
+input to the protocol is the PDAParam structure, configured out of band, which contains
 the corresponding batch window and size. For each collect invocation, collectors are
 required to keep state from the start of the protocol to the end as needed to produce
 the final aggregate output.
--- a/draft-pda-protocol.md
+++ b/draft-pda-protocol.md
@@ -1058,7 +1058,7 @@ In addition, for each PDAParam instance, leaders are required to:
 
 - Ensure that, for each report, all aggregator shares are either included or not in a
   given batch;
-- Implement and store state for some form of inter- and intra-batch replay mitigation; and
+- Implement and store state for the form of inter- and intra-batch replay mitigation in {{anti-replay}}; and
 - Store helper state associated with a given PDAParam batch.
 
 ### Collector capabilities
--- a/draft-pda-protocol.md
+++ b/draft-pda-protocol.md
@@ -193,7 +193,8 @@ because modular addition is commutative. I.e.,
       = F(x[1], x[2], x[3])
 ~~~
 
-**Prio.**
+### Prio {#prio-variant}
+
 This approach can be used to privately compute any function `F` that can be
 expressed as a function of the sum of the users' inputs. In Prio {{CB17}}, each
 user splits its input into shares and sends each share to one of the
@@ -216,7 +217,8 @@ below:
 This variety of aggregate types is sufficient to support a wide variety of
 data aggregation tasks.
 
-**Hits.**
+### Hits {#hits-variant}
+
 A common PDA task that can't be solved efficiently with Prio is the
 `t`-*heavy-hitters* problem {{BBCp21}}. In this setting, each user is in
 possession of a single `n`-bit string, and the goal is to compute the compute
@@ -320,6 +322,9 @@ made from one entity to another.
    uploaded. (See {{prio}}.) However, in general it is necessary for them to
    wait until (a) enough reports have been uploaded and (b) the collector has
    made a request. (See {{hits}}.)
+
+The operational capabilities of each entity are described further in
+{{entity-capabilities}}.
 
 # PDA protocols {#pa}
 
@@ -580,7 +585,8 @@ where `input_share` is the aggregator's input share and `task_id`, `time`, and
 
 The leader responds to well-formed requests to `[leader]/upload` with status 200
 and an empty body. Malformed requests are handled as described in
-{{pa-error-common-aborts}}.
+{{pa-error-common-aborts}}. Clients SHOULD NOT upload the same measurement value
+in more than one report if the leader responds with status 200 and an empty body.
 
 ### Upload Extensions {#upload-extensions}
 
@@ -703,7 +709,7 @@ zero knowledge that the input is valid. The exact procedure for doing so is
 protocol specific, but all protocols have the same basic structure. In
 particular, the protocol is comprised of a sequence of *aggregate requests* from
 the leader to the helper. At the end of this procedure, the leader and helper
-will have have aggregated a set of valid client inputs (though not necessarily a
+will have aggregated a set of valid client inputs (though not necessarily a
 complete batch).
 
 #### Aggregate Request
@@ -767,6 +773,7 @@ each i. The structure of each sub-response is specific to the PDA protocol:
 
 ~~~
 struct {
+  // XXX(caw): include running hash output
   opaque helper_state<0..2^16>;
   PDAAggregateSubResp seq<1..2^24-1>;
 } PDAAggregateResp;
@@ -1001,31 +1008,70 @@ indefinitely until a suitable output is found.
 
 [TODO: Define Hits-specific protocol messages.]
 
-# System design
+# Operational Considerations {#operational-capabilities}
 
-[[OPEN ISSUE: This section seems like a catch-all for things not in other
-sections. Perhaps there is a natural home for aggregator discovery, share
-uploading, open issues, and system parameters?]]
-
-## Aggregator discovery
-
-[[OPEN ISSUE: writeme]]
-
-## Share uploading
-
-[[OPEN ISSUE: writeme]]
-
-## Open questions and system parameters {#questions-and-params}
-
-[[OPEN ISSUE: discuss batch size parameter and thresholds]]
-[[OPEN ISSUE: discuss f^ leakage differences from [GB17]]]
-
-
-# Operational Considerations
-
-Prio has inherent constraints derived from the tradeoff between privacy
+PDA protocols have inherent constraints derived from the tradeoff between privacy
 guarantees and computational complexity. These tradeoffs influence how
 applications may choose to utilize services implementing the specification.
+
+## Protocol participant capabilities {#entity-capabilities}
+
+The design in this document has different assumptions and requirements for
+different protocol participants, including clients, aggregators, and
+collectors. This section describes these capabilities in more detail.
+
+### Client capabilities
+
+Clients have limited capabilities and requirements. Their only inputs to the protocol
+are (1) the PDAParam structure configured out of band and (2) a measurement. Clients
+are not expected to store any state across any upload
+flows, nor are they required to implement any sort of report upload retry mechanism.
+By design, the protocol in this document is robust against individual client upload
+failures since the protocol output is an aggregate over all inputs.
+
+### Aggregator capabilities
+
+Helpers and leaders have different operational requirements. The design in this
+document assumes an operationally competent leader, i.e., one that has no storage
+or computation limitations or constraints, but only a modestly provisioned helper, i.e., one that
+has computation, bandwidth, and storage constraints. By design, leaders must be
+at least as capable as helpers, where helpers are generally required to:
+
+- Support the collect protocol, which includes verifying and aggregating
+  sets of reports in a given batch; and
+- Publish and manage an HPKE configuration that can be used for the upload protocol;
+
+In addition, for each PDAParam instance, helpers are required to:
+
+- Implement some form of batch-to-report index, as well as inter- and intra-batch
+  replay mitigation storage, which includes some way of tracking batch report size
+  with optional support for state offloading. Some of this state may be used for
+  replay attack mitigation. One relay mitigation strategy is described in {{anti-replay}}.
+
+Beyond the minimal capabilities required of helpers, leaders are generally required to:
+
+- Support the upload protocol and store client reports for a given PDAParam instance,
+  where each report maps uniquely to a single batch, and index this storage by batch durations;
+- Track batch report size during each collect flow and request encrypted output shares
+  from helpers.
+
+In addition, for each PDAParam instance, leaders are required to:
+
+- Ensure that, for each report, all aggregator shares are either included or not in a
+  given batch;
+- Implement and store state for some form of inter- and intra-batch replay mitigation; and
+- Store helper state associated with a given PDAParam batch.
+
+### Collector capabilities
+
+Collectors statefully interact with aggregators to produce an aggregate output. Their
+input to the protocol is the PAParam structure, configured out of band, which contains
+the corresponding batch window and size. For each collect invocation, collectors are
+required to keep state from the start of the protocol to the end as needed to produce
+the final aggregate output.
+
+Collectors must also maintain state for the lifetime of each PDAParam value, which includes
+key material associated with the HPKE key configuration.
 
 ## Data resolution limitations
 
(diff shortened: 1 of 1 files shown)


draft-pda-protocol.md
```diff
@@ -950,31 +956,68 @@ indefinitely until a suitable output is found.
 
 [TODO: Define Hits-specific protocol messages.]
 
-# System design
+# Operational Considerations {#operational-capabilities}
+
+Prio has inherent constraints derived from the tradeoff between privacy
```
cjpatton, line 961:
```suggestion
PDA protocols have inherent constraints derived from the tradeoff between privacy
```

draft-pda-protocol.md
```diff
@@ -950,31 +956,68 @@ indefinitely until a suitable output is found.
...
+different protocol participants, including clients, aggregators, and
+collectors. This section describes these capabilities in more detail.
 
-[[OPEN ISSUE: writeme]]
+### Client capabilities
 
-## Share uploading
+Clients have limited capabilities and requirements. Their only inputs to the protocol
+are (1) the PDAParam structure configured out of band and (2) a corresponding value to be
+submitted in a report. Clients are not expected to store any state across any upload
```
cjpatton, line 975:
```suggestion
are (1) the PDAParam structure configured out of band and (2) a measurement. Clients
are not expected to store any state across any upload
```

draft-pda-protocol.md
```diff
@@ -950,31 +956,68 @@ indefinitely until a suitable output is found.
...
+flows, nor are required to implement any sort of report retry mechanism should it fail.
+By design, the protocol in this document is robust against individual client upload
+failures since the protocol output is an aggregate over all inputs.
```
cjpatton, line 978: Also worth mentioning here: The aggregators validate inputs before consuming them.

draft-pda-protocol.md
```diff
@@ -950,31 +956,68 @@ indefinitely until a suitable output is found.
...
 
-[[OPEN ISSUE: writeme]]
+### Aggregator capabilities
 
-## Open questions and system parameters {#questions-and-params}
+Helpers and leaders have different operational requirements. The design in this
+document assumes an operationally competent leader, i.e., one that has no storage
+or computation limitations or constraints, and minimalistic helper, i.e., one that
```
cjpatton, line 984:
```suggestion
or computation limitations or constraints, but only a modestly provisioned helper, i.e., one that
```

draft-pda-protocol.md
```diff
@@ -950,31 +956,68 @@ indefinitely until a suitable output is found.
...
+- Support the collect protocol, which includes verifying and aggregating
+  sets of reports in a given batch; and
+- Publish and manage an HPKE configuration that can be used for the upload protocol;
 
+In addition, for each PDAParam instance, helpers are required to:
 
-# Operational Considerations
+- Implement some form of batch-to-report and intra-batch replay mitigation storage,
+  which includes some way of tracking batch report size, with optional support for
+  state offloading.
```
cjpatton, line 996: Clarify here that some of the offloaded state can be used for anti-replay, but not all.

draft-pda-protocol.md
```diff
@@ -950,31 +956,68 @@ indefinitely until a suitable output is found.
...
 
-Prio has inherent constraints derived from the tradeoff between privacy
-guarantees and computational complexity. These tradeoffs influence how
-applications may choose to utilize services implementing the specification.
+[OPEN ISSUE: describe intra- and inter-batch replay attacks in the security considerations]
```
cjpatton, line 998: This will be resolved (at least partially) by #90.

draft-pda-protocol.md
```diff
@@ -950,31 +956,68 @@ indefinitely until a suitable output is found.
...
+- Support the upload protocol and store client reports for a given PDAParam instance,
+  where each report maps uniquely to a single batch, and index this storage by batch durations;
+- Track batch report size during each collect flow and request encrypted output shares
+  from helpers.
+
+In addition, for each PDAParam instance, leaders are required to:
+
+- Ensure that, for each report, all aggregator shares are either included or not in a
+  given batch;
+- Implement and store state for some form of inter- and intra-batch replay mitigation; and
```
cjpatton, line 1011: This is also a requirement for the helper.

draft-pda-protocol.md
```diff
@@ -950,31 +956,68 @@ indefinitely until a suitable output is found.
...
+- Store helper state associated with a given PDAParam batch.
+
+### Collector capabilities
+
+Collectors statefully interact with aggregators to produce an aggregate output. Their
+input to the protocol is the PAParam structure, configured out of band, which contains
+the corresponding batch window and size. For each collect invocation, collectors are
+required to keep state from the start of the protocol to the end as needed to produce
+the final aggregate output.
```
cjpatton, line 1020: They also need to keep around an HPKE secret key for the lifetime of the PDAParam.

draft-pda-protocol.md
```diff
@@ -1057,11 +1078,40 @@ order to achieve these goals:
...
 2. Aggregation function choice. Some aggregation functions leak slightly more
-   than the function output itself. {{questions-and-params}} discusses the
-   leakage profiles of various aggregation functions in more detail.
+   than the function output itself.
+
+[TODO: discuss these in more detail.]
+
+## Infrastructure diversity
+
+Prio deployments should ensure that aggregators do not have common dependencies
```
cjpatton, line 1088:
```suggestion
PDA deployments should ensure that aggregators do not have common dependencies
```

draft-pda-protocol.md
```diff
@@ -1057,11 +1078,40 @@ order to achieve these goals:
...
+system. The integrity of the output can be influenced by an incomplete round of
+aggregation caused by network partitions, or by bad actors attempting to cause
+inaccuracies in the aggregates. An example data integrity constraint is that
+every share must be processed exactly once by all aggregators. Data integrity
+constraints may be at odds with the threat model if meeting the constraints
+requires replaying data.
+
+Aggregator operators should expect to encounter invalid inputs during regular
+operation due to misconfigured or malicious clients. Low volumes of errors are
+tolerable; the input-verification protocol and AFEs are robust in the face of
```
cjpatton, line 1106:
```suggestion
tolerable; the input-verification procedure of the PDA protocol
```
The term "AFE" is specific to Prio.

draft-pda-protocol.md
```diff
@@ -950,31 +956,68 @@ indefinitely until a suitable output is found.
...
+The design in this document has different assumptions and requirements for
+different protocol participants, including clients, aggregators, and
+collectors. This section describes these capabilities in more detail.
 
-[[OPEN ISSUE: writeme]]
+### Client capabilities
 
-## Share uploading
+Clients have limited capabilities and requirements. Their only inputs to the protocol
+are (1) the PDAParam structure configured out of band and (2) a corresponding value to be
```
tgeoghegan: Corresponding to what?

draft-pda-protocol.md
```diff
@@ -950,31 +956,68 @@ indefinitely until a suitable output is found.
...
+submitted in a report. Clients are not expected to store any state across any upload
+flows, nor are required to implement any sort of report retry mechanism should it fail.
```
tgeoghegan, line 976:
```suggestion
flows, nor are they required to implement any sort of report upload retry mechanism.
```
It was unclear what "it" referred to, and I think retry implies a failure.

draft-pda-protocol.md
```diff
@@ -950,31 +956,68 @@ indefinitely until a suitable output is found.
...
+- Implement some form of batch-to-report and intra-batch replay mitigation storage,
+  which includes some way of tracking batch report size, with optional support for
+  state offloading.
 
-Prio has inherent constraints derived from the tradeoff between privacy
-guarantees and computational complexity. These tradeoffs influence how
-applications may choose to utilize services implementing the specification.
+[OPEN ISSUE: describe intra- and inter-batch replay attacks in the security considerations]
+
+Beyond the minimal capabiltiies required of helpers, leaders are generally required to:
```
tgeoghegan, line 1000:
```suggestion
Beyond the minimal capabilities required of helpers, leaders are generally required to:
```

draft-pda-protocol.md
```diff
@@ -1057,11 +1078,40 @@ order to achieve these goals:
...
+operation due to misconfigured or malicious clients. Low volumes of errors are
+tolerable; the input-verification protocol and AFEs are robust in the face of
+malformed data. Aggregators may need to detect and mitigate statistically
+significant floods of invalid or identical inputs that affect accuracy, e.g.,
+denial of service (DoS) events.
+
+Certain classes of errors do not exist in the input-validation protocol
+considered in this document. For example, packet loss errors when clients make
+requests directly to aggregators are not relevant when the leader proxies
+requests and controls the schedule for signaling aggregation rounds.
```
cjpatton, line 1114: I realize this is copy-pasted, but I'm questioning the value of this paragraph at this point. Is it saying anything that's new and not obvious?

draft-pda-protocol.md
```diff
@@ -1057,11 +1078,40 @@ order to achieve these goals:
...
+object storage service, then that cloud vendor would be able to reassemble all
+the input shares and defeat privacy.
+
+## Data integrity constraints
+
+Data integrity concerns the accuracy and correctness of the outputs in the
+system. The integrity of the output can be influenced by an incomplete round of
+aggregation caused by network partitions, or by bad actors attempting to cause
+inaccuracies in the aggregates. An example data integrity constraint is that
+every share must be processed exactly once by all aggregators. Data integrity
```
tgeoghegan, line 1100: Distributed systems people will hiss at you for saying "exactly once". Should this be "at most once"?

draft-pda-protocol.md
tgeoghegan, line 1106: "AFE" turns up in a handful of places in the doc now. Perhaps it merits an entry in the `Terminology` section at the top, especially since it's not obvious that "AFE" expands to "affine-aggregatable encodings".

draft-pda-protocol.md
```diff
@@ -950,31 +956,68 @@ indefinitely until a suitable output is found.
...
+
+- Support the upload protocol and store client reports for a given PDAParam instance,
+  where each report maps uniquely to a single batch, and index this storage by batch durations;
+- Track batch report size during each collect flow and request encrypted output shares
+  from helpers.
+
+In addition, for each PDAParam instance, leaders are required to:
+
+- Ensure that, for each report, all aggregator shares are either included or not in a
+  given batch;
```
tgeoghegan, line 1010: I find this a bit confusing. Is the intent here that for a report, the corresponding aggregator *output* shares must either both be included or both be omitted?

draft-pda-protocol.md
```diff
@@ -1057,11 +1078,40 @@ order to achieve these goals:
...
+
+## Data integrity constraints
+
+Data integrity concerns the accuracy and correctness of the outputs in the
+system. The integrity of the output can be influenced by an incomplete round of
+aggregation caused by network partitions, or by bad actors attempting to cause
+inaccuracies in the aggregates. An example data integrity constraint is that
+every share must be processed exactly once by all aggregators. Data integrity
+constraints may be at odds with the threat model if meeting the constraints
+requires replaying data.
```
cjpatton, line 1102: > Data integrity constraints may be at odds with the threat model if meeting the constraints requires replaying data.

I realize this statement was copy-pated, but I think we should remove it. The origin of this sentence is the observation that Hits requires multiple collect rounds over the same data. This fact isn't at odds with the need for anti-replay, since mutliple collect requests aren't "replaying" reports in an illegal way.

cjpatton (changes requested):

draft-pda-protocol.md
```diff
@@ -561,7 +566,8 @@ aggregator.]
 
 The leader responds to well-formed requests to `[leader]/upload` with status 200
 and an empty body. Malformed requests are handled as described in
-{{pa-error-common-aborts}}.
+{{pa-error-common-aborts}}. Clients MUST NOT upload the same measurement value in
+more than one report if the leader responds with status 200 and an empty body.
```
tgeoghegan, line 570: It seems like this would require clients to maintain a list of uploaded reports to avoid double-uploads, which contradicts the later statement:
>Clients are not expected to store any state across any upload flows

Besides which, I think leaders will need to implement detection of duplicate report uploads no matter what, in which case they can return an error to clients in the case of a double upload, which downgrades this to a SHOULD NOT.

draft-pda-protocol.md
chris-wood: That's just par for the course -- I don't think it changes any of the capacity or operational requirements?

draft-pda-protocol.md
chris-wood: Yep, noted it as such.

draft-pda-protocol.md
chris-wood: This is moved text, so let's change that separately.

draft-pda-protocol.md
chris-wood: Nope, we can delete.

draft-pda-protocol.md
chris-wood: Less text is better :-) I'll remove it.

draft-pda-protocol.md
chris-wood: Hah, yeah, fair :-) That's a good suggestion!

draft-pda-protocol.md
cjpatton: In fact, I'm not so sure we want to bolt ourselves to this term too much. First, it's not meaningful for every PDA protocol (for Hits in particular). Second, the term was coined in the original Prio paper, which we don't actually implement.

draft-pda-protocol.md
chris-wood: I don't think that's necessary to do here.

draft-pda-protocol.md
chris-wood: (Overcome by events)

draft-pda-protocol.md
chris-wood: Yeah, this seems reasonable.

draft-pda-protocol.md
```diff
@@ -1001,31 +1007,70 @@ indefinitely until a suitable output is found.
...
+Helpers and leaders have different operational requirements. The design in this
+document assumes an operationally competent leader, i.e., one that has no storage
+or computation limitations or constraints, but only a modestly provisioned helper, i.e., one that
+has computation, bandwidth, and storage constraints. By design, leaders must be
+at least as capable as helpers, where helpers are generally required to:
 
-## Open questions and system parameters {#questions-and-params}
+- Support the collect protocol, which includes verifying and aggregating
+  sets of reports in a given batch; and
+- Publish and manage an HPKE configuration that can be used for the upload protocol;
```
cjpatton, line 1041:
```suggestion
- Publish and manage an HPKE configuration that can be used for the upload protocol.
```

draft-pda-protocol.md
```diff
@@ -1001,31 +1007,70 @@ indefinitely until a suitable output is found.
...
 
-[[OPEN ISSUE: discuss batch size parameter and thresholds]]
-[[OPEN ISSUE: discuss f^ leakage differences from [GB17]]]
+In addition, for each PDAParam instance, helpers are required to:
 
+- Implement some form of batch-to-report index, as well as inter- and intra-batch
+  replay mitigation storage, which includes some way of tracking batch report size
+  with optional support for state offloading. Some of this state may be used for
+  replay attack mitigation. One relay mitigation strategy is described in {{anti-replay}}.
```
cjpatton, line 1048:
```suggestion
  replay attack mitigation. One replay mitigation strategy is described in {{anti-replay}}.
```
Also, isn't the strategy in {{anti-replay}} *the* strategy? Do we intend to allow deployments to skip this, or do something else?

draft-pda-protocol.md
cjpatton: +1

draft-pda-protocol.md
```diff
@@ -1001,31 +1007,70 @@ indefinitely until a suitable output is found.
...
+- Support the upload protocol and store client reports for a given PDAParam instance,
+  where each report maps uniquely to a single batch, and index this storage by batch durations;
+- Track batch report size during each collect flow and request encrypted output shares
+  from helpers.
+
+In addition, for each PDAParam instance, leaders are required to:
+
+- Ensure that, for each report, all aggregator shares are either included or not in a
+  given batch;
+- Implement and store state for some form of inter- and intra-batch replay mitigation; and
```
cjpatton, line 1061: Again, at this point it seems like there is one and only one replay mitigation, and it's spelled out in the doc.

cjpatton (commented):
Some nits, and a question: At this point, isn't it our intent to spell out the anti-replay mechanism? These changes suggest *some* mechanism is required, but isn't more prescriptive than that.

chris-wood:
> At this point, isn't it our intent to spell out the anti-replay mechanism? These changes suggest some mechanism is required, but isn't more prescriptive than that.

Indeed, the text doesn't require any specific mitigation, and I don't see why it should. Aggregators could do the thing described in the appendix, or they could keep a huge list of all reports seen thus far. There could even be some simpler variant in the future 🤷 I don't think we lose anything by eliding the implementation details here.

draft-pda-protocol.md
chris-wood: (Commented below.)

cjpatton (approved):

draft-pda-protocol.md
```diff
@@ -1001,31 +1007,70 @@ indefinitely until a suitable output is found.
...
+
+- Ensure that, for each report, all aggregator shares are either included or not in a
+  given batch;
+- Implement and store state for the form of inter- and intra-batch replay mitigation in {{anti-replay}}; and
+- Store helper state associated with a given PDAParam batch.
+
+### Collector capabilities
+
+Collectors statefully interact with aggregators to produce an aggregate output. Their
+input to the protocol is the PAParam structure, configured out of band, which contains
```
tgeoghegan, line 1067:
```suggestion
input to the protocol is the PDAParam structure, configured out of band, which contains
```

tgeoghegan (changes requested):
I flagged one typo and would like to see the text referred to in [this comment](https://github.com/abetterinternet/prio-documents/pull/93/files#r679473812) clarified.

draft-pda-protocol.md
chris-wood: This is indeed somewhat confusing, and probably not needed, so dropped.

tgeoghegan (approved):

Issue #89, opened 2021-07-21
Discuss client attestation in security considerations


cjpatton:
Many deployments will be in a position to have clients attest that (1) the report was generated by a trusted client or (2) the report was generated within a trusted execution environment. (1) allows you to mitigate Sybil attacks by only consuming reports generated by trusted clients (#86, #20). On top of that, with (2) you might choose to forego expensive MPC for validating inputs altogether (#45). We don't want to require these features for all deployments, but they're useful enough that we should make sure our protocol accommodates them.

With that in mind, it would be useful if folks that plan to do this would provide some details about how they envision composing (1) or (2) with PDA**. The question to answer is what changes, if any, are needed for the core spec.

cc/ @tgeoghegan, @csharrison, and @hostirosti

** 2022/1/18: Note that "PDA" was the working name for the PPM protocol at the time this issue was initially discussed.

csharrison:
For our use-case we were thinking of something like [Privacy Pass](https://datatracker.ietf.org/wg/privacypass/documents/) for anonymous attestation. However, this was in the context of a collector / proxy sitting between the client and the leader who is responsible for issuing and redeeming tokens. This design would put all attestation outside of the PDA protocol, and allow the collector to determine which clients are trusted. However, it only benefits the collector which has a different set of requirements from the PDA system.

Collectors want to make sure their data is accurate. The PDA system wants to ensure privacy. To that end I could imagine two layers of attestation:
1. A layer of attestation that records are trusted by the collector
2. A layer of attestation that records are trusted by PDA

However, I don't have any concrete ideas on how to accomplish (2) for our use-case realistically.

cjpatton:
> For our use-case we were thinking of something like [Privacy Pass](https://datatracker.ietf.org/wg/privacypass/documents/) for anonymous attestation. However, this was in the context of a collector / proxy sitting between the client and the leader who is responsible for issuing and redeeming tokens. This design would put all attestation outside of the PDA protocol, and allow the collector to determine which clients are trusted. 

This batched upload idea is still on the table, I think: see https://github.com/abetterinternet/prio-documents/issues/64. In particular, I think we're all in agreement that some collector/proxy should be able to opaquely carry out some client attestation thing and upload only "trusted" reports to the leader. (@tgeoghegan calls this the "ingestor, I believe.) The problem, as you point out, is the trust relationship between the leader and uploader.

[...]

PR #94, opened 2021-07-29
Add an extension slot to the upload request.

chris-wood:
One possible extension would be an opaque blob to carry client attestation information (for #89), if desired. The current set is empty, but I suspect we'll need freedom to extend here as this gets used without totally changing the protocol version.

If folks are generally supportive of this, it should probably land after #90, which touches and renames PDAUploadReq to PDAReport.

### DIFF
--- a/draft-pda-protocol.md
+++ b/draft-pda-protocol.md
@@ -509,6 +509,7 @@ struct {
   PDATaskID task_id;
   Time time;
   uint64 jitter;
+  Extension extensions<4..2^16-1>;
   PDAEncryptedInputShare encrypted_input_shares<1..2^16-1>;
 } PDAReport;
 ~~~
@@ -522,6 +523,8 @@ This message is called the client's *report*. It contains the following fields:
 * `jitter` is a random number chosen by the client generating the report. This
   and the timestamp field are used by the aggregators to ensure that each report
   appears at most once in a batch. (See {{anti-replay}}.)
+* `extensions` is a list of extensions to be included in the Upload flow; see
+  {{upload-extensions}}.
 * `encrypted_input_shares` contains the encrypted input shares of each of the
   aggregators.
 
@@ -558,7 +561,7 @@ the helper. `enc` is the encapsulate HPKE context and `context` is the HPKE
 context used by the client for encryption. The payload is encrypted as
 
 ~~~
-payload = context.Seal(input_share, task_id || time || jitter)
+payload = context.Seal(input_share, task_id || time || jitter || extensions)
 ~~~
 
 where `input_share` is the aggregator's input share and `task_id`, `time`, and
@@ -567,6 +570,27 @@ where `input_share` is the aggregator's input share and `task_id`, `time`, and
 The leader responds to well-formed requests to `[leader]/upload` with status 200
 and an empty body. Malformed requests are handled as described in
 {{pa-error-common-aborts}}.
+
+### Upload Extensions {#upload-extensions}
+
+Each PAUploadReq carries a list of extensions that clients may use to convey
+additional, authenticated information in the report. Each extension is a tag-length
+encoded value of the following form:
+
+~~~
+  struct {
+      ExtensionType extension_type;
+      opaque extension_data<0..2^16-1>;
+  } Extension;
+
+  enum {
+      TBD(0),
+      (65535)
+  } ExtensionType;
+~~~
+
+"extension_type" indicates the type of extension, and "extension_data" contains
+information specific to the extension.
 
 ## Collect {#pa-collect}
 
@@ -707,8 +731,9 @@ unique client report. Sub-requests are structured as follows:
 
 ~~~
 struct {
-  Time time;     // Equal to PDAReport.time.
-  uint64 jitter; // Equal to PDAReport.jitter.
+  Time time;                       // Equal to PDAReport.time.
+  uint64 jitter;                   // Equal to PDAReport.jitter.
+  Extension extensions<4..2^16-1>; // Equal to PDAReport.extensions.
   PDAEncryptedInputShare helper_share;
   select (PDAParam.proto) { // PDAParam for the PDA task
     case prio: PrioAggregateSubReq;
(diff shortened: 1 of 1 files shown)


draft-pda-protocol.md
```diff
@@ -508,14 +508,20 @@ the leader's endpoint URL. The payload is structured as follows:
 struct {
   PDATaskID task_id;
   Time time;
+  Extension extensions<8..2^16-1>;
```
ekr, line 511: Isn't this actually 4..2^16-1

ekr (approved):

draft-pda-protocol.md
chris-wood:
```suggestion
  Extension extensions<4..2^16-1>;
```

draft-pda-protocol.md
```diff
@@ -508,14 +508,20 @@ the leader's endpoint URL. The payload is structured as follows:
...
   PDATaskID task_id;
   Time time;
+  Extension extensions<4..2^16-1>;
   PDAEncryptedInputShare encrypted_input_shares<1..2^16-1>;
 } PDAUploadReq;
 ~~~
 
+[OPEN ISSUE: if we include extensions in the AAD then these extensions are required to be
+echoed in each aggregate subrequest. We could either not include this in the AAD, or replace
+the list with a hash of the list (that's sent separately), or something else.]
```
cjpatton, line 518: In fact, when the extension is used for the purpose of attestation, it probably *needs* to be echoed in each aggregate sub-request.

draft-pda-protocol.md
```diff
@@ -563,6 +569,27 @@ The leader responds to well-formed requests to `[leader]/upload` with status 200
 and an empty body. Malformed requests are handled as described in
 {{pa-error-common-aborts}}.
 
+### Upload Extensions {#upload-extensions}
+
+Each PAUploadReq carries a list of extensions that clients may use to convey
+additional, authenticated information in the report. Each extension is a tag-length
```
cjpatton, line 575:
```suggestion
additional, authenticated[TODO: Specify how] information in the report. Each extension is a tag-length
```

cjpatton (changes requested):
I'm supportive of solving #89 in a generic way, but it's not clear that this change has enough plumbing. In particular, there doesn't appear to be a way for the helper to verify either property of the report (1 - the report was generated by a trusted client; 2 - the report was generated in a trusted environment) Minimally, I think we would need to echo the extension in the aggregate sub-request.

I'd be happy with one of of the following outcomes:
- Copy the extension into the aggregate sub-request and add an OPEN ISSUE for deciding whether this is sufficient plumbing for client attestation
- Close this PR and park the issue

draft-pda-protocol.md
chris-wood: The contents of the extensions are presumably fed into the AAD, so closing.

draft-pda-protocol.md
chris-wood: That is where I lean, too, but it does mean extension bloat in each request. That's probably fine, though.

chris-wood:
Both (1) and (2) are _external to the protocol_. The purpose of extensions is to allow whatever mechanism exists for addressing them to be bound to the report, as is done here. 

> Copy the extension into the aggregate sub-request and add an OPEN ISSUE for deciding whether this is sufficient plumbing for client attestation

This is done in the issue. This is _not meant_ to close the attestation issue. It's meant to give us the mechanics to address it later.

cjpatton (approved):
Right you are. What threw me off was the following statement:
> [OPEN ISSUE: if we include extensions in the AAD then these extensions are required to be
> echoed in each aggregate subrequest. We could either not include this in the AAD, or replace
> the list with a hash of the list (that's sent separately), or something else.]

The "then these extensions are required to be echoed..." makes it sound as if they're *not* echoed, which they are. I'm happy with this PR modulo fixing this statement.

draft-pda-protocol.md
```diff
@@ -703,6 +730,7 @@ unique client report. Sub-requests are structured as follows:
 ~~~
 struct {
   Time time; // Equal to PDAReport.time.
+  Extension extensions<8..2^16-1>;
```
tgeoghegan, line 733: Any reason this is `<8..2^16-1>` here but `<4..2^16-1>` in `PDAUploadReq`?

draft-pda-protocol.md
```diff
@@ -1282,6 +1310,11 @@ the input shares and defeat privacy.
 
 # IANA Considerations
 
-TODO
+## Upload Extension Registry
+
+This document requests creation of a new registry for extensions to the Upload
+protocol. This registry should contain the following columns:
+
+[TODO: define how we want to structure this registry when the time comes]
```
tgeoghegan, line 1318: Do we need to provide this much structure for the extensions? IIUC at the PDA protocol layer, we will never examine or modify these extensions so why can't they be completely opaque blobs?

draft-pda-protocol.md
chris-wood: Me not making the change in two places :)

chris-wood:
```suggestion
  Extension extensions<4..2^16-1>;
```

draft-pda-protocol.md
chris-wood: This just creates the registry where extensions will go. We don't need to do anything more here, unless we decide we want to introduce some mandatory extensions.

tgeoghegan: Well, my point applies more to the `struct Extension` definition. The extension just be an opaque blob of bytes, and the responsibility of protocol implementations would be to relay the blob unmodified from `PDAReport` to `PDAAggregateSubReq`. Then the format of that blob is something to be agreed upon by the client that generated it and the aggregator that is examining it, out of band from the PDA protocol.

chris-wood: The contents of the extension, and how they're used, are up to the thing that defines the extension. So, yeah, I think what we have here aligns with what you envision.

PR #96, opened 2021-08-04
Draft charter

ekr:


charter.md
```diff
@@ -0,0 +1,45 @@
+There are many situations in which it is desirable to take
+measurements of data which people consider sensitive. For instance,
+one might want to measure the most common sites that people browse to
+or exposure to some disease. In these cases, the entity taking the
+measurement is not interested in people's individual responses but
+rather in aggregated data (e.g., how many users visit URL X), but
+conventional methods require collecting individual response and then
```
stpeter, line 7:
```suggestion
conventional methods require collecting individual responses and then
```

charter.md
```diff
@@ -0,0 +1,45 @@
...
+aggregating it, thus representing a threat to user privacy and
```
stpeter, line 8:
```suggestion
aggregating them, thus representing a threat to user privacy and
```

charter.md
```diff
@@ -0,0 +1,45 @@
...
+rendering many such measurements difficult and impractical.
+
+New cryptographic techniques such as Prio and Heavy Hitters address
+this gap by splitting up measurements between multiple servers which
+can jointly compute the aggregate value without either server learning
+the value of individual responses. The Private Data Aggregation (PDA)
+work will standardize a protocol for deployment of these techniques on
+the Internet. This protocol will include mechanisms for:
+
+- Client submission of individual reports, including proofs of validity.
```
stpeter, line 18:
```suggestion
- Client submission of individual reports, including proofs of validity, to a primary server.
```

charter.md
```diff
@@ -0,0 +1,45 @@
...
+- Server verification of validity proofs
```
stpeter, line 19:
```suggestion
- Verification of validity proofs by two or more servers
```

charter.md
```diff
@@ -0,0 +1,45 @@
...
+- Server computation of aggregate values and collection of results to
```
stpeter, line 20:
```suggestion
- Computation of aggregate values by two or more servers
```

charter.md
```diff
@@ -0,0 +1,45 @@
...
+  the entity taking the measurement
```
stpeter, line 21:
```suggestion
- Reporting of aggregate results to the entity taking the measurement
```

charter.md
ekr: I actually don't think we should commit to this in the charter (even though I think it's a good idea).

charter.md
ekr: Same here.

charter.md
ekr: Maybe "by the servers" here and below.

charter.md
```diff
@@ -0,0 +1,45 @@
...
+measurements of data which people consider sensitive. For instance,
+one might want to measure the most common sites that people browse to
+or exposure to some disease. In these cases, the entity taking the
+measurement is not interested in people's individual responses but
+rather in aggregated data (e.g., how many users visit URL X), but
+conventional methods require collecting individual responses and then
+aggregating them, thus representing a threat to user privacy and
+rendering many such measurements difficult and impractical.
+
+New cryptographic techniques such as Prio and Heavy Hitters address
```
cjpatton, line 11: "Heavy Hitters" isn't the name of the protocol in ePrint 2021/017. "Prio" is the name of a protocol, but we don't implement it. (We actually implement ePrint 2019/088.) Rather than name the protocols, I suggest we just cite the papers.

charter.md
```diff
@@ -0,0 +1,45 @@
...
+- Server verification of validity proofs
+- Server computation of aggregate values and collection of results to
+- Reporting of aggregate results to the entity taking the measurement
+  
+Configuration of clients and servers is out of scope for the working
+group. It is assumed that this happens out of band as part of the
+PDA service. 
+
+The WG will deliver a protocol which can accommodate multiple PDA
+algorithms, with the initial deliverable supporting both simple
```
cjpatton, line 28:
```suggestion
The WG will deliver a protocol which can accommodate multiple PDA
tasks, with the initial deliverable supporting both simple
```

charter.md
ekr: I think perhaps we just say "Prio". We're not saying we're going to standardize that, we're giving an example.

ekr: Also, I'm trying to get HCG to just rename all this stuff Prio

charter.md
```diff
@@ -0,0 +1,45 @@
...
+- Client submission of individual reports, including proofs of validity.
+- Server verification of validity proofs
+- Server computation of aggregate values and collection of results to
+- Reporting of aggregate results to the entity taking the measurement
+  
+Configuration of clients and servers is out of scope for the working
+group. It is assumed that this happens out of band as part of the
+PDA service. 
+
+The WG will deliver a protocol which can accommodate multiple PDA
```
martinthomson, line 27: a protocol?  seems like something that a working group might decide, but HH and Prio are very different things functionally.  my take: make two, share as many design elements as possible

Issue #98, opened 2021-08-10
The wrapper-core protocol interface


cjpatton:
On the last design call (2021/08/04) we decided to split our work into two (sets of) documents: one for the "core" cryptographic protocols (Prio, Hits, ...) and another for the "wrapper" protocol (basically the protocol bits we've been working on so far). The next step is to specify the "interface" between these (sets of) documents. This is our goal for this issue.

So far we've been reasoning about this core-wrapper interface by analogy of HPKE and its applications (ECH, OHTTP, etc.), which may indeed be an apt analogy, but it falls short of a precise definition. To that end, I'd like to suggest that the interface needs to accomplish two things.

- (1) Specify the syntax of every PDA protocol. The syntax should cover:
        - (1a) generation of input shares from a measurement;
        - (1b) the state machine of the leader and helper and how they respond to messages (in the core protocol);
        - (1c) accumulation of input shares into an output share; and
        - (1d) combination of output shares into the output.
- (2) Specify the security requirements of every PDA protocol. This has three main components:
        - (2a) The expected _communication model_, which determines how much control over the network the adversary has. (This may be different from the communication model of the wrapper protocol, as discussed below.)
        - (2b) A baseline notion of privacy in the expected communication model that allows any number of clients and all but one aggregator to be corrupted.
        - (2c) A baseline notion of robustness in the expected communication that allows any number of corrupt clients to be corrupted, but requires all aggregators to be honest.

Much of what we need is already worked out, but there are some gaps that require us to make some decisions. Two such gaps are discussed below.

## (2a) Communication model

As observed in #86, we've been considering attacks outside the formal model of either Prio or Hits (specifically Sybil and replay attacks). The communication model for the core protocol may therefore differ from (i.e., make stronger assumptions about the adversary than) the communication model for the wrapper protocol. I would argue that this is perfectly reasonable. In fact, the main job of the wrapper protocol should be to "emulate" whatever communication model is expected by the core protocol.

We are already well on our way here. The following mechanisms are used to provide properties that are implicit in the communication models of the papers:
- HPKE encryption creates a confidential channel between the client and each honest aggregator.
- The report's timestamp and jitter fields define a total ordering on reports, which the aggregators can use to ensure each report is counted in at most one batch.

Some bits are still missing, however:
- Who manages the report queue? It seems simplest if the core protocol can assume in-order, non-repeated reports. The wrapper protocol can then do this enforcement. 
[...]

PR #99, opened 2021-08-11
typos, formatting errors, obsolete todos

tgeoghegan:
Fixes a miscellany of typos, formatting errors and obsolete todos I encountered while reviewing the document.


draft-pda-protocol.md
```diff
@@ -479,13 +475,6 @@ Client          Leader         Helper
 ~~~~
 {: #pa-upload-flow title="Flow of the upload process"}
 
-[NOTE: @acmiyaguchi pointed out that the use of an anonymizing proxy for
-uploading shares might be easier to implement if the "upload" phase involved a
-single HTTP request. However, OHTTP
-(https//www.ietf.org/archive/id/draft-thomson-http-oblivious-01.html) allows
-clients to make multiple requests through a proxy, so these kinds of use cases
-should work.]
```
tgeoghegan, line 487: We have a section in security considerations about OHTTP and anonymizing proxies so I think this is settled.

draft-pda-protocol.md
```diff
@@ -504,12 +493,6 @@ request:
 * the key config specifies a KEM, KDF, or AEAD algorithm the client doesn't
   recognize.
 
-[OPEN ISSUE: @chris-wood: Can't the leader determine if helpers are "online"?
-This seems to reveal information that's specific to clients. Imagine, for
-example, that clients are prohibited from talking to helpers but not the leader.
-Is it OK that leaders learn that about a client? I'm not sure, so I'd be
-inclined to remove this unless we have a concrete use case.]
-
```
tgeoghegan, line 512: IIUC the concern here is that leaders or helpers can learn which clients are uploading data, and I believe deployments can address that by using OHTTP.

draft-pda-protocol.md
```diff
@@ -668,9 +651,6 @@ The batch window of the request is the interval `[batch_start, batch_end)`. A
 collect request is said to be valid if all of the following conditions hold (let
 `PDAParam` denote the parameters for the PDA task):
 
-1. The batch window of the request aligns with the size of the batch window for
-   the task, i.e., `batch_start` and `batch_end` are multiples of
-   `PDAParam.batch_window`.
```
tgeoghegan, line 673: I don't think this requirement is correct. If the minimum batch window is 8 hours, why does it matter if the collector wants to aggregate over 0400-1200 instead of 0000-0800? Maybe the intent here was that `batch_end - batch_start` be a multiple of `PDAParam.batch_window`, but I don't think that's necessary, either. I think it's OK if a collector wants to aggregate over a window larger than the minimum value in the PDAParams, so long as the next aggregate doesn't overlap. The aggregators implement anti-replay by keeping track of the newest report timestamp aggregated for a PDATask, so they don't need all the batch windows to be the same or to be multiples of the minimum, do they?

draft-pda-protocol.md
```diff
@@ -447,8 +443,8 @@ key.
 
 We assume the following conditions hold before execution of any PDA task begins:
 
-1. The aggregators agree on a set of PDA tasks, as well as the PDA parameters
-   associated to each task.
+1. The clients, aggregators and collector agree on a set of PDA tasks, as well
```
chris-wood, line 446:
```suggestion
1. The clients, aggregators, and collector agree on a set of PDA tasks, as well
```

draft-pda-protocol.md
chris-wood: Assuming the batch size criteria is met, this seems right.

draft-pda-protocol.md
```diff
@@ -984,18 +966,18 @@ for s in range(0,1000,1):
 
 ### Pseudorandom number generation {#prio-prng}
```
chris-wood, line 967: We might consider just punting this to the underlying CFRG document. Maybe we can move this section to the parking lot?

chris-wood (approved):

draft-pda-protocol.md
```diff
@@ -384,16 +380,16 @@ Time uint64; /* seconds elapsed since start of UNIX epoch */
...
 * `helper_url`: The helper's endpoint URL.
 * `collector_config`: The HPKE configuration of the collector (described in
   {{hpke-config}}). [OPEN ISSUE: Maybe the collector's HPKE config should be
   carried by the collect request?]
 * `batch_size`: The batch size, i.e., the minimum number of reports that are
   aggregated into an output.
-* `batch_window`: The batch window, i.e., the minimum time difference between
-  the oldest and newest report in a batch.
+* `batch_window`: The window of time covered by a batch, i.e., the maximum time
+  difference between the oldest and newest report in a batch.
```
cjpatton, line 392:
```suggestion
* `batch_window`: The window of time covered by a batch, i.e., the maximum interval
   between the oldest and newest report in a batch.
```

draft-pda-protocol.md
cjpatton: Yeah, this maybe wasn't very clear. The intent of this requirement is to divide time into intervals that the aggregators and collector agree on in advance of the collect request. This has the advantage of allowing aggregators to aggregate input shares in each window as they arrive, thereby reducing the aggregators' storage requirements. Note that this isn't possible for Hits, but it is for Prio.

draft-pda-protocol.md
```diff
@@ -937,13 +918,14 @@ criteria:
...
-   size of proofs, so it should be large enough to accommodate all foreseeable
-   use cases. Something like b >= 20 is probably good enough.
+   In particular, the prime modulus p should be chosen so that `(p-1) = 2^b * s`
+   for large `b` and odd `s`. Then `g^s` is a principle, `2^b`-th root of unity
+   (i.e., `g^(s\*2^b) = 1`), where `g` is the generator of the multiplicative
+   subgroup.
+   This fact allows us to quickly evaluate and interpolate polynomials at
+   `2^a`-th roots of unity for any `1 <= a <= b`. Note that `b` imposes an upper
+   bound on the size of proofs, so it should be large enough to accommodate all
+   foreseeable use cases. Something like `b >= 20` is probably good enough.
```
cjpatton, line 928: What changed here? Looks like we're just inserting a line break?

draft-pda-protocol.md
cjpatton: Yeah, I think we're going to end up removing the Prio stuff altogether. (See #98.) But we can do so in a future PR.

cjpatton (changes requested):
My only blocking comment is that I think we want (something like) the requirement you're removing. It's useful for the aggregators to be able to know the possible batch intervals before a collect request is made.

draft-pda-protocol.md
tgeoghegan: I'm putting `` around the math terms to make them fixed-width. I find it more readable if those stand out from the regular text, but I'm happy to revert.

cjpatton: Ack, no need!

draft-pda-protocol.md
tgeoghegan: Aggregators need to jointly evaluate the validity proof on a Prio input before it can be included in an aggregate, and proofs are not verified until the collect phase, which doesn't start until a collect request is emitted by the collector. I'm not sure how aggregators could speculatively aggregate input shares unless they are capable of removing an input share from their output if the proof is later found to be invalid, which would have its own storage requirements.

cjpatton: > Aggregators need to jointly evaluate the validity proof on a Prio input before it can be included in an aggregate, and proofs are not verified until the collect phase, which doesn't start until a collect request is emitted by the collector.

I don't this is quite right. My understanding is that the aggregators can verify an input's validity as long as they have the PDAParam. Waiting on the collect request is only necessary for Hits, where you need to know the set of candidate prefixes before you know how to proceed.

tgeoghegan: Following from our discussion in the design call earlier today, I've reverted this specific change so that @cjpatton can revisit this language later. I believe that we concluded that we want the batch windows in the collect requests to have the same size as the batch window in the PDAParams.

cjpatton (approved):

PR #119, opened 2021-08-17
PPMFoo->Foo. Fixed #116

### LINKED ISSUE #116: Drop "PPM" prefix for all protocol messages
Protocol messages are defined in the context of the PPM document, so the "PPM" seems redundant. I think we should just call things as they are, i.e., `TaskID` instead of `PPMTaskID`.

ekr:


cjpatton (changes requested):
Yesterday I spotted an old "PA" prefix that was missed when we did s/PA/PDA/ would you mind finding it and fixing it?

chris-wood (approved):

ekr:
@cjpatton I'm struggling with Github here and can't figure out what change you want.

cjpatton (approved):
Disregard, I found it and pushed myself.

ekr:
Thanks.

PR #128, opened 2021-08-18
Add boilerplate media type definitions for each protocol message.

### LINKED ISSUE #105: DefineMIME types for all of the things we are carrying over HTTP

chris-wood:
Closes #105

### DIFF
--- a/draft-pda-protocol.md
+++ b/draft-pda-protocol.md
@@ -495,7 +495,7 @@ keys for a reasonable period after key changes in order to avoid
 rejecting reports.
 [[OPEN ISSUE: https://github.com/abetterinternet/prio-documents/issues/106]]
 
-### Report
+### Upload Request
 
 Clients upload reports by using an HTTP POST `[leader]/upload`, where
 `[leader]` is the leader's endpoint URL. The payload is structured as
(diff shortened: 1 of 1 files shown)


chris-wood:
@mnot -- is it better practice to define media types per protocol message, or to define a single media type for all protocol messages?

mnot:
Generally, if they're different formats -- e.g., they have different schema, or just different syntax -- they need different media types (or an explicitly generic one like `application/octet-stream`). If it's one format that has different uses, you can give it a single media type and then switch contexts based upon internal flags.

The real question, though, is whether you have any potential for use cases where it's important for generic software that's handling these messages to do different things based upon the media type.

chris-wood:
> Generally, if they're different formats -- e.g., they have different schema, or just different syntax -- they need different media types (or an explicitly generic one like application/octet-stream). If it's one format that has different uses, you can give it a single media type and then switch contexts based upon internal flags.
>
> The real question, though, is whether you have any potential for use cases where it's important for generic software that's handling these messages to do different things based upon the media type.

I don't think there'd ever be generic software using these types (beyond the HpkeConfig?). Most messages are meant to be consumed by PPM and PPM alone. I moved protocol messages to the "message" registry space and kept the HpkeConfig in the application type space, and also left all intended usage fields as COMMON. I also added an open issue to have someone more familiar with these registries give these allocation requests a review.

@ekr, this should be good to go based on the above.

draft-pda-protocol.md
```diff
@@ -495,7 +495,7 @@ keys for a reasonable period after key changes in order to avoid
 rejecting reports.
 [[OPEN ISSUE: https://github.com/abetterinternet/prio-documents/issues/106]]
 
-### Upload Request
+### Report
```
ekr, line 498: This hed seems a bit short. What's your thinking here.

ekr (approved):
LGTM except for the hed.

draft-pda-protocol.md
chris-wood: Only that it was more clear, but it's a separate change, so I'll just revert.

PR #129, opened 2021-08-18
Cache using HTTP headers, invalidate and retry upon upload.

### LINKED ISSUE #106: Caching for HpkeConfig
It seems silly to have clients revalidating the HPKE key for each request, so we probably want some kind of caching. I'm not sure HTTP Caching is quite what we want, but something... 

@martinthomson @mnot recommendations?

chris-wood:
Closes #106.

### DIFF
--- a/draft-pda-protocol.md
+++ b/draft-pda-protocol.md
@@ -474,8 +474,8 @@ uint16 HpkeAeadId; // Defined in I-D.irtf-cfrg-hpke
 uint16 HpkeKemId;  // Defined in I-D.irtf-cfrg-hpke
 uint16 HpkeKdfId;  // Defined in I-D.irtf-cfrg-hpke
 ~~~
-[TODO: Decide whether to use the same config structure as OHTTP/ECH. This would
-add support for multiple cipher suites.]
+
+[OPEN ISSUE: Decide whether to expand the width of the id, or support multiple cipher suites (a la OHTTP/ECH).]
 
 The client issues a key configuration request to `Param.leader_url` and
 `Param.helper_url`. It aborts if any of the following happen for either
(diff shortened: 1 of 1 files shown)


draft-pda-protocol.md
```diff
@@ -485,15 +487,22 @@ request:
 * the key config specifies a KEM, KDF, or AEAD algorithm the client doesn't
   recognize.
 
+Aggregators SHOULD use the HTTP caching to permit client-side caching of this
```
tgeoghegan, line 490:
```suggestion
Aggregators SHOULD use HTTP caching to permit client-side caching of this
```

draft-pda-protocol.md
```diff
@@ -571,6 +580,12 @@ and an empty body. Malformed requests are handled as described in
 {{errors}}. Clients SHOULD NOT upload the same measurement value
 in more than one report if the leader responds with status 200 and an empty body.
 
+The leader responds to requests with out-of-date `HpkeConfig.id` values, indicated
+by `EncryptedInputShare.config_id`, with status 400 and an error of type
```
tgeoghegan, line 584: `HpkeConfig.id` is a uint8, meaning that any server can have 256 different keys, ever. If the lifetime of any HPKE key is on the order of days (which is what I infer from the suggested `Cache-Control` value above), and we don't allow recycling of `HpkeConfig.id`, then won't servers run out of key configs in just a few years? Which is a roundabout way of asking: should `HpkeConfig.id` be bigger than 8 bits?

tgeoghegan (changes requested):

draft-pda-protocol.md
ekr: That seems like probably the right approach, though I suppose we could recycle.

chris-wood: Indeed -- the current design requires aggregators to recycle. Since this is orthogonal to caching the config, let's file an open issue to address separately?

chris-wood: (We had an issue tracking HpkeConfig updates, so I just updated that issue accordingly.)

tgeoghegan (approved):

PR #134, opened 2021-08-18
use `urn:ietf:params:ppm` sub-namespace

tgeoghegan:
I think this is correct given the namespace reserved in `#ppm-urn-space`.


chris-wood (approved):

chris-wood:
Good catch!

Issue #133, opened 2021-08-18
How does the leader know which `EncryptedInputShare` in a report is which?


tgeoghegan:
We have this definition:
```
struct {
  TaskID task_id;
  Time time;
  uint64 nonce;
  Extension extensions<4..2^16-1>;
  EncryptedInputShare encrypted_input_shares<1..2^16-1>;
} Report;
<...>
* `encrypted_input_shares` contains the encrypted input shares of each of the aggregators.
```
The intent is to eventually allow an arbitrary number of aggregators, but how is the leader meant to know which `EncryptedInputShare` goes to which aggregator? Is share 0 understood to be the leader's, and share 1 is the helper's?

ekr:
Yeah, that's 0 and 1. But 3 and 4? LOL. 

Maybe we should do:

```
struct {
   AggregatorId aggregator;
   EncryptedInputShare encrypted_input_shares<1..2^16-1>;
} AnInputShare;
```

Where ```AggregatorId``` is  TBD.

tgeoghegan:
Ooh, vectors of labeled tuples are much better than my idea of requiring consistent ordering of per-aggregator values across `Report`, `Param` and `CollectResp`

tgeoghegan:
This is (going to be) resolved by #135.

Issue #104, opened 2021-08-12
Does PPMParam need to be defined at all?


ekr:
The current specification describes a PPMParam struct and defines a TaskID as SHA-256(PPMParam). However, because the PPMParam is never rendered on the wire. Rather it is just serialized in order to be fed into SHA-256. The parameters are provided to the endpoints out of band.

Consider a client implementor who behaves as follows:

- Create a JSON structure J that defines the parameters.
- Translate J into PPAParam P.
- Send the client the pair [J, TaskID = SHA-256(P)]

You'll note that the client never sees the PPAParam at all. As far as I can tell, this would be a completely conformant implementation. Similar comments apply to the aggregators as we also do not define how *they* are configured. So, similarly they could just be sent JSON and have a TaskID -> JSON lookup table. This isn't just a possible implementation, but it's likely to be a common one, because people have whatever central configuration system they have and don't necessarily want to carry around some binary blob.

In fact, the only reason people need to see the PPAParam structure at all as opposed to the information in it is to hash it to create TaskID (yes, i appreciate that I suggested the hashing thing).

So what I suggest is we merely make a list of the things that the client and the aggregators need to know respectively and not define any structure that contains that list. Then we can just pick the TaskID randomly.

When we originally discussed hashing, the idea was to have parameters be transitively incorporated into the protocol to avoid situations where each side is using different parameters. It's not clear to me that this is that important. A malicious client can just lie and use its own parameters, and if there is a bug, then it's just as likely it involves misinterpreting the parameters.

tgeoghegan:
I think our goal was to force clients to commit to a set of PPMParams, but you're right, nothing really ties the actual behavior of clients or aggregators to the PPMParams. We put a 16 byte nonce into PPMParam with the intention of guaranteeing that PPMTaskID would be unique, but I think we can just instead have PPMTaskID be that nonce. As for the sizes, we picked len(PPMTaskID) = 32 to match the length of a SHA-256, and len(PPMParam.nonce) == 16 to match a v4 UUID. If PPMTaskID is going to be randomly generated instead of derived, then 16 bytes should suffice.

cjpatton:
> So what I suggest is we merely make a list of the things that the client and the aggregators need to know respectively and not define any structure that contains that list. Then we can just pick the TaskID randomly.
> 
[...]

PR #103, opened 2021-08-11
clarify `PPMParam.collector_config` field

tgeoghegan:
Removes some TODOs, clarifies why we have `HpkeConfig
collector_config` in `struct PPMParam` instead of `Url collector_url`,
and amends collect protocol so that leader output shares are
encrypted to collector's HPKE config. Several response structures
no longer have `PPMProto` or `PPMTaskID` fields if the appropriate
task (and hence `PPMParam` and hence `PPMProto`) can be inferred
from the request.

PR #135, opened 2021-08-19
Amend PDUs to support multiple helpers

### LINKED ISSUE #117: Update protocol to generically allow >1 helper

tgeoghegan:
Modify `struct Param`, `struct Report` and `struct CollectResp` to
accommodate more than one helper. Note that the PPM protocol doesn't
actually support multiple helpers to enhance privacy (#68) but we
nonetheless want to define the PDUs so it is possible to introduce that
feature in the future. We do this by introducing an `AggregatorId` type
into various struct definitions, allowing participants to tell which
aggregator an input share or output share corresponds to. We also
introduce a special, reserved value of `AggregatorId` that allows
clients to tell which aggregator is the leader to which reports should
be uploaded. Finally, we use the aggregator ID in HPKE context
construction instead of the existing constants 0x00 and 0x01 for leader
and helper, respectively.

Along the way, we introduce the `HpkeConfigId` type alias used in few
different places.

Resolves #117, #133


tgeoghegan:
Note that this is stacked on #103 (sorry).

- [ ] make sure this PR targets `main` before merging

chris-wood:
@tgeoghegan I merged #103, so we can retarget this one now.

tgeoghegan:
> @tgeoghegan I merged #103, so we can retarget this one now.

Thanks! This is retargeted onto `main` and ready for review.

draft-pda-protocol.md
```diff
@@ -437,6 +434,33 @@ Time uint64; /* seconds elapsed since start of UNIX epoch */
...
+
+opaque Url<1..2^16-1>;
+~~~
+
+* `id`: An identifier for the aggregator, used in HPKE encryption and which
+  allows participants to resolve entries in `Param.aggregators` against values
+  in vectors in other protocol messages. The same `AggregatorId`s MUST be used
+  when consturcting other messages containing vectors with one value for each
+  aggregator (e.g., `Report.encrypted_input_shares` in {{uploading-reports}}).
+  Each aggregator knows its own `id`. `id` values must be unique in a `Param`.
```
chris-wood, line 457:
```suggestion
  Each aggregator knows its own `id`. `id` values MUST be unique in a `Param`.
  A `Param` structure with duplicate `id` values is considered malformed and MUST
  be ignored by clients.
```

draft-pda-protocol.md
```diff
@@ -437,6 +434,33 @@ Time uint64; /* seconds elapsed since start of UNIX epoch */
...
+* `endpoint`: The URL relative to which an aggregator's API endpoints can be
+  found.
+* `leader_aggregator_id` is a reserved `AggregatorId` identifying the leader.
+  `leader_aggregator_id` MUST occur exactly once among the `aggregators` in a
+  `Param`.
```
chris-wood, line 462:
```suggestion
  `Param`. A `Param` structure without the `leader_aggregator_id` is considered malformed
   and MUST be ignored by clients.
```

draft-pda-protocol.md
```diff
@@ -437,6 +434,33 @@ Time uint64; /* seconds elapsed since start of UNIX epoch */
...
 
+An `Aggregator` is structured as follows:
+
+~~~
+struct {
+  AggregatorId id;
+  Url endpoint;
+} Aggregator;
+
+uint8 AggregatorId;
```
ekr, line 445: I would consider making this larger. I could imagine there being > 256 aggregators.

draft-pda-protocol.md
```diff
@@ -437,6 +434,33 @@ Time uint64; /* seconds elapsed since start of UNIX epoch */
...
+
+AggregatorId leader_aggregator_id = 0;
+
+opaque Url<1..2^16-1>;
+~~~
+
+* `id`: An identifier for the aggregator, used in HPKE encryption and which
+  allows participants to resolve entries in `Param.aggregators` against values
+  in vectors in other protocol messages. The same `AggregatorId`s MUST be used
+  when consturcting other messages containing vectors with one value for each
```
ekr, line 455:
```suggestion
  when constructing other messages containing vectors with one value for each
```

draft-pda-protocol.md
ekr, line 457: Why do aggregators need to know their ID?

ekr:
```suggestion
  Each aggregator knows its own `id`. `id` values must be unique in a `Param` but may have different meanings in different Params.
```

draft-pda-protocol.md
ekr: Would it actually make sense to not have an integer in this field and only in the submissions, and just have those values refer to indexes here.

draft-pda-protocol.md
```diff
@@ -553,13 +578,13 @@ To encrypt an input share, the client first generates an HPKE
 {{!I-D.irtf-cfrg-hpke}} context for the aggregator by running
 
 ~~~
-enc, context = SetupBaseS(pk, "pda input share" || server_role)
+enc, context = SetupBaseS(pk, "pda input share" || aggregator_id)
```
ekr, line 581: Actually, why do we need this field at all?

chris-wood (approved):
Nice! This really cleans stuff up.

draft-pda-protocol.md
chris-wood: > Why do aggregators need to know their ID?

This is a fair question -- it seems like the aggregator ID is only used to coordinate messages, and might not need to be included in the AAD.

tgeoghegan: Yes, an aggregator only needs to know its own ID to put in the AAD when establishing HPKE context. That is also why I made that field `uint8`, to match the previous leader/helper `server_role` values. However we could also make the aggregator ID into a longer string of bytes, allowing the values to be human readable strings, which I think will be nice for logging purposes. But it sounds like you are doubting the value of including server role in the AAD at all? I would think that since a given aggregator could be acting as either a leader or a helper in different protocol instantiations, and maybe using the same HPKE config for both, including server role diversifies keys across protocol instantiations. Then again, if that's the goal, then we should probably also mix in the `Param.uuid` since `aggregator_id` values could be re-used across different protocol instantiations.

So my take here is we should:

* define `opaque AggregatorId<1..256>;`
* use `enc, context = SetupBase(S|R)(pk, "pda input share" || Param.uuid || aggregator_id)` and `enc, context = SetupBaseS(pk, "pda output share" || Param.uuid || aggregator_id)`

But I will defer to the superior crypto brains of @chris-wood and @ekr ... should we just drop `server_role`/`aggregator_id` from the HPKE application info?

tgeoghegan: Or we could go back to leader = `0x00` / helper = `0x01` in the info, which has the virtue of not entangling our mechanism for coordinating messages with our mechanism for key diversity, but then we will have to revisit the question of what application info should be provided by helpers beyond the first one, if and when we introduce multiple helpers. My goal in introducing the aggregator_id into the AAD was to avoid such a disruptive change.

draft-pda-protocol.md
tgeoghegan: I agree, there's no reason not to make this a 64 bit integer (or even a much longer byte string as I suggest in another thread on this PR). However I'm not sure if we want the IDs in other messages to refer to indexes here, because that imposes the new constraint on aggregator IDs that they be sequential, which is not required with the current `Aggregator` definition.

cjpatton (changes requested):
This change is basically right. The main thing I would change is to make more clear the intended meaning of aggregator ID. While it was intended to solve the indexing-into-input-shares-array problem, this PR also addresses the more fundamental problem of assigning roles to each of the aggregators in the protocol. I think this is a great idea, and that's what this should be about.

With that in mind, my high level ask is to be stricter about how ids are assigned. In particular, the ids should be contiguous (i.e., `0, 1, ..., len(Param.aggregators)-1`) and `Param.aggregators` should be sorted by id.

chris-wood:
> With that in mind, my high level ask is to be stricter about how ids are assigned. In particular, the ids should be contiguous (i.e., 0, 1, ..., len(Param.aggregators)-1) and Param.aggregators should be sorted by id.

What value does sorting these add?

draft-pda-protocol.md
```diff
@@ -437,6 +434,35 @@ Time uint64; /* seconds elapsed since start of UNIX epoch */
...
+
+~~~
+struct {
+  AggregatorId id;
+  Url endpoint;
+} Aggregator;
+
+uint64 AggregatorId;
+
+AggregatorId leader_aggregator_id = 0;
```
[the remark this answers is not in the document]
chris-wood: How would this value (of zero) different from having the leader appear first (also a value of zero)?

draft-pda-protocol.md
chris-wood: FWIW, I'd drop it from the AAD, and then mark an open issue asking whether or not we believe it's important to authenticate. (My impression is that this is not important to authenticate.)

draft-pda-protocol.md
```diff
@@ -446,29 +472,29 @@ task_id = SHA-256(param)
...
 
-Before the client can upload its report to the leader, it must know
-the keys of each of the aggregators. These are retrieved
-from each aggregator by sending a request to
-`[aggregator]/key_config`, where `[aggregator]` is the aggregator's
-endpoint URL, provided in the Config. The aggregator responds to
-well-formed requests with status 200 and an `HpkeConfig` value:
+Before the client can upload its report to the leader, it must know the keys of
+each of the aggregators. These are retrieved from each aggregator by sending a
+request to `Param.aggregators[i].endpoint/key_config` for `0 <= i <
```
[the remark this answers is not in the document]
tgeoghegan: I guess I'm not sure how clearly an RFC needs to spell out things like this for implementors. I agree it was pretty clear before so I think I'll put it back.

tgeoghegan:
After discussion with cjpatton, I tacked on a commit that attempts to
    specify this stuff without defining `AggregatorId` and inserting it into
    various messages. Instead, we now require a consistent ordering between
    `Param.aggregator_endpoints` and `Report.encrypted_input_shares` (the
    order of `CollectResp.encrypted_output_shares` doesn't matter; the
    collector doesn't need to do anything special with the leader's share).
    
    I'm not sure if this is better than the `AggregatorId` approach but now
    we can compare them.

chris-wood:
Taking a step back, is it true that we're simply trying to address aggregator routing at the leader here? In particular, upon receipt of a Report, the leader needs to know what aggregators should get the shares contained therein. The latest commit does this by saying "send the i-th share to the i-th helper that was agreed upon in the Param structure." Is this a correct interpretation of the problem?

tgeoghegan:
> Taking a step back, is it true that we're simply trying to address aggregator routing at the leader here? In particular, upon receipt of a Report, the leader needs to know what aggregators should get the shares contained therein.

This PR is about addressing #117: putting enough structure into the PDUs so that we could extend the protocol in the future to support multiple helpers without changing the wire format of messages, even if actually supporting >1 helper might require more rounds of communication and new, yet undefined messages. IMO, that also requires solving #133, the problem of matching an encrypted input share to an aggregator. I also think that now that we have introduced a means of distinguishing arbitrary numbers of aggregators from each other, we should use it to diversify key material (but see my question below).

The actual protocol extensions are #68, which this PR does not address, nor do I think we intend to do anything about that issue anytime soon.

> The latest commit does this by saying "send the i-th share to the i-th helper that was agreed upon in the Param structure." Is this a correct interpretation of the problem?

Yes.

I apologize for the back and forth on this. I think we have two important questions to resolve in this PR:

## Question 1
We currently diversify keys by "server role", which is either "leader" or "helper". Now that we admit that there can be more than two server roles, do we want to further diversify key material by that role? 

### Question 1.(a)
If yes to 1., do we want to do that in this PR or should we punt to a subsequent change and leave `server_role = 0x00` for leader, `0x01` for helper?

## Question 2
We want some protocol messages to be able to contain a vector of sub-messages, one for each participating aggregator. Do we want to achieve that with a vector of labeled tuples `(aggregator_id, value)` (as in commit d510e04ee4843b37e774cf3826af70f051272f7a), or with a simple vector of values, where the index into the vector is effectively the aggregator ID (as in commit 351d2a941369c048bc515cd8df32f91a7c48a3a8)?

I think both are workable, with different tradeoffs. I prefer the labeled tuples approach because I think it is more explicit and  imposing fewer requirements on the order of lists in different messages means less validation code and removes opportunities for implementation bugs. @cjpatton argues for the simple vector approach.

I'm not sure how to resolve that disagreement, except that consensus matters more to me than which one we choose. Should we, uh, vote on it or something?

chris-wood:
> I apologize for the back and forth on this. 

No worries -- we're in alignment on the problem here. 

As for the questions, I think:

1) Further diversification is not useful, and seems like it only complicates matters. In particular, a helper doesn't necessarily need to know that it's helper i or j -- it just needs to know that it's not the leader. Sticking with what we currently have makes sense.
2) I claim this is about routing messages to aggregators. As a silly example, imagine we split up Report into multiple HTTP requests, where each request carried one share for each aggregator. The request could identify the aggregator. This would also solve the problem here, but perhaps in a (much) worse way. 

Both proposals -- tagging each share by an ID, or sorting the report based on the parameters -- seem fine, but I lean towards something like the former. Sorting the report means that everyone needs to agree on the order of helpers everywhere, which seems like an unnecessary constraint. I would prefer a design wherein each share unambiguously, and without additional coordination, identifies the intended aggregator.

draft-pda-protocol.md
```diff
@@ -419,12 +418,15 @@ Duration uint64; /* Number of seconds elapsed between two instants */
 Time uint64; /* seconds elapsed since start of UNIX epoch */
 ~~~
 
-* `uuid`: A unique sequence of bytes used  to ensure that two otherwise
+* `uuid`: A unique sequence of bytes used to ensure that two otherwise
```
ekr, line 421: I wouldn't call this "uuid" because then some IETF nitpicker will complain that it's not actually a UUID.

Maybe ```task_nonce```

draft-pda-protocol.md
```diff
@@ -419,12 +418,15 @@ Duration uint64; /* Number of seconds elapsed between two instants */
...
   identical `Param` instances will have distinct `TaskID`s. It is
   RECOMMENDED that this be set to a random 16-byte string derived from a
   cryptographically secure pseudorandom number generator.
-* `leader_url`: The leader's endpoint URL.
-* `helper_url`: The helper's endpoint URL.
+* `aggregator_endpoints`: Vector of URLs relative to which an aggregator's API
+  endpoints can be found. The leader's endpoint MUST be the first in the vector.
+  The order in which aggregators appear in this vector MUST be consistent with
+  the order of the encrypted input shares in a `Report` (see
+  {{uploading-reports}}).
```
ekr, line 429: I would rewrite this because it's the reports which have to be consistent with this:
```suggestion
  endpoints can be found. The leader's endpoint MUST be the first in the vector.
  The order in encrypted input shares in a `Report` (see
  {{uploading-reports}}) MUST be the same as the
  order in which aggregators appear in this vector.
```

draft-pda-protocol.md
```diff
@@ -446,29 +448,29 @@ task_id = SHA-256(param)
 Where param is just the serialization of Param.
 
 
-## Uploading Reports
+## Uploading Reports {#uploading-reports}
```
ekr, line 451: FYI, you don't need this. kramdown automatically creates anchors by doing tr/A-Z /a-z\-/.

draft-pda-protocol.md
```diff
@@ -446,29 +448,29 @@ task_id = SHA-256(param)
...
 
-Before the client can upload its report to the leader, it must know
-the keys of each of the aggregators. These are retrieved
-from each aggregator by sending a request to
-`[aggregator]/key_config`, where `[aggregator]` is the aggregator's
-endpoint URL, provided in the Config. The aggregator responds to
+Before the client can upload its report to the leader, it must know the public
+key of each of the aggregators. These are retrieved from each aggregator by
+sending a request to `[aggregator]/key_config`, where `[aggregator]` is the
+aggregator's endpoint URL, provided in the `Param`. The aggregator responds to
```
ekr, line 461:
```suggestion
aggregator's endpoint URL, provided in the `Param` for the task. The aggregator responds to
```

draft-pda-protocol.md
```diff
@@ -477,9 +479,7 @@ uint16 HpkeKdfId;  // Defined in I-D.irtf-cfrg-hpke
 [TODO: Decide whether to use the same config structure as OHTTP/ECH. This would
 add support for multiple cipher suites.]
 
-The client issues a key configuration request to `Param.leader_url` and
-`Param.helper_url`. It aborts if any of the following happen for either
-request:
+The client aborts if any of the following happen for any `key_config` request:
```
ekr, line 482:
```suggestion
The client MUST abort if any of the following happen for any `key_config` request:
```

draft-pda-protocol.md
```diff
@@ -499,8 +499,8 @@ rejecting reports.
 
 ### Upload Request
 
-Clients upload reports by using an HTTP POST `[leader]/upload`, where
-`[leader]` is the leader's endpoint URL. The payload is structured as
+Let `[leader]` be the first entry in `Param.aggregators`. Clients upload reports
+by using an HTTP POST to `[leader]/upload`. The payload is structured as
```
ekr, line 503: FWIW, I prefer the phrasing before. "Let" is a bit stilted.

draft-pda-protocol.md
```diff
@@ -525,15 +525,18 @@ This message is called the client's *report*. It contains the following fields:
 * `extensions` is a list of extensions to be included in the Upload flow; see
   {{upload-extensions}}.
 * `encrypted_input_shares` contains the encrypted input shares of each of the
-  aggregators.
+  aggregators. The order in which the encrypted input shares appear MUST match
+  the order of the `aggregator_endpoints` in the `Param` corresponding to
+  `task_id` (i.e., the first share should be the leader's, the second share
+  should be for the second aggregator, and so on).
```
ekr, line 531:
```suggestion
  should be for the first helper, and so on).
```

draft-pda-protocol.md
```diff
@@ -717,9 +723,9 @@ sub-requests by ignoring any sub-request that does not follow the previous one
 (See {{anti-replay}}.)
 
 The response is an HTTP 200 OK with a body consisting of the helper's updated
-state and a sequence of *sub-responses*, where the i-th sub-response corresponds
-to the i-th sub-request for each i. The structure of each sub-response is
-specific to the PPM protocol:
+state and a sequence of *sub-responses*, where the j-th sub-response corresponds
+to the j-th sub-request for each `1 <= j <= len(AggregateReq.seq)`. The
+structure of each sub-response is specific to the PPM protocol:
```
ekr, line 728: This bit of mathiness seems a bit confusing. Instead of expanding the formula, I would consider "where each subresponse corresponds to the sub-request in the same position in AggregateReq"

ekr (commented):
I'm also a bit concerned about universal sorting, but I think we can resolve this separately, b/c it's really just about what goes into the HPKE functions. I suggest we land this once the more minor comments have been resolved.

draft-pda-protocol.md
tgeoghegan: I agree, but I think `struct Param` will go away altogether when someone (perhaps me, later this week) addresses #104, and we will be left with nothing but `task_id`. So I prefer to leave this text alone until it gets deleted in that PR (except, I guess, to delete this extra space).

ekr:
SGTM

tgeoghegan:
In the interest of continuing to move forward, I am going to keep the version where we keep the ordering consistent across different messages. Per @chris-wood's observation, I've punted the question of aggregator IDs or server roles beyond the existing 0x01/0x00 constants to #138.

draft-ppm-protocol.md
```diff
@@ -581,25 +585,26 @@ To encrypt an input share, the client first generates an HPKE
 {{!I-D.irtf-cfrg-hpke}} context for the aggregator by running
 
 ~~~
-enc, context = SetupBaseS(pk, "pda input share" || server_role)
+enc, context = SetupBaseS(pk, "pda input share" || task_id || server_role)
```
cjpatton, line 588: Does it make sense to replace `server_role` with the index of the input share in the report struct? This may be useful for protocols with more than two aggregators.

draft-ppm-protocol.md
```diff
@@ -728,11 +731,15 @@ struct {
...
-report uploaded by the client. Similarly, the `helper_share` field is the helper's
-encrypted input share as it appeared in the report. [OPEN ISSUE: We usually only
-need to send this in the first aggregate request. Shall we exclude it in
-subsequent requests somehow?] The remainder of the structure is dedicated to the
-protocol-specific request parameters.
+report uploaded by the client. Similarly, the `helper_share` field is the
+`EncryptedInputShare` from the `Report` whose index in
+`Report.encrypted_input_shares` is equal to the index of `[aggregator]` in
+`Param.aggregator_endpoints`.
+
```
cjpatton, line 738: Do you intend to start a new paragraph here? It's usually OK to inline OPEN ISSUEs with surrounding text.

draft-ppm-protocol.md
```diff
@@ -780,13 +787,14 @@ following procedure:
 
 ~~~
 context = SetupBaseR(helper_share.enc, sk,
-                     "pda input share" || server_role)
-input_share = context.Open(task_id || time || nonce || extensions,
-                           helper_share)
+                     "pda input share" || task_id || server_role)
```
cjpatton, line 790: See comment above.

cjpatton (commented):
Yup! This is looking good to me. A couple more questions in-line.

draft-ppm-protocol.md
chris-wood: We punted this to a separate issue.

PR #136, opened 2021-08-19
s/pda/ppm

cjpatton:
The draft name still has the acronym "pda" in it. Let's kill this for good.


ekr:
You also need to fix .targets.mk I believe

cjpatton:
nice catch, @ekr

tgeoghegan (approved):

Issue #138, opened 2021-08-23
Aggregator roles and key diversification


tgeoghegan:
Currently, the derivation of HPKE context entangles a constant for `server_role`, which is 0x01 for the leader and 0x00 for the helper. Some iterations of PR #135, which aimed to generalize message definitions to support more than one helper, introduced a notion of an aggregator ID or role, and then used that role in the key derivation. That notion wound up being a little more complicated than what needed to be addressed in that PR, but we should still consider whether and how we want to formalize the notion of an aggregator's role and whether it should be involved in encryption key derivation.

tgeoghegan:
I'm not sure if this should go in label parking-lot. The server role constants we have now are sufficient for the deployments that will realistically materialize in the near or medium term, and I think we could define server roles in the future in a manner that is compatible with the existing constants.

tgeoghegan:
One important question we must address is the maximum number of aggregators in a protocol instantiation. We are already encountering places where [aggregators and clients must agree on the representation of the ID](https://github.com/abetterinternet/libprio-rs/pull/76).

tgeoghegan:
We've long since introduced a formal `Role` enumeration whose layout was chosen to allow adding more helpers in the future. Closing.

PR #143, opened 2021-08-25
remove `struct Param`

### LINKED ISSUE #104: Does PPMParam need to be defined at all?

tgeoghegan:
`Param` is never transmitted over the wire, nor do we otherwise depend
on is structure, so this commit removes that struct definition and
instead enumerates the pre-negotiated task parameters each participant
must hold, leaving implementations free to store, represent and transmit
those parameters any way they wish. `TaskId` is still around, but
instead of being a SHA-256 over the serialiation of Param, it's now just
32 random bytes.

Resolves #104

### DIFF
--- a/draft-ppm-protocol.md
+++ b/draft-ppm-protocol.md
@@ -575,14 +575,14 @@ Encrypted input shares are structured as follows:
 
 ~~~
 struct {
-  HpkeConfigId config_id;
+  HpkeConfigId aggregator_config_id;
   opaque enc<1..2^16-1>;
   opaque payload<1..2^16-1>;
 } EncryptedInputShare;
 ~~~
 
-* `config_id` is equal to `HpkeConfig.id`, where `HpkeConfig` is the key config
-  of the aggregator receiving the input share.
+* `aggregator_config_id` is equal to `HpkeConfig.id`, where `HpkeConfig` is the
+  key config of the aggregator receiving the input share.
 * `enc` is the encapsulated HPKE context, used by the aggregator to decrypt its
   input share.
 * `payload` is the encrypted input share.
@@ -898,7 +898,7 @@ body consisting of the following structure:
 struct {
   HpkeConfigId collector_hpke_config_id;
   opaque enc<1..2^16-1>;
-  opaque encrypted_output_share<1..2^16>;
+  opaque payload<1..2^16>;
 } EncryptedOutputShare;
 ~~~
 
@@ -906,8 +906,7 @@ struct {
   corresponding to `CollectReq.task_id`.
 * `enc` is the encapsulated HPKE context, used by the collector to decrypt the
   output share.
-* `encrypted_output_share` is an encrypted `OutputShare`, whose structure is
-  given below.
+* `payload` is an encrypted `OutputShare`.
 
 The leader uses the helper's output share response to respond to the collector's
 collect request (see {{pa-collect}}).
--- a/draft-ppm-protocol.md
+++ b/draft-ppm-protocol.md
@@ -80,16 +80,6 @@ informative:
     author:
       -ins: S. Vadhan
 
-normative:
-
-  FIPS180-4:
-    title: NIST FIPS 180-4, Secure Hash Standard
-    author:
-      name: NIST
-      ins: National Institute of Standards and Technology, U.S. Department of Commerce
-    date: 2012-03
-    target: http://csrc.nist.gov/publications/fips/fips180-4/fips-180-4.pdf
-
 --- abstract
 
 There are many situations in which it is desirable to take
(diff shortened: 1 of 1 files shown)


draft-ppm-protocol.md
```diff
@@ -410,64 +410,45 @@ PPM has three major interactions which need to be defined:
...
+the configuration for each task. A task is uniquely identified by its task ID:
 
 ~~~
-task_id = SHA-256(param)
+opaque TaskId[32];
 ~~~
 
-Where param is the serialization of Param.
+A `TaskId` is a globally unique sequence of bytes. It is RECOMMENDED that this
+be set to a random string derived from a cryptographically secure pseudorandom
```
cjpatton, line 435:
```suggestion
be set to a random string output by a cryptographically secure pseudorandom
```

cjpatton:
Mind rebasing, @tgeoghegan?

tgeoghegan:
> This seems fine, although I'm a little concerned that the text is going to get a lot more confusing by dropping named parameters. PR #137 refers to the named parameters in `Param` quite a lot. Could we hold off on merging this until that PR is merged and see what it looks like after?

Sounds reasonable to me. I'll rebase after #137 is merged.

tgeoghegan:
Rebased on main to incorporate #137.

draft-ppm-protocol.md
```diff
@@ -415,69 +405,58 @@ PPM has three major interactions which need to be defined:
...
-  difference between the oldest and newest report in a batch. This defines the
-  boundaries with which the batch interval of each collect request must be
-  aligned. (See {{batch-parameter-validation}}.)
-* `proto`: The PPM protocol, e.g., Prio or Hits. The rest of the structure
-  contains the protocol specific parameters.
-
-The *task ID* is derived from the PPM parameters as:
+## Task Configuration {#task-configuration}
+
+Prior to the start of the protocol, each endpoint in the system must agree on
```
cjpatton, line 429:
```suggestion
Prior to the start of the protocol, each participant in the protocol must agree on
```

draft-ppm-protocol.md
```diff
@@ -415,69 +405,58 @@ PPM has three major interactions which need to be defined:
...
 
-Where param is the serialization of Param.
-
+A `TaskId` is a globally unique sequence of bytes. It is RECOMMENDED that this
+be set to a random string output by a cryptographically secure pseudorandom
+number generator. Each task has the following parameters associated with it:
+
+* aggregator endpoints: A list of URLs relative to which an aggregator's API
+  endpoints can be found. Each endpoint's list MUST be in the same order. The
+  leader's endpoint MUST be the first in the list. The order of the
```
cjpatton, line 442: These requirements make me think we should define `struct` for the list of aggregators that defines the order and which endpoint is the leader.

draft-ppm-protocol.md
```diff
@@ -415,69 +405,58 @@ PPM has three major interactions which need to be defined:
...
+* maximum batch lifetime: The maximum number of times a batch of reports may be
+  used in collect requests. Referred to as `max_batch_lifetime` elsewhere in
+  this document.
+* minimum batch size: The minimum number of reports that appear in a batch.
+* minimum batch duration: The minimum time difference between the oldest and
+  newest report in a batch. This defines the boundaries with which the batch
+  interval of each collect request must be aligned. (See
+  {{batch-parameter-validation}}.) Referred to as `min_batch_duration` elsewhere
+  i this document.
+* protocol: The PPM protocol, e.g., Prio or Hits.
```
cjpatton, line 459:
```suggestion
* protocol: The core PPM protocol, e.g., Prio or Hits.
```

draft-ppm-protocol.md
```diff
@@ -415,69 +405,58 @@ PPM has three major interactions which need to be defined:
...
-task_id = SHA-256(param)
+opaque TaskId[32];
 ~~~
 
-Where param is the serialization of Param.
-
+A `TaskId` is a globally unique sequence of bytes. It is RECOMMENDED that this
+be set to a random string output by a cryptographically secure pseudorandom
+number generator. Each task has the following parameters associated with it:
+
```
cjpatton, line 439: Bullets describing parameters with names used elsewhere in the doc should begin with the name. E.g.,:
> * `max_batch_lifetime` is the ...

draft-ppm-protocol.md
```diff
@@ -595,7 +574,8 @@ To encrypt an input share, the client first generates an HPKE
 {{!I-D.irtf-cfrg-hpke}} context for the aggregator by running
 
 ~~~
-enc, context = SetupBaseS(pk, "pda input share" || task_id || server_role)
+enc, context = SetupBaseS(pk,
+                          "pda input share" || task_id|| server_role)
```
cjpatton, line 578:
```suggestion
                          "pda input share" || task_id || server_role)
```
If this is too long, consider reducing the size of the tab.

draft-ppm-protocol.md
cjpatton, line 459: This is a "named parameter".

cjpatton (approved):
A few nits, but otherwise I'm good with landing this.

draft-ppm-protocol.md
tgeoghegan: I think you're right, but I think that will come with #138 when we formalize the notion of aggregator roles. Once we have that, I think this language and the discussion of `Report` gets much simpler.

PR #147, opened 2021-09-01
Add share-counting attacks to threat model

csharrison:
Aggregators merely counting the number of input shares could compromise
user privacy in some circumstances (namely, it can break differential privacy
if the number of records per user is not constant).

This attack only applies to aggregators who can count the number of shares,
so any protection still holds on the collector side.

To address this attack, there are two possible mitigations, which both involve
adding noise to the total number of input shares. This should be possible to
do with "null" shares that don't affect aggregate output (i.e. inserting 0 into
sums).

### DIFF
--- a/draft-ppm-protocol.md
+++ b/draft-ppm-protocol.md
@@ -1146,7 +1146,9 @@ absence of a share for a given user is sensitive.
 1. If computed over a sufficient number of input shares, output shares reveal
    nothing about either the inputs or the participating clients.
 1. Clients can ensure that aggregate counts are non-sensitive by generating
-   input independently of user behavior.
+   input independently of user behavior. For example, a client should periodically
+   upload a report even if the event that the task is tracking has not occurred, so
+   that the absence of reports cannot be distinguished from their presence.
 1. Bogus inputs can be generated that encode "null" shares that do not affect
    the aggregate output, but mask the total number of true inputs.
      * Either leaders or clients can generate these inputs to mask the total
--- a/draft-ppm-protocol.md
+++ b/draft-ppm-protocol.md
@@ -1155,7 +1155,7 @@ absence of a share for a given user is sensitive.
        indistinguishable from true inputs (size, metadata, etc).
       
 [OPEN ISSUE: Define what "null" shares are. They should be defined such that
-inserting null shares into an aggregation is effectively a no-op]
+inserting null shares into an aggregation is effectively a no-op. See issue#98.]
 
 ### Leader
 
--- a/draft-ppm-protocol.md
+++ b/draft-ppm-protocol.md
@@ -1145,10 +1145,17 @@ absence of a share for a given user is sensitive.
    shares.
 1. If computed over a sufficient number of input shares, output shares reveal
    nothing about either the inputs or the participating clients.
-1. Leaders can inject bogus "null" shares that do not affect the aggregate
-   output but mask the total number of shares to non-leader aggregators.
-1. Clients can inject bogus "null" shares that do not affect the aggregate
-   output but mask the total number of shares all aggregators.
+1. Clients can ensure that aggregate counts are non-sensitive by generating
+   input independently of user behavior.
+1. Bogus inputs can be generated that encode "null" shares that do not affect
+   the aggregate output, but mask the total number of true inputs.
+     * Either leaders or clients can generate these inputs to mask the total
+       number from non-leader aggregators or all the aggregators, respectively.
+     * In either case, care must be taken to ensure that bogus inputs are
+       indistinguishable from true inputs.
+      
+[OPEN ISSUE: Define what "null" shares are. They should be defined such that
+inserting null shares into an aggregation is effectively a no-op]
 
 ### Leader
 
--- a/draft-ppm-protocol.md
+++ b/draft-ppm-protocol.md
@@ -1134,6 +1134,9 @@ to emit output shares.
 1. Input validity proof forging. Any aggregator can collude with a malicious
 client to craft a proof that will fool honest aggregators into accepting
 invalid input.
+1. Aggregators can count the total number of input shares, which could
+compromise user privacy (and differential privacy {{dp}}) if the presence or
+absence of a share for a given user is sensitive.
 
 #### Mitigations
 
@@ -1142,6 +1145,10 @@ invalid input.
    shares.
 1. If computed over a sufficient number of input shares, output shares reveal
    nothing about either the inputs or the participating clients.
+1. Leaders can inject bogus "null" shares that do not affect the aggregate
+   output but mask the total number of shares to non-leader aggregators.
+1. Clients can inject bogus "null" shares that do not affect the aggregate
+   output but mask the total number of shares all aggregators.
 
 ### Leader
 
--- a/draft-ppm-protocol.md
+++ b/draft-ppm-protocol.md
@@ -1152,7 +1152,7 @@ absence of a share for a given user is sensitive.
      * Either leaders or clients can generate these inputs to mask the total
        number from non-leader aggregators or all the aggregators, respectively.
      * In either case, care must be taken to ensure that bogus inputs are
-       indistinguishable from true inputs.
+       indistinguishable from true inputs (size, metadata, etc).
       
 [OPEN ISSUE: Define what "null" shares are. They should be defined such that
 inserting null shares into an aggregation is effectively a no-op]
--- a/draft-ppm-protocol.md
+++ b/draft-ppm-protocol.md
@@ -1152,7 +1152,8 @@ absence of a share for a given user is sensitive.
      * Either leaders or clients can generate these inputs to mask the total
        number from non-leader aggregators or all the aggregators, respectively.
      * In either case, care must be taken to ensure that bogus inputs are
-       indistinguishable from true inputs (size, metadata, etc).
+       indistinguishable from true inputs (metadata, etc), especially when
+       constructing timestamps on reports.
       
 [OPEN ISSUE: Define what "null" shares are. They should be defined such that
 inserting null shares into an aggregation is effectively a no-op. See issue#98.]


tgeoghegan:
Specifically, this is about being able to count the number of contributions from an individual client, right? If so, this would also be mitigated by putting an anonymizing proxy between clients and the leader which would scrub identifying information from reports.

csharrison:
An anonymizing proxy mitigates attacks but not in the worst case. I didn't mention in the PR (I could if it is necessary), but the worst case attack involves the attackers having some side knowledge (e.g. via client collusion) such that if N reports arrive, then Tim did not contribute to this batch, and if N+1 reports arrive, then Tim did contribute to the batch. This is related to the Sybil style attacks which do the same thing but with the value embedded in the report. 

In these kinds of attacks, no auxiliary information in the report (like IP, etc) is needed.

In practice, I don't believe this attack does much to meaningfully reduce user privacy. However, I think calling it out is important if we want to make differential privacy claims since DP requires us to think through worst-case attacks.

draft-ppm-protocol.md
```diff
@@ -1142,6 +1145,10 @@ invalid input.
    shares.
 1. If computed over a sufficient number of input shares, output shares reveal
    nothing about either the inputs or the participating clients.
+1. Leaders can inject bogus "null" shares that do not affect the aggregate
+   output but mask the total number of shares to non-leader aggregators.
+1. Clients can inject bogus "null" shares that do not affect the aggregate
+   output but mask the total number of shares all aggregators.
```
cjpatton, line 1151: - The idea is pretty clear if you're as deep into this as we are, but I think someone new would have a hard time figuring out what a `bogus "null" share` is. We might stipulate that the core PPM protocol define this. If you'd like we can leave it as an open issue for #98.
- It sounds like what you want is that the clients or servers contribute "null" shares + some noise for DP. If this is correct, then let's say it.
- Instead of `inject bogus "null" shares` let's say "generate bogus reports that encode "null" shares`, as this is a bit more precise.

cjpatton (commented):
Can you say more about what adding "null" shares does for DP? As a working example, suppose we're computing the sum of the inputs `S = X[1] + ... + X[m]`, where `X[i]` is an element of a finite field and `m` is the number of real inputs.

It sounds like the scheme is to estimate `S` as
```
  S' = X[1] + ... + X[m] + // Real inputs
       e[1] + ... + e[m] + // error term for each real input
       e[m+1] ... + e[n]   // error term for each "null" input 
```
where `n` is the number of inputs, including the "null" inputs generated by the clients (resp. the leader). So we are including an error term for each real input and generating an error team for each "null" (i.e., `0`-valued) input. Is that the idea?

> [T]he worst case attack involves the attackers having some side knowledge (e.g. via client collusion) such that if N reports arrive, then Tim did not contribute to this batch, and if N+1 reports arrive, then Tim did contribute to the batch. This is related to the Sybil style attacks which do the same thing but with the value embedded in the report.

Suppose I have some a priori knowledge about the distribution of the inputs `X[i]`. (Maybe I have corrupted some small subset of the clients, and I know that each input is independently distributed.) Can't I make a pretty good guess about the value of `m` from `S'`? In fact, if i know that the expected value of `X[i]` is 1, and each `X[i]` is independent, then `S'` is going to be close to `m`.

csharrison:
> Can you say more about what adding "null" shares does for DP? As a working example, suppose we're computing the sum of the inputs `S = X[1] + ... + X[m]`, where `X[i]` is an element of a finite field and `m` is the number of real inputs.
> 
> It sounds like the scheme is to estimate `S` as
> 
> ```
>   S' = X[1] + ... + X[m] + // Real inputs
>        e[1] + ... + e[m] + // error term for each real input
>        e[m+1] ... + e[n]   // error term for each "null" input 
> ```
> 
> where `n` is the number of inputs, including the "null" inputs generated by the clients (resp. the leader). So we are including an error term for each real input and generating an error team for each "null" (i.e., `0`-valued) input. Is that the idea?

No this is not what I am talking about. This issue is not necessarily about attacks on `S`, but rather about attacks on revealing `m` to the helpers. In some settings, `m` could be just as sensitive as `S`. In your example you are using some local noise for each input (just one possible way of implementing DP). I don't think we would need this to protect `S`. However, to protect `m` we would only need to add false shares that truly sum to 0 (i.e. no noise embedded in them at all). So that `m` is hard to learn from `n` the total # of input records.

> > [T]he worst case attack involves the attackers having some side knowledge (e.g. via client collusion) such that if N reports arrive, then Tim did not contribute to this batch, and if N+1 reports arrive, then Tim did contribute to the batch. This is related to the Sybil style attacks which do the same thing but with the value embedded in the report.
> 
> Suppose I have some a priori knowledge about the distribution of the inputs `X[i]`. (Maybe I have corrupted some small subset of the clients, and I know that each input is independently distributed.) Can't I make a pretty good guess about the value of `m` from `S'`? In fact, if i know that the expected value of `X[i]` is 1, and each `X[i]` is independent, then `S'` is going to be close to `m`.

Great question. Let's simplify this and say that you know for a fact that the true dataset contains either:
- `m = N` records summing to `S = N`, OR
- `m = N+1` records summing to `S = N+1`
i.e. each record will always just have value 1. Also let's say that we've properly hidden `m` directly, and the adversary wants to learn `m` from `S`. I think it should be clear that if we can protect the true value of `S`, we can protect the true value of `m` as they are the same.

True, `S` will be close to N or N+1, but the noise added to the sum will be enough that both cases are reasonably probable, protecting both `S` and `m`. For example (with Laplace noise with epsilon ln2):
- `P(output >= N+1 | S = N) = 25%`
- `P(output >= N+1 | S = N+1) = 50%`

draft-ppm-protocol.md
csharrison: > It sounds like what you want is that the clients or servers contribute "null" shares + some noise for DP. If this is correct, then let's say it.

No, the null shares should be effectively "no-ops", and not introduce any change in the visible output. They only serve to protect the helpers from learning the true count.

Your other suggestions sound good, I can make an update

cjpatton:
> No this is not what I am talking about. This issue is not necessarily about attacks on `S`, but rather about attacks on revealing `m` to the helpers. In some settings, `m` could be just as sensitive as `S`. In your example you are using some local noise for each input (just one possible way of implementing DP). I don't think we would need this to protect `S`. However, to protect `m` we would only need to add false shares that truly sum to 0 (i.e. no noise embedded in them at all). So that `m` is hard to learn from `n` the total # of input records.

Sharpening this a bit more: The concern is not whether the collector learns `m` but whether the aggregators learn `m`. The clients can add `n-m` bogus reports to hide `m` from the aggregators, or the leader can add `n-m` bogus reports to hide `m` from the helper. Is that right? 

This would be a tad inconsistent with the current threat model, which assumes the collector may collude with an aggregator and, in particular, provide it with the final output. That said, I'm fine with adding it as a security consideration. We just want to make it clear that it only appears to be achievable under stronger non-collusion assumptions.


> > Suppose I have some a priori knowledge about the distribution of the inputs `X[i]`. (Maybe I have corrupted some small subset of the clients, and I know that each input is independently distributed.) Can't I make a pretty good guess about the value of `m` from `S'`? In fact, if i know that the expected value of `X[i]` is 1, and each `X[i]` is independent, then `S'` is going to be close to `m`.
> 
> Great question. Let's simplify this and say that you know for a fact that the true dataset contains either:
> 
>     * `m = N` records summing to `S = N`, OR
> 
>     * `m = N+1` records summing to `S = N+1`
>       i.e. each record will always just have value 1. Also let's say that we've properly hidden `m` directly, and the adversary wants to learn `m` from `S`. I think it should be clear that if we can protect the true value of `S`, we can protect the true value of `m` as they are the same.
> 
> True, `S` will be close to N or N+1, but the noise added to the sum will be enough that both cases are reasonably probable, protecting both `S` and `m`. For example (with Laplace noise with epsilon ln2):
> 
>     * `P(output >= N+1 | S = N) = 25%`
> 
>     * `P(output >= N+1 | S = N+1) = 50%`

I see, thanks for clarifying. I think we need to be careful about what we mean when we say that the number of real (i.e., non-bogus) inputs is "protected". The output of the distributed aggregation function may be differentially private, but it doesn't hide the number of real inputs.

csharrison:
> > No this is not what I am talking about. This issue is not necessarily about attacks on `S`, but rather about attacks on revealing `m` to the helpers. In some settings, `m` could be just as sensitive as `S`. In your example you are using some local noise for each input (just one possible way of implementing DP). I don't think we would need this to protect `S`. However, to protect `m` we would only need to add false shares that truly sum to 0 (i.e. no noise embedded in them at all). So that `m` is hard to learn from `n` the total # of input records.
> 
> Sharpening this a bit more: The concern is not whether the collector learns `m` but whether the aggregators learn `m`. The clients can add `n-m` bogus reports to hide `m` from the aggregators, or the leader can add `n-m` bogus reports to hide `m` from the helper. Is that right?

Yes that's right

> This would be a tad inconsistent with the current threat model, which assumes the collector may collude with an aggregator and, in particular, provide it with the final output. That said, I'm fine with adding it as a security consideration. We just want to make it clear that it only appears to be achievable under stronger non-collusion assumptions.

The collector colluding with an aggregator and providing the final output should be OK (similarly, the aggregators providing the noisy count of raw records to the collector). In this case, the adversary has 2 views of the data (noisy count and noisy output). You should still be able to prove DP protections in this case if each view is protected with DP (this comes from the fact that DP composes well: https://en.wikipedia.org/wiki/Differential_privacy#Composability).

> I see, thanks for clarifying. I think we need to be careful about what we mean when we say that the number of real (i.e., non-bogus) inputs is "protected". The output of the distributed aggregation function may be differentially private, but it doesn't hide the number of real inputs.

Yes, this is exactly the purpose of this PR. By default, we can't protect `m` from the aggregators unless we add the bogus records. However, if we can directly hide the number of real records (by adding bogus ones), _and also_ ensure `S` is DP, we can hide the true values of both `S` and `m`.

csharrison:
@cjpatton take another look. I've tried to keep this short + sweet but let me know if you think it can be compressed further.

draft-ppm-protocol.md
```diff
@@ -1142,6 +1145,17 @@ invalid input.
...
+   input independently of user behavior.
+1. Bogus inputs can be generated that encode "null" shares that do not affect
+   the aggregate output, but mask the total number of true inputs.
+     * Either leaders or clients can generate these inputs to mask the total
+       number from non-leader aggregators or all the aggregators, respectively.
+     * In either case, care must be taken to ensure that bogus inputs are
+       indistinguishable from true inputs (size, metadata, etc).
+      
+[OPEN ISSUE: Define what "null" shares are. They should be defined such that
+inserting null shares into an aggregation is effectively a no-op]
```
cjpatton, line 1158:
```suggestion
inserting null shares into an aggregation is effectively a no-op. See issue#98.]
```

cjpatton (changes requested):
Nice! Just a nit to add a reference to https://github.com/abetterinternet/prio-documents/issues/98.

cjpatton (approved):

draft-ppm-protocol.md
```diff
@@ -1142,6 +1145,17 @@ invalid input.
...
 1. If computed over a sufficient number of input shares, output shares reveal
    nothing about either the inputs or the participating clients.
+1. Clients can ensure that aggregate counts are non-sensitive by generating
+   input independently of user behavior.
+1. Bogus inputs can be generated that encode "null" shares that do not affect
+   the aggregate output, but mask the total number of true inputs.
+     * Either leaders or clients can generate these inputs to mask the total
+       number from non-leader aggregators or all the aggregators, respectively.
+     * In either case, care must be taken to ensure that bogus inputs are
+       indistinguishable from true inputs (size, metadata, etc).
```
tgeoghegan, line 1155:
```suggestion
       indistinguishable from true inputs (metadata, etc), especially when
       constructing timestamps on reports.
```
The size will necessarily be the same as a true input as we discussed earlier.

draft-ppm-protocol.md
```diff
@@ -1142,6 +1145,17 @@ invalid input.
    shares.
 1. If computed over a sufficient number of input shares, output shares reveal
    nothing about either the inputs or the participating clients.
+1. Clients can ensure that aggregate counts are non-sensitive by generating
+   input independently of user behavior.
```
tgeoghegan, line 1149:
```suggestion
   input independently of user behavior. For example, a client should periodically
   upload a report even if the event that the task is tracking has not occurred, so
   that the absence of reports cannot be distinguished from their presence.
```

tgeoghegan (approved):
Left a couple of suggestions that I don't think should block merging. @csharrison let us know when you've rebased or otherwise finished with this and we can merge it for you.

draft-ppm-protocol.md
csharrison: Yes that's right

csharrison:
This should be ready to go

PR #151, opened 2021-09-21
House cleaning

cjpatton:
Since there's no PR pending, I thought it would be a good time to fix up the doc a bit. No spec or even editorial changes, just some line breaks and a few typos.


cjpatton:
I'm assuming everyone trusts that this makes no substantive changes. time to merge.

Issue #150, opened 2021-09-16
Can `AggregateReq`/`VerifyReq`s span multiple batch windows, and what is the scope of a `helper_state` blob?


tgeoghegan:
Section 4.3 discusses the `helper_state` opaque blob, which is carried across successive `AggregateReq/Resp` messages as well as `OutputShareReq`. However we don't discuss the scope of a helper state blob. I think the only way it makes sense is if the leader maintains a helper state blob for each PPM task, but the text should be explicit about this.

tgeoghegan:
@ekr points out that we need to think about parallelism here. If, as I asserted, `helper_state` is per-task, then that means leader needs to guarantee that at most one `AggregateReq` is in flight for any task, to prevent concurrent modifications to the helper's state blob. We might want to constrain the aggregate protocol in a way that enables finer-grained state blobs to permit multiple aggregate protocol runs to occur in parallel. ekr's suggestion is to require that an `AggregateReq` only contain reports from a single batch interval, at which point the scope of a helper state blob can be (task, batch interval) which enables much greater parallelism.

tgeoghegan:
`helper_state` is gone as of #185, and aggregation job IDs provide more obvious answers for parallelizing input preparation.

PR #152, opened 2021-09-22
Specify the protocol in terms of VDAF evaluation

### LINKED ISSUE #98: The wrapper-core protocol interface

cjpatton:
Closes #98.

Along the way, this commit cleans up some ambiguous language that allows
for multiple helpers, but without fully specifying the protocol flow.

### DIFF
--- a/draft-ppm-protocol.md
+++ b/draft-ppm-protocol.md
@@ -33,7 +33,6 @@ author:
        organization: Cloudflare
        email: caw@heapingbits.net
 
-
 informative:
 
   CGB17:
@@ -79,6 +78,11 @@ informative:
     target: "https://privacytools.seas.harvard.edu/files/privacytools/files/complexityprivacy_1.pdf"
     author:
       - ins: S. Vadhan
+
+  VDAF:
+    title: "Verifiable Distributed Aggregation Functions"
+    date: 2021-09-21
+    target: "TODO"
 
 --- abstract
 
@@ -190,6 +194,8 @@ This document uses the protocol definition language of {{!RFC8446}}.
 
 # Overview {#overview}
 
+[TODO Frame the PPM protocol around evaluation of a VDAF.]
+
 The protocol is executed by a large set of clients and a small set of servers.
 We call the servers the *aggregators*. Each client's input to the protocol is a
 set of measurements (e.g., counts of some user behavior). Given the input set of
@@ -288,7 +294,7 @@ includes the following parameters:
 * The values to be measured;
 * The statistic to be computed (e.g., sum, mean, etc.);
 * The set of aggregators and necessary cryptographic keying material to use; and
-* The PPM scheme to use. This is to some extent dictated by the previous
+* The VDAF scheme to use. This is to some extent dictated by the previous
   choices.
 * The minimum "batch size" of reports which can be aggregated.
 * The rate at which measurements can be taken, i.e., the "minimum batch window".
@@ -306,7 +312,7 @@ time.
 
 The leader distributes the shares to the helpers and orchestrates the process of
 verifying them (see {{validating-inputs}}) and assembling them into a final
-measurement for the collector. Depending on the PPM scheme, it may be possible
+measurement for the collector. Depending on the VDAF scheme, it may be possible
 to incrementally process each report as it comes in, or may be necessary to wait
 until the entire batch of reports is received.
 
@@ -409,8 +415,14 @@ struct {
 
 ## Task Configuration {#task-configuration}
 
-Prior to the start of execution of the protocol, each participant must agree on
-the configuration for each task. A task is uniquely identified by its task ID:
+A task involves evaluation of a specific Verifiable Distributed Aggregation
+Function (VDAF, see [VDAF]) over a set of client inputs. The VDAF determines the
+type of input measurements and the type of the output of the aggregation
+function. It also determines how many aggregators are involved (this document
+currently supports just two) and how many rounds of communication are required
+to verify an input's validity.  Prior to the start of execution of the protocol,
+each participant must agree on the configuration for each task. A task is
+uniquely identified by its task ID:
 
 ~~~
 opaque TaskId[32];
@@ -424,7 +436,9 @@ number generator. Each task has the following parameters associated with it:
   endpoints can be found. Each endpoint's list MUST be in the same order. The
   leader's endpoint MUST be the first in the list. The order of the
   `encrypted_input_shares` in a `Report` (see {{uploading-reports}}) MUST be the
-  same as the order in which aggregators appear in this list.
+  same as the order in which aggregators appear in this list. The length of this
+  list MUST be two. [NOTE We intend to support multiple helpers in the future,
+  but for now there is just one.]
 * `collector_config`: The HPKE configuration of the collector (described in
   {{key-config}}). Having participants agree on this absolves collectors of the
   burden of operating an HTTP server. See
@@ -437,13 +451,10 @@ number generator. Each task has the following parameters associated with it:
   newest report in a batch. This defines the boundaries with which the batch
   interval of each collect request must be aligned. (See
   {{batch-parameter-validation}}.)
-* `protocol`: named parameter identifying the core PPM protocol, e.g., Prio or
-  Hits.
 
 ## Uploading Reports
 
-Clients periodically upload reports to the leader, which then distributes the
-individual shares to each helper.
+Clients generate and upload reports to the leader as described here.
 
 ### Key Configuration Request {#key-config}
 
@@ -548,11 +559,10 @@ struct {
   input share.
 * `payload` is the encrypted input share.
 
-To generate the report, the client begins by encoding its measurements as an
-input for the PPM scheme and splitting it into input shares. (Note that the
-structure of each input share depends on the PPM scheme in use, its parameters,
-and the role of aggregator, i.e., whether the aggregator is a leader or helper.)
-To encrypt an input share, the client first generates an HPKE
+To generate the report, the client begins by running the input-distribution
+algorithm of the VDAF scheme. This transforms its measurement into a pair of
+input shares, the first intended for the leader and the second intended for the
+helper. To encrypt each input share, the client first generates an HPKE
 {{!I-D.irtf-cfrg-hpke}} context for the aggregator by running
 
 ~~~
@@ -561,8 +571,8 @@ enc, context = SetupBaseS(pk,
 ~~~
 
 where `pk` is the aggregator's public key, `task_id` is `Report.task_id` and
-`server_role` is a byte whose value is `0x01` if the aggregator is the leader
-and `0x00` if the aggregator is the helper. `enc` is the encapsulated HPKE
+`server_role` is a byte whose value is `0x00` if the aggregator is the leader
+and `0x01` if the aggregator is the helper. `enc` is the encapsulated HPKE
 context and `context` is the HPKE context used by the client for encryption.
 The payload is encrypted as
 
@@ -611,27 +621,31 @@ information specific to the extension.
 
 ## Verifying and Aggregating Reports {#pa-aggregate}
 
-Once a set of clients have uploaded their reports to the leader, the leader can
-send them to the helpers to be verified and aggregated. In order to enable the
-system to handle very large batches of reports, this process can be performed
-incrementally. To aggregate a set of reports, the leader sends an AggregateReq
-to each helper containing those report shares. The helper then processes them
-(verifying the proofs and incorporating their values into the ongoing aggregate)
-and replies to the leader.
+Once a set of clients have uploaded their reports to the leader, the leader and
+helper begin verifying and aggregating them. In order to enable the system to
+handle very large batches of reports, this process can be performed
+incrementally. To aggregate a set of reports, the leader sends a sequence of
+aggregate request to the helper, the first of which contains the helper's
+encrypted input shares. After a number of successful requests, both aggregators
+have recovered a set of valid output shares, which they add into their local
+aggregate output share.
 
-The exact structure of the aggregation flow depends on the PPM scheme.
-Specifically:
+The structure of the aggregation flow is determined by the VDAF [VDAF] being
+executed. In particular, the VDAF specifies a constant number of rounds of
+communication that are required before the aggregators determine if their shares
+are valid. The flow requires as many aggregate requests as there are VDAF
+rounds.
 
-* Some PPM schemes (e.g., Prio) allow the leader to start aggregating reports
-  proactively before all the reports in a batch are received. Others (e.g.,
-  Hits) require all the reports to be present and must be initiated by the
-  collector.
+In addition, some VDAF schemes take an *output parameter* as input, which may
+only be known the aggregators once the previous collect request is completed.
+For example, the output parameter for Hits [BBCGGI21] is the set of candidate
+prefixes, which is only known once the collector has recovered the output from
+the previous collect request. On the other hand, PPM tasks that use prio3 [VDAF]
+can typically begin the process right away because there is no such parameter.
 
-* Processing the reports -- especially verifying the proofs -- may require
-  multiple round trips.
-
-Note that it is possible to aggregate reports from one batch while reports from
-the next batch are coming in.
+[TODO paragraph seems out-of-place. Maybe flesh it out and stick in operational
+considerations?] Note that it is possible to aggregate reports from one batch
+while reports from the next batch are coming in.
 
 This process is illustrated below in {{pa-aggregate-flow}}. In this example, the
 batch size is 20, but the leader opts to process the reports in sub-batches of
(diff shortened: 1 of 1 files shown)


draft-ppm-protocol.md
```diff
@@ -409,8 +415,14 @@ struct {
...
 
-Prior to the start of execution of the protocol, each participant must agree on
-the configuration for each task. A task is uniquely identified by its task ID:
+A task involves evaluation of a specific Verifiable Distributed Aggregation
+Function (VDAF, see [VDAF]) over a set of client inputs. The VDAF determines the
+type of input measurements and the type of the output of the aggregation
+function. It also determines how many aggregators are involved (this document
+currently supports just two) and how many rounds of communication are required
+to verify an input's validity.  Prior to the start of execution of the protocol,
+each participant must agree on the configuration for each task. A task is
```
tgeoghegan, line 424: Is this an appropriate place to discuss the PPM meta-protocol that would capture Prio's query randomness exchange, or is that a later PR?

draft-ppm-protocol.md
```diff
@@ -611,27 +621,31 @@ information specific to the extension.
...
-* Processing the reports -- especially verifying the proofs -- may require
-  multiple round trips.
-
-Note that it is possible to aggregate reports from one batch while reports from
-the next batch are coming in.
+Once a set of clients have uploaded their reports to the leader, the leader and
+helper begin verifying and aggregating them. In order to enable the system to
+handle very large batches of reports, this process can be performed
+incrementally. To aggregate a set of reports, the leader sends a sequence of
+aggregate request to the helper, the first of which contains the helper's
```
tgeoghegan, line 628:
```suggestion
aggregate requests to the helper, the first of which contains the helper's
```

draft-ppm-protocol.md
```diff
@@ -879,26 +915,23 @@ as follows:
...
-request may cause the leader to send a series of aggregate requests to the
-helpers in order to compute their share of the output. Alternately, the leader
-may already have made these requests and can respond immediately. In either case
-it responds to the collector's request as follows.
+* `output_param`, the output parameter with which the VDAF [VDAF] will be
+  evaluated.
+
+Depending on the VDAF and how the leader is configured, the collect request may
+cause the leader to send a series of aggregate requests to the helpers in order
+to compute their share of the output. Alternately, if `output_share` is empty or
```
tgeoghegan, line 931: I think this only makes sense if changed to:
```suggestion
to compute their share of the output. Alternately, if `output_param` is empty or
```
Also, what should the collector put in `output_param` if it is "empty or a well-known value"? Empty byte sequence?

tgeoghegan (changes requested):
I'm pleased to see the `protocol` named parameter go away as that simplifies several protocol messages.

draft-ppm-protocol.md
```diff
@@ -561,8 +571,8 @@ enc, context = SetupBaseS(pk,
 ~~~
 
 where `pk` is the aggregator's public key, `task_id` is `Report.task_id` and
-`server_role` is a byte whose value is `0x01` if the aggregator is the leader
-and `0x00` if the aggregator is the helper. `enc` is the encapsulated HPKE
+`server_role` is a byte whose value is `0x00` if the aggregator is the leader
+and `0x01` if the aggregator is the helper. `enc` is the encapsulated HPKE
```
ekr, line 575: Can we instead make this the index in the array?

draft-ppm-protocol.md
```diff
@@ -548,11 +559,10 @@ struct {
   input share.
 * `payload` is the encrypted input share.
 
-To generate the report, the client begins by encoding its measurements as an
-input for the PPM scheme and splitting it into input shares. (Note that the
-structure of each input share depends on the PPM scheme in use, its parameters,
-and the role of aggregator, i.e., whether the aggregator is a leader or helper.)
-To encrypt an input share, the client first generates an HPKE
+To generate the report, the client begins by running the input-distribution
```
ekr, line 562: This kind of comes out of nowhere. I think you need a section describing the functions in the VDAF in overview.

draft-ppm-protocol.md
ekr, line 628:
```suggestion
aggregate requests to the helper, the first of which contains the helper's
```

ekr: Why are you writing this as if it were one helper? Let's write all this text as if it were many and then have a single place where we say it's not.

draft-ppm-protocol.md
```diff
@@ -611,27 +621,31 @@ information specific to the extension.
...
+aggregate request to the helper, the first of which contains the helper's
+encrypted input shares. After a number of successful requests, both aggregators
+have recovered a set of valid output shares, which they add into their local
+aggregate output share.
+
+The structure of the aggregation flow is determined by the VDAF [VDAF] being
+executed. In particular, the VDAF specifies a constant number of rounds of
+communication that are required before the aggregators determine if their shares
+are valid. The flow requires as many aggregate requests as there are VDAF
+rounds.
```
ekr, line 637: This seems like an unnecessary assumption.

draft-ppm-protocol.md
```diff
@@ -611,27 +621,31 @@ information specific to the extension.
...
-Note that it is possible to aggregate reports from one batch while reports from
-the next batch are coming in.
+Once a set of clients have uploaded their reports to the leader, the leader and
+helper begin verifying and aggregating them. In order to enable the system to
+handle very large batches of reports, this process can be performed
+incrementally. To aggregate a set of reports, the leader sends a sequence of
+aggregate request to the helper, the first of which contains the helper's
+encrypted input shares. After a number of successful requests, both aggregators
+have recovered a set of valid output shares, which they add into their local
+aggregate output share.
```
ekr, line 631: I think this clearly reveals that these subsequent ones just shouldn't be aggregate requests but something else.

AggregateContinue perhaps.

draft-ppm-protocol.md
```diff
@@ -611,27 +621,31 @@ information specific to the extension.
...
+
+The structure of the aggregation flow is determined by the VDAF [VDAF] being
+executed. In particular, the VDAF specifies a constant number of rounds of
+communication that are required before the aggregators determine if their shares
+are valid. The flow requires as many aggregate requests as there are VDAF
+rounds.
+
+In addition, some VDAF schemes take an *output parameter* as input, which may
+only be known the aggregators once the previous collect request is completed.
```
ekr, line 640: Is this the first time a "collect" request has been mentioned.

draft-ppm-protocol.md
```diff
@@ -611,27 +621,31 @@ information specific to the extension.
...
+have recovered a set of valid output shares, which they add into their local
+aggregate output share.
+
+The structure of the aggregation flow is determined by the VDAF [VDAF] being
+executed. In particular, the VDAF specifies a constant number of rounds of
+communication that are required before the aggregators determine if their shares
+are valid. The flow requires as many aggregate requests as there are VDAF
+rounds.
+
+In addition, some VDAF schemes take an *output parameter* as input, which may
```
ekr, line 639: I would not call this "output" but rather "state"

draft-ppm-protocol.md
```diff
@@ -672,51 +686,70 @@ continuation of a previous sub-batch?]
 ### Aggregate Request
 
 The AggregateReq request is used by the leader to send a set of reports to the
-helper. These reports MUST all be associated with the same PPM task. [[OPEN
-ISSUE: And the same batch, right?]]
+helper. These reports MUST all be associated with the same PPM task. The reports
+MAY span multiple batches, but MUST be transmitted in the order specified in
+{{anti-replay}}.
```
ekr, line 691: Why are we letting this span multiple batches? That seems weird.

draft-ppm-protocol.md
```diff
@@ -672,51 +686,70 @@ continuation of a previous sub-batch?]
...
 
-For each aggregator endpoint `[aggregator]` in `AggregateReq.task_id`'s
-parameters except its own, the leader sends a POST request to
-`[aggregator]/aggregate` with the following message:
+For the second aggregator endpoint `[aggregator]` in `AggregateReq.task_id`'s
+parameters, the leader sends a POST request to `[aggregator]/aggregate` with the
+following message:
```
ekr, line 695: See above about cardinality.

draft-ppm-protocol.md
```diff
@@ -672,51 +686,70 @@ continuation of a previous sub-batch?]
...
   TaskID task_id;
   opaque helper_state<0..2^16>;
+  select (type) {
+    case start: opaque output_param<0..2^16-1>;
+    case next:  Empty;
+  }
   AggregateSubReq seq<1..2^24-1>;
 } AggregateReq;
+
+enum { start(0), next(1), } AggregateReqType;
```
ekr, line 709: See above. I think these should just be different messages.

draft-ppm-protocol.md
```diff
@@ -672,51 +686,70 @@ continuation of a previous sub-batch?]
...
 
-The `time`, `nonce`, and `extensions` fields have the same value as those in the
-report uploaded by the client. Similarly, the `helper_share` field is the
-`EncryptedInputShare` from the `Report` whose index in
-`Report.encrypted_input_shares` is equal to the index of `[aggregator]` in the
-task's aggregator endpoints. [OPEN ISSUE: We usually only need to send this in
-the first aggregate request. Shall we exclude it in subsequent requests
-somehow?] The remainder of the structure is dedicated to the protocol-specific
-request parameters.
+Each sub-request for a given report includes the report's timestamp, nonce, and
```
ekr, line 732: I don't understand "each sub-request for a given report". Is there some other structure?

"Each subrequest corresponds to a single report"

draft-ppm-protocol.md
```diff
@@ -672,51 +686,70 @@ continuation of a previous sub-batch?]
...
+extensions (`time`, `nonce`, and `extensions` respectively) and a value called
+the verification message (`verify_message`), generated as described in the next
+paragraph. The first sub-request also carries the helper's encrypted input share
+(`helper_share`).
```
ekr, line 736: This is not the first sub-request in the list but rather the first time you send one.

ekr (changes requested):
I think it would be helpful to separate out the change to talk about VBAF from the other technical changes you have made here.

draft-ppm-protocol.md
tgeoghegan: I agree with Chris here. I think there's no reason to require leaders to construct `AggregateReq`s around batch boundaries, and doing it this way enables really simple leader implementations that send an `AggregateReq` to helper every time some fixed number of reports is received.

A helper will consider all the `AggregateSubReq`s independently, and it's easy for it to decide what batch a particular subreq belongs to (it just has to truncate the subreq timestamp to the batch interval) so there's no need for all the subreqs in a req to belong to the same batch.

ekr: I don't understand the alleged semantics of this. Suppose that I send reqs from batch 1 and batch 2, what does the helper do?

tgeoghegan: Let's say it's Prio, in which scenario the aggregators can sum inputs as soon as the proofs check out without waiting for the collect protocol to begin. Further let's say an aggregator sends an `AggregateReq` that contains `AggregateSubReq`s whose timestamps span multiple batches.

A helper implementation would have a map where the keys are batch intervals and the values are the accumulated value for that batch window. Upon receipt of an `AggregateReq`, the helper iterates over the `AggregateSubReq`s and for each:

- decrypts the input share and proof share;
- verifies the proof (it has its own proof share and got leader's proof share in the `AggregateSubReq`);
- truncates the report timestamp to the batch interval (e.g. with [`chrono::DurationRound::duration_trunc`](https://docs.rs/chrono/0.4.19/chrono/trait.DurationRound.html#tymethod.duration_trunc) or [`time.Time.Truncate`](https://pkg.go.dev/time#Time.Truncate)) to figure out what batch the input belongs to;
- sums the input share into  the accumulated value for the batch interval from the map described earlier.

Later, a collector issues a `CollectReq` to the leader which specifies some `batch_interval`, and the leader sends a corresponding `OutputShareReq` to helper. Helper just looks up the `batch_interval` in its map of accumulators and responds with the value.

Enabling this kind of eager aggregation is a design goal, and I don't see that it requires that `AggregateReq`s be constructed on batch interval boundaries.

ekr: This is going to interact very badly with any kind of state offloading because you will end up with the state token containing multiple batches and it will be very unclear when you can abandon it or do anything in parallel.

More apropos to the current moment, the current text has an open issue here and this PR just decides that in the middle of making some other totally orthogonal change. If people want there to be an open issue where we discuss this design issue, then fine, but it doesn't belong in this PR.

tgeoghegan: >This is going to interact very badly with any kind of state offloading because you will end up with the state token containing multiple batches and it will be very unclear when you can abandon it or do anything in parallel.

Yes, I encountered this when implementing a prototype of this and tried to capture these questions in #150. My conclusion is that the opaque `helper_state` blob has to be per-task, and we need the PPM protocol to make that explicit so that the leader knows which state blob to send to which helper in which aggregate request.

Your point about parallelism is a really good one, though. My 2c is that we should put `OPEN ISSUE: And the same batch, right?` back and hash this question out in #150 + a later PR.

cjpatton:
Hi @tgeoghegan, @ekr, thanks for the super useful feedback! Based on your feedback, and in light of a super helpful conversation with @chris-wood today, I'm going to close this PR and open a new one after some major revisions. Specifically, @chris-wood came up with some changes to the VDAF syntax that will make the PPM spec a lot cleaner. I'll get cracking on this next Tuesday.

draft-ppm-protocol.md
cjpatton: This comment is solved by having the outputs of the VDAF setup algorithm be part of the task parameters.

draft-ppm-protocol.md
cjpatton: It's not an assumption. This follows from the VDAF definition and how its execution for two aggregators is specified here.

draft-ppm-protocol.md
cjpatton: I'm reverting this change and we'll punt.

draft-ppm-protocol.md
ekr: "Assumption" in this case means "that's how you've defined VDAF". But suppose I define a VDAF' which is like VDAF but has an indeterminate number of rounds.

draft-ppm-protocol.md
cjpatton: This will be a somewhat major refactor, so I'll wait to address until the next PR.

draft-ppm-protocol.md
cjpatton: No, it's mentioned above. I think it should already be mentioned in the overview section.

draft-ppm-protocol.md
cjpatton: "Output parameter" is the same term used in the VDAF syntax. "State" refers to the aggregator's state.

draft-ppm-protocol.md
cjpatton: Good catch. Should be "Each sub-request includes the report's ..."

draft-ppm-protocol.md
cjpatton: I think this might be fixed by splitting up the first aggregate request from the others, as you suggested above.

draft-ppm-protocol.md
ekr: I don't really think you can refer to whatever choices you happen to have made in the VDAF draft -- which has even less consensus than this document -- as authoritative for this kind of issue. My point is that this is misleading nomenclature and we should change it both here and in that document. It would be a different situation of the VDAF document were an RFC and so we were trying to conform to that, but it's not.

draft-ppm-protocol.md
cjpatton: Punting this discussion to the new PR.

PR #154, opened 2021-09-28
Specify the protocol in terms of VDAF evaluation

### LINKED ISSUE #98: The wrapper-core protocol interface

cjpatton:
Closes #98.

### DIFF
--- a/draft-gpew-priv-ppm.md
+++ b/draft-gpew-priv-ppm.md
@@ -33,7 +33,6 @@ author:
        organization: Cloudflare
        email: caw@heapingbits.net
 
-
 informative:
 
   CGB17:
@@ -79,6 +78,11 @@ informative:
     target: "https://privacytools.seas.harvard.edu/files/privacytools/files/complexityprivacy_1.pdf"
     author:
       - ins: S. Vadhan
+
+  VDAF:
+    title: "Verifiable Distributed Aggregation Functions"
+    date: 2021-09-21
+    target: "TODO"
 
 --- abstract
 
@@ -190,6 +194,12 @@ This document uses the protocol definition language of {{!RFC8446}}.
 
 # Overview {#overview}
 
+[TODO Give an overview of the VDAF syntax and frame the PPM protocol around
+evaluation of a VDAF.]
+
+[TODO Name the various requests, including collect requests, and say what they
+are.]
+
 The protocol is executed by a large set of clients and a small set of servers.
 We call the servers the *aggregators*. Each client's input to the protocol is a
 set of measurements (e.g., counts of some user behavior). Given the input set of
@@ -288,7 +298,7 @@ includes the following parameters:
 * The values to be measured;
 * The statistic to be computed (e.g., sum, mean, etc.);
 * The set of aggregators and necessary cryptographic keying material to use; and
-* The PPM scheme to use. This is to some extent dictated by the previous
+* The VDAF scheme to use. This is to some extent dictated by the previous
   choices.
 * The minimum "batch size" of reports which can be aggregated.
 * The rate at which measurements can be taken, i.e., the "minimum batch window".
@@ -306,7 +316,7 @@ time.
 
 The leader distributes the shares to the helpers and orchestrates the process of
 verifying them (see {{validating-inputs}}) and assembling them into a final
-measurement for the collector. Depending on the PPM scheme, it may be possible
+measurement for the collector. Depending on the VDAF scheme, it may be possible
 to incrementally process each report as it comes in, or may be necessary to wait
 until the entire batch of reports is received.
 
@@ -409,16 +419,23 @@ struct {
 
 ## Task Configuration {#task-configuration}
 
-Prior to the start of execution of the protocol, each participant must agree on
-the configuration for each task. A task is uniquely identified by its task ID:
+A task centers around evaluation of a specific Verifiable Distributed
+Aggregation Function (VDAF, see [VDAF]) over a set of client inputs. The VDAF
+determines the type of input measurements and the type of the output. It also
+determines how many aggregators are involved and the maximum number of rounds of
+communication among the aggregators that are required to verify the output's
+validity. Prior to the start of execution of the protocol, each participant must
+agree on the configuration parameters for the task. A task is uniquely
+identified by its task ID:
 
 ~~~
 opaque TaskId[32];
 ~~~
 
-A `TaskId` is a globally unique sequence of bytes. It is RECOMMENDED that this
-be set to a random string output by a cryptographically secure pseudorandom
-number generator. Each task has the following parameters associated with it:
+A `TaskId` is a unique identifier for a specific task among a set of tasks. It
+is RECOMMENDED that this be set to a random string output by a cryptographically
+secure pseudorandom number generator. Each task has the following parameters
+associated with it:
 
 * `aggregator_endpoints`: A list of URLs relative to which an aggregator's API
   endpoints can be found. Each endpoint's list MUST be in the same order. The
@@ -437,13 +454,21 @@ number generator. Each task has the following parameters associated with it:
   newest report in a batch. This defines the boundaries with which the batch
   interval of each collect request must be aligned. (See
   {{batch-parameter-validation}}.)
-* `protocol`: named parameter identifying the core PPM protocol, e.g., Prio or
-  Hits.
+* `public_param` and `verify_param`: The outputs of the VDAF setup algorithm,
+  referred to as the public parameter and verification parameter respectively.
+  The client uses the public parameter to split its input into input shares, and
+  the aggregators use the verification parameter to verify the input's validity.
+  The public parameter MUST be distributed to all clients and MAY be distributed
+  to the servers. The verification parameter MUST be distributed to all
+  aggregators and MUST NOT be revealed to the clients.
+
+The VDAF scheme associated with a task MUST be defined for two and only two
+aggregators. [NOTE: We plan to lift this restriction once we're ready to specify
+the protocol for multiple helpers.]
 
 ## Uploading Reports
 
-Clients periodically upload reports to the leader, which then distributes the
-individual shares to each helper.
+Clients generate and upload reports to the leader as described here.
 
 ### Key Configuration Request {#key-config}
 
@@ -499,9 +524,9 @@ the cache lifetime in order to avoid rejecting reports.
 
 ### Upload Request
 
-Clients upload reports by using an HTTP POST to `[leader]/upload`, where
-`[leader]` is the first entry in the task's aggregator endpoints. The payload is
-structured as follows:
+Clients upload reports by sending an HTTP POST request to `[leader]/upload`,
+where `[leader]` is the first entry in the task's aggregator endpoints. The
+payload is structured as follows:
 
 ~~~
 struct {
@@ -516,9 +541,7 @@ struct {
 This message is called the client's *report*. It contains the following fields:
 
 * `task_id` is the task ID of the task for which the report is intended.
-* `time` is the time at which the report was generated. This field is used by
-  the aggregators to ensure the report appears in at most one batch. (See
-  {{anti-replay}}.)
+* `time` is the time at which the report was generated.
 * `nonce` is a random number chosen by the client generating the report. This
   and the timestamp field are used by the aggregators to ensure that each report
   appears at most once in a batch. (See {{anti-replay}}.)
@@ -548,11 +571,10 @@ struct {
   input share.
 * `payload` is the encrypted input share.
 
-To generate the report, the client begins by encoding its measurements as an
-input for the PPM scheme and splitting it into input shares. (Note that the
-structure of each input share depends on the PPM scheme in use, its parameters,
-and the role of aggregator, i.e., whether the aggregator is a leader or helper.)
-To encrypt an input share, the client first generates an HPKE
+To generate the report, the client begins by running the VDAF input-distribution
+algorithm on the public parameter `public_param` and its measurement. This
+transforms the measurement into a sequence of input shares, one for each
+aggregator. To encrypt each input share, the client first generates an HPKE
 {{!I-D.irtf-cfrg-hpke}} context for the aggregator by running
 
 ~~~
@@ -561,10 +583,10 @@ enc, context = SetupBaseS(pk,
 ~~~
 
 where `pk` is the aggregator's public key, `task_id` is `Report.task_id` and
-`server_role` is a byte whose value is `0x01` if the aggregator is the leader
-and `0x00` if the aggregator is the helper. `enc` is the encapsulated HPKE
-context and `context` is the HPKE context used by the client for encryption.
-The payload is encrypted as
+`server_role` is a byte whose value is the index of the aggregator in the
+aggregator list. The bytestring `enc` is the encapsulated HPKE context and the
+bytestring `context` is the HPKE context used by the client for encryption. The
+payload is encrypted as
 
 ~~~
 payload = context.Seal(time || nonce || extensions, input_share)
@@ -580,9 +602,11 @@ the leader responds with status 200 and an empty body.
 
 The leader responds to requests with out-of-date `HpkeConfig.id` values,
 indicated by `EncryptedInputShare.config_id`, with status 400 and an error of
-type 'outdatedConfig'. Clients SHOULD invalidate any cached aggregator
-`HpkeConfig` and retry with a freshly generated Report. If this retried report
-does not succeed, clients MUST abort and discontinue retrying.
+type 'outdatedConfig'. [OPEN ISSUE: How does the leader know if the HPKE config
+of a helper is invalid? Does it have to make a key-config request?] Clients
+SHOULD invalidate any cached aggregator `HpkeConfig` and retry with a freshly
+generated Report. If this retried report does not succeed, clients MUST abort
+and discontinue retrying.
 
 ### Upload Extensions {#upload-extensions}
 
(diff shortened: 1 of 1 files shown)


draft-ppm-protocol.md
```diff
@@ -647,147 +676,222 @@ be cryptographically protected as described in {{helper-state}}.
...
 
-### Aggregate Request
+### Verify-Start Request
 
-The AggregateReq request is used by the leader to send a set of reports to the
-helper. These reports MUST all be associated with the same PPM task. [[OPEN
-ISSUE: And the same batch, right?]]
+The *verify-start request* request is used by the leader to send a set of
+reports to the helper. These reports MUST all be associated with the same PPM
+task. [[OPEN ISSUE: And the same batch, right?]]
```
tgeoghegan, line 702: A link to #150 would be appropriate in the `[[OPEN ISSUE]]`

draft-ppm-protocol.md
```diff
@@ -647,147 +676,222 @@ be cryptographically protected as described in {{helper-state}}.
...
+parameters, the leader sends a POST request to `[aggregator]/aggregate_start`
+with the following message:
 
 ~~~
 struct {
   TaskID task_id;
   opaque helper_state<0..2^16>;
-  AggregateSubReq seq<1..2^24-1>;
-} AggregateReq;
+  opaque output_param<0..2^16-1>;
```
tgeoghegan, line 712: Are there any risks with letting the leader insert arbitrary `output_param` into requests sent to helper? In the Priov3 case, it seems this should always be empty or absent (because the output param is already known to all the parties participating in a task).

In the Hits case, though, a leader could tamper with the string prefixes in the collector's queries. This might be beyond the scope of this particular PR, but maybe the helper and collector should establish a mutually authenticated channel using each other's HPKE configs to ensure leader can't tamper with the output param.

draft-ppm-protocol.md
```diff
@@ -647,147 +676,222 @@ be cryptographically protected as described in {{helper-state}}.
...
+
+Each sub-request includes the report's timestamp, nonce, and extensions (`time`,
+`nonce`, and `extensions` respectively) and a value called the verification
+message (`verify_message`), generated as described in the next paragraph. The
+first sub-request also carries the helper's encrypted input share
+(`helper_share`).
+
+The leader generates the verification message by decrypting its input share
+`leader_share`. To do so, it first looks up the HPKE secret key `sk` associated
+with `leader_share.aggregato_config_id`. Next, it runs
```
tgeoghegan, line 741:
```suggestion
with `leader_share.aggregator_config_id`. Next, it runs
```

draft-ppm-protocol.md
```diff
@@ -647,147 +676,222 @@ be cryptographically protected as described in {{helper-state}}.
...
+to the previous request and `state` is the aggregator's VDAF state.
+
+### Recovering the Output Share
+
+Once all rounds of the VDAF have been completed for a given report, each
+aggregator runs the VDAF verify-finish algorithm to recover its share of the
+output:
+
+~~~
+output_share = vdaf_finish(statre, inbound_message)
```
tgeoghegan, line 872:
```suggestion
output_share = vdaf_finish(state, inbound_message)
```

draft-ppm-protocol.md
```diff
@@ -647,147 +676,222 @@ be cryptographically protected as described in {{helper-state}}.
...
+(state, AggregateSubReq.verify_message) = vdaf_next(
+          state, inbound_message)
+~~~
+
+where `inbound_message` is the verification message sent by the peer in response
+to the previous request and `state` is the aggregator's VDAF state.
+
+### Recovering the Output Share
+
+Once all rounds of the VDAF have been completed for a given report, each
```
tgeoghegan, line 867: The helper has nowhere but the `helper_state` blob to keep track of how many rounds have been executed. How does it know whether to execute `vdaf_next` or `vdaf_finish` for any `VerifyNextSubReq`? Perhaps we are missing something like `vdaf_is_finished(state) -> int` (but with a good name instead of that bad one)?

draft-ppm-protocol.md
```diff
@@ -879,26 +980,23 @@ as follows:
...
 
-The remainder of the message is dedicated to the protocol-specific request
-parameters.
-
-Depending on the PPM scheme and how the leader is configured, the collect
-request may cause the leader to send a series of aggregate requests to the
-helpers in order to compute their share of the output. Alternately, the leader
-may already have made these requests and can respond immediately. In either case
-it responds to the collector's request as follows.
+Depending on the VDAF and how the leader is configured, th collect request may
```
tgeoghegan, line 994:
```suggestion
Depending on the VDAF and how the leader is configured, the collect request may
```

draft-ppm-protocol.md
```diff
@@ -879,26 +980,23 @@ as follows:
 struct {
   TaskID task_id;
   Interval batch_interval;
-  select (protocol) { // Protocol corresponding to task_id
-    case prio: PrioCollectReq;
-    case hits: HitsCollectReq;
-  }
+  opaque output_param<0..2^16-1>;
```
tgeoghegan, line 983: What should a collector put here in the Prio case? Empty string?

tgeoghegan (changes requested):
First bunch of comments. I want to (re-)read https://github.com/cjpatton/ppm and then possibly comment some more!

draft-ppm-protocol.md
```diff
@@ -562,9 +585,9 @@ enc, context = SetupBaseS(pk,
 
 where `pk` is the aggregator's public key, `task_id` is `Report.task_id` and
 `server_role` is a byte whose value is `0x01` if the aggregator is the leader
-and `0x00` if the aggregator is the helper. `enc` is the encapsulated HPKE
-context and `context` is the HPKE context used by the client for encryption.
-The payload is encrypted as
+and `0x00` if the aggregator is the helper. The bytestring `enc` is the
```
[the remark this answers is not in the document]
cjpatton: Good catch.

draft-ppm-protocol.md
cjpatton: Yes.

draft-ppm-protocol.md
cjpatton: The VDAF fixes the number of rounds. The first round uses vdaf_start and subsequent rounds use vdaf_next.

draft-ppm-protocol.md
cjpatton: In terms of soundness, there is a risk of a network attacker tampering with this. For that we'll need some sort of sender authentication in the leader<->helper channel. (We've discussed this type of threat before, but I'm not sure we have a ticket for it.)

In terms of privacy, mutual authentication doesn't buy us anything because the leader can still cheat. That said, privacy ought to hold here regardless of what the cheating leader does (so long as the helper itself is honest). Of course, we need to prove this for every VDAF we want to implement. (For Prio this is trivial since the only valid output parameter is "".)

tgeoghegan: I wrote this up in #155. I think we should just put a TODO or a NOTE here referencing this issue and move on with this PR.

draft-ppm-protocol.md
tgeoghegan: I buy that the number of rounds is a constant determined by the VDAF. But how does helper know what round it's on when it receives a `VerifyNextReq`? Implicitly a helper would have to keep a count of executed rounds in its `helper_state`, or we could require the VDAF to encode that into the `state` that gets passed into `vdaf_next`.

draft-ppm-protocol.md
tgeoghegan: Sounds fine, but I think we should have explicit text instructing implementations to do this.

draft-ppm-protocol.md
cjpatton: I'd also be fine with adding a "round_number" field to the request body.

draft-ppm-protocol.md
cjpatton: The instruction is "the leader copies the output parameter verbatim from the collect request to verify-start request". Is that not already explicitly there?

draft-ppm-protocol.md
tgeoghegan: That could work, but then we have to consider what if any risks there are to allowing the leader to control the round number. If we put it in helper-controlled state, then it's protected by the encryption we already require from helper implementations.

draft-ppm-protocol.md
tgeoghegan: So it is! I did notice one typo in the paragraph below. Resolving this thread.

draft-ppm-protocol.md
```diff
@@ -879,26 +980,23 @@ as follows:
...
-parameters.
-
-Depending on the PPM scheme and how the leader is configured, the collect
-request may cause the leader to send a series of aggregate requests to the
-helpers in order to compute their share of the output. Alternately, the leader
-may already have made these requests and can respond immediately. In either case
-it responds to the collector's request as follows.
+Depending on the VDAF and how the leader is configured, the collect request may
+cause the leader to send a series of requests to the helper in order to compute
+their share of the output. Alternately, if `output_share` is empty or a
```
tgeoghegan, line 996:
```suggestion
their share of the output. Alternately, if `output_param` is empty or a
```

draft-ppm-protocol.md
cjpatton: Good point! Either way the helper is going to have to keep track of its state, and thus what message to expect next. I think we should make no changes here and just make it clear that the helper has to keep track of this.

draft-ppm-protocol.md
```diff
@@ -647,147 +676,222 @@ be cryptographically protected as described in {{helper-state}}.
...
-somehow?] The remainder of the structure is dedicated to the protocol-specific
-request parameters.
+where `task_id`, `time`, `nonce`, and `extensions` are the fields of the report
+with the same name. If decryption succeeds, it runs the VDAF verify-start
+algorithm to compute its local VDAF state and the verification message:
 
-In order to provide replay protection, the leader is required to send aggregate
+~~~
+(state, VerifyStartSubReq.verify_message) = vdaf_start(
+      verify_param, aggregation_param, time || nonce, input_share)
```
tgeoghegan, line 756: Do we not need a parameter indicating whether the entity running `vdaf_start` is the leader or helper, so that it can tell whether the input share is compressed or not (if the VDAF is Prio)? Or do we encode that information into `input_share`?

tgeoghegan (approved):

draft-ppm-protocol.md
cjpatton: It's encoded by the `input_share`. I know you didn't like this redundancy, but it makes the VDAF syntax a lot simpler.

draft-ppm-protocol.md
```diff
@@ -409,22 +419,31 @@ struct {
 
 ## Task Configuration {#task-configuration}
 
-Prior to the start of execution of the protocol, each participant must agree on
-the configuration for each task. A task is uniquely identified by its task ID:
+A task centers around evaluation of a specific Verifiable Distributed
+Aggregation Function (VDAF, see [VDAF]) over a set of client inputs. The VDAF
+determines the type of input measurements and the type of the output of the
+aggregation function. It also determines how many aggregators are involved (this
+document currently supports just two) and how many rounds of communication among
```
ekr, line 426: Please file an issue for this rounds question. I am not convinced that we should commit to a fixed number of rounds as a protocol invariant.

draft-ppm-protocol.md
```diff
@@ -409,22 +419,31 @@ struct {
...
+secure pseudorandom number generator. Each task has the following parameters
+associated with it:
 
 * `aggregator_endpoints`: A list of URLs relative to which an aggregator's API
   endpoints can be found. Each endpoint's list MUST be in the same order. The
   leader's endpoint MUST be the first in the list. The order of the
   `encrypted_input_shares` in a `Report` (see {{uploading-reports}}) MUST be the
-  same as the order in which aggregators appear in this list.
+  same as the order in which aggregators appear in this list. The length of this
+  list MUST be two. [NOTE We intend to support multiple helpers in the future,
```
ekr, line 445: This restriction should appear in precisely one place so that we do not have to change it everywhere if we add multiple helpers. The first location is probably the right one.

draft-ppm-protocol.md
```diff
@@ -437,13 +456,17 @@ number generator. Each task has the following parameters associated with it:
   newest report in a batch. This defines the boundaries with which the batch
   interval of each collect request must be aligned. (See
   {{batch-parameter-validation}}.)
-* `protocol`: named parameter identifying the core PPM protocol, e.g., Prio or
-  Hits.
+* `public_param` and `verify_param`: The outputs of the VDAF setup algorithm,
+  referred to as the public parameter and verification parameter respectively.
+  The client uses the poblic parameter to split its input into input shares, and
```
ekr, line 461:
```suggestion
  The client uses the public parameter to split its input into input shares, and
```

draft-ppm-protocol.md
```diff
@@ -437,13 +456,17 @@ number generator. Each task has the following parameters associated with it:
...
+  the aggregators use the verification parameter to verify the input's validity.
+  The public parameter MUST be distributed to all clients and MAY be distributed
+  to the servers. The verification parameter MUST be distributed to all
+  aggregators and MUST NOT be revealed to the clients.
```
ekr, line 465: It's quite hard to understand the role of the verification parameter without a description of the vdaf system

draft-ppm-protocol.md
```diff
@@ -548,23 +571,23 @@ struct {
...
-input for the PPM scheme and splitting it into input shares. (Note that the
-structure of each input share depends on the PPM scheme in use, its parameters,
-and the role of aggregator, i.e., whether the aggregator is a leader or helper.)
-To encrypt an input share, the client first generates an HPKE
-{{!I-D.irtf-cfrg-hpke}} context for the aggregator by running
+To generate the report, the client begins by running the VDAF input-distribution
+algorithm on the public parameter `public_param` and its measurement. This
+transforms the measurement into a pair of input shares, the first intended for
+the leader and the second intended for the helper. To encrypt each input share,
+the client first generates an HPKE {{!I-D.irtf-cfrg-hpke}} context for the
```
ekr, line 578:
```suggestion
transforms the measurement into a set of input shares, one for each aggregator. To encrypt each input share,
the client first generates an HPKE {{!I-D.irtf-cfrg-hpke}} context for the
```

draft-ppm-protocol.md
```diff
@@ -548,23 +571,23 @@ struct {
...
                           "pda input share" || task_id || server_role)
 ~~~
 
 where `pk` is the aggregator's public key, `task_id` is `Report.task_id` and
-`server_role` is a byte whose value is `0x01` if the aggregator is the leader
-and `0x00` if the aggregator is the helper. `enc` is the encapsulated HPKE
-context and `context` is the HPKE context used by the client for encryption.
-The payload is encrypted as
+`server_role` is a byte whose value is `0x00` if the aggregator is the leader
+and `0x01` if the aggregator is the helper. The bytestring `enc` is the
```
ekr, line 588:
```suggestion
`server_role` is a byte whose value is the index of the aggregator in the aggregator list. The bytestring `enc` is the
```

Or some such...

draft-ppm-protocol.md
```diff
@@ -611,33 +635,38 @@ information specific to the extension.
...
-Note that it is possible to aggregate reports from one batch while reports from
-the next batch are coming in.
-
-This process is illustrated below in {{pa-aggregate-flow}}. In this example, the
-batch size is 20, but the leader opts to process the reports in sub-batches of
-10. Each sub-batch takes two round-trips to process. Once both sub-batches have
-been processed, the leader can issue an OutputShareReq in order to retrieve the
-helper's aggregated result.
+Once a set of clients have uploaded their reports to the leader, the leader and
+helper begin verifying and aggregating them. In order to enable the system to
```
ekr, line 639:
```suggestion
helper(s) begin verifying and aggregating them. In order to enable the system to
```

ekr (changes requested):
Only got partway through, but I have a number of comments.

draft-ppm-protocol.md
cjpatton: Good call. I removed this MUST here and the paranthetical above saying the protocol is only defined for two aggregators. I added a MUST for this at the end of {{task-configuration}}.

draft-ppm-protocol.md
cjpatton: Before I do, I'd like to chat in person about the implication of this restriction and why I feel it's a safe assumption. Just to be clear: the intention is that the VDAF scheme determines the number of rounds, not the wrapper protocol (i.e., this specification).

draft-ppm-protocol.md
cjpatton: Works for me, I thought we were still blocking on this change.

draft-ppm-protocol.md
cjpatton: I think if we're going to make this change, we'll have to actually specify how multiple helpers works. We can do this here, but my preference would be to wait for antoher PR.

draft-ppm-protocol.md
```diff
@@ -647,147 +676,222 @@ be cryptographically protected as described in {{helper-state}}.
...
+
+The leader generates the verification message by decrypting its input share
+`leader_share`. To do so, it first looks up the HPKE secret key `sk` associated
+with `leader_share.aggregator_config_id`. Next, it runs
+
+~~~
+context = SetupBaseR(leader_share.enc, sk,
+                     "pda input share" || task_id || 0x00)
+input_share = context.Open(time || nonce || extensions,
+                           leader_share.payload)
```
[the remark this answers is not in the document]
cjpatton: `SetupBaseR` has a different info string, but yeah, we should unify this when we allow for multiple helpers.

draft-ppm-protocol.md
cjpatton: Ack, I'll add a description of VDAFs in the overview for this PR.

draft-ppm-protocol.md
```diff
@@ -611,33 +635,38 @@ information specific to the extension.
...
-This process is illustrated below in {{pa-aggregate-flow}}. In this example, the
-batch size is 20, but the leader opts to process the reports in sub-batches of
-10. Each sub-batch takes two round-trips to process. Once both sub-batches have
-been processed, the leader can issue an OutputShareReq in order to retrieve the
-helper's aggregated result.
+Once a set of clients have uploaded their reports to the leader, the leader and
+helper begin verifying and aggregating them. In order to enable the system to
+handle very large batches of reports, this process can be performed
+incrementally. To aggregate a set of reports, the leader sends a sequence of
+requests to the helper, the first of which contains the helper's encrypted input
```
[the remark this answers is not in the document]
cjpatton: This is not currently specified.

cjpatton:
@tgeoghegan note that I'm closing this in favor of #168. Let me know if you'd like me to recreate it so that we have something to reference for abetterinternet/ppm-prototype.

PR #156, opened 2021-10-03
Clarify timing of report aggregations & purpose of anti-replay.

### LINKED ISSUE #153: Clarify need for client-provided timestamps & replay protection
Clients in the PPM send a client-generated `(timestamp, nonce)` tuple with their upload requests, and encrypt their reports using a nonce that is derived from this `(timestamp, nonce)` tuple. [section 4.2.2]

The aggregators use the `(timestamp, nonce)` tuple is used as a unique identifier for an individual report, to enable an "anti-replay" mechanism described in section 4.4.2. The reasoning for this anti-replay mechanism is described as:

> Using a report multiple times within a single batch, or using the same report in multiple batches, is considered a privacy violation.

I argue that this mechanism does not practically stop any of the actors from submitting some arbitrary report multiple times in a single batch: malicious clients can effectively replay any of their reports by re-submitting the report content re-encrypted with a new `(timestamp, nonce)` tuple; malicious servers can effectively pose as a client and perform the described "malicious client" attack themselves.

But per discussion with timg@, the purpose of the anti-replay mechanism is not to stop a malicious server from doing arbitrary ballot-stuffing; it is to protect a malicious leader from using the same legitimate *client-provided* report multiple times, which could enable certain attacks on client privacy.

Separately, there is currently nothing in the spec placing any requirements or server validations on the timestamp value. A misconfigured or malicious client could send a "poisoned" report with a timestamp of `MAX_UINT64`. As the leader must send reports to the helper ordered by `(timestamp, nonce)` [sections 4.3.1, 4.4.2], once this report was sent to the helper as part of an aggregation request the leader would no longer be able to submit any additional reports (and if it tried, the helper would silently filter them).

Separately (again), message deduplication based on a total ordering of message timestamps will lead to messages that are received in a delayed fashion being dropped, since the leader will no longer be able to include them in aggregation requests.

Proposals:
* Update the explanation of the anti-replay mechanism to describe the intended protection more explicitly, i.e. that it covers only replay of messages sent by a client, with replay attempted by the server (I think).
* Think through and add server validation on the client-submitted timestamp. For example, we might decide that only reports in the current batch, or a previous batch falling within a "grace window", are allowed. (I am unsure if this would be acceptable for Heavy Hitters or other PPM algorithms that require a more "online" query capability.)
* Consider switching clients from reporting a timestamp to reporting a `batch_id` counter. The current timestamp-based report could be mapped to a batch ID by something like `timestamp / batch_duration` (though using a counter opens up the possibility of other methods of determining the current `batch_id`). Reports should still be deduplicated by the `(batch_id, nonce)` tuple. The ordering-based deduplication mechanism would no longer work, but as it stands this ordering-based deduplication will cause delayed messages to be dropped anyway; whatever replaces the ordering-based deduplication mechanism will likely require more storage than the ordering-based mechanism, but this might be mitigated by bounding what reports can be accepted by `batch_id`, per the previous proposal.

BranLwyd:
As previously written, the timing of report aggregations would allow
aggregation of reports with timestamps far into the future, which would
allow a single client reporting a timestamp of INT_MAX to DOS the
aggregation system due to the way timestamps are used in the anti-replay
mechanism. Given the severity of not following this portion of the
specification, I upgrade the SHOULD to a MUST as well.

Fixes #153.

### DIFF
--- a/draft-ppm-protocol.md
+++ b/draft-ppm-protocol.md
@@ -776,10 +776,13 @@ and constructs its response to the aggregate request.
 #### Leader State
 
 The leader is required to issue aggregate requests in order, but reports are
-likely to arrive out-of-order. The leader SHOULD buffer reports for a time
-period proportional to the batch window before issuing the first aggregate
-request. Failure to do so will result in out-of-order reports being dropped by
-the helper.
+likely to arrive out-of-order. The leader MUST store reports until their
+timestamp is sufficiently far in the past (proportional to the batch window)
+before including them in an aggregate request. Failure to do so will result in
+reports being dropped by the helper. The leader SHOULD NOT accept reports from
+clients timestamped too far into the future (proportional to the batch window).
+Failure to do so will result in reports from malicious, misconfigured, or
+malfunctioning clients being stored for an arbitrary amount of time.
 
 #### Helper State
 
@@ -955,10 +958,11 @@ aggregator has validated and which fall in the batch interval of the request.
 
 ### Anti-replay {#anti-replay}
 
-Using a report multiple times within a single batch, or using the same report in
-multiple batches, is considered a privacy violation. To prevent such replay
-attacks, this specification defines a total ordering on reports that aggregators
-can use to ensure that reports are aggregated once.
+Using a client-provided report multiple times within a single batch, or using
+the same report in multiple batches, may allow a server to learn information
+about the client's measurement. To prevent such replay attacks, this
+specification defines a total ordering on reports that aggregators can use to
+ensure that reports are aggregated once.
 
 Aggregate requests are ordered as follows: We say that a report `R2` follows
 report `R1` if either `R2.time > R1.time` or `R2.time == R1.time` and `R2.nonce


cjpatton (approved):
Yup, I agree this is a good idea.

draft-ppm-protocol.md
```diff
@@ -776,10 +776,13 @@ and constructs its response to the aggregate request.
 #### Leader State
 
 The leader is required to issue aggregate requests in order, but reports are
-likely to arrive out-of-order. The leader SHOULD buffer reports for a time
-period proportional to the batch window before issuing the first aggregate
-request. Failure to do so will result in out-of-order reports being dropped by
-the helper.
+likely to arrive out-of-order. The leader MUST store reports until their
+timestamp is sufficiently far in the past (proportional to the batch window)
```
tgeoghegan, line 780: Given that this is a `MUST`, I'd like us to be more explicit than saying "proportional". I wonder if we should have an explicit grace period parameter, or perhaps a submission deadline, so that clients can decide how long to hold onto inputs before uploading them. I can imagine clients wanting to defer uploads until they are on a WiFi network, or maybe until they are not drawing power from a battery, but if they know that the deadline is just about to elapse then they might opt to do the uploads a little sooner.

I also wonder if we should define an error code and error document so that the leader can explicitly reject a report for being too old, and then tell the client what the oldest report it would accept is. This would allow the client to skip over potentially many queued reports that will be rejected by leader for being too old.

tgeoghegan: I'm open to punting some or all of the above to a later PR in the interest of getting this change merged.

BranLwyd: I like the idea of an explicit task parameter: even though only the leader _needs_ to know this parameter for the protocol to work, as you point out, clients might wish to know the parameter to make decisions about which reports are worthy of submission. With an explicit `grace_window` parameter, clients could assume that any reports more than a `grace_window` into the past is unlikely to be accepted and drop them, without dropping too many reports that would have been accepted.

An explicit error type would work too, and would permit clients to make more precise report-or-drop decisions at the cost of more coordination/specification. I'll punt on specifying that unless someone objects.

BranLwyd:
Based on last week's discussion, I dropped the `grace_window` behavior from a `MUST` to a `SHOULD`.

Please note this now-allowed bad behavior, copied from the commit message:
> Specifically, an implementation that chooses to ignore the `grace_window` SHOULD, and accepts reports clock-skewed a few minutes in the future, could end up dropping non-skewed reports if a skewed report arrives just before an aggregate request is made.

I don't think this caveat is bad enough to be worthy of specifying grace_window with a MUST, I only note it for posterity.

PR #159, opened 2021-10-08
Clarify configuration

ekr:


charter.md
```diff
@@ -19,9 +19,11 @@ will include mechanisms for:
 - Verification of validity proofs by the servers
 - Computation of aggregate values by the servers and reporting of
   results to the entity taking the measurement
-  
-Configuration of clients and servers is out of scope for the working
-group. It is assumed that this happens out of band. 
+
+A successful PRIV system assumes that clients and the various servers
```
cjpatton, line 23: s/PRIV/PPM/? This is a new acronym to me

chris-wood: Some IETF leadership folks objected to the collision with IPPM, so PRIV was a fallback: https://datatracker.ietf.org/group/priv/about/

tgeoghegan (approved):

PR #164, opened 2021-11-08
Ietf112 slides

ekr:


ietf112-slides/ppm-ietf-112.tex
```diff
@@ -0,0 +1,455 @@
...
+    \item ``I am 180cm tall'' when I am actually 175cm
+    \item A problem with any surveying technique
+    \item Solution: live with somewhat noisy data
+    \end{itemize}
+  \item Completely ridiculous
+    \begin{itemize}
+    \item ``I am 1km tall'' (or worse, ``I am -1km tall'')
+    \item Easy to remove with standard systems by filtering
+      \begin{itemize}
+      \item ... but with Prio the data is encrypted
```
tgeoghegan, line 329: "encrypted" isn't the right word here, especially since the secret shares constructed by clients *do* get encrypted, but not to conceal them from aggregators. I can't come up with a better single word to express this, though.

ekr:
I'm going to claim they are encrypted, with a one time pad.

Issue #141, opened 2021-08-25
Recovering after batch mismatch


ekr:
Note from today's call: the CFRG documents need to specify that after the aggregate phase, the helper and leader need to agree to which shares are valid. Depending on the crypto, this may require the leader to tell the helper which ones were valid, if the helper doesn't get the proof result.

cjpatton:
Options discussed so far:
1. Have the helper wait for an ACK from the leader before aggregating its output share
2. If the helper fails to respond to the request for some reason, have the leader "rewind" the helpers state, either by replaying the previous helper state (may impact anti-replay) or with a special "compensation request", which has the helper deduct the last output share form the total (Robert's idea).
3. Run the protocol with multiple sets of helpers so that, if a fault occurs in one run, another run can be used.

cjpatton:
As of DAP-01, there is a checksum in the AggregateShareReq that's used to confirm that the Aggregators agree on the set of aggregated reports. This is sufficient for "error detection", but we might want to do something for "error correction". I'm going to re-purpose this issue and put it in the parking lot.

cjpatton:
At IETF 118 we decided to close this without doing anything. We already have error detection, and batch mismatch should be exceptionally rare. It's also not clear right now how to do error correction.

PR #169, opened 2021-12-07
Revise anti-replay mechanism

cjpatton:
As proposed on the list: https://mailarchive.ietf.org/arch/msg/ppm/WP7PF6y_dr_VY9yHWc8hZOuIZWQ/


draft-gpew-priv-ppm.md
```diff
@@ -959,21 +954,34 @@ aggregator has validated and which fall in the batch interval of the request.
...
+are required to ignore reports with noncee whose timestamp are too far in the
+past or too far in the future.
+
+The procedure for filtering requests is defined as follows.
+
+1. For each task `task_id`, create a set `N[task_id]` of observed nonces.
+1. On input of a report (resp. encrypted input share) with task ID `task_id` and
+   nonce `nonce`:
+   * If `nonce.time < now() - max_report_offset` then ignore it.
+   * Else if `nonce.time > now() + max_report_offset` then ignore it.
```
BranLwyd, line 978: In previous discussion, the "look-back" window we considered was on the order of a few hours; the "look-forward" window was effectively as small as tolerable given likely clock skews between the clients & the leader, on the order of a few minutes.

In this PR, the look-forward & look-back are identical (i.e. `max_report_offset`).

WDYT about making the look-forward and look-back windows into two separate parameters, or otherwise not proscribing that they be identical?

draft-gpew-priv-ppm.md
```diff
@@ -959,21 +954,34 @@ aggregator has validated and which fall in the batch interval of the request.
...
+   * Else if `nonce` in `N[task_id]` then ignore it.
+   * Else add `nonce` to `N[task_id]` and accept.
+
+To prevent the adversary from tampering with the ordering of reports, honest
+clients incorporate the nonce into the AAD for HPKE encryption. Note that this
+strategy may result in dropping reports that happen to have the same nonce
```
tgeoghegan, line 984:
```suggestion
strategy may result in dropping reports that happen to have the same nonce.
```
This is unlikely given that the nonce is a 64 bit integer though, right?

tgeoghegan (commented):
If we're revisiting the helper's storage capabilities to the extent of expecting them to maintain a set of nonces, should we also revisit the notion of the leader storing the helper's encrypted input shares, too?

draft-gpew-priv-ppm.md
```diff
@@ -959,21 +954,34 @@ aggregator has validated and which fall in the batch interval of the request.
...
+large. To ensure they do not need to store these indefinitely, the aggregators
+are required to ignore reports with noncee whose timestamp are too far in the
+past or too far in the future.
+
+The procedure for filtering requests is defined as follows.
+
+1. For each task `task_id`, create a set `N[task_id]` of observed nonces.
+1. On input of a report (resp. encrypted input share) with task ID `task_id` and
+   nonce `nonce`:
+   * If `nonce.time < now() - max_report_offset` then ignore it.
```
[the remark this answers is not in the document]
cjpatton: > This means that skewed clocks (or non-zero network latency) between different aggregators could lead to the aggregators disagreeing about which reports are acceptable and which should be ignored. My understanding is that such a disagreement can be catastrophic to the accuracy of the eventual aggregation.

Clock skew is indeed a problem, but the worst that can happen if a clock is off is that a report gets dropped needlessly. The kind of catastrophic failure you're talking about would happen only if one aggregator decided to aggregate its input share but the other did not. If a report is ignored, then an input share cannot be validated and therefore cannot be aggregated.

> To avoid the need to keep precisely-synchronized clocks between the leader & helper, I propose that the time-based filtering continue to be done by the leader alone. (The nonce-based filtering can/should still be done by both the leader & the helper.)

Both aggregators need to enforce anti-replay, so the helper would have to store the entire nonce set for the task indefinitely. This is the main problem we're trying to avoid with time-based filtering.

draft-gpew-priv-ppm.md
cjpatton: Good catch, I forgot about this! (This is why we have reviews!) I'll break this into two parameters.

draft-gpew-priv-ppm.md
cjpatton: It's not terribly unlikely. Say you have an upload rate of 10,000,000 reports/sec. Then the probability of a nonce collision is roughly 1,000,000^2/2^64, i.e., not negligible.

cjpatton:
> If we're revisiting the helper's storage capabilities to the extent of expecting them to maintain a set of nonces, should we also revisit the notion of the leader storing the helper's encrypted input shares, too?

Yeah, I think we could do that, but let's punt to a future PR.

draft-gpew-priv-ppm.md
```diff
@@ -959,21 +956,34 @@ aggregator has validated and which fall in the batch interval of the request.
...
+To detect replay attacks, each aggregator keeps track of the set of nonces
+pertaining to reports that were previously processed for a given task. If the
+leader receives a report from a client whose nonce is in this set, it simply
+ignores it. Similarly, a helper ignores an encrypted input share whose nonce is
+in this set.
+
+The size of the nonce set is unbounded, and depending the task, can grow quite
+large. To ensure they do not need to store these indefinitely, the aggregators
+are required to ignore reports with nonces whose timestamp are too far in the
+past or too far in the future.
```
cjpatton, line 972: I suppose we could make this a MAY.

cjpatton:
Rebased and squashed.

cjpatton:
Removed the time-sensitive parameters

.targets.mk
```diff
@@ -1,4 +1,9 @@
-TARGETS_DRAFTS := draft-gpew-priv-ppm
-TARGETS_TAGS := 
-draft-gpew-priv-ppm-00.md: draft-gpew-priv-ppm.md
-	sed -e 's/draft-gpew-priv-ppm-latest/draft-gpew-priv-ppm-00/g' $< >$@
+TARGETS_DRAFTS := draft-gpew-priv-ppm
```
martinthomson, line 1: You shouldn't be checking this file in.  Especially as it is listed in .gitignore.

draft-gpew-priv-ppm.md
```diff
@@ -947,26 +938,30 @@ aggregator has validated and which fall in the batch interval of the request.
...
+
+To detect replay attacks, each aggregator keeps track of the set of nonces
+pertaining to reports that were previously processed for a given task. If the
+leader receives a report from a client whose nonce is in this set, it simply
+ignores it. Similarly, a helper ignores an encrypted input share whose nonce is
+in this set.
+
+To prevent the adversary from tampering with the ordering of reports, honest
+clients incorporate the nonce into the AAD for HPKE encryption. Note that this
+strategy may result in dropping reports that happen to have the same nonce.
```
martinthomson, line 953: This note being attached to the AAD statement is confusing.  It's a note about the whole scheme, not the AAD construction.

draft-gpew-priv-ppm.md
```diff
@@ -771,13 +765,10 @@ and constructs its response to the aggregate request.
...
-likely to arrive out-of-order. The leader SHOULD store reports for a time
-period proportional to the batch window before including them in an aggregate
-request. Failure to do so may result in out-of-order reports being dropped by
-the helper. The leader MUST NOT accept reports whose timestamps are in the
-future. Implementors MAY provide for some small leeway, usually no more than a
-few minutes, to account for clock skew.
+The leader is required to buffer reports while waiting to aggregate them. The
+leader MUST NOT accept reports whose timestamps are too far in the future.
+Implementors MAY provide for some small leeway, usually no more than a few
+minutes, to account for clock skew.
```
martinthomson, line 771: The point isn't that it is too far into the future, it is that you either track nonces, or you reject the input.

.targets.mk
cjpatton: Ugh, thanks for flagging! Reverted.

However, after `make clean && make` I see the file in `git status`, despite the fact that it's listed as ignored:
```
$ git status
On branch cjpatton/replay
Your branch is up to date with 'origin/cjpatton/replay'.

Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
	modified:   .targets.mk

no changes added to commit (use "git add" and/or "git commit -a")
```

draft-gpew-priv-ppm.md
cjpatton: Good call. I replaced this with a paragraph that, hopefully, makes it more clear what we're trying to accomplish here.

draft-gpew-priv-ppm.md
cjpatton: Fair enough, but this text isn't new to this PR. What's changed here is the need to re-order reports.

draft-gpew-priv-ppm.md
```diff
@@ -568,11 +569,11 @@ context and `context` is the HPKE context used by the client for encryption.
 The payload is encrypted as
 
 ~~~
-payload = context.Seal(time || nonce || extensions, input_share)
+payload = context.Seal(nonce || extensions, input_share)
```
tgeoghegan, line 572: Do we need to spell out how a `struct Nonce` gets mashed into an HPKE context? i.e. endianness

draft-gpew-priv-ppm.md
```diff
@@ -585,6 +586,13 @@ type 'outdatedConfig'. Clients SHOULD invalidate any cached aggregator
 `HpkeConfig` and retry with a freshly generated Report. If this retried report
 does not succeed, clients MUST abort and discontinue retrying.
 
+The leader MUST ignore any report whose timestamp falls in a batch interval that
+has been agregated at least once. Otherwise, comparing the aggregate result to
```
tgeoghegan, line 590:
```suggestion
has been aggregated at least once. Otherwise, comparing the aggregate result to
```

draft-gpew-priv-ppm.md
```diff
@@ -712,25 +719,22 @@ the first aggregate request. Shall we exclude it in subsequent requests
...
 requests are handled as described in {{errors}}.) It first looks for PPM
-parameters corresponding to `AggregateReq.task_id`. It then filters out
-out-of-order sub-requests by ignoring any sub-request that does not follow the
-previous one (See {{anti-replay}}.)
+parameters corresponding to `AggregateReq.task_id`. It then preprocesses the
+sub-requests as described in {{anti-replay}}. Any sub-requests filtered out by
+this procedure MUST be ignored.
+
+In addition, the helper MUST ignore any report whose timestamp falls in a batch
+interval that has been agregated at least once.
```
tgeoghegan, line 733:
```suggestion
interval that has been aggregated at least once.
```

draft-gpew-priv-ppm.md
tgeoghegan, line 590: What does it mean for a *batch interval* to have been *aggregated*? I don't think that verb is defined for that noun in the protocol right now. As a reader, if I want to see where things become "aggregated", then the `AggregateReq/AggregateResp` exchange seems like the obvious place, but an `AggregateReq` is not defined in terms of a batch interval. `AggregateSubReq` does include parameters that uniquely identify a `Report`, which I think lets us conclude that a `Report` is aggregated when the leader gets an `AggregateSubResp` from the helper for a `Report`. `OutputShareReq` does have a `batch_interval` member, so is the receipt of the corresponding `EncryptedOutputShare` when a batch interval becomes "aggregated"?

I think we should write out a state machine for reports and batch intervals, especially since it seems like they enter the "aggregated" state at different moments. We maybe want to use a different verb than "aggregate" for batch interval to avoid confusion, too. How about "collect", since the `OutputShareReq/EncryptedOutputShare` process should be triggered by a collector issuing `CollectReq` to leader?

Also I think we should not do something like add a `batch_interval` member to `AggregateReq` and declare that receipt of `AggregateResp` is when a batch interval is aggregated, because I think that would prevent leaders and helpers from summing Prio inputs ahead of receipt of all the reports in a batch interval.

draft-gpew-priv-ppm.md
```diff
@@ -947,26 +947,50 @@ aggregator has validated and which fall in the batch interval of the request.
...
-from tampering with the ordering of reports, honest clients incorporate the
-ordering-sensitive parameters `(time, nonce)` into the AAD for HPKE encryption.
-Note that this strategy may result in dropping reports that happen to have the
-same timestamp and nonce value.
+about the client's measurement, violating the privacy goal of PPM. To
+prevent such replay attacks, this specification requires the aggregators to
+detect and filter out replayed reports.
+
+To detect replay attacks, each aggregator keeps track of the set of nonces
+pertaining to reports that were previously processed for a given task. If the
```
tgeoghegan, line 955: Similarly to my point about the verb "aggregate" and the noun "batch interval", I think the notion of a report being "processed" is ambiguous. Given that there are two processing stages for an aggregator, the `AggregateReq` process and then the `OutputShareReq` process, when is a report "processed"? I suspect it's the former. Once again I think spelling out the state machine for reports and batch intervals would be helpful.

draft-gpew-priv-ppm.md
```diff
@@ -947,26 +947,50 @@ aggregator has validated and which fall in the batch interval of the request.
...
 as required by the protocol) by only responding to valid collect requests, as
 described in {{batch-parameter-validation}}.
 
+The size of the nonce set is unbounded, and depending the task, can grow quite
+large. To ensure they do not need to store these indefinitely, each aggregator
+MAY at its own discretion ignore reports with timestamps that are too far in the
+past, too far in the future, or would fall in a batch interval that has already
+been aggregated (see {{batch-parameter-validation}}).
+
+[OPEN ISSUE: Allowing aggregators to ignore reports at-will is likely to lead
```
tgeoghegan, line 976: Could you write up a GH issue for this for visibility?

tgeoghegan (changes requested):

draft-gpew-priv-ppm.md
ekr: I don't think so. The HPKE function takes a blob and the TLS language specifies a bigendian serialization.

draft-gpew-priv-ppm.md
ekr: Why are you saying "ignore" as opposed to "generate error"

draft-gpew-priv-ppm.md
ekr: Again, why ignore and not error?

draft-gpew-priv-ppm.md
cjpatton: TLS-syntax specifies the byte encoding of `struct Nonce`, including endianness. See also my comment here: https://github.com/abetterinternet/ppm-specification/issues/139#issuecomment-1002795321. (If I've understood you correctly, we should be able to close that issue without action.)

draft-gpew-priv-ppm.md
```diff
@@ -947,26 +947,50 @@ aggregator has validated and which fall in the batch interval of the request.
...
+generated by the client with a nonce its peer has not yet seen. To prevent this,
+clients incorporate the nonce into the AAD for HPKE encryption, ensuring that
+the input share is only recovered if the aggregator is given the correct nonce.
+(See {{upload-request}}.)
 
 Aggregators prevent the same report from being used in multiple batches (except
 as required by the protocol) by only responding to valid collect requests, as
 described in {{batch-parameter-validation}}.
 
+The size of the nonce set is unbounded, and depending the task, can grow quite
```
ekr, line 970: It's not unbounded. It's of size 2^{96}.

draft-gpew-priv-ppm.md
```diff
@@ -947,26 +947,50 @@ aggregator has validated and which fall in the batch interval of the request.
...
+large. To ensure they do not need to store these indefinitely, each aggregator
+MAY at its own discretion ignore reports with timestamps that are too far in the
+past, too far in the future, or would fall in a batch interval that has already
+been aggregated (see {{batch-parameter-validation}}).
```
ekr, line 974: It's not at your discretion whether you ignore those which have been aggregated.

draft-gpew-priv-ppm.md
```diff
@@ -947,26 +947,50 @@ aggregator has validated and which fall in the batch interval of the request.
...
+the aggregators to make the same decision about the time intervals for which
+nonces are tracked and for which nonces are ignored. There are two ways to this:
+
+1. Pick a duration `d_min` for reports too far in the future and a duration
+   `d_max` for reports too far in the past, and ignore a report if `now() - time
+   > d_dmin` or `time - now() > d_max`, where `time` is the report timestamp.
+   This requires the aggregators to have a roughly synchronized clock.
+
+2. @martinthomson made the following suggestion, which doesn't require clock
+   sync. The leader and helper could periodically negotiate the time `t_min` in
```
ekr, line 987: Why is it useful not to require clock sync? The leader and the helper are servers, and they're quite capable of having accurate clocks.

ekr (changes requested):

draft-gpew-priv-ppm.md
cjpatton: Sure thing, however I want to wait until folks are happy with this change. After all, the issue is triggered by this PR and only makes sense if it's merged.

draft-gpew-priv-ppm.md
martinthomson: Good sync != perfect sync.  Aggregation will fail if anti-replay is independently enforced by helpers (which is something I argue we want) and any amount of relative skew could produce disagreements.

draft-gpew-priv-ppm.md
```diff
@@ -947,26 +947,50 @@ aggregator has validated and which fall in the batch interval of the request.
...
+   the past before which reports are deemed too old and the time `t_max` in the
+   future after which reports are deemed too new. This is fairly straight
+   forward to piggy back on aggregate requests, but would get a bit trickier in
+   the case of multiple helper.]
```
martinthomson, line 993: I don't see how this is difficult.  The leader starts a run and tells helpers what t_min and t_max are as part of that.

draft-gpew-priv-ppm.md
martinthomson: Definitely.  This is an important point on which all helpers (and the leader) need to agree.  Discretion can't come into it or aggregation will fail.

draft-gpew-priv-ppm.md
```diff
@@ -585,6 +586,13 @@ type 'outdatedConfig'. Clients SHOULD invalidate any cached aggregator
 `HpkeConfig` and retry with a freshly generated Report. If this retried report
 does not succeed, clients MUST abort and discontinue retrying.
 
+The leader MUST ignore any report whose timestamp falls in a batch interval that
+has been aggregated at least once. Otherwise, comparing the aggregate result to
+the previous aggregate result may result in a significant privacy violation.
+
+[NOTE: Differential privacy provides a generic protection against these kinds of
+"batch faults".]
```
martinthomson, line 595:
```suggestion
```

This is not a helpful statement.  It could be misinterpreted as saying that differential privacy is enough.  But anti-replay is critical to ensuring that differential privacy protections are adequate.  This says the exact opposite.

draft-gpew-priv-ppm.md
```diff
@@ -712,25 +719,22 @@ the first aggregate request. Shall we exclude it in subsequent requests
...
 requests are handled as described in {{errors}}.) It first looks for PPM
-parameters corresponding to `AggregateReq.task_id`. It then filters out
-out-of-order sub-requests by ignoring any sub-request that does not follow the
-previous one (See {{anti-replay}}.)
+parameters corresponding to `AggregateReq.task_id`. It then preprocesses the
+sub-requests as described in {{anti-replay}}. Any sub-requests filtered out by
+this procedure MUST be ignored.
+
+In addition, the helper MUST ignore any report whose timestamp falls in a batch
+interval that has been aggregated at least once.
```
martinthomson: You should include all the anti-replay text in the one place.  I read the previous piece and guessed that only the leader looks at timestamps.  But that isn't it.

Note also s/timestamp/nonce or "nonce contains a timestamp that"

draft-gpew-priv-ppm.md
cjpatton: @tgeoghegan good call. I've changed "aggregated" to be more precise. (Similarly below.)

draft-gpew-priv-ppm.md
cjpatton: I made this more precise along the same lines as above.

draft-gpew-priv-ppm.md
cjpatton: Actually it's 2^128, but point well taken! Fixed to not imply unboundedness.

draft-gpew-priv-ppm.md
cjpatton: Nice catch.

draft-gpew-priv-ppm.md
cjpatton: > The leader and the helper are servers, and they're quite capable of having accurate clocks.

I tend to agree. If we align `d_max` and `d_min` with `min_batch_duration`, then all they have to do is ensure their clocks are within `min_batch_duration` seconds of one another. However I think @martinthomson's broader point is this requirement isn't strictly necessary. The true requirement is that they agree on what set of nonces they have to track at any given time.

This was discussed extensively here: https://mailarchive.ietf.org/arch/msg/ppm/oDCOJ-cUR8NQFG6At3byH1fdcd4/

draft-gpew-priv-ppm.md
cjpatton: @ekr good question.

One reason to not *require* an error here is that, if we did, the leader would have to check if the batch has been collected before completing the HTTP request. The current text allows it to make this determination later on.

One reason to *not permit* an error at all is that it volunteers information to the client that the client doesn't really need, namely, that the collector has requested the aggregate result for the interval. This doesn't seem all that damaging right now, but who knows.

ekr: And one reason to provide an error is so clients can debug themselves.

draft-gpew-priv-ppm.md
cjpatton: @martinthomson AFAICT a disagreement only results needlessly dropping reports, I don't see how it can result in a replay (i.e., privacy violation). I'm probably missing something though.

ekr: I'm not sure what you mean by "perfect sync". NTP is capable of providing 10ms error. Given that these timestamps are in seconds, this seems fine. I don't see a real problem here.

draft-gpew-priv-ppm.md
cjpatton: Great point, I will just axe this.

The broader point is that we've uncovered a number of potential "batch faults", all of which fall more or less into the umbrella of privacy leaks that DP aims to mitigate. To be clear, I don't think we should *require* DP to be used in the core PPM protocol, but I thought it might be worth noting the connection for later on when we really think through how to incorporate DP.

draft-gpew-priv-ppm.md
```diff
@@ -712,25 +720,23 @@ the first aggregate request. Shall we exclude it in subsequent requests
...
 requests are handled as described in {{errors}}.) It first looks for PPM
-parameters corresponding to `AggregateReq.task_id`. It then filters out
-out-of-order sub-requests by ignoring any sub-request that does not follow the
-previous one (See {{anti-replay}}.)
+parameters corresponding to `AggregateReq.task_id`. It then preprocesses the
+sub-requests as described in {{anti-replay}}. Any sub-requests filtered out by
+this procedure MUST be ignored.
+
+In addition, the helper MUST ignore any report whose timestamp falls in a batch
+interval for which it has received at least one output-shrae request from the
```
tgeoghegan, line 731:
```suggestion
interval for which it has received at least one output-share request from the
```

draft-gpew-priv-ppm.md
tgeoghegan: Is there perhaps a change here you forgot to push here? This text is unchanged since I reviewed. I think at least we should change "processed" to "aggregated".

tgeoghegan (changes requested):

draft-gpew-priv-ppm.md
cjpatton: Shouldn't be too bad, but you do have to negotiate `t_min, t_max` that works for all of the helpers. You have to make sure that whatever `t_min` you pick is newer than than the oldest value that is tolerated by any aggregator. (Similarly for `t_max`, you'd have to pick it so that it's older than the newest value tolerated by any aggregator.)

draft-gpew-priv-ppm.md
cjpatton: The bug in this case is clockskew, right?

ekr: Or any number of other things that cause the clock to be wrong.

draft-gpew-priv-ppm.md
cjpatton: Woops, totally missed this. I thought this was about something else. s/processed/aggregated/g is the right change here.

draft-gpew-priv-ppm.md
cjpatton: > You should include all the anti-replay text in the one place. I read the previous piece and guessed that only the leader looks at timestamps. But that isn't it.

Agreed. I tried but couldn't figure out how to do so without a more major refactor. One idea I had was to change the {{anti-replay}} section to speak more generally about "batch faults", i.e., instances where a report was included or excluded from a batch in a way that leaked more information about the measurement than I intended.

What I did instead was note above that the helper enforces the same thing.

Note that this text gets de-duplicated in https://github.com/abetterinternet/ppm-specification/pull/174.

> Note also s/timestamp/nonce or "nonce contains a timestamp that"

Done.

draft-gpew-priv-ppm.md
cjpatton: In any case, the benefit seems marginal to me. But I don't feel that strongly about it. I'd go for permitting an error here if you'd prefer that.

draft-gpew-priv-ppm.md
```diff
@@ -519,22 +524,18 @@ struct {
...
 * `task_id` is the task ID of the task for which the report is intended.
-* `time` is the time at which the report was generated. This field is used by
-  the aggregators to ensure the report appears in at most one batch. (See
-  {{anti-replay}}.)
-* `nonce` is a random number chosen by the client generating the report. This
-  and the timestamp field are used by the aggregators to ensure that each report
-  appears at most once in a batch. (See {{anti-replay}}.)
+* `nonce` is a nonce generated by the client and used by the aggregators to
+  uniquely identify the report in the context of a PPM task. The nonce includes
+  the time at which the report was generated and a random 64-bit integer. This
```
tgeoghegan, line 529: nit: since the `Nonce` type appears in a few places, I think the discussion of its `time` and `rand` members should be moved to the definition of `struct Nonce`, around line 417.

draft-gpew-priv-ppm.md
```diff
@@ -585,6 +586,12 @@ type 'outdatedConfig'. Clients SHOULD invalidate any cached aggregator
 `HpkeConfig` and retry with a freshly generated Report. If this retried report
 does not succeed, clients MUST abort and discontinue retrying.
 
+The leader MUST ignore any report whose nonce contains a timestamp that falls in
+a batch interval for which it has received at least one collect request from the
+collector. (See {{pa-collect}}.) Otherwise, comparing the aggregate result to
+the previous aggregate result may result in a significant privacy violation.
```
tgeoghegan, line 592: nit: this adjective suggests that we have some kind of classification of privacy violations.
```suggestion
the previous aggregate result may result in a privacy violation.
```

draft-gpew-priv-ppm.md
```diff
@@ -712,25 +718,23 @@ the first aggregate request. Shall we exclude it in subsequent requests
...
-sub-requests in ascending order, where the ordering on sub-requests is
-determined by the algorithm defined in {{anti-replay}}. Specifically, the leader
-constructs its request such that:
-
-* each sub-request follows the previous sub-request; and
-* the first sub-request follows the last sub-request in the previous aggregate
-  request.
+In order to provide replay protection, the leader preprocesses the set of
+reports carried by the AggregateReq as described in {{anti-replay}}. Any reports
+filtered out by this procedure MUST be ignored.
```
tgeoghegan, line 723: I find this confusing because the leader should be applying the anti-replay logic at the time of receiving uploads. By the time it gets to constructing an `AggregateReq`, it should already have discarded any reports with duplicate nonces by this point. It's also confusing to state that the leader should "preprocess the set of reports carried by the AggregateReq" because the leader constructs the `AggregateReq`.

tgeoghegan (approved):
A few editorial nits but this LGTM.

draft-gpew-priv-ppm.md
cjpatton: Addressed the latter request. For the former, an implementation of the leader might opt to wait to filter reports until it's ready to begin the aggregation flow. This makes sense if it wants to limit the amount of work it has to do for each upload request, where latency matters a lot more.

cjpatton:
There seems to be consensus for the main change introduced by this PR, which is to not require the Leader to sort reports. The only outstanding question is how to mitigate data loss that results from aggregators ignoring reports "at will". This is definitely a solvable problem, but there's not yet consensus on how to solve it. I'd like to suggest that we merge this PR and open an issue to hash out this problem.

@chris-wood, @ekr: Please review and approve.

draft-gpew-priv-ppm.md
```diff
@@ -713,25 +719,23 @@ the first aggregate request. Shall we exclude it in subsequent requests
...
-parameters corresponding to `AggregateReq.task_id`. It then filters out
-out-of-order sub-requests by ignoring any sub-request that does not follow the
-previous one (See {{anti-replay}}.)
+parameters corresponding to `AggregateReq.task_id`. It then preprocesses the
+sub-requests as described in {{anti-replay}}. Any sub-requests filtered out by
+this procedure MUST be ignored.
+
+In addition, the helper MUST ignore any report whose nonce contains a timestamp
+that falls in a batch interval for which it has received at least one
+output-share request from the leader. (See {{output-share-request}}.)
```
chris-wood, line 734:
```suggestion
output-share request from the leader. (See {{output-share-request}}.)
This means leaders cannot interleave a sequence of aggregate and 
output-share requests for a single batch.
```

We should probably add guidance for the leader here as to _when_ they should issue their output-share request. If they do it as soon as the minimum batch count is hit, then nothing else in the batch can be aggregated. (We can do this in a followup PR, I think.)

draft-gpew-priv-ppm.md
```diff
@@ -772,13 +775,10 @@ and constructs its response to the aggregate request.
...
 
-The leader is required to issue aggregate requests in order, but reports are
-likely to arrive out-of-order. The leader SHOULD store reports for a time
-period proportional to the batch window before including them in an aggregate
-request. Failure to do so may result in out-of-order reports being dropped by
-the helper. The leader MUST NOT accept reports whose timestamps are in the
-future. Implementors MAY provide for some small leeway, usually no more than a
-few minutes, to account for clock skew.
+The leader is required to buffer reports while waiting to aggregate them. The
+leader MUST NOT accept reports whose timestamps are too far in the future.
```
chris-wood, line 779:
```suggestion
leader SHOULD NOT accept reports whose timestamps are too far in the future.
```

I don't see how we can enforce a MUST without a specific definition of "too far in the future."

draft-gpew-priv-ppm.md
chris-wood: I agree with @martinthomson here -- it's best if we can simply avoid skew altogether, and that seems doable with the agreed upon min/max limits.

chris-wood (approved):
_Really sorry_ for the delay here. This LGTM. I think we need to address #141 next.

draft-gpew-priv-ppm.md
```diff
@@ -586,6 +587,12 @@ type 'outdatedConfig'. Clients SHOULD invalidate any cached aggregator
 `HpkeConfig` and retry with a freshly generated Report. If this retried report
 does not succeed, clients MUST abort and discontinue retrying.
 
+The leader MUST ignore any report whose nonce contains a timestamp that falls in
+a batch interval for which it has received at least one collect request from the
+collector. (See {{pa-collect}}.) Otherwise, comparing the aggregate result to
+the previous aggregate result may result in a privacy violation.
+(Note that the helpers enforce this as well; see {{aggregate-request}}.)
```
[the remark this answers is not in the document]
chris-wood: (And we would need to add this new error, if this suggestion lands.)

draft-gpew-priv-ppm.md
```diff
@@ -586,6 +587,12 @@ type 'outdatedConfig'. Clients SHOULD invalidate any cached aggregator
 `HpkeConfig` and retry with a freshly generated Report. If this retried report
 does not succeed, clients MUST abort and discontinue retrying.
 
+The leader MUST ignore any report whose nonce contains a timestamp that falls in
+a batch interval for which it has received at least one collect request from the
+collector. (See {{pa-collect}}.) Otherwise, comparing the aggregate result to
```
ekr, line 592: This seems like a recipe for silent failure in case of clock skew.

draft-gpew-priv-ppm.md
```diff
@@ -713,25 +719,23 @@ the first aggregate request. Shall we exclude it in subsequent requests
...
 requests are handled as described in {{errors}}.) It first looks for PPM
-parameters corresponding to `AggregateReq.task_id`. It then filters out
-out-of-order sub-requests by ignoring any sub-request that does not follow the
-previous one (See {{anti-replay}}.)
+parameters corresponding to `AggregateReq.task_id`. It then preprocesses the
+sub-requests as described in {{anti-replay}}. Any sub-requests filtered out by
+this procedure MUST be ignored.
+
+In addition, the helper MUST ignore any report whose nonce contains a timestamp
+that falls in a batch interval for which it has received at least one
```
ekr, line 733: This also is a recipe for silent error.

draft-gpew-priv-ppm.md
ekr: +1

draft-gpew-priv-ppm.md
ekr: This seems much more complicated than simply having sync.

draft-gpew-priv-ppm.md
ekr: This just seems unspecified.

draft-gpew-priv-ppm.md
cjpatton: I'm not convinced is as easy as that. Would you be OK with merging without a change and moving the discussion to an issue? (As I proposed above.)

draft-gpew-priv-ppm.md
cjpatton: I agree, but the spec currently says MUST here, so this isn't something that's changed in this PR. Can we resolve this by leaving an OPEN ISSUE here?

draft-gpew-priv-ppm.md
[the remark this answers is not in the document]
cjpatton: This is a duplicate of @ekr's point here: https://github.com/abetterinternet/ppm-specification/pull/169#discussion_r776503058.

draft-gpew-priv-ppm.md
chris-wood: Why not just make the change here? It's obviously related to this PR.

draft-gpew-priv-ppm.md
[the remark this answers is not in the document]
chris-wood: I agree, and I think we need to be clear with the protocol interaction here. I am not supportive of "ignore behavior is left unspecified."

draft-gpew-priv-ppm.md
cjpatton: People often ask to keep spec PRs minimal, but I don't think anyone will object so I'll just change this. Done.

draft-gpew-priv-ppm.md
chris-wood: I see no reason to just let this behavior go unspecified. What if some leaders reply with 200 OK and some reply with an error in this case? At the end of the day, the report upload was unsuccessful, and the signal sent back to the client should not be 200 OK.

draft-gpew-priv-ppm.md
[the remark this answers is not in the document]
cjpatton: As noted there I don't feel strongly about this. There are two in favor, so I'll make the change. I'm resolving this and will bump the other thread.

draft-gpew-priv-ppm.md
chris-wood: I don't really see how it would be complicated. This is the same as aggregators agreeing on batch interval boundaries, no?

draft-gpew-priv-ppm.md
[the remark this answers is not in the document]
cjpatton: I'm going to close this thread and quote you in the other thread. That way we don't have two threads on the same topic.

ekr (changes requested):
OK. I've gone through this again and I'm no more convinced than I was the last time.

I think the basic structure of the nonce as time + random ID is fine, but the actual description of how endpoints are supposed to behave is not.

1. As a practical matter, if clocks aren't roughly synced we are going to have some serious problems because 
OK. I've gone through this again and I'm no more convinced than I was the last time.

I think the basic structure of the nonce as time + random ID is fine,
but the actual description of how endpoints are supposed to behave is
not.

First, as a practical matter, if clocks aren't roughly synced, then we
are going to have some serious problems about out of window
rejection. So, I don't think trying to avoid clock sync is
helpful. Second, I think we can simply not worry about a small rate of
lossage from data too far in the future. IOW,

I think this leads to a simpler design in which only the leader needs
to keep buffered state outside of its window.

1. Aggregators are expected to have accurate, synced clocks.

2. You can't aggregate an event of time T till T + epsilon. This is
   enforced by the helper, which rejects if it sees it. The leader
   needs to wait till T + epsilon + skew where skew is the allowed
   clock skew.

3. You can't aggregate the block T1 - T2 after T3 - T4 where T3 > T2.

This means that the leader buffers data between the end of the last
aggregation block and now() + Buffer where Buffer is configurable, but
on the order of the minimum aggregation block window. The helper
doesn't know Buffer because it will never see anything that's out of
window. The helper just needs to know when the last aggregation block
ended, the data it is currently working on, and the current time and
reject (with an error) anything out of that window or that is
duplicated.

This is a more straightforward design and also detect errors, whereas
the current design just fails silently.


If we want to allow data to be sent directly to the helper. In that
case, each aggregator needs to buffer data between the end of the
last aggregation block and now + Buffer(). This means that there
is some possibility of disagreement on both edges (because of
race conditions and clock sync). This means that we need a way
for each side to tell the other that there was an error, and we
just need a "rejected because too early" or "rejected because too late"
error, just like for any other error. However, with this much buffer,
this should occur infrequently. Note that this easily handles the
case where the helper rejects but if the leader does, it may never
learn there is something the helper accepted.

draft-gpew-priv-ppm.md
cjpatton: @chris-wood supports the idea of having the leader issue an error here. Quote: 
> I agree, and I think we need to be clear with the protocol interaction here. I am not supportive of "ignore behavior is left unspecified."

As I mentioned above, it may not always be feasible for the leader to check for the error condition on-the-fly. It requires less internal-state coordination to just ingest the report and prune it later. I am fine with a MAY or even a SHOULD here but I don't think it should be a MUST. I'm fine being in the rough, just let me know if you think this absolutely needs to be a MUST.

chris-wood: Sure, I'm operating under the assumption that the leader can detect this immediately upon upload. If the leader defers the check to some point later on, then obviously it can't reply with an error on the fly. A SHOULD seems like it'd work here.

draft-gpew-priv-ppm.md
cjpatton: Added text saying the leader SHOULD reply with an error. As I noted in a previous thread, it may not be feasible for the leader to decide if the batch interval has been collected. It's much simpler if all the server has to do is ingest the report at this step and wait to prune it when it's ready to start aggregating.

chris-wood:
> First, as a practical matter, if clocks aren't roughly synced, then we are going to have some serious problems about out of window rejection.

I'm not sure I agree with this. My understanding of the design principle here -- which I support -- is that aggregators just agree on a list of timestamps, just like they agree on other parameters for configuration of a measurement task. That means the only real requirement for clock sync then falls on the leader, as they control what is effectively "the current batch." The leader controls what things are sent to the aggregators, so they can track the current batch based on what the leader asks them to aggregate and then output. (I _think_ they could do this without any notion of a clock whatsoever, but I may be missing something.)

ekr:
@chris-wood I think you may be correct, at least in the "everything gets sent to the leader" model. The helper can just accept whatever the leader says as the "right edge" of the window. And if the leader is dumb enough to do something way in the future, then it can't aggregate for a long time.

chris-wood:
Yeah, this seems like it gets more complicated if we move to the "send directly to aggregators" model. 🤷

draft-gpew-priv-ppm.md
cjpatton: Yeah I think you're right that failing silently here is a bad idea. There are other ways in which the helper might fail for a given report, and all of these should be handled by relaying an error back to the leader. I have an idea for how to address this in a way that handles all such errors in a uniform manner (see #179 if you're curious), but it would be hard to specify very concretely given the state the document is in. I've updated the text to make sure the helper sends an error here. It's a bit awkward, but it'll do the job for now.

cjpatton:
For my part, I actually think @ekr's suggestion is quite nice. However not everyone agrees that we should require the aggregators to be roughly time-synced, and we'll need to discuss it further. I'm trying to avoid scope creep with the current PR. @ekr I've addressed your inline comments ...  please let me know if you're happy merging this without your suggestion and we can discuss it in a new issue.

draft-gpew-priv-ppm.md
cjpatton: Totally agree here. I've added some text saying that the helper MUST reply with an error in this case rather than its next VDAF message. The text is a bit awkward, but I think it's hard to do much better in the current state of the document. In fact, there are a number of errors that might happen between now and when the helper is ready to reply. In #179 I've tried to solve this problem in a more generic way.

ekr:
I'm not sure why it makes sense to land a PR changing things when we
don't have consensus.

I would suggest instead that we just land the protocol piece that
includes the nonce and say that the precise replay suppression
algorithm is out of scope. As a practical matter, this will just mean
that interop is a little messy because the helper may silently
reject, but that can happen anyway for other reasons, and none
of this will be much of a problem if clocks are synced, which we
can just do for now.

cjpatton:
I don't think we disagree on the replay-suppression algorithm. In fact, the only piece we haven't converged on is how to avoid storing nonce sets indefinitely while minimizing data loss.

We already require that the nonces be stored as long as they might be used (i.e., for as long as a batch isn't collected) It seems like all we need to do is remove the following text:

> Depending on the rate of upload requests, the size of the nonce set can grow to     
> be quite large. To ensure they do not need to store these indefinitely, each     
> aggregator MAY at its own discretion ignore reports with timestamps that are too     
> far in the past or too far in the future (see {{batch-parameter-validation}}).     

and replace it with an open issue where we discuss this problem.

ekr:
> I don't think we disagree on the replay-suppression algorithm. In fact, the only piece we haven't converged on is how to avoid storing nonce sets indefinitely while minimizing data loss.

I was including that as part of the algorithm, but OK.


> We already require that the nonces be stored as long as they might be used (i.e., for as long as a batch isn't collected) It seems like all we need to do is remove the following text:
> 
> > Depending on the rate of upload requests, the size of the nonce set can grow to
> > be quite large. To ensure they do not need to store these indefinitely, each
> > aggregator MAY at its own discretion ignore reports with timestamps that are too
> > far in the past or too far in the future (see {{batch-parameter-validation}}).
> 
> and replace it with an open issue where we discuss this problem.

Seems like you should also remove the rest of the OPEN ISSUE text following.

cjpatton:
Done, and squashed the commits.

cjpatton:
Here's the issue: https://github.com/abetterinternet/ppm-specification/issues/180
I'll merge this once @ekr approves.

ekr (approved):
LGTM but can you put in an open issue marker for #180. Just like one line should be fine.

cjpatton:
Done.

Issue #45, opened 2021-05-10
Mode to make proof optional 


siyengar:
Some applications might not need as much protection against malicious users and might be okay with a malicious user changing the values by a larger amount.  They might be using Prio purely for privacy and might have other ways to guard against malicious inputs, for example by verify the user account or by attesting the application. 
It might be useful for the Validity proof to be optional depending on the application. This would reduce overheads perhaps of  communication between servers to validate the proofs and also in communication of the proofs from the client -> server.  

These applications could still benefit from:
1. the multi-server deployment model
2. AFE encoding specifications

cjpatton:
This seems reasonable to me. Right now we have several pending design decisions floating around, so I'd like to suggest we wait on text until more of the protocol is settled.

cjpatton:
Something to note here: if we have have a bit in the PA parameters that disables verification, we'll need to make sure this bit is explicitly authenticated.

cjpatton:
Sounds good to me. Here is the issue for reference: https://github.com/cjpatton/vdaf/issues/20

PR #174, opened 2021-12-29
Have clients upload report shares to helpers

### LINKED ISSUE #130: Cost implications of single leader model
Assuming `Aggregators` and `Servers` are running on distinct cloud providers / hosting environments there is an additional egress cost incurred using the `Leader` model where one `Aggregator` coordinates input validation and data collection from all clients and then forwards data shares  / batches to respective `Aggregators` that are part of the setup.

 As of 2021-08-18 there are the following network costs of major cloud providers. *No claim of 100% accuracy*

TL;DR Assuming `Aggregators` / `Servers` running on distinct providers using the Leader model we can assume $50 - $230 / TB added cost for egress network traffic.

//TODO add how many input shares per TB approx.

# Network Cost Comparison across providers

## Ingress Network Cost
*>>cost incurred by clients sending data to `Aggregator(s)`<<*
*all prices per GB*

AWS | Azure | GCP 
----|-------|------
 free | free | free 

## Egress Network Cost
*>>cost incurred by sending data out leaving the provider (e.g. server communication, leader aggregator sending data to other aggregators<<*
*all prices per GB*

AWS | Azure | GCP
-----|-------|------
$0.05-$0.154<sup>1</sup> |   $0.0875 - $0.181<sup>2</sup> |  $0.045 - $0.23<sup>3</sup>

<sup>1</sup> Tiered discount 1GB free, 1GB-10TB, 10TB-50TB, 50TB-150TB, 150TB+ 
<sup>2</sup> Tiered discount 5GB free, 5GB-10TB, 10TB-50TB, 50TB-150TB, 150TB-500TB, 500TB+
<sup>3</sup> Tiered discount 0-1TB, 1-10TB, 10+TB Premium Tier starts at $0.08, Standard Tier goes down to $0.045 Tiered 0-10TB, 10-150TB, 150-500TB

## Intra-Provider Network Cost
*>>cost incurred by sending data within a provider between different regions<<*

### Intra-Continental
*all prices per GB*
Region | AWS | Azure | GCP
---- | ---- | ---- | ----
NA | $0.01-$0.02  |  $0.02 | $0.01
Europe | $0.02 |  $0.02 | $0.02
Asia | $0.09 | $0.08 | $0.05
Oceania | $0.098 | $0.08 | $0.08
MEAA | $0.1105-$0.147 | $0.08 | $0.08
South America | $0.138 | $0.16 | $0.08

### Inter-Continental
*all prices per GB*
Region | AWS | Azure | GCP
---- | ---- | ---- | ----
NA -> * | $0.01-$0.02 |  $0.05 | $0.08-$0.15<sup>1</sup>
Europe -> * | $0.02 |  $0.05 | $0.08-$0.15<sup>1</sup>
Asia, Oceania, Africa -> * | $0.086-$0.098 | $0.08 | $0.08-$0.15<sup>1</sup>
South America -> * | $0.138 | $0.16 | $0.08-$0.15<sup>1</sup>

<sup>1</sup> from/to Indonesia or Oceania2 $0.15 for GCP

cjpatton:
Closes #130 (cc/ @hostirosti).
Addresses a bug pointed out in https://github.com/abetterinternet/ppm-specification/issues/183#issuecomment-1016975853 (cc/ @ekr).

Taking #169 effectively increases the storage requirements for the helper to `O(n)`, where `n` is the number of reports. This PR would increase storage requirements for the helper by a constant factor, but would reduce the leader <-> helper communication overhead by the same factor.

2022-05-06: Moved back to draft. There is not yet consensus on this change.
2022-01-24: Moved from draft to review. The consensus reached on [this thread](https://mailarchive.ietf.org/arch/msg/ppm/sCJ1oKR0KMHlZ65FCPUbw3FrGxY/) seems to be to take this change.
2021-12-29: This PR is currently marked as "Draft" because I don't yet intend for it to be merged. Rather the goal of this PR is to make the discussion about this trade-off more concrete.


draft-gpew-priv-ppm.md
```diff
@@ -586,33 +600,31 @@ payload = context.Seal(nonce || extensions, input_share)
...
-the previous aggregate result may result in a privacy violation. (Note that the
-helpers enforce this as well; see {{aggregate-request}}.) In addition, the
-leader SHOULD abort the upload protocol and alert the client with error
+To prevent reports from being misused, the aggreagtors filter out replayed or
+otherwise stale reports during the aggregation protocol using the procedure
+described in {{anti-replay}}. In addition, the aggregator that receives a stale
+report SHOULD abort the upload protocol and alert the client with error
 "staleReport".
 
 ### Upload Extensions {#upload-extensions}
```
tgeoghegan, line 620: The reason we had upload extensions in the first place was so that they could be tunnelled through the leader into the helper. Now that we do direct uploads, can we remove these? The obvious remaining use case is tunnelling extensions through a client proxy/intermediary/batching client, but if deployments use OHTTP as we recommend, then PPM shouldn't need to specify anything as things like HTTP request headers will get securely relayed.

draft-gpew-priv-ppm.md
```diff
@@ -692,8 +704,9 @@ continuation of a previous sub-batch?]
 
 ### Aggregate Request
 
-The AggregateReq request is used by the leader to send a set of reports to the
-helper. These reports MUST all be associated with the same PPM task and batch.
+The AggregateReq request is used by the leader to coordinate aggregation of a
+set of reports with the helpers. These reports incident to a particular
+AggregateReq MUST all be associated with the same PPM task and batch.
```
tgeoghegan, line 709: Now that the helper is responsible for its own nonce, report and aggregate accumulator storage, do we need aggregate requests to not span batch intervals anymore? I'm not sure but I think the reason we had this constraint on aggregate requests was so that the state blob could be "sharded" by batch interval and allow parallelism in the helper. Now, though, we might expect the helper to have its own storage where it keeps track of aggregate accumulators for several different batch intervals.

draft-gpew-priv-ppm.md
```diff
@@ -704,49 +717,39 @@ struct {
...
-
-In addition, for any report whose nonce contains a timestamp that falls in a
-batch interval for which it has completed at least one aggregate-share request
-(see {{aggregate-share-request}}), the helper MUST send an error messsage in
-response rather than its next VDAF message. Note that this means leaders cannot
-interleave a sequence of aggregate and aggregate-share requests for a single
-batch.
+reports incident to the AggregateReq as described in {{anti-replay}}. For any
+report filtered out by this procedure, the helper MUST respond with an error
+message rather than its next VDAF message.
```
tgeoghegan, line 752:
```suggestion
sub-response rather than its next VDAF message.
```
To make it clear that filtering out an individual report does not fail the overall aggregate transaction.

draft-gpew-priv-ppm.md
```diff
@@ -805,10 +802,10 @@ carries is up to the helper implementation.
 
 ### Aggregate Share Request {#aggregate-share-request}
 
-Once the aggregators have verified at least as many reports as required for the
-PPM task, the leader issues an "aggregate-share request" to each helper. The
-helper responds to this request by extracting its aggregate share from its state
-and encrypting it under the collector's HPKE public key.
+Once it has verified at least as many reports as required for the PPM task, the
+leader may issue an "aggregate-share request" to each helper. The helper
+responds to this request by extracting its aggregate share from its state and
```
tgeoghegan, line 807: nit: "extracting its aggregate share from its state" implies the helper will have an aggregate share ready to go in its state, which won't be true for `poplar1`. We could just say "The helper responds to this request with its aggregate share, encrypted under the collector's HPKE public key."

draft-gpew-priv-ppm.md
```diff
@@ -862,21 +859,17 @@ body consisting of the following structure:
...
-  aggregate share.
-* `payload` is an encrypted `AggregateShare`.
-
-The leader uses the helper's aggregate share response to respond to the
-collector's collect request (see {{pa-collect}}).
+The leader uses the helper's AggregateShareResp to respond to the collector's
+collect request (see {{pa-collect}}).
 
+After issuing an aggregate-share request for a given batch interval, it is an
+error for the leader to issue any more aggreagte requests for reports in the
```
tgeoghegan, line 870:
```suggestion
error for the leader to issue any more aggregate requests for reports in the
```

draft-gpew-priv-ppm.md
```diff
@@ -862,21 +859,17 @@ body consisting of the following structure:
...
+batch interval. These reports will be rejected by helpers as described in
+{{aggregate-request}}.
```
tgeoghegan, line 872: It's an error for the leader to issue more than `max_batch_lifetime` requests for a given batch interval, right? As written this states the leader can issue exactly one request.

draft-gpew-priv-ppm.md
```diff
@@ -969,11 +960,27 @@ prevent such replay attacks, this specification requires the aggregators to
...
+the leader and has completed a collect request for a batch containing the report
+(see {{pa-collect}}) or if the aggregator is a helper and has completed an
+aggregate-share request for a batch containing the report (see
+{{aggregate-share-request}}). It also checks if the report has been
+"aggregated". This is the case if a report with the same nonce was previously
+used in an aggregate request.
+
+* If the report has not yet been aggregated but is contained by a batch that has
+  been collected, then the aggregator MUST ignore it. This prevents additional
+  reports from being aggregated after its batch has already been collected.
```
tgeoghegan, line 975:
```suggestion
  been collected, then the aggregator MUST ignore it. This prevents additional
  reports from being aggregated after their batch has already been collected.
```

draft-gpew-priv-ppm.md
```diff
@@ -969,11 +960,27 @@ prevent such replay attacks, this specification requires the aggregators to
...
+
+* If the report has been aggregated but has not been collected, then the
+  aggregator MUST ignore it. This prevents a report from being used more than
+  once in a batch.
```
tgeoghegan, line 979: I think but am not certain that this needs to reference the task's `max_batch_lifetime` rather than allowing exactly once, because successive collector queries will change the aggregation parameter (e.g., different prefixes in `poplar1`).

draft-gpew-priv-ppm.md
```diff
@@ -304,17 +327,16 @@ aggregators. Each task is identified by a unique 32-byte ID which is used to
...
-The leader distributes the shares to the helpers and orchestrates the process of
-verifying them (see {{validating-inputs}}) and assembling them into a final
-measurement for the collector. Depending on the VDAF, it may be possible to
-incrementally process each report as it comes in, or may be necessary to wait
-until the entire batch of reports is received.
+The leader orchestrates the process of verifying them (see
+{{validating-inputs}}) and assembling them into a final aggregate for the
+collector. Depending on the VDAF, it may be possible to incrementally process
+each report as it comes in, or it may be necessary to wait until the entire
+batch of reports is received.
```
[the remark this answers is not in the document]
cjpatton: No that's not possible. Yeah, it's the aggregation parameter that matters, but here (in the overview) we have so far avoided talking about the aggregation parameter.

Note that this change isn't new to this PR.

draft-gpew-priv-ppm.md
```diff
@@ -521,49 +549,35 @@ the cache lifetime in order to avoid rejecting reports.
 
 ### Upload Request
 
-Clients upload reports by using an HTTP POST to `[leader]/upload`, where
-`[leader]` is the first entry in the task's aggregator endpoints. The payload is
-structured as follows:
+Report shares are usually uploaded by clients dirctly, but MAY instead be
+uploaded by an intermediary that ingests and buffers reports on behalf of the
+collector. A report share is uploaded using an HTTP POST to
```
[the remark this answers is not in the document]
cjpatton: IMO this is out-of-scope. The way we should try to spell this PR is to allow either clients to upload report shares directly or allow report shares to be uploaded by an intermediary. How the report shares get to the intermediary isn't relevant for us.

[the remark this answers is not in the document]
cjpatton: Good point, that's just one possible case. I dropped "on behalf of the collector".

draft-gpew-priv-ppm.md
cjpatton: #89 will likely make use of upload extensions, hence we'll need to keep them around.

draft-gpew-priv-ppm.md
cjpatton: Yeah, it may not be strictly necessary. However I think we should keep this as-is for now in order to keep this PR minimal. Plus, it may be helpful for something we're not seeing right now.

draft-gpew-priv-ppm.md
cjpatton: Actually at this point it will have computed an aggregate share. However I took your suggestion because your text is a lot clearer.

draft-gpew-priv-ppm.md
cjpatton: I don't think want to change the batch after it has been collected with *any* aggregation parameter. Imagine, for example, that the collector requests the aggregate for candidate prefixes ["ab"]. Then add another measurement and re-run with candidate prefixes ["abc"]. Then I know immediately whether the additional measurement was prefixed by "abc".

draft-gpew-priv-ppm.md
```diff
@@ -257,6 +259,28 @@ The overall system architecture is shown in {{pa-topology}}.
...
+      |   +--------------------+   |
+      |   |                        |
+      |   |                        | collect
++-----V---V--+              +------+-----+
+|            |              |            |
+|   Helper   |              | Collector  |
+|            |              |            |
++------------+              +------------+
+~~~
+{: #pa-topology-alt title "System Architecture"}
```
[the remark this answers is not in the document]
chris-wood: Agree with Tim here -- I'm not seeing the value of keeping both.

draft-gpew-priv-ppm.md
```diff
@@ -445,9 +474,7 @@ number generator. Each task has the following parameters associated with it:
 
 * `aggregator_endpoints`: A list of URLs relative to which an aggregator's API
   endpoints can be found. Each endpoint's list MUST be in the same order. The
-  leader's endpoint MUST be the first in the list. The order of the
-  `encrypted_input_shares` in a `Report` (see {{uploading-reports}}) MUST be the
-  same as the order in which aggregators appear in this list.
+  leader's endpoint MUST be the first in the list.
```
chris-wood, line 477: Is this requirement necessary anymore? It seems like the client doesn't need to care which aggregator is the leader if both are treated as separate aggregators to store shares.

draft-gpew-priv-ppm.md
```diff
@@ -464,16 +491,17 @@ number generator. Each task has the following parameters associated with it:
...
 
 ### Key Configuration Request {#key-config}
 
-Before the client can upload its report to the leader, it must know the public
-key of each of the aggregators. These are retrieved from each aggregator by
-sending a request to `[aggregator]/key_config`, where `[aggregator]` is the
-aggregator's endpoint URL, obtained from the task parameters. The aggregator
-responds to well-formed requests with status 200 and an `HpkeConfig` value:
+Before the client can upload a report share to an aggregator, it must first
+obtain the aggregator's public key. This is retrieved by sending a request to
```
chris-wood, line 501:
```suggestion
Before the client can upload a report share to an aggregator, it first fetches
the aggregator's public key. This is done by sending a request to
```

draft-gpew-priv-ppm.md
chris-wood: We may also need general extensibility outside of new VDAFs in the future -- a lesson learned from TLS.

draft-gpew-priv-ppm.md
```diff
@@ -521,49 +549,34 @@ the cache lifetime in order to avoid rejecting reports.
 
 ### Upload Request
 
-Clients upload reports by using an HTTP POST to `[leader]/upload`, where
-`[leader]` is the first entry in the task's aggregator endpoints. The payload is
-structured as follows:
+Report shares are usually uploaded by clients directly, but MAY instead be
+uploaded by an intermediary that ingests and buffers reports. A report share is
+uploaded using an HTTP POST to `[aggregator]/upload`, where `[aggregator]` is
```
chris-wood, line 554: This seems to preclude sending both reports to the same endpoint, even though they're encrypted under separate aggregator public keys. We could address this by allowing the task configuration to specify the upload endpoint(s) _separately_ from the aggregation endpoints. Would that work?

draft-gpew-priv-ppm.md
chris-wood: Agreed. Let's note this and revisit in a followup change. (Good catch, Tim!)

draft-gpew-priv-ppm.md
```diff
@@ -704,49 +716,39 @@ struct {
...
-The `nonce` and `extensions` fields have the same value as those in the
-report uploaded by the client. Similarly, the `helper_share` field is the
-`EncryptedInputShare` from the `Report` whose index in
-`Report.encrypted_input_shares` is equal to the index of `[aggregator]` in the
-task's aggregator endpoints. [OPEN ISSUE: We usually only need to send this in
-the first aggregate request. Shall we exclude it in subsequent requests
-somehow?] The remainder of the structure is dedicated to VDAF-specific request
-parameters.
+The `nonce` field has the same value as those in the leader's report share
+uploaded by the client. The remainder of the structure is dedicated to
```
chris-wood, line 739: Hmm, now that the nonce<>client uniqueness is no longer enforced by a single party, what happens if, say, two benign clients happen to generate colliding nonce values, and each aggregator picks a different one while filtering for anti-replay? We could maybe increase the width of the nonce random value to hedge against this collision, but it's not clear to me that's sufficient.

draft-gpew-priv-ppm.md
```diff
@@ -969,11 +959,27 @@ prevent such replay attacks, this specification requires the aggregators to
...
+  been collected, then the aggregator MUST ignore it. This prevents additional
+  reports from being aggregated after their batch has already been collected.
+
+* If the report has been aggregated but has not been collected, then the
+  aggregator MUST ignore it. This prevents a report from being used more than
+  once in a batch.
+
+In particular, an aggregator only process a report if either it has been
+aggregated and collected or it has never been aggregated and is not contained by
+a batch that has been collected.
```
chris-wood, line 982: This new requirement is pretty subtle to me. Can we drop an open issue here to further investigate if this is indeed the correct thing to do?

chris-wood (commented):
Thanks, @cjpatton! I have a couple questions and comments. The biggest unknown for me is the new set of report processing requirements, but I think we can work through that.

draft-gpew-priv-ppm.md
cjpatton: Duplicate of https://github.com/abetterinternet/ppm-specification/pull/174#discussion_r791074329?

draft-gpew-priv-ppm.md
cjpatton: Oh interesting! To be clear, before #169 we still had both aggregators enforce anti-replay. The difference is that we no longer of single party pick the set of reports that are processed. Nice catch.

This basically amounts to an attack on correctness. What will happen for prio3 or poplar1 is that the aggregators will derive different verification randomness, which will cause them to deem the report invalid and reject it (with high probability).

However, where this edge case *would* matter is if a VDAF were used that didn't have any verifiability. (We should probably not get in the habbit of calling such a scheme a VDAF: see https://github.com/cjpatton/vdaf/issues/20.) This was already requested in #45, so it's within the realm of possibility.

Note that a collision fairly likely, since the nonce only contains 64 random bits. (The remaining bits are a timestamp.) We may consider bumping this to 128 just to be on the safe side.

How about addressing this by leaving an OPEN ISSUE in the text? It would reference #45 and we would leave a note there to make sure we keep track of this.

draft-gpew-priv-ppm.md
cjpatton: I agree it's subtle. I'd be good with that outcome, although it was already discussed here in a different context: https://github.com/abetterinternet/ppm-specification/issues/183#issuecomment-1016978627.

chris-wood: Let's at least leave an OPEN ISSUE referencing #183, then, since that's still open. (I will review #183 more carefully soon.)

draft-gpew-priv-ppm.md
chris-wood: That's a fine resolution, yep. Let's also note that this type of collision will cause verification failures with overwhelming probability?

draft-gpew-priv-ppm.md
chris-wood: Well, consider the case where there is an ingestor service. Certainly, that ingestor service can't be the _client_ in PPM, since that would require it see inputs and then encrypt them under the aggregator public keys. Rather, the ingestor sees encrypted things that it then relays onward to the aggregator endpoints. What we're specifying here is the client behavior, and I think we need to make sure that we the client can send both its encrypted shares to a single ingestor if needed. Right now that seems prohibited?

draft-gpew-priv-ppm.md
chris-wood: Oh, yeah, that was totally not clear to me! How is this expressed in terms of VDAF?

draft-gpew-priv-ppm.md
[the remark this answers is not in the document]
tgeoghegan: OK -- there's enough to chew on in this PR that we don't have to wordsmith this right now.

draft-gpew-priv-ppm.md
cjpatton: Will do.

draft-gpew-priv-ppm.md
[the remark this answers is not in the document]
tgeoghegan: This PR, sure, but I'm not the only person left wondering whether or how intermediates can batch up reports: ekr asked about this on the PPM list. Anyway, I filed #188 so we don't lose track of the idea.

draft-gpew-priv-ppm.md
tgeoghegan: Well, I don't want to open the can of worms of client attestation in the margins of this PR, but I'm happy to accept TLS as a precedent.

draft-gpew-priv-ppm.md
tgeoghegan: Then I think there's something I must not be understanding about how `max_batch_lifetime` works, or how PPM is meant to solve t-heavy-hitters, which I thought involved the collector sending a series of collect requests with longer candidate prefixes each time. But that doesn't need to be solved in this PR.

cjpatton: You understand correctly how PPM is supposed to solve heavy hitters. All `max_batch_lifetime` is supposed to say is that you can't make any more collect requests than `max_batch_lifetime`. Otherwise it doesn't make any restrictions on the aggregation parameter. In particular, Poplar can be implemented as you describe.

All that this line is saying is that you can't change the set of reports in a batch after you've collected it. Otherwise, this could lead to a privacy violation, regardless of the VDAF in use.

draft-gpew-priv-ppm.md
cjpatton: It's not explicit in the syntax, but for schemes that needs to differentiate aggregator roles, they can do so via the verification parameter. For example, see https://cjpatton.github.io/vdaf/draft-patton-cfrg-vdaf.html#name-setup-2

draft-gpew-priv-ppm.md
cjpatton: Ok I see. Then I think this section would need to be clear that it specifies both the behavior of the client generating report shares and the entity (either the client or someone else) that uploads shares to the aggregators. Would you prefer these be different sections perhaps?

cjpatton:
We can revisit this later, but this is pretty stale and it woiuld be nice to clear the decks. Closing.

PR #177, opened 2022-01-04
Rename "Heavy Hitters" to "Poplar"

### LINKED ISSUE #176: "Heavy Hitters" is now "Poplar"
@henrycg has informed me that the authors of [[BBCGGI21](https://eprint.iacr.org/2021/017.pdf)] have given their protocol a proper name so as to avoid confusing conflating the protocol with the problem it solves. We ought to adopt the name in the PPM and [VDAF](https://github.com/cjpatton/vdaf) specs.

cjpatton:
Closes #176.

The corresponding change in the VDAF draft is here: https://github.com/cjpatton/vdaf/pull/16

### DIFF
--- a/draft-gpew-priv-ppm.md
+++ b/draft-gpew-priv-ppm.md
@@ -205,10 +205,11 @@ schemes that implement the VDAF interface specified in
   etc. This class of VDAFs is based on Prio [CGB17] and includes improvements
   described in [BBCGGI19].
 
-* `hits`, which allows for finding the most common strings among a collection
-  of clients (e.g., the URL of their home page) as well as counting the number
-  of clients that hold a given string. This VDAF is the basis of the
-  privacy preserving heavy hitters protocol of [BBCGGI21].
+* `poplar1`, which allows for finding the most popular strings among a
+  collection of clients (e.g., the URL of their home page) as well as counting
+  the number of clients that hold a given string. This VDAF is the basis of the
+  Poplar protocol of [BBCGGI21], which is designed to solve the heavy hitters
+  problem in a privacy preserving manner.
 
 This protocol is designed to work with schemes that use secret sharing. Rather
 than send its input in the clear, each client shards its measurements into a
@@ -624,7 +625,7 @@ The exact structure of the aggregation flow depends on the VDAF.  Specifically:
 
 * Some VDAFs (e.g., `prio3`) allow the leader to start aggregating reports
   proactively before all the reports in a batch are received. Others (e.g.,
-  `hits`) require all the reports to be present and must be initiated by the
+  `poplar1`) require all the reports to be present and must be initiated by the
   collector.
 
 * Processing the reports -- especially validating them -- may require multiple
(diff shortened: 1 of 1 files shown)


draft-gpew-priv-ppm.md
```diff
@@ -205,10 +205,11 @@ schemes that implement the VDAF interface specified in
   etc. This class of VDAFs is based on Prio [CGB17] and includes improvements
   described in [BBCGGI19].
 
-* `hits`, which allows for finding the most common strings among a collection
+* `pops`, which allows for finding the most popular strings among a collection
```
tgeoghegan, line 208: As @bifurcation noted in https://github.com/cjpatton/vdaf/pull/16, I think `poplar` is nicer.

tgeoghegan (changes requested):

draft-gpew-priv-ppm.md
cjpatton: "Poplar" is the name of the full-blown protocol, whereas this name is supposed to refer to the underlying VDAF. I'd like to avoid conflating the two.

ekr: I guess we're just voting now, but I would prefer poplar.

cjpatton: Well then, this just got interesting. I wonder if others have an opinion or ideas for alternatives. I don't care much about the name for the VDAF, my only criterion is that it is distinct from "Poplar". A more verbose possibility is `poplar-vdaf`. 
cc/ @csharrison @eriktaubeneck @martinthomson @stpeter @bifurcation @BranLwyd

csharrison: You could align with the naming scheme for the Prio VDAF and call it `poplar1` :) Other than that I don't have a strong opinion.

eriktaubeneck: Keeping it consistent with the naming scheme for Prio seems positive, and a version number doesn't seem like a terrible addition as it seems likely the VDAF may evolve like the prio VDAF did.

cjpatton: I like your thinking @eriktaubeneck  and @csharrison ! I vote for `poplar1`.

ekr:
SGTM.

henrycg (approved):
Looks good! Thanks.

cjpatton:
s/pops/poplar1/ as discussed. @tgeoghegan and @ekr can I get you to approve this?

draft-gpew-priv-ppm.md
cjpatton: s/pops/poplar1/.

ekr (approved):
LGTM

tgeoghegan (approved):
This change is good and consistent with the corresponding change to vdaf. However I see there are two other occurrences of "Hits" in the doc, on lines 1335 and 1356:

```
The PPM parameters also specify the maximum number of times a report can be
used. Some protocols, such as *Hits*, require reports to be used in multiple
batches spanning multiple collect requests.
<...>
Most PPM protocols, including Prio and *Hits*, are robust against malicious
clients, but are not robust against malicious servers.
```
I think we could change those to "Poplar"

cjpatton:
s/Hits/Poplar/. Thanks @tgeoghegan !

chris-wood (approved):

PR #181, opened 2022-01-18
Align some terminology with the VDAF draft

cjpatton:
What PPM calls an "output share" is called an "aggregate share" in the
VDAF draft. Furthermore, in the VDAF draft, an "output share" is what
an aggregator recovers from running the MPC with its peers. This change
fixes this ambiguity without changing the PPM protocol itself (modulo
changing the info string for HPKE).


draft-gpew-priv-ppm.md
```diff
@@ -732,10 +741,11 @@ sub-requests as described in {{anti-replay}}. Any sub-requests filtered out by
 this procedure MUST be ignored.
 
 In addition, for any report whose nonce contains a timestamp that falls in a
-batch interval for which it has received at least one output-share request (see
-{{output-share-request}}), the helper MUST send an error messsage in response
-rather than its next VDAF message. Note that this means leaders cannot
-interleave a sequence of aggregate and output-share requests for a single batch.
+batch interval for which it has coimpleted at least one aggregate-share request
```
tgeoghegan, line 744:
```suggestion
batch interval for which it has completed at least one aggregate-share request
```

draft-gpew-priv-ppm.md
```diff
@@ -165,12 +162,17 @@ Leader:
 : A distinguished aggregator that coordinates input validation and data
    collection.
 
-Output:
+Aggregate Result:
```
tgeoghegan, line 165:
```suggestion
Aggregate result:
```
To be consistent with other glossary entries

draft-gpew-priv-ppm.md
```diff
@@ -317,19 +319,26 @@ until the entire batch of reports is received.
...
-multi-party computation, specified by the chosen VDAF, to ensure that consuming
-the report will not corrupt the aggregate result. The input shares
-generated by the client include information used by the aggregators during
-aggregation. For example, `prio3` includes a distributed zero-knowledge proof of
-the input's validity [BBCGGI19] which the aggregators can jointly verify
-and reject the report if it cannot be verified. However, they do not
-learn anything about the individual report other than that it is valid.
+multi-party computation specified by the chosen VDAF
+{{?I-D.draft-cfrg-patton-vdaf}} in order to prepare a report for aggregation. At
+the beginning of this computation, each aggregator is in posession of an input
```
tgeoghegan, line 330:
```suggestion
the beginning of this computation, each aggregator is in possession of an input
```

tgeoghegan (approved):
LGTM, modulo typos

draft-gpew-priv-ppm.md
```diff
@@ -317,19 +319,26 @@ until the entire batch of reports is received.
...
+{{?I-D.draft-cfrg-patton-vdaf}} in order to prepare a report for aggregation. At
+the beginning of this computation, each aggregator is in possession of an input
+share uploaded by the client. At the end of the computation, each aggregator is
+in posession of either an "output share" that is ready to be aggregated or an
+indication that a valid output share could not be recovered.
+
+To facilitiate this computation, the input shares generated by the client
+include information used by the aggregators during aggregation in order to
+validate their recovered output shares. For example, `prio3` includes a
+distributed zero-knowledge proof of the input's validity [BBCGGI19] which the
```
chris-wood, line 338:
```suggestion
distributed zero-knowledge proof of the input's validity {{BBCGGI19}} which the
```

chris-wood (approved):

draft-gpew-priv-ppm.md
cjpatton: Hmm, the tool that renders the markdown into XML seems to interpret this correctly. Plus we've been using `{{XXX}}` rather than `{{blah}}` for paper references fairly consistently throughout the doc. Do you think, we should change this back?

chris-wood: `[..]` does nothing in the translation process. It literally duplicates the text inline. You could write `[bananas]` and it would render `[bananas]` in the XML/TXT/HTML output. `{{...}}`` actually invokes the translation process to cite things correctly. So, yeah, we should be using `{{...}}`.

cjpatton: That's not true, as far as I can tell. take a look at the rendered HTML. A link is put there that navigates to the informational references at the bottom. And we don't use the `{{...}}` syntax anywhere for this particular reference.

cjpatton: As discussed offline, I'll follow up with a PR that changes all the [...] style references with {{...}} style references.

draft-gpew-priv-ppm.md
```diff
@@ -165,12 +162,17 @@ Leader:
...
+Aggregate share:
+: A share of the aggregate result emitted by an aggregator. Aggregate shares are
+   reassembled by the collector into the final output.
+
 Output share:
-: The share of an output emitted by an aggregator. Output shares can be
-   reassembled by the leader into the final output.
+: An aggregator's share of the output of the VDAF
+   {{?I-D.draft-cfrg-patton-vdaf}} preparation step. Many output shares are
+   combined into an aggregate share via the VDAF aggregation algorithm.
```
ekr, line 175: This may be the VDAF term, but it seems kind of unfortunate. Why  isn't this called "input"?

draft-gpew-priv-ppm.md
```diff
@@ -317,19 +319,26 @@ until the entire batch of reports is received.
...
+multi-party computation specified by the chosen VDAF
+{{?I-D.draft-cfrg-patton-vdaf}} in order to prepare a report for aggregation. At
+the beginning of this computation, each aggregator is in possession of an input
+share uploaded by the client. At the end of the computation, each aggregator is
+in posession of either an "output share" that is ready to be aggregated or an
+indication that a valid output share could not be recovered.
+
+To facilitiate this computation, the input shares generated by the client
+include information used by the aggregators during aggregation in order to
+validate their recovered output shares. For example, `prio3` includes a
```
ekr, line 337:
```suggestion
validate their corresponding output shares. For example, `prio3` includes a
```

I think this might help

draft-gpew-priv-ppm.md
```diff
@@ -317,19 +319,26 @@ until the entire batch of reports is received.
...
-aggregation. For example, `prio3` includes a distributed zero-knowledge proof of
-the input's validity [BBCGGI19] which the aggregators can jointly verify
-and reject the report if it cannot be verified. However, they do not
-learn anything about the individual report other than that it is valid.
+multi-party computation specified by the chosen VDAF
+{{?I-D.draft-cfrg-patton-vdaf}} in order to prepare a report for aggregation. At
+the beginning of this computation, each aggregator is in possession of an input
+share uploaded by the client. At the end of the computation, each aggregator is
+in posession of either an "output share" that is ready to be aggregated or an
+indication that a valid output share could not be recovered.
```
ekr, line 333:
```suggestion
indication that a valid output share could not be computed.
```
"recovered" is a weird word.

draft-gpew-priv-ppm.md
```diff
@@ -792,57 +802,57 @@ usually need to be encrypted in order to protect user privacy. However, the
...
 
-Next, the helper encrypts the output share `output_share` under the collector's
+Next, the helper encrypts the aggregate share `agg_share` under the collector's
 public key as follows:
 
 ~~~
 enc, context = SetupBaseS(pk,
-                          "pda output share" || task_id || server_role)
-encrypted_output_share = context.Seal(batch_interval, output_share)
+                          "pda agg share" || task_id || server_role)
```
ekr, line 847: Is "aggregate" not gonna fit?

ekr (approved):
Generally looks good. I would change the "agg" in the label if you can as it is confusing.

draft-gpew-priv-ppm.md
cjpatton: So I'm thinking it might simplify analysis if we make the info string the same length for each invocation of HPKE the same length. I'm thinking "ppm inp share" for the encrypted input shares and "ppm agg share" for the encrypted aggregate shares. I sort of half-assed it here, but I was planning on doing this in a future PR. I'll just change it to "aggregate" in this PR and defer that discussion.

cjpatton: Done

cjpatton:
Squashed.

draft-gpew-priv-ppm.md
cjpatton: As discussed offline, the "prepare step" is the MPC that maps input shares to output shares (using the aggregation parameter to control the output).

cjpatton:
Thanks fellas!

PR #182, opened 2022-01-19
Convert [..]-style refs to {{..}}

cjpatton:
Based on #181 (merge that first).

@chris-wood pointed out that, if there is no reference for "banana",
then `[banana]` gets rendered as text. On the other hand `{{banana}}`
will fail to build because the reference is missing. The latter behavior
is preferable.


chris-wood (approved):

ekr:
LGTM

PR #184, opened 2022-01-21
Redefine `Interval` to invalid values impossible

tgeoghegan:
While reviewing #179, @cjpatton and I figured out a way to redefine
`struct Interval` so that invalid values cannot be represented[1].

The existing representation contains `Time start` and `Time end`, which
means an interval could be defined with `end < start`, which is invalid.
Redefining `Interval` as a start instant and a duration makes it
impossible to represent invalid `Interval`s (since `Duration` is an
unsigned integer).

Additionally, this implicitly fixes a bug in the batch interval validation
logic: we asserted that `batch_interval.end` should be divisible by
`min_batch_interval`, but since `end` is _excluded_ from the `Interval`, I
think that was incorrect. Instead, `batch_interval.end - 1` should be divisible
by `min_batch_duration`. In any case, this is simpler with the new
definition of `Interval`: `Interval.duration` needs to be a multiple of
`min_batch_duration`.

[1]: https://github.com/abetterinternet/ppm-specification/pull/179#discussion_r789340693

### DIFF
--- a/draft-gpew-priv-ppm.md
+++ b/draft-gpew-priv-ppm.md
@@ -407,17 +407,18 @@ PPM has three major interactions which need to be defined:
 We start with some basic type definitions used in other messages.
 
 ~~~
- /* ASCII encoded URL. e.g., "https://example.com" */
+/* ASCII encoded URL. e.g., "https://example.com" */
 opaque Url<1..2^16-1>;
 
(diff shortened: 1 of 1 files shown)


cjpatton (approved):

chris-wood (approved):
Nice!

PR #193, opened 2022-03-01
Make collect requests asynchronous

### LINKED ISSUE #111: Do something about long-running requests
It's easy to imagine that a request will take a long time, especially if you are using Collect and that has to trigger a bunch of work from the leader. We need some way of handling this.

tgeoghegan:
Make the handling of collect requests on the leader asynchronous, since
servicing a collect request could entail running the preparation
protocol over all the reports that fall into the batch interval,
especially in VDAFs like poplar1 where the aggregation parameter is not
available until the collector provides it.

I've only made the collect request handling asynchronous. My reasoning is that handling other requests synchronously should be feasible:

- upload requests should only contain shares of a single report and thus never be all that big. Additionally, the client needs an explicit acknowledgement from the aggregator that the uploaded report was accepted and has been persisted.
- the requests in the aggregation/preparation sub-protocol should also never get that big: even when preparing a large number of reports, the leader can choose to carve that work up into any number of aggregation jobs, each of which covers a small enough number of reports that the message sizes and processing times should be reasonable to handle in synchronous request/response exchanges. 

Resolves #111

### DIFF
--- a/draft-gpew-priv-ppm.md
+++ b/draft-gpew-priv-ppm.md
@@ -886,6 +886,9 @@ request, the collector issues a POST request to `[leader]/collect`, where
 `[leader]` is the leader's endpoint URL. The body of the request is structured
 as follows:
 
+[OPEN ISSUE: Decide if and how the collector's request is authenticated. If not,
+then we need to ensure that collect job URIs are resistant to enumeration
+attacks.]
 ~~~
 struct {
   TaskID task_id;
@@ -900,36 +903,63 @@ The named parameters are:
 * `batch_interval`, the request's batch interval.
 * `agg_param`, an aggregation parameter for the VDAF being executed.
 
-Depending on the VDAF scheme and how the leader is configured, the collect
-request may cause the leader to send a series of aggregate requests to the
-helpers in order to compute their share of the aggregate. Alternately, the
-leader may already have made these requests and can respond immediately. In
-either case it responds to the collector's request as follows.
+Depending on the VDAF scheme and how the leader is configured, the leader and
+helper may already have prepared all the reports falling within `batch_interval`
+and be ready to return the aggregate shares right away, but this cannot be
+guaranteed. In fact, for some VDAFs, it is not be possible to begin preparing
+inputs until the collector provides the aggregation parameter in the
+`CollectReq`. For these reasons, collect requests are handled asynchronously.
 
-It begins by checking that the request meets the requirements of the batch
-parameters using the procedure in {{batch-parameter-validation}}. If so, it
-obtains the helper's encrypted aggregate share for the batch interval by sending
-an AggregateShareReq to the helper as described in {{aggregate-share-request}}.
-(This request may too have been made in advance.)
+Upon receipt of a `CollectReq`, the leader begins by checking that the request
+meets the requirements of the batch parameters using the procedure in
+{{batch-parameter-validation}}. If so, it immediately sends the collector a
+response with HTTP status 303 See Other and a Location header containing a URI
+identifying the collect job that can be polled by the collector, called the
+"collect job URI".
 
-Next, the leader computes its own aggregate share by aggregating all of the
-verified output shares that fall within the batch interval. Finally, it responds
-with HTTP status 200 and a body consisting of a CollectResp message:
+The leader then begins working with the helper to prepare the shares falling
+into `CollectReq.batch_interval` (or continues this process, depending on the
+VDAF) as described in {{pa-aggregate}}.
 
-[OPEN ISSUE: What happens if this all takes a really long time.]
+After receiving the response to its CollectReq, the collector makes an HTTP GET
+request to the collect job URI to check on the status of the collect job and
+eventually obtain the result. If the collect job is not finished yet, the leader
+responds with HTTP status 202 Accepted. The response MAY include a Retry-After
+header field to suggest a pulling interval to the collector.
 
-[TODO: Decide if and how the collector's request is authenticated.]
+Once all the necessary reports have been prepared, the leader obtains the
+helper's encrypted aggregate share for the batch interval by sending an
+`AggregateShareReq` to the helper as described in {{aggregate-share-request}}.
+The leader then computes its own aggregate share by aggregating all of the
+prepared output shares that fall within the batch interval.
+
+When both aggregators' shares are successfully obtained, the leader responds to
+subsequent HTTP GET requests to the collect job's URI with HTTP status 200 OK
+and a body consisting of a `CollectResult`:
 
 ~~~
 struct {
   EncryptedAggregateShare shares<1..2^16-1>;
-} CollectResp;
+} CollectResult;
 ~~~
 
 * `shares` is a vector of `EncryptedAggregateShare`s, as described in
   {{aggregate-share-request}}, except that for the leader's share, the `task_id`
   and `batch_interval` used to encrypt the `AggregateShare` are obtained from
   the `CollectReq`.
+
+If obtaining aggregate shares fails, then the leader responds to subsequent HTTP
+GET requests to the collect job URI with an HTTP error status and a problem
+document as described in {{errors}}.
+
+The leader MUST retain a collect job's results until the collector sends an HTTP
+DELETE request to the collect job URI, in which case the leader responds with
+HTTP status 204 No Content.
+
+[OPEN ISSUE: Allow the leader to drop aggregate shares after some reasonable
+amount of time has passed, but it's not clear how to specify that. ACME doesn't
+bother to say anything at all about this when describing how subscribers should
+fetch certificates: https://datatracker.ietf.org/doc/html/rfc8555#section-7.4.2]
 
 [OPEN ISSUE: Describe how intra-protocol errors yield collect errors (see
 issue#57). For example, how does a leader respond to a collect request if the
@@ -1412,7 +1442,7 @@ corresponding media types types:
 - AggregateShareReq {{aggregate-share-request}}: "message/ppm-aggregate-share-req"
 - AggregateShareResp {{aggregate-share-request}}: "message/ppm-aggregate-share-resp"
 - CollectReq {{pa-collect}}: "message/ppm-collect-req"
-- CollectResp {{pa-collect}}: "message/ppm-collect-req"
+- CollectResult {{pa-collect}}: "message/ppm-collect-result"
 
 The definition for each media type is in the following subsections.
 


divergentdave:
With this change, we will have two differing definitions of structs named `CollectResp`. My understanding is that clients will know which one to expect based on the URL of the request, and the status code in the response. (they will receive a URL from /collect on success, and the `collect_job` URL will return a URL with status code 202, or shares with status code 200) I think we should change one of these for clarity to the reader, i.e. `CollectCompleteResp` or `CollectJobResp`.

draft-gpew-priv-ppm.md
```diff
@@ -900,37 +900,70 @@ The named parameters are:
...
+the client.
+
+The leader then begins working with the helper to prepare the shares falling
+into `CollectReq.batch_interval` (or continues this process, depending on the
+VDAF) as described in {{pa-aggregate}}. After receiving a `CollectResp`, the
+collector periodically makes HTTP GET requests to `CollectResp.collect_job` to
+check on the status of the collect job and eventually obtain the result. If the
+collect job is not finished yet, the leader responds with HTTP status 202
+Accepted and a `CollectResp`. The `CollectResp` MAY contain a different
+`collect_job` than previous `CollectResp`s and the HTTP response MAY include a
```
BranLwyd, line 934: Q: what's the practical reason for allowing the `collect_job` to change?

draft-gpew-priv-ppm.md
```diff
@@ -900,37 +900,70 @@ The named parameters are:
...
+
+When both aggregator's shares are successfully obtained, then the leader
+responds to subsequent HTTP GET requests to `CollectResp.collect_job` with HTTP
+status 200 OK and a body consisting of a `CollectResult`:
 
 ~~~
 struct {
-  EncryptedAggregateShare shares<1..2^16-1>;
-} CollectResp;
+  EncryptedAggregateShare shares<1..2^8-1>;
```
BranLwyd, line 949: This seems low -- do we really only want to support shares of up to 255 bytes? Notably, each EncryptedAggregateShare structure has several fields (`enc`, `payload`) that may be up to 64 KiB each, so currently a maximally-sized EncryptedAggregateShare may not fit in this vector.

BranLwyd (changes requested):

draft-gpew-priv-ppm.md
tgeoghegan: Ah, that's the second time I've made that mistake! I thought this meant up to 2^8-1 `EncryptedAggregateShare` structures, but you're right, it means number of bytes. I'll make this 2^24-1 to be safe.

draft-gpew-priv-ppm.md
tgeoghegan: I didn't see a reason to require that it be the same, but I suppose we could instead insist that it always be the same. Or we could say nothing, implicitly permitting `collect_job` to change.

BranLwyd: WDYT about having the `collect_job` handler return an HTTP 202 with an unspecified (or empty) body if the results aren't yet available? This removes the unnecessary flexibility of changing `collect_jobs` without requiring specification language.

This is pretty nit-picky; I'd personally go with an unspecified body to leave things maximally flexible while allowing us to specify later, but LGTM either way.

BranLwyd (approved):

draft-gpew-priv-ppm.md
```diff
@@ -900,37 +900,70 @@ The named parameters are:
...
 
-Depending on the VDAF scheme and how the leader is configured, the collect
-request may cause the leader to send a series of aggregate requests to the
-helpers in order to compute their share of the aggregate. Alternately, the
-leader may already have made these requests and can respond immediately. In
-either case it responds to the collector's request as follows.
+Depending on the VDAF scheme and how the leader is configured, the leader and
+helper may already have prepared all the reports falling within `batch_interval`
+and be ready to return the aggregate shares right away, but this cannot be
+guaranteed and in some VDAFs, it may not be possible to begin preparing inputs
```
cjpatton, line 906:
```suggestion
guaranteed.  In fact, for some VDAFs, it is not be possible to begin preparing inputs
```

draft-gpew-priv-ppm.md
```diff
@@ -900,37 +900,70 @@ The named parameters are:
...
+until the collector provides the aggregation parameter in the `CollectReq`. For
+those reasons, collect requests are handled asynchronously.
```
cjpatton, line 908:
```suggestion
these reasons, collect requests are handled asynchronously.
```

draft-gpew-priv-ppm.md
```diff
@@ -900,37 +900,70 @@ The named parameters are:
...
+`collect_job` than previous `CollectResp`s and the HTTP response MAY include a
+`Retry-After` header field to suggest a polling interval to the collector.
+
+Once all the necessary reports have been prepared, the leader obtains the
+helper's encrypted aggregate share for the batch interval by sending an
+`AggregateShareReq` to the helper as described in {{aggregate-share-request}}.
+The leader then computes its own aggregate share by aggregating all of the
+prepared output shares that fall within the batch interval.
+
+When both aggregator's shares are successfully obtained, then the leader
```
cjpatton, line 943:
```suggestion
When both aggregator's shares are successfully obtained, the leader
```

draft-gpew-priv-ppm.md
```diff
@@ -900,37 +900,70 @@ The named parameters are:
...
-Next, the leader computes its own aggregate share by aggregating all of the
-verified output shares that fall within the batch interval. Finally, it responds
-with HTTP status 200 and a body consisting of a CollectResp message:
-
-[OPEN ISSUE: What happens if this all takes a really long time.]
+~~~
+struct {
+  Url collect_job;
+} CollectResp;
+~~~
```
cjpatton, line 919: Is sending a binary-encoded response with the URL necessary/desirable? What about just adding the URL as an HTTP header? What does ACME do here?

draft-gpew-priv-ppm.md
cjpatton: Allowing the collect job URL to change seems to add some complexity. If we don't need this complexity, than I suggest we remove it. In particular,  I'd propose removing this optional behavior.

draft-gpew-priv-ppm.md
```diff
@@ -900,37 +900,70 @@ The named parameters are:
...
+
+When both aggregator's shares are successfully obtained, then the leader
+responds to subsequent HTTP GET requests to `CollectResp.collect_job` with HTTP
+status 200 OK and a body consisting of a `CollectResult`:
 
 ~~~
 struct {
-  EncryptedAggregateShare shares<1..2^16-1>;
-} CollectResp;
+  EncryptedAggregateShare shares<1..2^24-1>;
```
cjpatton, line 949: Why the size increase here? It seems unlikely to me that we'll ever have aggregate shares larger than 2^16 - 1 bytes.

cjpatton (changes requested):
Mostly editorial changes. One substantive question: Would it be more appropriate to add the collect job URL as an HTTP header?

draft-gpew-priv-ppm.md
ekr: I would add a TODO just to get this done.

draft-gpew-priv-ppm.md
```diff
@@ -900,37 +900,70 @@ The named parameters are:
...
-[OPEN ISSUE: What happens if this all takes a really long time.]
+~~~
+struct {
+  Url collect_job;
+} CollectResp;
+~~~
 
-[TODO: Decide if and how the collector's request is authenticated.]
+`CollectResp.collect_job` is a URL identifying the collect job that can be
+polled by the collector. It SHOULD contain an unpredictable component to
```
ekr, line 922: This SHOULD seems odd. If this is a real attack then there should be a MUST requiring high entropy.

I'm also not clear what an "enumeration" attack is. Presumably the leader has an access control mechanism, so what does it matter if I try /results/1 and get permission denied? And if it's supposed to prevent me knowing how many jobs there are, then I think this needs to be entirely unpredictable.

draft-gpew-priv-ppm.md
ekr: I prefer @BranLwyd's suggestion here. Note that otherwise we have weird race conditions.

draft-gpew-priv-ppm.md
```diff
@@ -900,37 +900,70 @@ The named parameters are:
...
   the `CollectReq`.
 
+If obtaining aggregate shares fails, then the leader responds to subsequent HTTP
+GET requests to `CollectResp.collect_job` with an HTTP error status and a
+problem document as described in {{errors}}.
+
+The leader SHOULD retain aggregate shares for a reasonable amount of time. Even
+after serving a `CollectResult` to the collector, leaders SHOULD continue to
+respond to further GET requests to `CollectResp.collect_job` in case the
+collector does not receive the response the first time.
```
ekr, line 965: I see why you are doing this and you have to for E2E reasons, but this seems hard to operationalize. I sugegest that instead we want a delete.

ekr (changes requested):

draft-gpew-priv-ppm.md
tgeoghegan: I will incorporate Bran's idea.

draft-gpew-priv-ppm.md
tgeoghegan: You're right, it makes more sense to assume this is authenticated. I'll delete this text.

draft-gpew-priv-ppm.md
tgeoghegan: I consulted ACME for inspiration here: [Section 7.4.2 Downloading the Certificate](https://datatracker.ietf.org/doc/html/rfc8555#section-7.4.2). ACME doesn't specify how long the certificate should be available for at all (and technically allows an ACME server to _never_ provide the cert). Maybe we can get away with saying nothing, too. In any case I wrote up the DELETE idea.

tgeoghegan (commented):
Per @chris-wood's observation, I put the `OPEN ISSUE` about collector to leader auth back in. I also went with HTTP 303 See Other over the previous CollectResp. I suspect we still need to say something about how long a leader should retain aggregate shares, in case the collector never sends the DELETE. @divergentdave suggested that we might put retention period parameters for reports, aggregate shares and other persistently stored data into a PPM task's parameters, which would allow us to punt the problem of picking retention problems to deployments and to write clear protocol text about what to do.

cjpatton (approved):
Nicely done.

divergentdave (approved):

PR #194, opened 2022-03-01
Add checksum & report_count to aggregation share requests.

### LINKED ISSUE #163: Aggregators should inform collector of contribution count in collect response
Collectors will want to know how many inputs are included in the aggregations they receive, for a number of reasons: they may wish to compute a mean over a privately aggregated sum, or track metrics on how many clients are providing inputs. struct `CollectResp` should include a count of how many contributions went into an aggregation.

BranLwyd:
This is intended to allow PPM implementations to discover if the leader
& helper(s) have fallen out-of-sync in their views of what shares are
included in the aggregation.

Notes: it would be trivial for an adversary to generate sets of nonces that collide with one another. Since the goal is to provide a check against server malfunction, rather than a security check against malicious clients or other actors in the system, I think this is acceptable. I evaluated a number of collision-resistant multiset-hash constructions[1], but none of them were currently standardized & specifying them would be more effort than seems necessary given the goals of this checksum.

[1] the most promising of which was [`LtHash`](https://eprint.iacr.org/2019/227.pdf)

Closes #163.

### DIFF
--- a/draft-gpew-priv-ppm.md
+++ b/draft-gpew-priv-ppm.md
@@ -811,9 +811,9 @@ helper responds to this request by extracting its aggregate share from its state
 and encrypting it under the collector's HPKE public key.
 
 First, the leader computes a checksum over the set of output shares included in
-the batch window. The checksum is computed by taking the SHA256 of the nonces
-for the client reports included in the aggregation, then combining the hash
-values with a bytewise-XOR operation.
+the batch window. The checksum is computed by taking the SHA256 hash of each
+nonce from the client reports included in the aggregation, then combining the
+hash values with a bitwise-XOR operation.
 
 Then, for each aggregator endpoint `[aggregator]` in the parameters associated
 with `CollectReq.task_id` (see {{pa-collect}}) except its own, the leader sends
--- a/draft-gpew-priv-ppm.md
+++ b/draft-gpew-priv-ppm.md
@@ -810,6 +810,9 @@ PPM task, the leader issues an "aggregate-share request" to each helper. The
 helper responds to this request by extracting its aggregate share from its state
 and encrypting it under the collector's HPKE public key.
 
+[OPEN ISSUE: consider updating the `checksum` algorithm to not permit
+collisions]
+
 First, the leader computes a checksum over the set of output shares included in
 the batch window. The checksum is computed by taking the SHA256 hash of each
 nonce from the client reports included in the aggregation, then combining the
--- a/draft-gpew-priv-ppm.md
+++ b/draft-gpew-priv-ppm.md
@@ -810,29 +810,42 @@ PPM task, the leader issues an "aggregate-share request" to each helper. The
 helper responds to this request by extracting its aggregate share from its state
 and encrypting it under the collector's HPKE public key.
 
-For each aggregator endpoint `[aggregator]` in the parameters associated with
-`CollectReq.task_id` (see {{pa-collect}}) except its own, the leader sends a
-POST request to `[aggregator]/aggregate_share` with the following message:
+First, the leader computes a checksum over the set of output shares included in
+the batch window. The checksum is computed by taking the SHA256 of the nonces
+for the client reports included in the aggregation, then combining the hash
+values with a bytewise-XOR operation.
+
+Then, for each aggregator endpoint `[aggregator]` in the parameters associated
+with `CollectReq.task_id` (see {{pa-collect}}) except its own, the leader sends
+a POST request to `[aggregator]/aggregate_share` with the following message:
 
 ~~~
 struct {
   TaskID task_id;
   Interval batch_interval;
+  uint64 report_count;
+  opaque checksum[32];
   opaque helper_state<0..2^16>;
 } AggregateShareReq;
 ~~~
 
 * `task_id` is the task ID associated with the PPM parameters.
 * `batch_interval` is the batch interval of the request.
+* `report_count` is the number of reports included in the aggregation.
+* `checksum` is the checksum computed over the set of client reports, computed
+  as described above.
 * `helper_state` is the helper's state, which is carried across requests from
   the leader.
 
 To respond to an AggregateShareReq message, the helper first looks up the PPM
 parameters associated with task `task_id`. Then, using the procedure in
 {{batch-parameter-validation}}, it ensures that the request meets the
-requirements of the batch parameters. If so, it aggregates all valid output
-shares that fall in the batch interval into an aggregate share. The response
-contains an opaque, VDAF-specific message:
+requirements of the batch parameters. It also computes a checksum based on its
+view of the output shares included in the batch window, and checks that the
+`report_count` and `checksum` included in the request match its computed
+values. If so, it aggregates all valid output shares that fall in the batch
+interval into an aggregate share. The response contains an opaque,
+VDAF-specific message:
 
 ~~~
 struct {


tgeoghegan (approved):
This would close #163. This LGTM, though I'm eager to hear what @cjpatton thinks about using XORed SHA256 here.

draft-gpew-priv-ppm.md
```diff
@@ -810,29 +810,42 @@ PPM task, the leader issues an "aggregate-share request" to each helper. The
 helper responds to this request by extracting its aggregate share from its state
 and encrypting it under the collector's HPKE public key.
 
-For each aggregator endpoint `[aggregator]` in the parameters associated with
-`CollectReq.task_id` (see {{pa-collect}}) except its own, the leader sends a
-POST request to `[aggregator]/aggregate_share` with the following message:
+First, the leader computes a checksum over the set of output shares included in
+the batch window. The checksum is computed by taking the SHA256 of the nonces
+for the client reports included in the aggregation, then combining the hash
+values with a bytewise-XOR operation.
```
divergentdave, line 816: nit:
```suggestion
values with a bitwise-XOR operation.
```

draft-gpew-priv-ppm.md
```diff
@@ -810,29 +810,42 @@ PPM task, the leader issues an "aggregate-share request" to each helper. The
...
 {{batch-parameter-validation}}, it ensures that the request meets the
-requirements of the batch parameters. If so, it aggregates all valid output
-shares that fall in the batch interval into an aggregate share. The response
-contains an opaque, VDAF-specific message:
+requirements of the batch parameters. It also computes a checksum based on its
+view of the output shares included in the batch window, and checks that the
+`report_count` and `checksum` included in the request match its computed
+values. If so, it aggregates all valid output shares that fall in the batch
+interval into an aggregate share. The response contains an opaque,
+VDAF-specific message:
```
divergentdave, line 848: I think we will want a separate error enumeration value to signify mismatches in either the report count or the checksum. However, I note that the rest of this aggregate share request section doesn't define how errors are signaled (i.e. for invalid batch parameters, or invalid helper state) so I think it's okay to punt on this for now.

draft-gpew-priv-ppm.md
```diff
@@ -810,29 +810,42 @@ PPM task, the leader issues an "aggregate-share request" to each helper. The
 helper responds to this request by extracting its aggregate share from its state
 and encrypting it under the collector's HPKE public key.
 
-For each aggregator endpoint `[aggregator]` in the parameters associated with
-`CollectReq.task_id` (see {{pa-collect}}) except its own, the leader sends a
-POST request to `[aggregator]/aggregate_share` with the following message:
+First, the leader computes a checksum over the set of output shares included in
+the batch window. The checksum is computed by taking the SHA256 of the nonces
+for the client reports included in the aggregation, then combining the hash
```
divergentdave, line 815: Slight rewording for clarity:
```suggestion
the batch window. The checksum is computed by taking the SHA256 hash of each
nonce from the client reports included in the aggregation, then combining the hash
```

divergentdave (approved):

draft-gpew-priv-ppm.md
BranLwyd: Indeed, I worded this PR to continue to punt on error reporting as much as error reporting was previously punted on. (but FWIW, I agree we'd want to be able to discriminate the checksum/count-mismatch error case from other error cases)

draft-gpew-priv-ppm.md
BranLwyd: Huh, I always thought "bytewise" was more common than "bitwise", but a quick Googling shows this is not the case.

cjpatton (commented):
> Notes: it would be trivial for an adversary to generate sets of nonces that collide with one another. Since the goal is to provide a check against server malfunction, rather than a security check against malicious clients or other actors in the system, I think this is acceptable.

I agree that we don't care about a malicious aggregator here, but we do care about malicious clients. We need to think through what an attacker can do if it controls a subset of the clients and can disrupt the transmission of messages between the aggregators.

Specifically, suppose the leader has computed the checksum

```
chucksum_leader = hash(N1) ^ ... ^ hash(Nb)
```

and the helper has computed the checksum

```
checksum_helper = hash(M1) ^ ... ^ hash(Mc]
```

where `N1, ..., Nb` and `M1, ..., Mc` are distinct sets of nonces (i.e., they differ by at least one element).  Suppose further that the attacker controls the values of the nonces. The attacker's goal is to choose the nonce sets so that `checksum_leader == checksum_helper`. Thus the question is what we need to assume about `hash` to ensure that this is hard.

It's not clear to me that this amounts to a standard assumption about cryptographic hash functions like SHA256. We might be better off using a MAC here instead.

That said, I think it would be best to keep this change as simple as possible. I would be OK with merging as-is, so long as we add a note about the attack above in an "OPEN ISSUE" in the text.

divergentdave:
The hash function wouldn't matter in that setting, as the attacker could generate a pool of a few hundred/thousand nonces and their hashes, fix one set of nonces, and use a system of equations over GF(2)^(digest length) to determine what the other set of nonces should be to match the checksum. Equivalently, they could start with generating a pool of nonces and hashes, treat them as vectors in GF(2)^(digest length), put them in a matrix, and find the kernel of that matrix. Pick two unique vectors that are in the kernel and have the same Hamming weight, then each entry determines whether the corresponding nonce from the pool should be included. These two vectors will give you two unique sets of nonces from the pool, such that they form two batches of the same size with the same checksum.

Using a MAC would require sorting at some point, and would preclude accumulating the checksum as each prepare request comes in, like with LtHash or this XOR construction.

Long term, we need to more rigorously handle ordering of prepare and collect interactions, gracefully handling communication failures and retries between the leader and helper, and agreement on reports to be collected, but I agree this is fine for now as an error heuristic.

draft-gpew-priv-ppm.md
```diff
@@ -810,29 +810,42 @@ PPM task, the leader issues an "aggregate-share request" to each helper. The
 helper responds to this request by extracting its aggregate share from its state
 and encrypting it under the collector's HPKE public key.
 
-For each aggregator endpoint `[aggregator]` in the parameters associated with
-`CollectReq.task_id` (see {{pa-collect}}) except its own, the leader sends a
-POST request to `[aggregator]/aggregate_share` with the following message:
+First, the leader computes a checksum over the set of output shares included in
+the batch window. The checksum is computed by taking the SHA256 hash of each
+nonce from the client reports included in the aggregation, then combining the
+hash values with a bitwise-XOR operation.
```
ekr, line 816: I don't object to this, really, but it seems like kind of overkill to use SHA-256.

ekr (approved):
LGTM at least provisionally.

draft-gpew-priv-ppm.md
BranLwyd: I agree, a non-cryptographic hash would be fine here IMO since we think we don't need a cryptographic hash, since the overall construction permits attackers to produce collisions anyway, and since using a cryptographic hash may give the false impression that the overall construction is actually secure.

I'm going to leave the OPEN ISSUE for now but we should consider addressing this -- either switching to a `checksum` algorithm that actually does provide collision-resistance over the set of nonces, or switch away from using cryptographic-strength primitives entirely.

BranLwyd:
> > Notes: it would be trivial for an adversary to generate sets of nonces that collide with one another. Since the goal is to provide a check against server malfunction, rather than a security check against malicious clients or other actors in the system, I think this is acceptable.
> 
> I agree that we don't care about a malicious aggregator here, but we do care about malicious clients. We need to think through what an attacker can do if it controls a subset of the clients and can disrupt the transmission of messages between the aggregators.
> 
> Specifically, suppose the leader has computed the checksum
> 
> ```
> chucksum_leader = hash(N1) ^ ... ^ hash(Nb)
> ```
> 
> and the helper has computed the checksum
> 
> ```
> checksum_helper = hash(M1) ^ ... ^ hash(Mc]
> ```
> 
> where `N1, ..., Nb` and `M1, ..., Mc` are distinct sets of nonces (i.e., they differ by at least one element). Suppose further that the attacker controls the values of the nonces. The attacker's goal is to choose the nonce sets so that `checksum_leader == checksum_helper`. Thus the question is what we need to assume about `hash` to ensure that this is hard.
> 
> It's not clear to me that this amounts to a standard assumption about cryptographic hash functions like SHA256. We might be better off using a MAC here instead.
> 
> That said, I think it would be best to keep this change as simple as possible. I would be OK with merging as-is, so long as we add a note about the attack above in an "OPEN ISSUE" in the text.

I think any hash function would permit an attacker to create a collision. Treating the hash function as a random oracle, the attacker could try random nonces until they find hash values that produce a basis for `Z_2^n` (where `n` is the output size of the hash in bits). I think this is expected to take right around `n` hash computations if I remember my linear algebra correctly. Then they can use this basis to select a set of nonces producing any desired checksum value. By repeating the process and finding a different basis, they can then produce a distinct set of nonces producing any desired checksum value. (I think this is similar to/the same as David's argument above.)

Swapping out SHA256 for HMAC-SHA256 with the key kept secret by the aggregators would fix that particular issue since clients could no longer predict how a given nonce would contribute to the checksum; but as far as I'm aware there are no security proofs for this construction so there may be other attacks.

The lack of collision-resistance may be acceptable for two reasons IMO: (a) there are other mechanisms in PPM to make sure the aggregators are aggregating the same set of client reports, this is intended as a check against server malfunction/distributed systems errors; (b) since we are currently implementing the client-uploads-to-leader-only model, clients can't provide different sets of nonces to different aggregators. I'm still not entirely happy with it, however.

I added an OPEN ISSUE and I think we should revisit this later.

PR #196, opened 2022-03-04
Small terminology fixes

divergentdave:
Here are a few small editorial changes:

- Refer to TLS's "presentation language", per section 3 of the RFC.
- The output `enc` from the HPKE setup function is an encapsulated key, not an encapsulated context. The term context is defined as "A context is an implementation-specific structure...", whereas `enc` is alternately described as "an encapsulated KEM shared secret", an "encapsulated key", or a "KEM encapsulated key".
- Remove stray whitespace from a hyphenated word.

### DIFF
--- a/draft-gpew-priv-ppm.md
+++ b/draft-gpew-priv-ppm.md
@@ -188,7 +188,7 @@ Server:
 {:br}
 
 
-This document uses the protocol definition language of {{!RFC8446}}.
+This document uses the presentation language of {{!RFC8446}}.
 
 # Overview {#overview}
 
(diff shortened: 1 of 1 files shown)


draft-gpew-priv-ppm.md
```diff
@@ -559,7 +559,7 @@ struct {
 
 * `aggregator_config_id` is equal to `HpkeConfig.id`, where `HpkeConfig` is the
   key config of the aggregator receiving the input share.
-* `enc` is the encapsulated HPKE context, used by the aggregator to decrypt its
+* `enc` is the HPKE encapsulated key, used by the aggregator to decrypt its
```
cjpatton, line 562: This is sometimes referred to as the "encapsulated context", since it's used to construct an HPKE context. That said, I'm fine with this change.

cjpatton (approved):
One non-blocking comment.

PR #197, opened 2022-03-04
Add HMAC-SHA256 tag to all Request/Response messages.

BranLwyd:
Previously these were inconsistently applied between different messages.
I introduce a new `collect_auth_key`, distinct from the existing
`agg_auth_key`, for collect messages.

I think we could get away with only authenticating the request messages
(leaning on TLS to provide authentication for the responses); I'd be
amenable to that, but my understanding is that we don't want to rely on
the properties of the underlying channel, so I went ahead and specified
authentication tags for both request & response messages.

### DIFF
--- a/draft-gpew-priv-ppm.md
+++ b/draft-gpew-priv-ppm.md
@@ -493,9 +493,6 @@ of the aggregators is configured with following parameters:
   specified. See issue#161.]
 * `agg_auth_key`: A key for HMAC-SHA256 {{?RFC2104}} shared by the aggregators and
   used for message authentication during the aggregation flow.
-* `collect_auth_key`: A key for HMAC-SHA256 {{?RFC2104}} shared by the
-  aggregators and the collector and used for message authentication during the
-  collect flow.
 
 Finally, the collector is configured with the HPKE secret key corresponding to
 `collector_hpke_config`.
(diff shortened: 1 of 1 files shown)


draft-gpew-priv-ppm.md
```diff
@@ -493,6 +493,9 @@ of the aggregators is configured with following parameters:
   specified. See issue#161.]
 * `agg_auth_key`: A key for HMAC-SHA256 {{?RFC2104}} shared by the aggregators and
   used for message authentication during the aggregation flow.
+* `collect_auth_key`: A key for HMAC-SHA256 {{?RFC2104}} shared by the
+  aggregators and the collector and used for message authentication during the
+  collect flow.
```
cjpatton, line 498: This key would only need to be shared by the collector and leader. The helper doesn't need this key, does it?

cjpatton (changes requested):
Tagging the AggregateShareReq with the key shared by the aggregators makes sense. (The fact that this was missing was an oversight.)

However I think we should sync up before making changes to the collect flow. I envisioned using mutual TLS for this, Mozilla might have a different idea.

BranLwyd:
> Tagging the AggregateShareReq with the key shared by the aggregators makes sense. (The fact that this was missing was an oversight.)
> 
> However I think we should sync up before making changes to the collect flow. I envisioned using mutual TLS for this, Mozilla might have a different idea.

OK, I've removed the `tag` from the `Collect{Req,Resp}` for now. mTLS would work too, and I'll be happy to go with whatever is determined by group consensus.

divergentdave (approved):

cjpatton (approved):

PR #198, opened 2022-03-09
fix HPKE info strings and specify agg share handling

tgeoghegan:
The info strings used in HPKE context construction used the prefix
"pda", referring to the old "Private Data Aggregation" name of this
protocol. This commit also adds text instructing the collector how to
decrypt the aggregate shares it receives and to use the VDAF's
unsharding algorithm to obtain aggregate results.

### DIFF
--- a/draft-gpew-priv-ppm.md
+++ b/draft-gpew-priv-ppm.md
@@ -570,7 +570,7 @@ the aggregator by running
 
 ~~~
 enc, context = SetupBaseS(pk,
-                          "pda input share" || task_id || server_role)
+                          "ppm input share" || task_id || server_role)
 ~~~
 
 where `pk` is the aggregator's public key, `task_id` is `Report.task_id` and
@@ -773,7 +773,7 @@ following procedure:
 
 ~~~
 context = SetupBaseR(helper_share.enc, sk,
-                     "pda input share" || task_id || server_role)
+                     "ppm input share" || task_id || server_role)
 input_share = context.Open(nonce || extensions, helper_share)
 ~~~
 
@@ -861,7 +861,7 @@ public key as follows:
 
 ~~~
 enc, context = SetupBaseS(pk,
-                       "pda aggregate share" || task_id || server_role)
+                       "ppm aggregate share" || task_id || server_role)
 encrypted_agg_share = context.Seal(batch_interval, agg_share)
 ~~~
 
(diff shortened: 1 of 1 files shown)


tgeoghegan:
Question for the reviewers: VDAF defines labels like `vdaf-00 prio3` ([VDAF sharding spec](https://cjpatton.github.io/vdaf/draft-patton-cfrg-vdaf.html#section-6.2.2)). Should we include the draft version number in info strings like this? e.g. `ppm-01 input share`?

cjpatton (approved):
LGTM!

> Question for the reviewers: VDAF defines labels like `vdaf-00 prio3` ([VDAF sharding spec](https://cjpatton.github.io/vdaf/draft-patton-cfrg-vdaf.html#section-6.2.2)). Should we include the draft version number in info strings like this? e.g. `ppm-01 input share`?

I'd be in favor of this, though we might want to go with `ppm-00` since `00` will be the first draft of the WG doc when it gets adopted.

draft-gpew-priv-ppm.md
```diff
@@ -981,6 +981,21 @@ fetch certificates: https://datatracker.ietf.org/doc/html/rfc8555#section-7.4.2]
...
+
+~~~
+context = SetupBaseR(share.enc, sk,
+                     "ppm aggregate share" || task_id || server_role)
+input_share = context.Open(batch_interval, share.payload)
+~~~
+
+where `sk` is the HPKE secret key, `task_id` is `AggregateReq.task_id` and
+`server_role` is the role of the server that send the aggregate share (`0x01`
+for the leader and `0x00` for the helper). `batch_interval` is the `Interval`
```
BranLwyd, line 995: While this works, wouldn't it be more natural to specify `0x00` for the leader and `0x01` for the helper? This would permit a more natural extension to multiple helpers (which would be able to use `0x02`, `0x03`, ... without placing the leader between the first and second helpers) and would cause these role identifier numbers to match up with the ordering of the input shares in a client report, aggregate shares in an collect result, etc.

(I think this may be somewhat beyond the scope of this individual PR, is this an issue that has been discussed? I notice the interop PR specifies a Role message that has not only the leader & helper but also the collector & client.)

cjpatton: Yeah I think this would be a good change, but it should go in a different PR as you suggest.

tgeoghegan: I'd be open to back-porting the language around this from PR #179. There we put leader and helper as `0x02` and `0x03` so that we can allow further helpers to be `0x04` and above, and also extend the info string to include the roles of both sender and recipient. I think that change should be fairly non-controversial, and it'll make it easier to eventually land #179 if we chip little bits of it out.

cjpatton: Good catch, @tgeoghegan

BranLwyd: Backporting this from the interop PR when it is merged into main SGTM.

BranLwyd (approved):

chris-wood (approved):

PR #203, opened 2022-03-14
Interop: Error handling and editorial things

cjpatton:


draft-gpew-priv-ppm.md
```diff
@@ -820,28 +832,28 @@ been collected. This is the case if the aggregator is the leader and has
...
-{{aggregate-share-request}}). It also checks if the report has been observed.
+{{aggregate-share-request}}). It also checks if the report has been aggregated.
 This is the case if the report was used in a previous aggregate request (see
 {{aggregate-init-request}} and {{aggregate-request}}).
 
-* If the report has not yet been observed but is contained by a batch that has
-  been collected, then the aggregator fails with error `batch-collected`. This
-  prevents additional reports from being aggregated after its batch has already
-  been collected.
+* If the report has never been aggregated but is contained by a batch that has
```
BranLwyd, line 839: I'm not sure I agree with the change of wording from "observed" to "aggregated" here.

There will be a gap of time from when a client upload completes to when it is first included in an aggregation job; my interpretation would be that those reports were "observed" (or perhaps "received") but not yet "aggregated". I think we want this error logic to cover reports that have not yet been included in an aggregation job, right?

draft-gpew-priv-ppm.md
```diff
@@ -820,28 +832,28 @@ been collected. This is the case if the aggregator is the leader and has
...
+  fail with error `report-replayed`. This prevents a report from being used more
+  than once in a batch.
 
 Note that detecting whether a report has been replayed (i.e., it has been
-aggregated but not yet collected) requires each aggregator to store the set of
-nonces of reports that have been aggregated in uncollected batch intervals. So
-that the aggregator does not have to maintain this storage indefinitely, it MAY
+aggregated but not yet collected) requires each aggregator to store the nonces
+of reports that have been aggregated in uncollected batch intervals. So that the
+aggregator does not have to maintain this storage indefinitely, it MAY insetead
```
BranLwyd, line 851:
```suggestion
aggregator does not have to maintain this storage indefinitely, it MAY instead
```

(nit, typo fix)

draft-gpew-priv-ppm.md
```diff
@@ -854,29 +866,28 @@ input_share = context.Open(ReportShare.nonce || ReportShare.extensions,
...
-Otherwise, `out` is interpreted as follows. If the VDAF is 0-round, then `out`
-is the aggregator's output share, in which case the aggregator finishes and
-stores its output share for further processing as described in
-{{out-to-agg-share}}. Otherwise, if the VDAF consists of one round or more, then
-the aggregator continues and `out` is the aggregator's next VDAF message.
+If either step fails, then the aggregator MUST fail with error
+`vdaf-prep-error`. Otherwise, `out` is interpreted as follows. If the VDAF is
+0-round, then `out` is the aggregator's output share, in which case the
+aggregator finishes and stores its output share for further processing as
+described in {{out-to-agg-share}}. Otherwise, if the VDAF consists of one round
```
BranLwyd, line 887: This splitting of VDAF output into two parts is new to me -- how is this to be done? e.g. is the definition of how to split the output into a tuple part of the VDAF?

draft-gpew-priv-ppm.md
```diff
@@ -943,33 +954,33 @@ from the helper. Upon receiving it:
...
 ~~~
 
 If either of these operations fails, then the leader moves to state FAILED
 without sending a message to the helper. Otherwise it interprets `out` as
 follows. If this is the last round of VDAF preparation phase, then `out` is the
 leader's output share, in which case the leader finishes. Otherwise, it
-continues with `out` as its next VDAF message. Either way, it moves to state
-WAITING.
+interpresets `out` as the tuple `(new_state, outbound)`, where `new_state` is
+its new preparation state and `outobund` is its next VDAF messaage, and
```
BranLwyd, line 981:
```suggestion
its new preparation state and `outbound` is its next VDAF messaage, and
```

(nit, typo fix)

draft-gpew-priv-ppm.md
```diff
@@ -943,33 +954,33 @@ from the helper. Upon receiving it:
...
+continues wiht `outbound` as its next VDAF message. Either way, it moves to
```
BranLwyd, line 982:
```suggestion
continues with `outbound` as its next VDAF message. Either way, it moves to
```

(nit, typo fix)

draft-gpew-priv-ppm.md
cjpatton: Yeah, this is just a matter of aligning the notation with the current draft: https://cjpatton.github.io/vdaf/draft-patton-cfrg-vdaf.html#name-preparation

draft-gpew-priv-ppm.md
BranLwyd: That makes sense; my only note is that the client's Upload Request section (4.2.2) refers to this section as part of a `SHOULD`, so I suppose this logic may also apply to client upload endpoints. I think this is OK since the relevant bullet point (this one) is not meaningfully changed by the wording change. This would matter for the `report-replayed` point, but that's not mentioned by the client upload section for now; we might need to deal with it later but it's fine for now.

BranLwyd (approved):

draft-gpew-priv-ppm.md
```diff
@@ -1136,28 +1151,26 @@ struct {
...
 
-After verifying the authentication tag, the leader handles the AggregateResp as
-follows. It first checks that the sequence of Transition messages corresponds to
-the ReportShare sequence of the AggregateInitReq. If the messages appear out of
-order, any are missing, or a nonce is not recognized, then the leader MUST abort
-and alert its peer with error "XXX".
+The leader handles the AggregateResp as follows. It first verifies the tag. If
+this check fails, it MUST abort with error "invalidHmac".
+
+Next, it first checks that the sequence of Transition messages corresponds to
```
divergentdave, line 1162: nit
```suggestion
Next, it checks that the sequence of Transition messages corresponds to
```

draft-gpew-priv-ppm.md
```diff
@@ -1136,28 +1151,26 @@ struct {
...
+the ReportShare sequence of the AggregateInitReq. If any message appears out of
+order or is missing, or a transition has an unrecognized nonce, then the leader
```
divergentdave, line 1164: I believe that "or is missing" conflicts with section 4.4.4.1, assuming a multi-round VDAF. That section says that the leader will filter out reports that correspond to a state of FAILED, in which case the next AggregateContinueReq would have fewer transitions/nonces/reports than the AggregateInitReq and AggregateResp that came before it.

draft-gpew-priv-ppm.md
```diff
@@ -450,6 +448,19 @@ struct {
...
+
+This documented uses the verb "abort with error ERROR", where "ERROR" is an
+error type enumerated in {{error-types}}, to prescribe the behavior of a
+protocol participant when it encounters a fatal error:
+
+* If a fatal error is encountered while processing an HTTP request, the HTTP
+  server responds with an HTTP error status as described in {{errors}}.
+
+* If a fatal error is encountered while processing an HTTP request, the HTTP
+  client... [TODO: Figure out what behavior is called for here. An HTTP request
```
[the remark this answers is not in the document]
cjpatton: Yes, good catch.

draft-gpew-priv-ppm.md
```diff
@@ -943,33 +954,33 @@ from the helper. Upon receiving it:
...
 If either of these operations fails, then the leader moves to state FAILED
 without sending a message to the helper. Otherwise it interprets `out` as
 follows. If this is the last round of VDAF preparation phase, then `out` is the
 leader's output share, in which case the leader finishes. Otherwise, it
-continues with `out` as its next VDAF message. Either way, it moves to state
-WAITING.
+interpresets `out` as the tuple `(new_state, outbound)`, where `new_state` is
+its new preparation state and `outbound` is its next VDAF messaage, and
+continues with `outbound` as its next VDAF message. Either way, it moves to
+state WAITING.
```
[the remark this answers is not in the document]
cjpatton: For the leader, "WAITING" means I'm waiting for the helper's Transition before either (1) I can compute my next VDAF message or (2) I commit to the output share I've just computed. Does that make sense or am I misunderstanding something?

draft-gpew-priv-ppm.md
cjpatton: Hmm, maybe we just need to sharpen this a little. This paragraph pertains to the leader processing `AggregateResp`. The helper is not allowed to change the sequence of nonces relative to the previous `AggregateReq`. In particular, if a nonce is missing (relative to the previous request), then we abort.

draft-gpew-priv-ppm.md
cjpatton: Hmm, interesting. We should to try to work this out now, if possible. I think what we want to say that, before a batch is collected, two reports with the same nonce MUST NOT be aggregated twice. Furthermore, the leader SHOULD report to the client if it attempted to upload a report with a previously observed nonce.

draft-gpew-priv-ppm.md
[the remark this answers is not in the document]
cjpatton: Yw, maybe we can be describing the state machine better...

tgeoghegan (commented):
While implementing upload today, I also noticed some inconsistencies between the handling of reports in the leader during the upload protocol and the helper

PR #204, opened 2022-03-15
Extend HPKE info strings

tgeoghegan:
When constructing HPKE contexts, we now include both sender and
recipient roles, and the application label also includes the PPM
version. I used `ppm-02` here because we recently submitted draft-01 of
`draft-gpew-priv-ppm` and so the next submitted draft, which would include
this change, will be `-02`. I also backported the definition of `struct
HpkeCiphertext` from the interop PR because it makes everything tidier
and should be non-controversial we can get a head-start on bringing those
changes into main.

### DIFF
--- a/draft-gpew-priv-ppm.md
+++ b/draft-gpew-priv-ppm.md
@@ -428,6 +428,24 @@ struct {
   Time time;
   uint64 rand;
 } Nonce;
+
+/* The various roles in the PPM protocol. */
+enum {
+  collector(0),
+  client(1),
+  leader(2),
+  helper(3),
+} Role;
+
+/* Identifier for a server's HPKE configuration */
+uint8 HpkeConfigId;
+
+/* An HPKE ciphertext. */
+struct {
+  HpkeConfigId config_id;    // config ID
+  opaque enc<1..2^16-1>;     // encapsulated HPKE context
+  opaque payload<1..2^16-1>; // ciphertext
+} HpkeCiphertext;
 ~~~
 
 ## Task Configuration {#task-configuration}
@@ -484,7 +502,6 @@ struct {
   HpkePublicKey public_key;
 } HpkeConfig;
 
-uint8 HpkeConfigId;
 opaque HpkePublicKey<1..2^16-1>;
 uint16 HpkeAeadId; // Defined in I-D.irtf-cfrg-hpke
 uint16 HpkeKemId;  // Defined in I-D.irtf-cfrg-hpke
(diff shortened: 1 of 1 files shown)


tgeoghegan:
> There's one more instance of `EncryptedInputShare.config_id` that needs to be changed to `HpkeCiphertext.config_id`. Otherwise, LGTM.

Fixed, thank you! I also rewrote the relevant paragraph to acknowledge that there's two encrypted input shares and that the leader should examine the HPKE config ID of the one intended for itself. Further, I recall @cjpatton pointing out to me that should this draft get adopted by the WG, we'll reset to something like `draft-authors-ppm-ppm-01`, so `ppm-01` is the more appropriate application label.

draft-gpew-priv-ppm.md
```diff
@@ -547,55 +564,45 @@ This message is called the client's *report*. It contains the following fields:
...
-
 To generate the report, the client begins by sharding its measurement into a
 sequence of input shares as specified by the VDAF in use. To encrypt an input
 share, the client first generates an HPKE {{!I-D.irtf-cfrg-hpke}} context for
 the aggregator by running
 
 ~~~
-enc, context = SetupBaseS(pk,
-                          "ppm input share" || task_id || server_role)
+enc, context = SetupBaseS(pk, Report.task_id || "ppm-01 input share" ||
```
cjpatton, line 573: Assuming this gets adopted by the WG, the first draft will be "draft-ietf-ppm-00" (or something like that). If you take this suggestion, don't forget to apply it below.

```suggestion
enc, context = SetupBaseS(pk, Report.task_id || "ppm-00 input share" ||
```

draft-gpew-priv-ppm.md
```diff
@@ -772,13 +779,13 @@ this means. See issue#57.] Next, it attempts to decrypt the payload with the
...
+context = SetupBaseR(helper_share.enc, sk, task_id ||
+                     "ppm-01 input share" || 0x01 || server_role)
 input_share = context.Open(nonce || extensions, helper_share)
 ~~~
 
 where `sk` is the HPKE secret key, `task_id` is `AggregateReq.task_id` and
-`server_role` is the role of the server (`0x01` for the leader and `0x00` for
+`server_role` is the role of the server (`0x02` for the leader and `0x03` for
 the helper). `nonce` and `extensions` are obtained from the
 corresponding fields in `AggregateSubReq`. If decryption fails, then the
```
cjpatton, line 790: Might be good to remind the reader here that "0x01" stands for the client.

cjpatton (commented):
Just a few minor things.

cjpatton (approved):

PR #205, opened 2022-03-15
Define leader-to-helper abort procedure.

branlwyd:
This solution optimizes for the size of the specification edit; if it
works out well, we should clean it up & attempt to unify this abort
procedure with the helper-to-leader abort procedure.

### DIFF
--- a/draft-gpew-priv-ppm.md
+++ b/draft-gpew-priv-ppm.md
@@ -458,8 +458,13 @@ protocol participant when it encounters a fatal error:
   server responds with an HTTP error status as described in {{errors}}.
 
 * If a fatal error is encountered while processing an HTTP response, the HTTP
-  client... [TODO: Figure out what behavior is called for here. An HTTP request
-  should be sent that conveys the error.]
+  client sends a followup "abort" request with a message specific to the particular
+  interaction. (For example, in an aggregation flow, the leader would send an
+  `AggregateAbortReq` as described in {{aggregate-message-auth}}.)
+
+[TODO: consider unifying the format of the request/response for errors, or
+providing a unified way for HTTP clients to transmit errors for all interaction
+types]
 
 ## Task Configuration {#task-configuration}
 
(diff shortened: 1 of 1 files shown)


cjpatton:
So I had a thought about this this morning.  I'm questioning whether we actually need a mechanism for the Leader->Heper alert. IIUC the motivation for this signal that we discussed is that helper either (1) needs to tear down the HTTP session and all associated state or (2) “revert” any output shares it just accepted.

For (1) I wonder if a sufficient signal is that the HTTP session has been torn down, but the aggregation run isn’t complete.

We’re intentionally leaving (2) out of scope for now. Regardless, I think we would handle it similar to what we're doing here, but we don't need to do it now. What do y'all think?

BranLwyd:
> I'm questioning whether we actually need a mechanism for the Leader->Heper alert.

I think we can "do nothing" & ignore leader->helper alerts as long as we're OK with the helper never finding out a given aggregation job has been aborted. I don't think that causes problems with the spec & I don't think it will cause serious problems in implementations -- it seems like, at worst, we'd end up with the helper holding on to its aggregation data longer than needed.

cjpatton:
What do @tgeoghegan and @divergentdave think?

tgeoghegan:
What we discussed yesterday is that in the scope of the interop target, the only thing a helper would do with this alert is log it. I'm comfortable with not doing anything for now, on the premise that the aggregate share checksums will let us detect mismatches and we'll have a tight communication loop between aggregator operators.

I do think we should keep thinking about whether we need  a Leader->Helper `AggregateCommit` message, but part of the point of the interop experiment is to find out how valuable that would be.

If you need my vote one way or the other: I say let's do nothing for interop.

divergentdave:
Doing nothing is fine with me. Our underlying motivation is to resolve the issue that "alert its peer" is not implementable in the context of a leader processing an HTTP response from a helper, and I think not sending the error message to the helper at all would be a fine way to solve this for now. The only sensible action the helper could take is to delete some of its stored state. We can address deletion of saved prepare messages in the spec at a later point, when we address adding a "time to live" concept for reports, etc.

BranLwyd:
Doing nothing is fine for me too; retracting this PR, I'll leave it as a draft for a bit in case we want to pull something out of it in the future.

Issue #146, opened 2021-08-31
How should servers figure out which task a message pertains to?


tgeoghegan:
Aggregators are expected to have an endpoint `hpke_config` from which they vend the HPKE public key and parameters so that clients may encrypt the contents of reports. There are no parameters to the `hpke_config` request, so there is no opportunity for an aggregator to provide a different config based on the task in play.

Now, the `hpke_config` endpoint is relative to the aggregator's base endpoint URL, obtained from the per-task parameters, so an aggregator could simply provide base URLs that have the task ID in their path and use different key material on that basis. But all the other protocol messages like `Report` or `CollectReq` explicitly include task ID, suggesting that aggregators are supposed to be able to service multiple tasks from a single endpoint. Something feels inconsistent there.

tgeoghegan:
Just for consistency, I think we could change the endpoint to `[aggregator]/hpke_config/<task_id>`. We would probably have to redefine task ID to make sure it's safe to use in URL construction (URL encoded byte strings look terrible).

cjpatton:
Good catch, and I like your suggested fix.

tgeoghegan:
I could swear we talked about this a long time ago. Anyway, I am putting this in `parking-lot` for now because I don't think any progress is blocked on it, and it strikes me as the kind of thing that we will go back and forth on a lot during the IETF process.

tgeoghegan:
I think the real issue here is that when handling a message, collectors, leaders and helpers need to know which task it pertains to in order to select the appropriate HPKE context (among other parameters). We should decide whether the task ID appears in protocol messages or whether it can be inferred from the endpoint and be consistent about it.

BranLwyd:
nit: IMO it would be more consistent to put the task ID in a `/hpke_config` request into the request body -- all other request/response pairs include the task ID in the request body, rather than including it in the request path.

(There's no technical downside to putting the task ID in the path, this is only a "consistency" concern.)

divergentdave:
I think the issue with putting the task ID for a /hpke_config request in the body is that we have to switch from GET to POST, and lose cache-ability.

BranLwyd:
That's a good point -- it's still pretty unsatisfying that this one request/response pair will have to be structured in this way, while all other request/response pairs (Upload, Aggregate, AggregateShare, Collect) use a different structuring.

However, given that some clients will literally be web browsers, where HTTP caching is "free", I think it's worthwhile to keep HTTP caching semantics working.

[...]

PR #212, opened 2022-04-04
add a `task_id` query parameter for `hpke_config`

tgeoghegan:
# This PR targets the interop target branch

All protocol request messages contain a `task_id` field, which allows
aggregators to service multiple tasks from a single set of endpoints.
The exception is `hpke_config`: since it is an HTTP GET request, there
is no body, and thus no place for the client to indicate the desired
`task_id` to the aggregator. We don't want to introduce a request body
to `hpke_config`, because then it would have to be a POST instead of a
GET, and that would make it harder for implementations to cache HPKE
config values. So instead we have clients encode the task ID in RFC 4648
Base16 (a.k.a. hexadecimal) and provide it as a query parameter.

See issue #146

### DIFF
--- a/draft-gpew-priv-ppm.md
+++ b/draft-gpew-priv-ppm.md
@@ -389,8 +389,9 @@ not listed in the appropriate IANA registry (see {{ppm-urn-space}}). Clients
 SHOULD display the "detail" field of all errors. The "instance" value MUST be
 the endpoint to which the request was targeted. The problem document MUST also
 include a "taskid" member which contains the associated PPM task ID, encoded
-with base64 using the standard alphabet {{!RFC4648}} (this value is always
-known, see {{task-configuration}}).
+in Base 64 using the URL and filename safe alphabet with no padding as defined
+in sections 5 and 3.2 of {{!RFC4648}} (this value is always known, see
+{{task-configuration}}).
 
 In the remainder of this document, we use the tokens in the table above to refer
 to error types, rather than the full URNs. For example, an "error of type
@@ -480,7 +481,8 @@ with which all of the parties are configured:
   endpoints can be found. Each endpoint's list MUST be in the same order. The
   leader's endpoint MUST be the first in the list. The order of the
   `encrypted_input_shares` in a `Report` (see {{uploading-reports}}) MUST be the
-  same as the order in which aggregators appear in this list.
+  same as the order in which aggregators appear in this list. Aggregators MAY
+  use the same endpoint for multiple tasks.
 * `max_batch_lifetime`: The maximum number of times a batch of reports may be
   used in collect requests.
 * `min_batch_size`: The minimum number of reports that appear in a batch.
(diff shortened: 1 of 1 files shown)


tgeoghegan:
My goal here is to enable our aggregator implementations in the interop target to use one HPKE config per task. The alternative to this would be for us to use aggregator endpoints that somehow incorporate the task ID, as I described [here](https://github.com/divviup/janus/issues/48#issuecomment-1085209198), which is already permitted by the spec as written.

BranLwyd (approved):

cjpatton (commented):
SGTM, but why not base64 encode the taskID for consistency with https://github.com/abetterinternet/ppm-specification/pull/208?

tgeoghegan:
> SGTM, but why not base64 encode the taskID for consistency with #208?

The Base64 alphabet includes `/`, `+` and `=`, which means you have to URL encode (a.k.a. percent encode) Base64 strings before you can use them in a URL, which is a chore and less readable. An alternative would be to use RFC 4648's [`base64url` encoding](https://datatracker.ietf.org/doc/html/rfc4648#section-5).

cjpatton:
> > SGTM, but why not base64 encode the taskID for consistency with #208?
> 
> The Base64 alphabet includes `/`, `+` and `=`, which means you have to URL encode (a.k.a. percent encode) Base64 strings before you can use them in a URL, which is a chore and yields less readable URLs (not that 64 character hex strings is especially user-friendly). An alternative would be to use RFC 4648's [`base64url` encoding](https://datatracker.ietf.org/doc/html/rfc4648#section-5).

Gotcha. I think the main thing is to be consistent. I would either:

1. Change this PR to use base64url; or
2. Tack on a commit that makes all encodings of the taskID (except those that appear in the body, of course) the same (i.e., hexadecimal).

I think (2.) would be my preference.

tgeoghegan:
I agree with the consistency argument. I'm leaning towards using `base64url` everywhere because it's more compact than Base16/hex, and also I think implementations are less likely to be surprised by case requirements in base64url (i.e., it's obvious that base64 is case sensitive, but you have to read RFC 4648 attentively to notice that it mandates uppercase hex strings).

cjpatton:
Base64Url sounds good to me. I took a quick look and it appears that most standard base64 libraries also support the URL encoding.

tgeoghegan:
I will clean this up and merge it once I have implemented it in Janus and convinced myself it's sound.

draft-gpew-priv-ppm.md
```diff
@@ -517,9 +519,17 @@ individual shares to each helper.
 
 Before the client can upload its report to the leader, it must know the public
 key of each of the aggregators. These are retrieved from each aggregator by
-sending a request to `[aggregator]/hpke_config`, where `[aggregator]` is the
-aggregator's endpoint URL, obtained from the task parameters. The aggregator
-responds to well-formed requests with status 200 and an `HpkeConfig` value:
+sending a request to `[aggregator]/hpke_config?task_id=[task-id]`, where
+`[aggregator]` is the aggregator's endpoint URL, obtained from the task
+parameters, and `[task-id]` is the task ID obtained from the task parameters,
+encoded in Base 64 with URl and filename safe alphabet with no padding, as
```
divergentdave, line 525: FYI, typo here
```suggestion
encoded in Base 64 with URL and filename safe alphabet with no padding, as
```

PR #219, opened 2022-04-18
First cut of editorial changes in prep for aggregation flow

chris-wood:
Extracted and slightly modified from chris-wood-cjpatton/agg-flow.

### DIFF
--- a/draft-gpew-priv-ppm.md
+++ b/draft-gpew-priv-ppm.md
@@ -378,6 +378,7 @@ in the "type" field (within the PPM URN namespace "urn:ietf:params:ppm:error:"):
 | unrecognizedMessage     | The message type for a response was incorrect or the payload was malformed. |
 | unrecognizedTask        | An endpoint received a message with an unknown task ID. |
 | outdatedConfig          | The message was generated using an outdated configuration. |
+| batchInvalid            | A collect or aggregate-share request was made with invalid batch parameters. |
 
 This list is not exhaustive. The server MAY return errors set to a URI other
 than those defined above. Servers MUST NOT use the PPM URN namespace for errors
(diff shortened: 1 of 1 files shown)


chris-wood:
(Build failure appears to be some GitHub API issue.)

divergentdave:
Ah, this looks like we're hitting the patch for https://github.blog/2022-04-12-git-security-vulnerability-announced/. I'll check if the upstream template repository has a fix.

chris-wood:
Thanks @divergentdave!

cjpatton (approved):
🚢 !

draft-gpew-priv-ppm.md
```diff
@@ -467,19 +468,27 @@ number generator. Each task has the following parameters associated with it:
...
-  discussion.
 * `max_batch_lifetime`: The maximum number of times a batch of reports may be
   used in collect requests.
 * `min_batch_size`: The minimum number of reports that appear in a batch.
 * `min_batch_duration`: The minimum time difference between the oldest and
   newest report in a batch. This defines the boundaries with which the batch
   interval of each collect request must be aligned. (See
   {{batch-parameter-validation}}.)
-* `protocol`: named parameter identifying the VDAF scheme in use.
+* A unique identifier for the VDAF being used for the task.
```
tgeoghegan, line 478: I think we need a bit more than this. Suppose that a PPM task executes [`Prio3Aes128Histogram`](https://github.com/divviup/libprio-rs/blob/1bd54185d01110ff4486b8d71ed8f17f1ea77b78/src/vdaf/prio3.rs#L123). Besides that label, everyone also needs to know what the bucket boundaries are. Perhaps the PPM text should refer to VDAF's description of a [`Measurement` type](https://github.com/cjpatton/vdaf/blob/6904d4a6295da82daf7ecd09c094326db4fb8ccd/draft-patton-cfrg-vdaf.md?plain=1#L404)?

chris-wood, line 478:
```suggestion
* A unique identifier for the VDAF instance used for the task, including the type of measurement associated with the task.
```

chris-wood: I think this needs to be "something that fully defines the VDAF instance you're using, modulo the parameters generated from VDAF.setup() (since those change per invocation)," so I elaborated on this a bit.

PR #223, opened 2022-04-21
Add VDAF mapping for aggregation flow

chris-wood:
This wires up PPM to drive the underlying VDAF state machine for individual reports. It splits the aggregate flow into three phases: initialization, progression (where the aggregators exchange messages to perform validation), and finalization. It also moves the process of fetching the aggregate share value for each aggregator to the collect flow, since that seems more closely aligned with the process of collection than it is about aggregation.

Open question (there may be more): Can we come up with a better name than `Process` for the message that conveys VDAF intermediate results?

### DIFF
--- a/draft-gpew-priv-ppm.md
+++ b/draft-gpew-priv-ppm.md
@@ -379,6 +379,7 @@ in the "type" field (within the PPM URN namespace "urn:ietf:params:ppm:error:"):
 | unrecognizedTask        | An endpoint received a message with an unknown task ID. |
 | outdatedConfig          | The message was generated using an outdated configuration. |
 | batchInvalid            | A collect or aggregate-share request was made with invalid batch parameters. |
+| invalidReportShare      | The report share was invalid and could not be processed. |
 
 This list is not exhaustive. The server MAY return errors set to a URI other
 than those defined above. Servers MUST NOT use the PPM URN namespace for errors
@@ -615,9 +616,9 @@ retrying.
 
 The leader MUST ignore any report whose nonce contains a timestamp that falls in
 a batch interval for which it has received at least one collect request from the
-collector. (See {{pa-collect}}.) Otherwise, comparing the aggregate result to
+collector. (See {{collect-flow}}.) Otherwise, comparing the aggregate result to
 the previous aggregate result may result in a privacy violation. (Note that the
-helpers enforce this as well; see {{aggregate-request}}.) In addition, the
+helpers enforce this as well; see {{collect-flow}}.) In addition, the
 leader SHOULD abort the upload protocol and alert the client with error
 "staleReport".
 
@@ -645,18 +646,24 @@ encoded value of the following form:
 "extension_type" indicates the type of extension, and "extension_data" contains
 information specific to the extension.
 
+### Leader State {#leader-state}
+
+The leader is required to buffer reports while waiting to aggregate them. The
+leader SHOULD NOT accept reports whose timestamps are too far in the future.
+Implementors MAY provide for some small leeway, usually no more than a few
+minutes, to account for clock skew.
 
 ## Verifying and Aggregating Reports {#pa-aggregate}
 
 Once a set of clients have uploaded their reports to the leader, the leader can
 send them to the helpers to be verified and aggregated. In order to enable the
 system to handle very large batches of reports, this process can be performed
-incrementally. To aggregate a set of reports, the leader sends an AggregateReq
-to each helper containing those report shares. The helper then processes them
-(verifying the proofs and incorporating their values into the ongoing aggregate)
-and replies to the leader.
+incrementally. To aggregate a set of reports, the leader sends a message to each
+helper containing those report shares. The helper then processes them (verifying
+the proofs and incorporating their values into the ongoing aggregate) and replies
+to the leader.
 
-The exact structure of the aggregation flow depends on the VDAF.  Specifically:
+The exact structure of the aggregation flow depends on the VDAF. Specifically:
 
 * Some VDAFs (e.g., `prio3`) allow the leader to start aggregating reports
   proactively before all the reports in a batch are received. Others (e.g.,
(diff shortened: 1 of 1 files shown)


draft-gpew-priv-ppm.md
```diff
@@ -672,242 +679,366 @@ independently.
...
+
+Let `[aggregator]` denote the helper's API endpoint. The leader sends a POST
+request to `[aggregator]/aggregate` with its AggregateInitReq message as
+the payload. The media type is "message/ppm-aggregate-init-req".
+
+To process the leader's request, the helper first checks that the nonces in
+`AggregateInitReq.seq` are all distinct. If two InputReportShare messages have the
+same nonce, then the helper MUST abort with error "unrecognizedMessage".
+
+Next, the helper initializates the preparation state for each report share in
```
ekr, line 802: How does it do that? Show the API call.

draft-gpew-priv-ppm.md
```diff
@@ -672,242 +679,366 @@ independently.
...
+process of aggregating this candidate set. This message is structured as follows:
 
 ~~~
 struct {
+  TaskID task_id;
+  opaque agg_param<0..2^16-1>;
   opaque helper_state<0..2^16>;
-  AggregateSubResp seq<1..2^24-1>;
-} AggregateResp;
+  ReportShare seq<1..2^16-1>;
```
ekr, line 782: Can we rename this to "reports"?

draft-gpew-priv-ppm.md
```diff
@@ -672,242 +679,366 @@ independently.
...
-The helper handles each sub-request `AggregateSubReq` as follows. It first looks
-up the HPKE config and corresponding secret key associated with
-`helper_share.config_id`. If not found, then the sub-response consists of an
-"unrecognized config" alert. [TODO: We'll want to be more precise about what
-this means. See issue#57.] Next, it attempts to decrypt the payload with the
-following procedure:
+Processing a report share begins by first checking if the report has already
+been aggregated. This is the case if the report was used in a previous aggregate
+request and is therefore a replay. In this case, the helper marks the report as
+invalid with the `report-replayed` error. Note that detecting whether a report
```
ekr, line 820: Can we just merge this with the uniquness check above?

draft-gpew-priv-ppm.md
```diff
@@ -672,242 +679,366 @@ independently.
...
 ~~~
-context = SetupBaseR(helper_share.enc, sk, task_id ||
-                     "ppm-00 input share" || 0x01 || server_role)
-input_share = context.Open(nonce || extensions, helper_share)
+context = SetupBaseR(encrypted_input_share.enc, sk, task_id ||
+                     "ppm-00 input share" || 0x01 || 0x03)
+
+input_share = context.Open(ReportShare.nonce || ReportShare.extensions,
+                           ReportShare.encrypted_input_share.payload)
 ~~~
```
ekr, line 843: See above about merging this.

draft-gpew-priv-ppm.md
```diff
@@ -672,242 +679,366 @@ independently.
...
+struct {
+  opaque helper_state<0..2^16>;
+  ProcessShare seq<1..2^16-1>;
+} AggregateInitResp;
+~~~
 
-Then, for each aggregator endpoint `[aggregator]` in the parameters associated
-with `CollectReq.task_id` (see {{pa-collect}}) except its own, the leader sends
-a POST request to `[aggregator]/aggregate_share` with the following message:
+The `helper_state` parameter contains the helper's initial state. This is an
```
ekr, line 898: This section has gotten very long.

draft-gpew-priv-ppm.md
```diff
@@ -672,242 +679,366 @@ independently.
...
+as finished if the VDAF computation is finished.
+
+The helper's response to the leader is an HTTP 200 OK whose body is the
+AggregateInitResp and media type is "message/ppm-aggregate-init-resp".
+
+Upon receipt of a helper's AggregateInitResp message, the leader checks that the
+sequence of ProcessShare messages corresponds to the ReportShare sequence of the
+AggregateInitReq. If any message appears out of order, is missing, has an
+unrecognized nonce, or if two messages have the same nonce, then the leader MUST
+abort with error "unrecognizedMessage".
```
ekr, line 920: Why are we both having order and the nonce? Seems like a great opportunity for things to go wrong.

draft-gpew-priv-ppm.md
```diff
@@ -672,242 +679,366 @@ independently.
...
+### Aggregate Continuation {#agg-continue-flow}
+
+In the continuation phase, the leader drives aggregation of each share in the candidate
+report set until the underlying VDAF moves into a terminal state, which happens either
+when aggregation completes or an error occurs. The leader does this by repeating the
+following procedure.
+
+* If the helper failed, then mark the report as failed, remove it from the candidate
+  report set, and do not process it further.
+* If the leader continued but the helper finished, or if the leader finished but
```
ekr, line 931: Can we get a forward ref to how things ifnish.

draft-gpew-priv-ppm.md
```diff
@@ -672,242 +679,366 @@ independently.
...
+  the helper continued, then mark the report as failed and do not process it further.
+* If the leader and helper finished, then continuation is complete. At this point the
+  report has been completely processed and both aggregators have recovered an
+  output share. The leader proceeds as described in {{agg-complete}}.
+* If the leader and helper continued, then the leader computes its next state
+  transition ("prep_next") and sends the helper a ProcessShare message with the
```
ekr, line 937: Where is this defined?

draft-gpew-priv-ppm.md
```diff
@@ -672,242 +679,366 @@ independently.
...
+~~~
+
+where [leader_outbound, helper_outbound] is a vector of two elements. The next state
+transition is then computed as:
+
+~~~
+out = VDAF.prep_next(prep_state, inbound)
+~~~
+
+If either of these operations fails, then the leader marks the report as invalid.
```
ekr, line 956: If it's invalid does it have to tell the helper?

draft-gpew-priv-ppm.md
```diff
@@ -958,8 +1089,91 @@ responds with HTTP status 202 Accepted. The response MAY include a Retry-After
...
+} AggregateShareReq;
+~~~
+
+* `task_id` is the task ID associated with the PPM parameters.
+* `batch_interval` is the batch interval of the request.
+* `report_count` is the number of reports included in the aggregation.
+* `checksum` is the checksum computed over the set of client reports. The
+  checksum is computed by taking the SHA256 hash of each nonce from the client
+  reports included in the aggregation, then combining the hash values with a
+  bitwise-XOR operation.
```
ekr, line 1120: Why are we including the nonces and not the values?

draft-gpew-priv-ppm.md
```diff
@@ -958,8 +1089,91 @@ responds with HTTP status 202 Accepted. The response MAY include a Retry-After
...
+{{agg-complete}}, obtaining an opaque byte string `agg_share` whose
+structure is specific to the VDAF in use. The helper encrypts the
+`agg_share` as follows:
+
+~~~
+enc, context = SetupBaseS(pk, AggregateShareReq.task_id ||
+                              "ppm-00 aggregate share" || 0x03 || 0x00)
+
+encrypted_agg_share = context.Seal(AggregateShareReq.batch_interval,
+                                   agg_share)
```
ekr, line 1144: Here too, let's break out the crypto.

ekr (changes requested):
Technically this seems mostly OK. Editorially, I think it needs a lot of restructure to make it clear. At a higher level, it seems like the leader sends AggregateInitReq prior to engaging with the VDAF to compute the initial state? If so, I would restructure the text to show things in parallel. With that said, this is also a missed opportunity to save a round trip in some cases, I imagine

draft-gpew-priv-ppm.md
cjpatton: Our thinking here is that the leader would just skip reports that can't be processed further. The motivation for this is to keep the helper simple. It also has the advantage of avoiding communication overhead for reports that can't be aggregated. One potential downside is that the helper doesn't get to find out *why* processing failed --- VDAF evaluation may fail in different ways, for example --- whereas the leader does.

draft-gpew-priv-ppm.md
```diff
@@ -672,242 +679,366 @@ independently.
...
+~~~
+agg_share = VDAF.out_shares_to_agg_share(agg_param, out_shares)
+~~~
 
-## Collecting Results {#pa-collect}
+Note that for most VDAFs, it is possible to aggregate output shares as they
+arrive rather than wait until the batch is collected. To do so however, it is
+necessary to enforce the batch parameters as described in
+{{batch-parameter-validation}} so that the aggregator knows which aggregate
+share to update.
```
tgeoghegan, line 1035: This is kind of awkward. The implication is that batch parameters don't need to be validated when using a VDAF without an aggregation parameter. Also I don't understand how validation of batch parameters informs which aggregate is used?

draft-gpew-priv-ppm.md
```diff
@@ -958,8 +1089,91 @@ responds with HTTP status 202 Accepted. The response MAY include a Retry-After
...
 Once all the necessary reports have been prepared, the leader obtains the
-helper's encrypted aggregate share for the batch interval by sending an
-`AggregateShareReq` to the helper as described in {{aggregate-share-request}}.
+helper's encrypted aggregate share for the batch interval by requesting an
+aggregate share from each helper.
+
+To do this, the leader first computes a checksum over the set of output shares
+included in the batch window. The checksum is computed by taking the SHA256 hash of
+each nonce from the client reports included in the aggregation, then combining the
+hash values with a bitwise-XOR operation.
```
tgeoghegan, line 1098: This is identical to the definition of `checksum`, a couple paragraphs later. We should delete one.

draft-gpew-priv-ppm.md
```diff
@@ -958,8 +1089,91 @@ responds with HTTP status 202 Accepted. The response MAY include a Retry-After
...
+`encrypted_aggregate_share.config_id` is set to the collector's HPKE config ID.
+`encrypted_aggregate_share.enc` is set to the encapsulated HPKE context `enc`
+computed above and `encrypted_aggregate_share.ciphertext` is the ciphertext
+`encrypted_agg_share` computed above.
+
+After receiving the helper's response, the leader uses the HpkeCiphertext to
+respond to a collect request (see {{collect-flow}}).
+
+The leader MAY make multiple aggregate-share requests for a given batch interval
+and aggregation parameter MUST get the same result each time.
```
tgeoghegan, line 1170:
```suggestion
The leader MAY make multiple aggregate-share requests for a given batch interval
and aggregation parameter and MUST get the same result each time.
```
The intent here is that the helper should use the same set of reports to service multiple aggregate-share requests, right? Could that be stated explicitly.

draft-gpew-priv-ppm.md
```diff
@@ -379,6 +379,7 @@ in the "type" field (within the PPM URN namespace "urn:ietf:params:ppm:error:"):
 | unrecognizedTask        | An endpoint received a message with an unknown task ID. |
 | outdatedConfig          | The message was generated using an outdated configuration. |
 | batchInvalid            | A collect or aggregate-share request was made with invalid batch parameters. |
+| invalidReportShare      | The report share was invalid and could not be processed. |
```
tgeoghegan, line 382: This error type doesn't seem to be used anywhere.

tgeoghegan (changes requested):

draft-gpew-priv-ppm.md
```diff
@@ -672,242 +679,366 @@ independently.
...
-[OPEN ISSUE: Should there be an indication of whether a given AggregateReq is a
+[OPEN ISSUE: Should there be an indication of whether a given aggregate request is a
 continuation of a previous sub-batch?]
 
-[TODO: Decide if and how the collector's request is authenticated.]
+The aggregation flow can be thought of as having three phases:
 
+- Initialization: Begin the aggregation flow by sharing report shares with each
+  helper. Each aggregator, including the leader, initializates the underlying
+  VDAF instance using these report shares and the public VDAF configured for
```
tgeoghegan, line 713: What is a "public VDAF"?

draft-gpew-priv-ppm.md
chris-wood: The intent here is that if the parameters of the aggregation request are the same then the output should be the same.

draft-gpew-priv-ppm.md
```diff
@@ -672,242 +679,366 @@ independently.
...
-This encryption prevents the leader from learning the actual result, as it only
-has its own share and not the helper's share, which is encrypted for the
-collector. The helper responds to the leader with HTTP status 200 OK and a body
-consisting of an `HpkeCiphertext` where `config_id` is set to the collector's
-HPKE config ID, `enc` is set to the encapsulated HPKE context `enc` and
-`ciphertext` is `encrypted_agg_share`.
+Once processing of a report share is finished, each aggregator stores the the
+recovered output shares until the batch to which it pertains is collected.
+To aggregate the output shares, denoted `out_shares`, the aggregator runs
+the aggregation algorithm specified by the VDAF:
```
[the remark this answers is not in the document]
chris-wood: > "output share" should be singular to agree with "report share" on the line above, right?

Yep!

> I've argued before we should spell out the state machine for report shares and for batches. That shouldn't happen in this change, but since a lot of subtle stuff like anti-replay depends on the state of a report vs. the state of a batch, the document would benefit from more clarity about this.

I don't think we have a clear set of requirements for batch validation yet (https://github.com/ietf-wg-ppm/ppm-specification/issues/195), but once we do, yes, we should definitely add more detail.

draft-gpew-priv-ppm.md
chris-wood: Leftover 👍

draft-gpew-priv-ppm.md
chris-wood: Strange leftover. Deleting "public".

draft-gpew-priv-ppm.md
```diff
@@ -672,242 +679,366 @@ independently.
...
+The aggregation flow can be thought of as having three phases:
 
+- Initialization: Begin the aggregation flow by sharing report shares with each
+  helper. Each aggregator, including the leader, initializates the underlying
+  VDAF instance using these report shares and the public VDAF configured for
+  the corresponding measurement task.
+- Continuation: Continue the aggregation flow by exchanging messages produced
+  by the underlying VDAF instance until aggregation completes or an error occurs.
+- Completion: Finish the aggregate flow, yielding an aggregate share corresponding
+  to all output shares in the batch.
```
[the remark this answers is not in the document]
chris-wood: Yep, good catch. This should be "output shares."

draft-gpew-priv-ppm.md
```diff
@@ -672,242 +679,366 @@ independently.
...
+from the set of candidate reports. Moreover, if the report has never been aggregated
+but is contained by a batch that has been collected, then the aggregator marks
+this report as invalid and removes it from the set of candidate reports.
 
-The helper handles well-formed requests as follows. (As usual, malformed
-requests are handled as described in {{errors}}.) It first looks for PPM
-parameters corresponding to `AggregateReq.task_id`. It then preprocesses the
-sub-requests as described in {{anti-replay}}. Any sub-requests filtered out by
-this procedure MUST be ignored.
+Next, the leader runs the preparation-state initialization algorithm for the
```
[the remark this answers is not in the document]
chris-wood: That's right 👍 it would be pointless for the leader to kick off aggregation for any share that it would eventually throw away.

draft-gpew-priv-ppm.md
chris-wood: That's better!

draft-gpew-priv-ppm.md
```diff
@@ -672,242 +679,366 @@ independently.
...
-The helper state is an optional parameter of an aggregate request that the
-helper can use to carry state across requests. At least part of the state will
-usually need to be encrypted in order to protect user privacy. However, the
-details of precisely how the state is encrypted and the information that it
-carries is up to the helper implementation.
+The value `out` is interpreted as follows. If the VDAF is 0-round, then `out` is the
+aggregator's output share, in which case the aggregator finishes and stores its
+output share for further processing as described in {{agg-complete}}. Otherwise,
+if the VDAF consists of one round or more, then the aggregator interprets `out` as
+the pair `(new_state, agg_msg)`, where `new_state` is its updated state and `agg_msg`
```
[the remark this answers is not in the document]
chris-wood: Yep, that's better!

draft-gpew-priv-ppm.md
chris-wood: Right now, they shouldn't actually aggregate until the leader issues an AggregateShareReq. We can change how that's done if we want to optimize things, but let's do so separately.

draft-gpew-priv-ppm.md
```diff
@@ -672,242 +679,366 @@ independently.
...
+parameters except its own, the leader sends a POST request to
+`[aggregator]/aggregate` with AggregateContinueReq as the payload and the media
+type set to "message/ppm-aggregate-continue-req".
 
-To handle the leader's request, the helper first ensures that the request meets
-the requirements for batch parameters following the procedure in
-{{batch-parameter-validation}}. If the batch parameters are invalid, then it
-MUST abort with error "batchInvalid".
+For each ProcessShare in AggregateContinueReq.seq received from the leader, the helper
+proceeds as follows:
```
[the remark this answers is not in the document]
chris-wood: It's "do this check and then proceed accordingly to either continue processing, invalidate the report, or finish."

draft-gpew-priv-ppm.md
```diff
@@ -672,242 +679,366 @@ independently.
...
 struct {
-  TaskID task_id;
-  opaque agg_param<0..2^16-1>;  // VDAF aggregation parameter
-  opaque helper_state<0..2^16>; // helper's opaque state
-  AggregateSubReq seq<1..2^24-1>;
-} AggregateReq;
+  Nonce nonce;
+  Extension extensions<0..2^16-1>;
+  HpkeCiphertext encrypted_input_share;
+} ReportShare;
```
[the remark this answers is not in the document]
chris-wood: Yeah, I think you're right. I'll move it later.

draft-gpew-priv-ppm.md
```diff
@@ -672,242 +679,366 @@ independently.
...
+report set until the underlying VDAF moves into a terminal state, which happens either
+when aggregation completes or an error occurs. The leader does this by repeating the
+following procedure.
+
+* If the helper failed, then mark the report as failed, remove it from the candidate
+  report set, and do not process it further.
+* If the leader continued but the helper finished, or if the leader finished but
+  the helper continued, then mark the report as failed and do not process it further.
+* If the leader and helper finished, then continuation is complete. At this point the
+  report has been completely processed and both aggregators have recovered an
```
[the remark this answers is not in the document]
chris-wood: Yeah, this is a good suggestion. Let's converge on "prepare" as the verb of choice here.

draft-gpew-priv-ppm.md
```diff
@@ -672,242 +679,366 @@ independently.
...
+  the corresponding measurement task.
+- Continuation: Continue the aggregation flow by exchanging messages produced
+  by the underlying VDAF instance until aggregation completes or an error occurs.
+- Completion: Finish the aggregate flow, yielding an aggregate share corresponding
+  to all output shares in the batch.
 
-### Aggregate Request
+Once both aggregation is complete and a sufficient number of reports have
+been aggregated, the aggregate shares can be collected to produce the final,
+aggregate output. That process is described in {{collect-flow}}.
```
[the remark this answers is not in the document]
chris-wood: I think we can remove it, since this is prematurely describing the collect flow.

draft-gpew-priv-ppm.md
chris-wood: This is shown below.

draft-gpew-priv-ppm.md
chris-wood: I think we can probably relax this. Let's address it in a followup issue.

draft-gpew-priv-ppm.md
chris-wood: Yep, done.

draft-gpew-priv-ppm.md
BranLwyd: Followup issue: https://github.com/ietf-wg-ppm/ppm-specification/issues/217 (FWIW, I agree this can be relaxed, I included some potential considerations in the issue.)

draft-gpew-priv-ppm.md
```diff
@@ -672,242 +679,365 @@ independently.
...
+#### Leader Continuation
+
+The leader begins each round of continuation for a report share based on a previous
+response from the helper, denoted `helper_outbound`, and a previous message produced
+from the leader, denoted `leader_outbound`. Each helper message carries a PrepareResult,
+indicating if preparation for the report share should continue, has failed, or is finished.
+
+If the helper message failed, then the leadaer marks the report as failed and removes it
+from the candidate report set and does not process it further. Otherwise, the leader
+computes its next state transition as follows:
```
cjpatton, line 955:
```suggestion
The leader begins each round of continuation for a report share based on its locally computed 
prepare message and the previous PrepareShare from the helper. If PrepareShare is of type "failed", 
then the leader marks the report as failed and removes it from the candidate report set and does not
process it further. If the type is "finished", then the leader aborts with "unrecognizedMessage".
[[OPEN ISSUE: This behavior is not specified.]] If the type is "continued", then the leader proceeds as
follows.

Let `leader_outbound` denote the leader's prepare message and `helper_outbound` denote the
helper's. The leader computes the next state transition as follows:
```

draft-gpew-priv-ppm.md
```diff
@@ -672,242 +679,365 @@ independently.
...
 ~~~
-struct {
-  opaque message<0..2^16-1>; // VDAF message
-} AggregateShare;
+out = VDAF.prep_next(prep_state, inbound)
 ~~~
 
-Next, the helper encrypts the aggregate share `agg_share` under the collector's
-public key as follows:
+where `inbound` is the previous VDAF message sent by the leader and `prep_state` is
```
cjpatton, line 1014:
```suggestion
where `inbound` is the previous VDAF preapre message sent by the leader and `prep_state` is
```

draft-gpew-priv-ppm.md
```diff
@@ -672,242 +679,365 @@ independently.
...
+is the AggregateContinueResp and media type is "message/ppm-aggregate-continue-resp". The helper
+then awaits the next message from the leader.
 
-This encryption prevents the leader from learning the actual result, as it only
-has its own share and not the helper's share, which is encrypted for the
-collector. The helper responds to the leader with HTTP status 200 OK and a body
-consisting of an `HpkeCiphertext` where `config_id` is set to the collector's
-HPKE config ID, `enc` is set to the encapsulated HPKE context `enc` and
-`ciphertext` is `encrypted_agg_share`.
+[[OPEN ISSUE: consider relaxing this ordering constraint]]
```
cjpatton, line 1038:
```suggestion
[[OPEN ISSUE: consider relaxing this ordering constraint. See issue#217.]]
```

draft-gpew-priv-ppm.md
```diff
@@ -672,242 +679,365 @@ independently.
...
-For each aggregator endpoint `[aggregator]` in `AggregateReq.task_id`'s
-parameters except its own, the leader sends a POST request to
-`[aggregator]/aggregate` with the following message:
+* Each report in the set MUST pertain to the same PPM task.
+* It is an error to allow a report to be replayed. If the leader has already aggregated
+  a report, but the report does not pertain to a batch that has been collected, then
+  the report MUST be excluded.
+* It is an error to include a new report in a batch that has already been collected.
+  If the report pertains to a batch that has been collected, but the leader has not yet
+  aggregated the report, then it MUST be excluded.
```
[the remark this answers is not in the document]
chris-wood: Agreed that it's redundant. I think we can omit this text and the forward reference entirely.

draft-gpew-priv-ppm.md
```diff
@@ -672,242 +679,365 @@ independently.
...
+The helper's response to the leader is an HTTP 200 OK whose body is the
+AggregateInitResp and media type is "message/ppm-aggregate-init-resp".
+
+Upon receipt of a helper's AggregateInitResp message, the leader checks that the
+sequence of PrepareShare messages corresponds to the ReportShare sequence of the
+AggregateInitReq. If any message appears out of order, is missing, has an
+unrecognized nonce, or if two messages have the same nonce, then the leader MUST
+abort with error "unrecognizedMessage".
+
+[[OPEN ISSUE: the leader behavior here is sort of bizarre -- to whom does it abort?]]
```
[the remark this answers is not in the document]
chris-wood: Indeed -- it's above.

draft-gpew-priv-ppm.md
```diff
@@ -672,242 +679,365 @@ independently.
...
+Once the helper has processed each valid report share in `AggregateInitReq.report_shares`, the
+helper then creates an AggregateInitResp message to complete its initialization. This message is
+structured as follows:
+
+~~~
+enum {
+  continued(0),
+  finished(1)
+  failed(2),
+} PrepareResult;
```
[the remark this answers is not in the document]
chris-wood: How about `PrepareStepResult`? (We can bike shed the name later)

draft-gpew-priv-ppm.md
```diff
@@ -672,242 +679,365 @@ independently.
...
   Nonce nonce;
-  opaque message<0..2^16-1>; // VDAF message
-} AggregateSubResp;
+  PrepareResult prepare_result;
+  select (PrepareShare.prepare_result) {
+    case continued: opaque prep_msg<0..2^16-1>; // VDAF preparation message
+    case finished:  // empty
+    case failed:    ReportShareError;
+  }
+} PrepareShare;
```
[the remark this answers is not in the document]
chris-wood: That works for me, and am happy to bike shed later.

draft-gpew-priv-ppm.md
chris-wood: Headers included!

draft-gpew-priv-ppm.md
```diff
@@ -672,247 +677,377 @@ independently.
...
+Let `[aggregator]` denote the helper's API endpoint. The leader sends a POST
+request to `[aggregator]/aggregate` with its AggregateInitReq message as
+the payload. The media type is "message/ppm-aggregate-init-req".
+
+#### Helper Initialization
+
+Each helper begins their portion of the aggregate initialization phase with the set
+of candidate report shares obtained in an `AggregateInitReq` message from the leader.
+It attempts to recover and validate the corresponding input shares similar to the leader,
+and eventually returns a response to the leader carrying the preparation state for each
```
cjpatton, line 789:
```suggestion
and eventually returns a response to the leader carrying a VDAF-specific message for each
```

draft-gpew-priv-ppm.md
```diff
@@ -672,247 +677,377 @@ independently.
...
+out = VDAF.prep_next(prep_state, None)
+~~~
+
+`vdaf_verify_param` is the public VDAF parameter, and `agg_param` is the opaque aggregation
+parameter. If either step fails, the aggregator marks the report as invalid with error
+`vdaf-prep-error`.
+
+Otherwise, the value `out` is interpreted as follows. If the VDAF is 0-round, then `out`
+is the aggregator's output share. Otherwise, if the VDAF consists of one round or more,
+then the aggregator interprets `out` as the pair `(prep_state, prep_msg)`.
```
tgeoghegan, line 925:
```suggestion
Otherwise, the value `out` is interpreted as follows. If this is the last round of the VDAF,
then `out` is the aggregator's output share. Otherwise, `out` is the pair `(prep_state, prep_msg)`.
```
To match the corresponding paragraph describing the leader's behavior on L952.

draft-gpew-priv-ppm.md
```diff
@@ -672,247 +677,377 @@ independently.
...
-struct {
-  opaque message<0..2^16-1>; // VDAF message
-} AggregateShare;
+out = VDAF.prep_next(prep_state, inbound)
 ~~~
 
-Next, the helper encrypts the aggregate share `agg_share` under the collector's
-public key as follows:
+where `inbound` is the previous VDAF preapre message sent by the leader and `prep_state` is
+its current preparation state. If this operation fails, then the helper fails
```
tgeoghegan, line 1005:
```suggestion
the helper's current preparation state. If this operation fails, then the helper fails
```
"its" is ambiguous since the sentence earlier refers to the leader.

draft-gpew-priv-ppm.md
```diff
@@ -672,242 +679,365 @@ independently.
...
+   input share as invalid with the  error `report-dropped` under the conditions
+   prescribed in {{anti-replay}}. The helper also checks to see if the report
+   has never been aggregated but is contained by a batch that has been collected.
 
-The helper state is an optional parameter of an aggregate request that the
-helper can use to carry state across requests. At least part of the state will
-usually need to be encrypted in order to protect user privacy. However, the
-details of precisely how the state is encrypted and the information that it
-carries is up to the helper implementation.
+If both checks succeed, the input share is not marked as invalid.
```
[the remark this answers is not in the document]
cjpatton: Bump.

draft-gpew-priv-ppm.md
```diff
@@ -672,247 +677,377 @@ independently.
...
+  PrepareStep prepare_shares<1..2^16-1>;
+} AggregateContinueResp;
 ~~~
 
-where `pk` is the HPKE public key encoded by the collector's HPKE key
-configuration and `task_id` is `AggregateShareReq.task_id`. `agg_share` is the
-serialized `AggregateShare`, and `batch_interval` is obtained from the
-`AggregateShareReq`.
+The order of AggregateContinueResp.prepare_shares matches that of the PrepareStep values in
+`AggregateContinueReq.prepare_shares`. The helper's response to the leader is an HTTP 200 OK whose body
```
tgeoghegan, line 1024:
```suggestion
The order of AggregateContinueResp.prepare_shares MUST match that of the PrepareStep values in
`AggregateContinueReq.prepare_shares`. The helper's response to the leader is an HTTP 200 OK whose body
```

draft-gpew-priv-ppm.md
```diff
@@ -672,247 +677,377 @@ independently.
...
+This output message for each report in AggregateContinueReq.prepare_shares is then sent
+to the leader in an AggregateContinueResp message, structured as follows:
 
 ~~~
-enc, context = SetupBaseS(pk, task_id,
-                          "ppm-00 aggregate share" || 0x03 || 0x00)
-encrypted_agg_share = context.Seal(batch_interval, agg_share)
+struct {
+  opaque helper_state<0..2^16>;
+  PrepareStep prepare_shares<1..2^16-1>;
```
cjpatton, line 1019: nit: for consistency with similar abbreviation elsewhere
```suggestion
  PrepareStep prep_shares<1..2^16-1>;
```

draft-gpew-priv-ppm.md
```diff
@@ -1004,21 +1125,139 @@ fetch certificates: https://datatracker.ietf.org/doc/html/rfc8555#section-7.4.2]
...
+`encrypted_aggregate_share.config_id` is set to the collector's HPKE config ID.
+`encrypted_aggregate_share.enc` is set to the encapsulated HPKE context `enc`
+computed above and `encrypted_aggregate_share.ciphertext` is the ciphertext
+`encrypted_agg_share` computed above.
+
+After receiving the helper's response, the leader uses the HpkeCiphertext to
+respond to a collect request (see {{collect-flow}}).
+
+The leader MAY make multiple aggregate-share requests for a given batch interval
+and aggregation parameter and MUST get the same result each time.
```
cjpatton, line 1204: I don't understand why this behaviour is useful. Is this new, or is this already in the spec?

draft-gpew-priv-ppm.md
```diff
@@ -1004,21 +1125,139 @@ fetch certificates: https://datatracker.ietf.org/doc/html/rfc8555#section-7.4.2]
...
+~~~
+
+### Aggregate Share Encryption {#aggregate-share-encrypt}
+
+Encrypting an aggregate share `agg_share` for a given `AggregateShareReq` is done
+as follows:
+
+~~~
+enc, context = SetupBaseS(pk, AggregateShareReq.task_id ||
+                              "ppm-00 aggregate share" || 0x03 || 0x00)
```
cjpatton, line 1239: Replac
```suggestion
                              "ppm-00 aggregate share" || server_role || 0x00)
```
where `server_role` is `0x02` for the leader and `0x03` for a helper.

cjpatton (approved):
Just a few editorial comments. Also, I'd like us to consider addressing https://github.com/ietf-wg-ppm/ppm-specification/pull/223/files#r856672481 here, but this can wait if you feel it needs discussion.

draft-gpew-priv-ppm.md
```diff
@@ -1004,21 +1125,139 @@ fetch certificates: https://datatracker.ietf.org/doc/html/rfc8555#section-7.4.2]
...
+~~~
+struct {
+    HpkeCiphertext encrypted_aggregate_share;
+} AggregateShareResp;
+~~~
+
+`encrypted_aggregate_share.config_id` is set to the collector's HPKE config ID.
+`encrypted_aggregate_share.enc` is set to the encapsulated HPKE context `enc`
+computed above and `encrypted_aggregate_share.ciphertext` is the ciphertext
+`encrypted_agg_share` computed above.
```
tgeoghegan, line 1198: I think this discussion of how to populate the fields of `encrypted_aggregate_share` would be more clear in the `aggregate-share-encrypt` section, below.

tgeoghegan (changes requested):

draft-gpew-priv-ppm.md
tgeoghegan: I'm not sure if this is what Chris W. had in mind but I agree with this text. Suppose we have a task with `max_batch_lifetime = 1`. Consider:

1. Leader makes an `AggregateShareReq` for some batch interval.
2. Helper services the request, and marks that batch interval as having been collected once.
3. Helper attempts to transmit an `AggregateShareResp` to leader, but the message is truncated.

Now, if the leader retries its `AggregateShareReq`, the helper will refuse the request because the batch interval's lifetime has been consumed. So, to allow the leader to retry this request, the helper has to be willing to resend previously computed aggregate shares.

draft-gpew-priv-ppm.md
```diff
@@ -672,247 +677,377 @@ independently.
...
-where `sk` is the HPKE secret key, `task_id` is `AggregateReq.task_id` and
-`server_role` is the role of the server (`0x02` for the leader and `0x03` for
-the helper; `0x01`, the client role, is used as the sender). `nonce` and
-`extensions` are obtained from the corresponding fields in `AggregateSubReq`. If
-decryption fails, then the sub-response consists of a "decryption error" alert.
-[See issue#57.] Otherwise, the helper handles the request for its plaintext
-input share `input_share` and updates its state as specified by the PPM
-protocol.
+where `sk` is the HPKE secret key, `task_id` is the task ID, `nonce` and
+`extensions` are the nonce and extensions of the report share respectively,
```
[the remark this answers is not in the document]
chris-wood: Yep, that's right. I left it here to be specific. We can omit in the future if desired.

draft-gpew-priv-ppm.md
```diff
@@ -672,247 +677,377 @@ independently.
...
+The validation checks are as follows.
 
-#### Helper State
+1. Check if the report has never been aggregated but is contained by
+   a batch that has been collected. If this check fails, the input share
+   is marked as invalid with the error `batch-collected`. This prevents
+   additional reports from being aggregated after its batch has already
+   been collected.
+2. Check if the report has already been aggregated. If this check fails,
+   the input share is marked as invalid with the error `report-replayed`.
```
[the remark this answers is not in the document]
chris-wood: Let's address this in subsequent changes. We _still_ don't have batch validation requirements nailed down, so this is going to likely change anyway.

draft-gpew-priv-ppm.md
```diff
@@ -672,247 +677,377 @@ independently.
...
+   This is the case if the report was used in a previous aggregate request
+   and is therefore a replay. An aggregator may also choose to mark an
+   input share as invalid with the  error `report-dropped` under the conditions
+   prescribed in {{anti-replay}}. The helper also checks to see if the report
+   has never been aggregated but is contained by a batch that has been collected.
```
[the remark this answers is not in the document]
chris-wood: Indeed, and it's also helper-specific, so I'll remove it.

draft-gpew-priv-ppm.md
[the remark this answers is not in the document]
chris-wood: Let's do this in a separate PR. We still don't have these requirements nailed down and the text is subject to change.

draft-gpew-priv-ppm.md
chris-wood: Keeping this as-is for now. We can bash the name in a followup change.

draft-gpew-priv-ppm.md
chris-wood: Yanking this out to discuss in a separate issue: https://github.com/ietf-wg-ppm/ppm-specification/issues/226

tgeoghegan (approved):

draft-gpew-priv-ppm.md
```diff
@@ -1004,21 +1123,137 @@ fetch certificates: https://datatracker.ietf.org/doc/html/rfc8555#section-7.4.2]
...
+
+To handle the leader's request, the helper first ensures that the request meets
+the requirements for batch parameters following the procedure in
+{{batch-parameter-validation}}. If the batch parameters are invalid, then it
+MUST abort with error "batchInvalid".
+
+Next, it computes a checksum based on its view of the output shares included in
+the batch window, and checks that the `report_count` and `checksum` included in
+the request match its computed values. If not, then it MUST abort with error
+"batchMismatch".
```
divergentdave, line 1164: Should we add this to the errors table above? (I see it says "this list is not exhaustive", but I'm not sure if that means other errors may be in the same URI namespace, or other errors would have entirely different URIs)

draft-gpew-priv-ppm.md
```diff
@@ -672,247 +677,375 @@ independently.
...
+shares as follows:
+
+1. Decrypt the input share for each report share as described in {{input-share-decryption}}.
+1. Check that the resulting input share is valid as described in {{input-share-batch-validation}}.
+1. Initialize VDAF preparation as described in {{input-share-prep}}.
+
+If any step yields an invalid report share, the leader removes the report share from
+the set of candidate reports. Once the leader has initialized this state for all valid
+candidate report shares, it then creates an AggregateInitReq message for each helper to
+initialize the preparation of this candidate set. This message is structured as follows:
```
ekr, line 755: Does this have to be done serially? Could the leader instead send all of the shares and then discover later that one is bogus? Surely yes, because the proof could fail.

draft-gpew-priv-ppm.md
```diff
@@ -672,247 +677,375 @@ independently.
...
 ~~~
 
-The structure contains the PPM task, an opaque, VDAF-specific aggregation
-parameter, an opaque *helper state* string, and a sequence of *sub-requests*,
-each corresponding to a unique client report. Sub-requests are structured as
-follows:
+The `nonce` and `extensions` fields of each ReportShare match that in the Report
+uploaded by the client. The `encrypted_input_share` field is the `HpkeCiphertext`
+whose index in `Report.encrypted_input_shares` is equal to the index of the aggregator
+in the task's `aggregator_endpoints`. The `agg_param` field is an opaque, VDAF-specific
```
ekr, line 775: ...to which the share is being sent...

draft-gpew-priv-ppm.md
```diff
@@ -672,247 +677,375 @@ independently.
...
+request to `[aggregator]/aggregate` with its AggregateInitReq message as
+the payload. The media type is "message/ppm-aggregate-init-req".
+
+#### Helper Initialization
+
+Each helper begins their portion of the aggregate initialization phase with the set
+of candidate report shares obtained in an `AggregateInitReq` message from the leader.
+It attempts to recover and validate the corresponding input shares similar to the leader,
+and eventually returns a response to the leader carrying a VDAF-specific message for each
+report share.
```
ekr, line 790: I wonder if it would be better to factor out the common behaviors for all aggregators including the leader.

draft-gpew-priv-ppm.md
```diff
@@ -672,247 +677,375 @@ independently.
...
+
+To begin this process, the helper first checks that the nonces in `AggregateInitReq.report_shares`
+are all distinct. If two ReportShare values have the same nonce, then the helper MUST abort with
+error "unrecognizedMessage". If this check succeeds, the helper then attempts to recover each
+input share in `AggregateInitReq.report_shares` as follows:
+
+1. Decrypt the input share for each report share as described in {{input-share-decryption}}.
+1. Check that the resulting input share is valid as described in {{input-share-batch-validation}}.
+1. Initialize VDAF preparation and initial outputs as described in {{input-share-prep}}.
```
ekr, line 799: For instance, this is common to leader and helper.

draft-gpew-priv-ppm.md
```diff
@@ -672,247 +677,375 @@ independently.
...
+   a batch that has been collected. If this check fails, the input share
+   is marked as invalid with the error `batch-collected`. This prevents
+   additional reports from being aggregated after its batch has already
+   been collected.
+2. Check if the report has already been aggregated. If this check fails,
+   the input share is marked as invalid with the error `report-replayed`.
+   This is the case if the report was used in a previous aggregate request
+   and is therefore a replay. An aggregator may also choose to mark an
+   input share as invalid with the  error `report-dropped` under the conditions
+   prescribed in {{anti-replay}}.
```
ekr, line 895: Shouldn't the nonce check go here?

draft-gpew-priv-ppm.md
ekr: :+1:

ekr (changes requested):

draft-gpew-priv-ppm.md
```diff
@@ -1004,21 +1123,137 @@ fetch certificates: https://datatracker.ietf.org/doc/html/rfc8555#section-7.4.2]
...
+
+~~~
+agg_share = VDAF.out_shares_to_agg_share(agg_param, out_shares)
+~~~
+
+Note that for most VDAFs, it is possible to aggregate output shares as they
+arrive rather than wait until the batch is collected. To do so however, it is
+necessary to enforce the batch parameters as described in
+{{batch-parameter-validation}} so that the aggregator knows which aggregate
+share to update.
```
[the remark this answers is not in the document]
tgeoghegan: We're discussing adding a couple of methods to VDAF here: https://github.com/cjpatton/vdaf/issues/47

draft-gpew-priv-ppm.md
chris-wood: Yep, done.

draft-gpew-priv-ppm.md
chris-wood: I think this done right now. The commonality is these steps, each of which have their own sections:

```
1. Decrypt the input share for each report share as described in {{input-share-decryption}}.
1. Check that the resulting input share is valid as described in {{input-share-batch-validation}}.
1. Initialize VDAF preparation and initial outputs as described in {{input-share-prep}}.
```

draft-gpew-priv-ppm.md
```diff
@@ -672,247 +677,375 @@ independently.
...
+process it further. If the type is "finished", then the leader aborts with "unrecognizedMessage".
+[[OPEN ISSUE: This behavior is not specified.]] If the type is "continued", then the leader proceeds as
+follows.
+
+Let `leader_outbound` denote the leader's prepare message and `helper_outbound` denote the
+helper's. The leader computes the next state transition as follows:
+
+~~~
+inbound = VDAF.prep_shares_to_prep(agg_param, [leader_outbound, helper_outbound])
+out = VDAF.prep_next(prep_state, inbound)
```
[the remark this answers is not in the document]
chris-wood: I think we had Process at some point in this PR, but that was confusing since it didn't align with the VDAF terminology. I suggest we punt this to the VDAF draft and then just use whatever term makes sense there.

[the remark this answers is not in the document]
chris-wood: https://github.com/cjpatton/vdaf/issues/48

draft-gpew-priv-ppm.md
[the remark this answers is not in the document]
chris-wood: Cool! I'll resolve this then.

draft-gpew-priv-ppm.md
```diff
@@ -672,247 +677,375 @@ independently.
...
+Once the helper has processed each valid report share in `AggregateInitReq.report_shares`, the
+helper then creates an AggregateInitResp message to complete its initialization. This message is
+structured as follows:
 
 ~~~
+enum {
+  continued(0),
+  finished(1)
+  failed(2),
+} PrepareStepResult;
```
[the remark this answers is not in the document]
chris-wood: Only to match VDAF. See https://github.com/cjpatton/vdaf/issues/48.

draft-gpew-priv-ppm.md
chris-wood: Yep, it can. I added a note to that effect.

draft-gpew-priv-ppm.md
chris-wood: Sorry, what nonce check?

ekr: The check that the nonces aren't internally duplicated.

draft-gpew-priv-ppm.md
```diff
@@ -672,247 +678,381 @@ independently.
...
+If decryption fails, the aggregator marks the report share as invalid with the
+error `hpke-decrypt-error`. Otherwise, it outputs the resulting `input_share`.
 
-After processing all of the sub-requests, the helper encrypts its updated state
-and constructs its response to the aggregate request.
+#### Input Share Validation {#input-share-batch-validation}
 
-#### Leader State
+Validating an input share will either succeed or fail. In the case of failure,
+the input share is marked as invalid with a ReportShareError of type `vdaf-prep-error`.
```
divergentdave, line 888: I think "vdaf-prep-error" here is incorrect. The two checks below each specify different ReportShareError codes. The next section, "Input Share Preparation", correctly says it may produce a "vdaf-prep-error" error.

chris-wood, line 888:
```suggestion
the input share is marked as invalid with a corresponding ReportShareError error.
```

chris-wood: Yep, resolved. Thanks!

ekr (approved):

draft-gpew-priv-ppm.md
```diff
@@ -672,247 +678,381 @@ independently.
...
+  HpkeCiphertext encrypted_input_share;
+} ReportShare;
+
+struct {
+  TaskID task_id;
+  opaque agg_param<0..2^16-1>;
+  opaque helper_state<0..2^16>;
+  ReportShare report_shares<1..2^16-1>;
+} AggregateInitReq;
 ~~~
```
chris-wood, line 775:
```suggestion
~~~

[[OPEN ISSUE: consider sending report shares separately (in parallel) to the aggregate instructions. RIght now, aggregation parameters and the corresponding report shares are sent at the same time, but this may not be strictly necessary. ]]
```

PR #224, opened 2022-04-25
interop: Add aggregation parameter to AggregateShareReq

cjpatton:
Without this, the scope of the request is ambiguous.

### DIFF
--- a/draft-gpew-priv-ppm.md
+++ b/draft-gpew-priv-ppm.md
@@ -1240,6 +1240,7 @@ PPM task, the leader sends to each helper a POST request to
 struct {
   TaskID task_id;
   Interval batch_interval;
+  opaque agg_param<0..2^16-1>;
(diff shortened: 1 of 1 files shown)


tgeoghegan:
I had the same thought while implementing `/aggregate_share` but concluded that it's not needed. The scope of the `AggregateShareReq` is clear based on the `batch_interval`. The aggregation parameter gets provided to helper during `AggregateInitReq`. What do you need it for while handling `AggregateShareReq`, which should just be a matter of compiling the results of all the aggregation jobs described by `AggregateShareReq.batch_interval`?

cjpatton:
Consider the case where the leader simultaneously runs two aggregation flows for the same batch interval but with different aggregation parameters.

tgeoghegan (approved):

divergentdave (approved):

PR #227, opened 2022-04-27
Add problem type for batch checksum/report count

tgeoghegan:
# This targets the interop branch

Adds a problem document type for the case where the helper and leader
compute different checksums or report counts when constructing aggregate
shares. The error is chosen to match what was added in #223.


divergentdave (approved):

BranLwyd (approved):

cjpatton (approved):
Beat me to it!

PR #229, opened 2022-04-27
Update reference to VDAF draft

### LINKED ISSUE #189: fix VDAF reference
Would be helpful to readers if this reference was fixed in the Editor's Copy:

[I-D.draft-cfrg-patton-vdaf]
    "*** BROKEN REFERENCE ***".

cjpatton:
Closes #189.


chris-wood (approved):
Maybe change each reference to `{{?VDAF=I-D.draft-irtf-cfrg-vdaf}}`? That way it'll render like `[VDAF]` in the text, which reads a bit nicer.

PR #232, opened 2022-04-28
Replace helper_state with index into per-aggregator storage

### LINKED ISSUE #185: Replace helper state blob with an aggregation job UUID
@cjpatton and I were debating the [helper state blob](https://abetterinternet.github.io/ppm-specification/draft-gpew-priv-ppm.html#name-helper-state). Now that helpers have to store all the report nonces and are also likely to be expected to store their own report shares, I argue that we should get rid of it: we don't need it and it entails a bunch of complicated security considerations. Helpers need to encrypt their state so leaders can't see it, need to protect themselves against leaders replaying old state blobs and they need to version it so that a helper can deal with a blob constructed by an older version of itself.

Chris' rebuttal is that even if you don't use the blob to store something like a set of nonces or a counter or a partial aggregate share, it's still useful as a key into the storage where the helper keeps its state. This is helpful for a helper running on something like a Cloudflare Worker, where ephemeral servers are launched in response to incoming messages, and so a different helper instance might handle the sequence of `AggregateInit` and `Aggregate` requests.

We agreed that we don't need an arbitrarily large opaque blob shipped back and forth between leader and helper to solve that problem. All you need is a unique value like a UUID so that a stateless helper can recover context from storage. This will let us delete a whole bunch of SHOULD and MAY discussion from the spec.

chris-wood:
Since aggregators already need to store nonce sets for the purpose of replay prevention, it doesn't make much sense to also offload part of their state to the leader. Instead, aggregators might benefit from a unique index into some local storage for keeping track of per-aggregation job state. This change drops the helper_state in favor of such an index.

Closes #185.

### DIFF
--- a/draft-ietf-ppm-dap.md
+++ b/draft-ietf-ppm-dap.md
@@ -661,12 +661,18 @@ minutes, to account for clock skew.
 Once a set of clients have uploaded their reports to the leader, the leader can
 send them to the helpers to be verified and aggregated. In order to enable the
 system to handle very large batches of reports, this process can be performed
-incrementally. To aggregate a set of reports, the leader sends a message to each
-helper containing those report shares. The helper then processes them (verifying
-the proofs and incorporating their values into the ongoing aggregate) and replies
-to the leader.
+incrementally. Verification of a set of reports is referred to as an aggregation
+job. Each aggregation job is associated with exactly one PPM task, and a PPM
+task can have many aggregation jobs. Each job is associated with an ID that is
+unique within the context of a PPM task in order to distinguish different jobs
+from one another. Each aggregator uses this ID as an index into per-job storage,
+e.g., to keep track of report shares that belong to a given aggregation job.
 
-The exact structure of the aggregation flow depends on the VDAF. Specifically:
+To run an aggregation job, the leader sends a message to each helper containing
+the report shares in the job. The helper then processes them (verifying the proofs
+and incorporating their values into the ongoing aggregate) and replies to the leader.
+
+The exact structure of the aggregation job flow depends on the VDAF. Specifically:
 
 * Some VDAFs (e.g., `prio3`) allow the leader to start aggregating reports
   proactively before all the reports in a batch are received. Others (e.g.,
@@ -677,8 +683,7 @@ The exact structure of the aggregation flow depends on the VDAF. Specifically:
   round trips.
 
 Note that it is possible to aggregate reports from one batch while reports from
-the next batch are coming in. This is because each report is validated
-independently.
+the next batch are coming in. This is because each report is validated independently.
 
 This process is illustrated below in {{aggregation-flow-illustration}}. In this
 example, the batch size is 20, but the leader opts to process the reports in
@@ -692,16 +697,16 @@ be cryptographically protected as described in {{agg-init}}.
 ~~~
 Leader                                                 Helper
 
-Aggregate request (Reports 1-10) ------------------------>  \
-<----------------------------- Aggregate response (State 1) | Reports
-Aggregate request (continued, State 1) ------------------>  | 1-10
-<----------------------------- Aggregate response (State 1) /
+Aggregate request (Reports 1-10, Job = N) --------------->  \
+<----------------------------- Aggregate response (Job = N) | Reports
+Aggregate request (continued, Job = N) ------------------>  | 1-10
+<----------------------------- Aggregate response (Job = N) /
 
 
-Aggregate request (Reports 11-20, State 1) -------------->  \
-<----------------------------- Aggregate response (State 1) | Reports
-Aggregate request (continued, State 2) ------------------>  | 11-20
-<----------------------------- Aggregate response (State 2) /
+Aggregate request (Reports 11-20, Job = M) -------------->  \
+<----------------------------- Aggregate response (Job = M) | Reports
+Aggregate request (continued, Job = M) ------------------>  | 11-20
+<----------------------------- Aggregate response (Job = M) /
 ~~~
 {: #aggregation-flow-illustration title="Aggregation Flow (batch size=20).
 Multiple aggregation flows can be executed at the same time."}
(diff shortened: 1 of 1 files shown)


ekr:
A few points:
1. This seems strictly less flexible. The helpers could already store an index if they wanted, but now they are required to do so.
2. Can't you store nonce sets in the helper state?

chris-wood:
> This seems strictly less flexible. The helpers could already store an index if they wanted, but now they are required to do so.

Yeah, they could store an index in helper_state. However, helper_state is a "per-aggregation" storage mechanism, not a "per-task" mechanism, and (I think) the replay state needs to cover the entire task. So we could (a) change helper_state to be a per-task thing, in which case it's not clear 2^16 is enough space, or if we want to be paying that bandwidth for the duration of the task, or (b) go with the layer of indirection as in here. On balance, I think (b) is probably better?

cjpatton:
>     2. Can't you store nonce sets in the helper state?

This can get really hairy. Think of this way: Storing nonce sets is required for report replay protection, which is needed for privacy.  Depending on what state the helper has locally for managing the "version" of the state blob, it may be possible for the leader to "replay" an old state and bypass replay protection for reports.

There are other ways that helper state gets hairy. For example, we haven't been precise yet about how handling concurrent aggregation flows: If the helper state is "per-task", then concurrent flows would be a problem.

Overall the goal of this change is to remove this complexity altogether. We can always add it back later if we need to.

ekr:
I don't agree that this change reduces complexity. AFAICT, this is two changes:

1. It has the leader assign the ID.
2. It restricts the ID space to uint32.

ISTM that it's just as straightforward for the helper to assign the id at the start of the job, and then use any format it chooses, which the current system allows. I get that you think that certain ways of assigning and using that ID are complex, but that's not our problem because the helper can figure it out for themselves.

cjpatton:
>     2. It restricts the ID space to uint32.

You mean 32-byte strings, right? uint32 is just 8 bytes, whereas the AggregationJobID is 32 bytes.

Anther important thing it does is remove the state blob from the AggregateShareReq. Here's the bit I think you're missing: Right now we require the leader to send the same `helper_state` it received in the helper's previous response. This means that all aggregate requests, i.e., AggregateInitReq, AggregateContReq, and AggregateShareReq, need to be issued one after the other. In particular, it's not possible to run multiple aggregation flows concurrently. You're forced to do

```
AggregateInitReq(report_set_1, helper_state_0)
<- helper_state_1
AggregateContReq(report_set_1, helper_state_1)
<- helper_state_2
```

*before* you do

```
AggregateInitReq(report_set_2, helper_state_2)
<- helper_state_3
AggregateContReq(report_set_2, heper_state_3)
<- helper_state_4
```

(Note that the "aggregation flow" excludes the AggregateShareReq, which is used to combine the results across multiple aggregation flows.) It's going to be important for scalability that we can run multiple aggregation flows simultaneously. That means that we need, at a minimum, the following things:
1. Don't carry state across aggregation flows
2. Make sure there's a way for a helper to "link" aggregate requests pertaining to the same aggregation flow

This PR accomplishes both of these, but as you point out, is more strict than necessary. Perhaps we could keep the helper state blob as-is, but remove it from the AggregateShareReq and mandate only that the helper state is maintained for the duration of a single aggregation flow.

ekr:
> Here's the bit I think you're missing: Right now we require the leader to send the same helper_state it received in the helper's previous response.

I'm undecided about whether this is a good or bad change, but it should be decided on its own merits, not buried in a PR entitled "Replace helper_state with index into per-aggregator storage"

tgeoghegan:
I believe this change does yield a significant reduction in complexity, because we no longer have to spell out how the helper needs to protect its state from the leader. Aggregation job IDs do not need to be kept secret from the leader, nor does the helper need to need to implement any anti-replay protections for them (because all the sensitive state is now assumed to be in the helper's database).

draft-gpew-priv-ppm.md
```diff
@@ -659,12 +659,18 @@ minutes, to account for clock skew.
 Once a set of clients have uploaded their reports to the leader, the leader can
 send them to the helpers to be verified and aggregated. In order to enable the
 system to handle very large batches of reports, this process can be performed
-incrementally. To aggregate a set of reports, the leader sends a message to each
-helper containing those report shares. The helper then processes them (verifying
-the proofs and incorporating their values into the ongoing aggregate) and replies
-to the leader.
+incrementally. Verification of a set of reports is referred to as an aggregation
+job. Each aggregation job is associated with at most one PPM task, and a PPM
```
tgeoghegan, line 663: "at most one" suggests an aggregation job could be associated with zero PPM tasks.
```suggestion
job. Each aggregation job is associated with exactly one PPM task, and a PPM
```

draft-gpew-priv-ppm.md
```diff
@@ -690,16 +695,16 @@ be cryptographically protected as described in {{agg-init}}.
...
+<----------------------------- Aggregate response (Job = N) | Reports
+Aggregate request (continued, Job = N) ------------------>  | 1-10
+<----------------------------- Aggregate response (Job = N) /
 
 
-Aggregate request (Reports 11-20, State 1) -------------->  \
-<----------------------------- Aggregate response (State 1) | Reports
-Aggregate request (continued, State 2) ------------------>  | 11-20
-<----------------------------- Aggregate response (State 2) /
+Aggregate request (Reports 11-20, Job = N) -------------->  \
```
tgeoghegan, line 704: I think the idea is that reports 11-20 would be job `M` where `M !=N`.

tgeoghegan (changes requested):

cjpatton:
@ekr: The reasoning for this was originally discussed on issue #185. This issue didn't get much attention, so we didn't expect the change to be controversial. We certainly didn't intend to bury it. We tried implementing both and thought this approach solved several problems compared to the status quo. Would you prefer that we take this too the list?

chris-wood:
Let's just move this discussion to the list 😃

ekr:
> 

I don't understand these points. Modulo the parallelism points @cjpatton raises, nothing prevents the helper from simply issuing each task a fresh ID and sending that as the helper state. The change in this PR is that it prevents the helper from doing something fancier because the leader issues the ID. Yes, in the latter case, it might need some kind of crypto to protect the state, but those design choices are out of our purview.

ekr:
> @ekr: The reasoning for this was originally discussed on issue https://github.com/ietf-wg-ppm/ppm-specification/issues/185. This issue didn't get much attention, so we didn't expect the change to be controversial. We certainly didn't intend to bury it. We tried implementing both and thought this approach solved several problems compared to the status quo. Would you prefer that we take this too the list?

I've just read that issue several times and I don't really see much discussion of issuing aggregation requests in parallel.

The question here is not about the list versus not the list. It's about the process we are going to follow for landing changes. In general, substantive changes should be getting the consensus of the WG prior to being landed. 

At this relatively early phase, IETF sometimes allows editors some latitude to make design decisions ahead of time that will then be reviewed by the group when the next draft is published. However, typically the way to do this would be to email the list with a clear explanation of what the changes were and a notification that you were going to land a set of PRs unless someone objected by (say) a week away.

draft-gpew-priv-ppm.md
```diff
@@ -690,16 +695,16 @@ be cryptographically protected as described in {{agg-init}}.
...
+<----------------------------- Aggregate response (Job = N) | Reports
+Aggregate request (continued, Job = N) ------------------>  | 11-20
+<----------------------------- Aggregate response (Job = N) /
```
chris-wood, line 707:
```suggestion
Aggregate request (Reports 11-20, Job = M) -------------->  \
<----------------------------- Aggregate response (Job = M) | Reports
Aggregate request (continued, Job = M) ------------------>  | 11-20
<----------------------------- Aggregate response (Job = M) /
```

cjpatton:
@chris-wood mind rebasing to fix conflicts?

cjpatton (approved):

PR #233, opened 2022-04-29
HTTP problem document construction details

tgeoghegan:
Defines more error types to represent various failure modes, updates
protocol text to indicate when the new error types should be used, and
affirms that the task ID should be encoded as base64url with no padding
(since task ID has a fixed size).

### DIFF
--- a/draft-ietf-ppm-dap.md
+++ b/draft-ietf-ppm-dap.md
@@ -380,7 +380,11 @@ in the "type" field (within the PPM URN namespace "urn:ietf:params:ppm:error:"):
 | unrecognizedMessage     | The message type for a response was incorrect or the payload was malformed. |
 | unrecognizedTask        | An endpoint received a message with an unknown task ID. |
 | outdatedConfig          | The message was generated using an outdated configuration. |
+| staleReport             | Report could not be processed because it arrived too late. |
+| reportFromTheFuture     | Report could not be processed because its timestamp is too far in the future. |
 | batchInvalid            | A collect or aggregate-share request was made with invalid batch parameters. |
+| insufficientBatchSize   | There are not enough reports in the batch interval. |
+| batchLifetimeExceeded   | The batch lifetime has been exceeded for one or more reports included in the batch interval. |
 | batchMismatch           | Aggregators disagree on the report shares that were aggregated in a batch. |
 
 This list is not exhaustive. The server MAY return errors set to a URI other
@@ -389,8 +393,9 @@ not listed in the appropriate IANA registry (see {{ppm-urn-space}}). Clients
 SHOULD display the "detail" field of all errors. The "instance" value MUST be
 the endpoint to which the request was targeted. The problem document MUST also
 include a "taskid" member which contains the associated PPM task ID, encoded
-with base64 using the standard alphabet {{!RFC4648}} (this value is always
-known, see {{task-configuration}}).
+in Base 64 using the URL and filename safe alphabet with no padding as defined
+in sections 5 and 3.2 of {{!RFC4648}} (this value is always known, see
+{{task-configuration}}).
 
 In the remainder of this document, we use the tokens in the table above to refer
 to error types, rather than the full URNs. For example, an "error of type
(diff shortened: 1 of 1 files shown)


draft-gpew-priv-ppm.md
```diff
@@ -378,7 +378,11 @@ in the "type" field (within the PPM URN namespace "urn:ietf:params:ppm:error:"):
 | unrecognizedMessage     | The message type for a response was incorrect or the payload was malformed. |
 | unrecognizedTask        | An endpoint received a message with an unknown task ID. |
 | outdatedConfig          | The message was generated using an outdated configuration. |
+| staleReport             | Report could not be processed because it arrived too late. |
+| reportFromTheFuture     | Report could not be processed because its timestamp is too far in the future. |
```
ekr, line 382: I would probably call these `reportTooLate` and `reportTooEarly`

draft-gpew-priv-ppm.md
```diff
@@ -378,7 +378,11 @@ in the "type" field (within the PPM URN namespace "urn:ietf:params:ppm:error:"):
...
 | batchInvalid            | A collect or aggregate-share request was made with invalid batch parameters. |
+| insufficientBatchSize   | There are not enough reports in the batch interval. |
```
ekr, line 384: Are all of these fatal?

draft-gpew-priv-ppm.md
```diff
@@ -387,8 +391,9 @@ not listed in the appropriate IANA registry (see {{ppm-urn-space}}). Clients
 SHOULD display the "detail" field of all errors. The "instance" value MUST be
 the endpoint to which the request was targeted. The problem document MUST also
 include a "taskid" member which contains the associated PPM task ID, encoded
-with base64 using the standard alphabet {{!RFC4648}} (this value is always
-known, see {{task-configuration}}).
+in Base 64 using the URL and filename safe alphabet with no padding as defined
+in sections 5 and 3.2 of {{!RFC4648}} (this value is always known, see
+{{task-configuration}}).
```
ekr, line 396: I would put the parenthetical after the word "task ID" rather than after the text about b64.

draft-gpew-priv-ppm.md
tgeoghegan: Yes, inasmuch as they represent conditions that the aggregators cannot recover from. `insufficientBatchSize` for instance means there aren't enough reports in `CollectReq.batch_interval` to satisfy the task's `min_batch_size`, which means the resulting aggregate would violate client privacy. The description text could be improved on this one.

ekr: Well, but could I aggregate some more into the batch and then ask for collect?

tgeoghegan: Yes, absolutely. Getting this error is how a collector would know to either try again with a bigger `batch_interval` or wait for more reports to arrive.

tgeoghegan: That makes sense. And for this error, we should add some text to the section on collecting explaining what a collector could do if they encounter it.

- [ ] explain in `collect-flow` what collectors should do in the face of `insufficientBatchSize`

BranLwyd (approved):

cjpatton (approved):
Looks great!

divergentdave (approved):

draft-gpew-priv-ppm.md
chris-wood: @tgeoghegan can we file an issue to track this?

tgeoghegan: Done: #239

draft-ietf-ppm-dap.md
```diff
@@ -623,7 +628,14 @@ collector. (See {{collect-flow}}.) Otherwise, comparing the aggregate result to
 the previous aggregate result may result in a privacy violation. (Note that the
 helpers enforce this as well; see {{collect-flow}}.) In addition, the
 leader SHOULD abort the upload protocol and alert the client with error
-"staleReport".
+"reportTooLate".
+
+The leader MUST buffer reports while waiting to aggregate them. The
```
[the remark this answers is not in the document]
tgeoghegan: Sorry, I mashed "MERGE" before I noticed this. I captured this as #240

PR #220, opened 2022-04-20
rename document

tgeoghegan:
The draft has been adopted by the WG and now needs a new name.
Additionally, this commit replaces references to the old repository on
`abetterinternet` with the current GitHub organization.

PR #235, opened 2022-05-02
Update security considerations

### LINKED ISSUE #211: Sybil attack distinction
From dkg in Vienna: 

> maybe distinguish between the two kinds of sybil attacks as a sybil attack against the collector (stats poisoning), vs. a sybil attack against a reporting client (privacy violation)

cjpatton:
Closes #211.

This change is primarily editorial and includes the following changes:

* Replace references to "Prio" with references to a generic VDAF. (This
  section was written long ago when we had Prio in mind.)
* Elaborate on known issues for collect requests.
* Discuss Sybil attacks, including enumerating the different types
  (#211).

### DIFF
--- a/draft-ietf-ppm-dap.md
+++ b/draft-ietf-ppm-dap.md
@@ -1444,34 +1444,63 @@ defined by the PPM parameters. (See {{batch-parameter-validation}}.)
 
 # Security Considerations {#sec-considerations}
 
-Prio assumes a powerful adversary with the ability to compromise an unbounded
-number of clients. In doing so, the adversary can provide malicious (yet
-truthful) inputs to the aggregation function. Prio also assumes that all but one
-server operates honestly, where a dishonest server does not execute the protocol
-faithfully as specified. The system also assumes that servers communicate over
-secure and mutually authenticated channels. In practice, this can be done by TLS
-or some other form of application-layer authentication.
+PPM assumes an active attacker that controls the network and has the ability to
+statically corrupt any number of clients, aggregators, and collectors. That is,
+the attacker can learn the secret state of any party prior to the start of its
+attack. For example, it may coerce a client into providing malicious input
+shares for aggregation or coerce an aggregator into diverting from the
+protocol specified (e.g., by divulging its input shares to the attacker).
 
-In the presence of this adversary, Prio provides two important properties for
-computing an aggregation function F:
+In the presence of this adversary, PPM aims to achieve the following high-level
+secure aggregation goals:
 
-1. Privacy. The aggregators and collector learn only the output of F computed
-   over all client inputs, and nothing else.
-1. Robustness. As long as the aggregators execute the input-validation protocol
-   correctly, a malicious client can skew the output of F only by reporting
-   false (untruthful) input. The output cannot be influenced in any other way.
+1. Privacy. Clients trust that some aggregator is honest. That is, as long as at
+   least one aggregator executes the protocol faithfully, the parties learn nothing
+   beyond the aggregate result (i.e., the output of the aggregation function computed over
+   the honest measurements).
+1. Correctness. The collector trusts that the aggregators execute the protocol
+   correctly. That is, as long as the aggregators execute the protocol faithfully,
+   a malicious client can skew the aggregate result only by reporting
+   a false (untruthful) measurement. The result cannot be influenced in any
+   other way.
 
-There are several additional constraints that a Prio deployment must satisfy in
-order to achieve these goals:
+Currently, the specification does not achieve these goals. In particular, there are several open
+issues that need to be addressed before these goals are met. Details for each issue are below.
 
-1. Minimum batch size. The aggregation batch size has an obvious impact on
-   privacy. (A batch size of one hides nothing of the input.)
-2. Aggregation function choice. Some aggregation functions leak slightly more
-   than the function output itself.
-
-[TODO: discuss these in more detail.]
+1. When crafted maliciously, collect requests may leak more information about
+   the measurements than the system intends. For example, the spec currently
+   allows sequences of collect requests to reveal an aggregate result for a
+   batch smaller than the minimum batch size. [OPEN ISSUE: See issue#195. This
+   also has implications for how we solve issue#183.]
+1. Even benign collect requests may leak information beyond what one might
+   expect intuitively. For example, the Poplar1 VDAF
+   {{?VDAF=I-D.draft-irtf-cfrg-vdaf}} can be used to compute the set of heavy
+   hitters among a set of arbitrary bit strings uploaded by clients. This
+   requires multiple evaluations of the VDAF, the results of which reveal
+   information to the aggregators and collector beyond what follows from the
+   heavy hitters themselves. Note that this leakage can be mitigated using
+   differential privacy. [OPEN ISSUE: We have yet not specified how to add DP.]
+1. The core PPM spec does not defend against Sybil attacks. In this type of
+   attack, the adversary adds to a batch a number of reports that skew the
+   aggregate result in its favor. For example: The result may reveal additional
+   information about the honest measurements, leading to a privacy violation; or
+   the result may have some property that is desirable to the adversary ("stats
+   poisoning"). The upload sub-protocol includes an extensions mechanism that
+   can be used to prevent --- or at least mitigate --- these types of attacks.
+   See {{upload-extensions}}. [OPEN ISSUE: No such extension has been
+   implemented, so we're not yet sure if the current mechanism is sufficient.]
+1. Attacks may also come from the network. Thus, it is required that the
+   aggregators and collector communicate with one another over mutually
+   authenticated and confidential channels. The core PPM spec does not specify
+   such a mechanism beyond requiring server authentication for HTTPS sessions.
+   Note that clients are not required to authenticate themselves. [OPEN ISSUE:
+   It might be better to be prescriptive about leader authentication in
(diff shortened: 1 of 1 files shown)


draft-gpew-priv-ppm.md
```diff
@@ -1442,35 +1442,62 @@ defined by the PPM parameters. (See {{batch-parameter-validation}}.)
...
-There are several additional constraints that a Prio deployment must satisfy in
-order to achieve these goals:
-
-1. Minimum batch size. The aggregation batch size has an obvious impact on
-   privacy. (A batch size of one hides nothing of the input.)
-2. Aggregation function choice. Some aggregation functions leak slightly more
-   than the function output itself.
-
-[TODO: discuss these in more detail.]
+PPM assumes an active attacker that controls the network and has the ability to
```
tgeoghegan, line 1445:
```suggestion
DAP assumes an active attacker that controls the network and has the ability to
```
Since #220

draft-gpew-priv-ppm.md
```diff
@@ -1442,35 +1442,62 @@ defined by the PPM parameters. (See {{batch-parameter-validation}}.)
...
+statically corrupt any number of clients, aggregators, and collectors. That is,
+the attacker can learn the secret state of any party prior to the start of its
+attack. For example, it may coerce a client into providing malicious input
+shares for aggregation or coerce an aggregator into not diverting from the
```
tgeoghegan, line 1449:
```suggestion
shares for aggregation or coerce an aggregator into diverting from the
```

draft-gpew-priv-ppm.md
```diff
@@ -1442,35 +1442,62 @@ defined by the PPM parameters. (See {{batch-parameter-validation}}.)
...
+   information about the honest measurements, leading to a privacy violation; or
+   the result may have some property that is desirable to the adversary ("stats
+   poisoning"). The upload sub-protocol includes an extensions mechanism that
+   can be used to prevent --- or at least mitigate --- these types of attacks.
+   See {{upload-extensions}}. [OPEN ISSUE: No such extension has been
+   implemented, so we're not yet sure if the current mechanism is sufficient.]
+1. Attacks may also come from the network. Thus, it is required that the
+   aggregators and collector communicate with one another over mutually
+   authenticated and confidential channels. The core PPM spec does not specify
+   such a mechanisn beyond requiring server authentication for HTTPS sessions.
```
tgeoghegan, line 1494:
```suggestion
   such a mechanism beyond requiring server authentication for HTTPS sessions.
```

draft-gpew-priv-ppm.md
```diff
@@ -1442,35 +1442,62 @@ defined by the PPM parameters. (See {{batch-parameter-validation}}.)
...
+   also has implications for how we solve issue#183.]
+1. Even benign collect requests may leak information beyond what one might
+   expect intuitively. For example, the Poplar1 VDAF
+   {{?VDAF=I-D.draft-irtf-cfrg-vdaf}} can be used to compute the set of heavy
+   hitters among a set of arbitrary bit strings uploaded by clients. This
+   requires multiple evaluations of the VDAF, the results of which reveal
+   information to the aggregators and collector beyond what follows from the
+   heavy hitters themselves. Note that this leakage can be mitigated using
+   differential privacy. [OPEN ISSUE: We have yet not specified how to add DP.]
+1. The core PPM spec does not defend against Sybil attacks. In this type of
```
tgeoghegan, line 1482: Does "core" mean a deployment that doesn't use report extensions or any kind of client authentication mechanism?

csharrison: Beyond the protocol extensions it is also the case that some VDAFs have Sybil protections (against privacy) built-in e.g. with noise addition / differential privacy. PPM _could_ enforce that the aggregation function provide this kind of privacy but we choose not to.

draft-gpew-priv-ppm.md
cjpatton: This change needs to be applied elsewhere, better to do it it one change I think.

draft-gpew-priv-ppm.md
cjpatton: @tgeoghegan "core" refers to the bits of the spec that apply to all implementations. In particular, not every implementation will be required to interpret extensions the same way.

@csharrison It's not clear to me that DP always provides adequate protection against Sybil attacks. I think it ultimately depends on how you define privacy. That said, requiring DP is totally on the table, but before doing this I think we should start by spelling out what this would look like as an optional feature of the protocol. I think the first question to ask is whether this mechanism lives here or in the underlying VDAF. I would love to have your help on this.

csharrison: > @csharrison It's not clear to me that DP always provides adequate protection against Sybil attacks. I think it ultimately depends on how you define privacy. That said, requiring DP is totally on the table, but before doing this I think we should start by spelling out what this would look like as an optional feature of the protocol. I think the first question to ask is whether this mechanism lives here or in the underlying VDAF. I would love to have your help on this.

I think if you can prove that the system achieves user-level DP you should be able to show that even with worst-case Sybil attacks, a user's privacy is still protected to some degree. That being said, this is a property of the whole system, not just the VDAF.

I am fairly sure that DAP / PPM does not want to require DP in its deployments. This brings on a lot of baggage, and for instances where these kinds of attacks are truly not a problem (e.g. with airtight client authentication) it might not be that useful. The main point here is that we might want to call out that the particular VDAF used may mitigate some of the risks of Sybil.

cjpatton: If achievable, I think specifying an optional, generic DP mechanism here (perhaps for some class of VDAFs with a specific property) would have significant advantages. IIUC, tuning and enforcing the privacy budget will require knowledge of how measurements are generated by clients over time. This is something that will be dictated by the deployment, and I think it would be best to avoid bleeding deployment details into VDAF wherever possible.

That said, VDAF may need to at least say something about how/when to add noise into an input or aggregate share.

csharrison: > If achievable, I think specifying an optional, generic DP mechanism here (perhaps for some class of VDAFs with a specific property) would have significant advantages

I think this is possible. We can probably have some tunable, exposed property of a VDAF that is a necessary but not sufficient condition for achieving DP. I think we can describe the rest of the conditions needed in the DAP deployment to get the rest of the way there.

I think this will not be super easy to do, but I am happy to help think through what it will take (maybe in a separate issue).

draft-ietf-ppm-dap.md
```diff
@@ -1444,35 +1444,62 @@ defined by the PPM parameters. (See {{batch-parameter-validation}}.)
...
+   the result may have some property that is desirable to the adversary ("stats
+   poisoning"). The upload sub-protocol includes an extensions mechanism that
+   can be used to prevent --- or at least mitigate --- these types of attacks.
+   See {{upload-extensions}}. [OPEN ISSUE: No such extension has been
+   implemented, so we're not yet sure if the current mechanism is sufficient.]
+1. Attacks may also come from the network. Thus, it is required that the
+   aggregators and collector communicate with one another over mutually
+   authenticated and confidential channels. The core PPM spec does not specify
+   such a mechanism beyond requiring server authentication for HTTPS sessions.
+   Note that clients are not required to authenticate themselves.
```
tgeoghegan, line 1497: I think we're currently debating this point and thus will need to revisit this text. Perhaps include a bracketed note referencing https://github.com/ietf-wg-ppm/ppm-specification/issues/155 or something else that captures the discussion about mutual auth?

draft-gpew-priv-ppm.md
cjpatton: Kicking off an issue is probably a good way forward. I'm happy to help with the protocol design, but will need to lean on you to make sure we have all the plumbing we need to implement. A couple seed questions to start:
1. Who adds noise and when? Do we want to support local DP, centralized DP, or both? Also would be helpful to define these (and point to references where needed).
2. What syntactic changes are necessary for VDAF?

Once you create the issue, it would also be helpful to ping the list with a brief description of the issue and a link.

tgeoghegan (approved):

draft-ietf-ppm-dap.md
```diff
@@ -1444,35 +1444,64 @@ defined by the PPM parameters. (See {{batch-parameter-validation}}.)
...
+   beyond the aggregate result (i.e., the output of the aggregation function computed over
+   the honest measurements).
+1. Correctness. The collector trusts that the aggregators execute the protocol
+   correctly. That is, as long as the aggregators execute the protocol faithfully,
+   a malicious client can skew the aggregate result only by reporting
+   a false (untruthful) measurement. The result cannot be influenced in any
+   other way.
+
+To achieve these goals, there are a variety of attack vectors that a PPM
+deployment must consider:
```
chris-wood, line 1468:
```suggestion
Currently, the specification does not achieve these goals. In particular, there are several open
issues that need to be addressed before these goals are met. Details for each issue are below.
```

chris-wood (approved):
The list of attacks is really a list of things we ought to fix in the protocol, so I would not even bother including them in the security considerations. If we want to use this as an opportunity to identify open issues that need to be addressed before we consider the protocol secure, I would simply list the open issues and keep discussion in GitHub. (That is, the text here seems redundant with the issue 🤷 )

draft-gpew-priv-ppm.md
csharrison: commented on https://github.com/ietf-wg-ppm/ppm-specification/issues/19, I doesn't answer all the questions but it's a start :)

PR #236, opened 2022-05-03
include task ID in request to hpke_config

tgeoghegan:
When obtaining an aggregator's HPKE configuration, clients now specify
the task ID they are interested in. Aggregators are not required to use
a distinct HPKE configuration for each task, but now it's possible for
them to do so. The endpoint is also renamed to `hpke_config` from
`key_config` for clarity.

### DIFF
--- a/draft-ietf-ppm-dap.md
+++ b/draft-ietf-ppm-dap.md
@@ -360,7 +360,7 @@ client could not report 10^6s or -20s.
 
 Communications between PPM entities are carried over HTTPS {{!RFC2818}}. HTTPS
 provides server authentication and confidentiality. In addition, report shares
-are encrypted directly to the aggregators using HPKE {{!I-D.irtf-cfrg-hpke}}.
+are encrypted directly to the aggregators using HPKE {{!RFC9180}}.
 
 ## Errors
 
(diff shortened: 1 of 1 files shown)


BranLwyd (approved):

cjpatton (approved):

chris-wood (approved):
Nice 👍

draft-ietf-ppm-dap.md
```diff
@@ -485,7 +485,7 @@ In addition, in order to facilitate the aggregation and collect protocols, each
 of the aggregators is configured with following parameters:
 
 * `collector_config`: The HPKE configuration of the collector (described in
-  {{key-config}}).
+  {{hpke-config}}).
```
divergentdave, line 488: This referenced section is by and large particular to the client requesting an HPKE configuration from the leader, but it does include relevant structure definitions. How's this?

```
... The HPKE configuration of the collector (`HpkeConfig` defined in {{hpke-config}}).
```

divergentdave (approved):

draft-ietf-ppm-dap.md
tgeoghegan: IMO we should fix this by moving the definition of `struct HpkeConfig` and its attendant types from `{{hpke-config}}` up to the common type definitions in "Protocol Definition" to eliminate this awkward forward reference. I'll do that in a separate change.

PR #238, opened 2022-05-10
Add bearer token for HTTP request authentication

### LINKED ISSUE #155: Authentication of Collector's query
During execution of the collect protocol, the leader is responsible for relaying some parameters of the collect request to the helper, represented by the opaque `output_param` in `CollectReq` and `OutputShareReq` messages introduced by #154. That parameter will always be null/empty/absent for Priov3, but in Hits, it would contain the string prefixes that the collector wants a count of in the result set (and who knows what else in yet-undefined VDAFs).

It's currently possible for the leader (or any actor that can tamper with messages on the network) to insert arbitrary `output_param` into `OutputShareReq` messages. @cjpatton reckons [this is a threat to soundness](https://github.com/abetterinternet/ppm-specification/pull/154#discussion_r718917818). We should think about whether this enables the leader to undermine soundness in any ways beyond the capabilities the threat model already grants it, and if so, introduce collector->helper authentication+integrity. ~Confidentiality~Privacy may not be needed on here because the `output_param` should be identical for helper and leader.

cjpatton:
Defines an HTTP request authentication mechanism and requires its usage
in all collector->leader and leader->helper requests. We intend for this
mechanism to be replaced eventually by something more secure, e.g., TLS
client authentication.

For details see https://mailarchive.ietf.org/arch/msg/ppm/z65FK8kOU27Dt38WNhpI6apc2so/.

~Closes issue 155.~ As pointed out in the comments, this change may not be sufficient.

### DIFF
--- a/draft-ietf-ppm-dap.md
+++ b/draft-ietf-ppm-dap.md
@@ -359,8 +359,9 @@ client could not report 10^6s or -20s.
 # Message Transport
 
 Communications between PPM entities are carried over HTTPS {{!RFC2818}}. HTTPS
-provides server authentication and confidentiality. In addition, report shares
-are encrypted directly to the aggregators using HPKE {{!RFC9180}}.
+provides server authentication and confidentiality. When client authenticaiton
+is also required, the client uses the mechanism described in
+{{https-sender-auth}}.
 
 ## Errors
 
@@ -386,6 +387,7 @@ in the "type" field (within the PPM URN namespace "urn:ietf:params:ppm:error:"):
 | insufficientBatchSize   | There are not enough reports in the batch interval to satisfy the task's minimum batch size. |
 | batchLifetimeExceeded   | The batch lifetime has been exceeded for one or more reports included in the batch interval. |
 | batchMismatch           | Aggregators disagree on the report shares that were aggregated in a batch. |
+| unauthorizedRequest     | Authentication of an HTTP request failed (see {{https-sender-auth}}). |
 
 This list is not exhaustive. The server MAY return errors set to a URI other
 than those defined above. Servers MUST NOT use the PPM URN namespace for errors
@@ -405,6 +407,29 @@ to error types, rather than the full URNs. For example, an "error of type
 This document uses the verbs "abort" and "alert with `[some error message]`" to
 describe how protocol participants react to various error conditions.
 
+## HTTPS Sender Authentication {#https-sender-auth}
+
+Some HTTP requests in the PPM protocol require the sender to authenticate its
+request. It does so as described here.
+
+Prior to the start of the protocol, the sender and receiver arrange to share a
+secret sender-specific API token, which MUST be suitable for representation in
+an HTTP header.
+
+For requests requiring authentication, the sender includes a "DAP-Auth-Token"
+header in its HTTP request containing the API token.
+
+To authenticate the request, the receiver looks up the token for the
+sender as determined by the task configuration. (See {{task-configuration}}.) If
+the value of the "DAP-Auth-Token" header does not match the token, then
+the receiver MUST abort with error "unauthorizedRequest". The error code of the
+HTTP response MUST be 403.
+
+[OPEN ISSUE: This simple bearer-token scheme is meant to unblock interop
+testing. Eventually it should be replaced with a more secure authentication
+mechanism, e.g., TLS client certificates. See
+https://mailarchive.ietf.org/arch/msg/ppm/z65FK8kOU27Dt38WNhpI6apc2so/ for
+details.]
 
 # Protocol Definition
 
@@ -495,6 +520,11 @@ of the aggregators is configured with following parameters:
   the setup algorithm computed jointly by the aggregators before the start of the
   PPM protocol {{?VDAF=I-D.draft-irtf-cfrg-vdaf}}). [OPEN ISSUE: This is yet to be
   specified. See issue#161.]
+
+The helper stores a bearer token used to authenticate HTTP requests from the
+leader. Likewise, the leader stores a bearer token to authenticate HTTP request
+from the collector. The authentication mechanism is described in
+{{https-sender-auth}}.
 
 Finally, the collector is configured with the HPKE secret key corresponding to
 `collector_hpke_config`.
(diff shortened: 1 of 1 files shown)


chris-wood (commented):
I've no objections to this, but I don't think it needs to be codified in the spec. I think experimental deployments should be free to choose whatever header and bearer token content they want to make this work. 

Also, are we sure we want to close #155 with this? I'd suggest we keep that issue open, as it tracks the "real" solution to this problem.

draft-ietf-ppm-dap.md
```diff
@@ -400,6 +402,29 @@ to error types, rather than the full URNs. For example, an "error of type
...
+## HTTPS Sender Authentication {#https-sender-auth}
+
+Some HTTP requests in the PPM protocol require the sender to authenticate its
+request. It does so as described here.
+
+Prior to the start of the protocol, the sender and receiver arrange to share a
+bearer token. The bearer token is an opaque, 32-byte string that MUST be
+generated using a cryptographically secure random number generator.
+
+For requests requiring authentication, the sender includes a "PPM-Auth-Token"
```
tgeoghegan, line 414:
```suggestion
For requests requiring authentication, the sender includes a "DAP-Auth-Token"
```

draft-ietf-ppm-dap.md
```diff
@@ -400,6 +402,29 @@ to error types, rather than the full URNs. For example, an "error of type
...
+header in its HTTP request whose value is the bearer token encoded in Base64 and
+using the URL-safe alphabet defined in {{!RFC4648}}.
+
+To authenticate the request, the receiver looks up the bearer token for the
+sender as determined by the task configuration. (See {{task-configuration}}.) If
+the value of the "PPM-Auth-Token" header does not match the bearer token, then
+the receiver MUST abort with error "unauthorizedRequest".
```
tgeoghegan, line 421: I think we should require servers to respond with HTTP 403 Forbidden here (our definition of "abort with error ..." allows servers to use any 4xx or 5xx status).

tgeoghegan (changes requested):
I agree with Chris W. that we should keep #155 open, especially since this change does not solve the problem of integrity or authenticity of collect parameters from the collector to the helper.

draft-ietf-ppm-dap.md
```diff
@@ -400,6 +402,29 @@ to error types, rather than the full URNs. For example, an "error of type
...
 describe how protocol participants react to various error conditions.
 
+## HTTPS Sender Authentication {#https-sender-auth}
+
+Some HTTP requests in the PPM protocol require the sender to authenticate its
+request. It does so as described here.
+
+Prior to the start of the protocol, the sender and receiver arrange to share a
+bearer token. The bearer token is an opaque, 32-byte string that MUST be
+generated using a cryptographically secure random number generator.
```
ekr, line 412:
```suggestion
Prior to the start of the protocol, the sender and receiver arrange to share a
secret sender-specific API token, which MUST be suitable for representation in
an HTTP header.
```

draft-ietf-ppm-dap.md
```diff
@@ -400,6 +402,29 @@ to error types, rather than the full URNs. For example, an "error of type
...
+
+For requests requiring authentication, the sender includes a "PPM-Auth-Token"
+header in its HTTP request whose value is the bearer token encoded in Base64 and
+using the URL-safe alphabet defined in {{!RFC4648}}.
```
ekr, line 416:
```suggestion
header in its HTTP request containing the API token.
```

draft-ietf-ppm-dap.md
ekr, line 421:
```suggestion
To authenticate the request, the receiver looks up the  token for the
sender as determined by the task configuration. (See {{task-configuration}}.) If
the value of the "PPM-Auth-Token" header does not match the token, then
the receiver MUST abort with error "unauthorizedRequest".
```

ekr: This seems inconsistent with the guidance of BCP 56 S 8.

ekr (changes requested):

draft-ietf-ppm-dap.md
ekr: First, the actual guidance is not good:
1. 32 bytes is far longer than needed
2. Binary strings are hard to cut and paste, so people will want to actually have them in ASCII, which this prohibits, and even if it didn't then would then have to re-encode them per the rules below.
3. It forbids structure (e.g., the account ID) which might be convenient.
4. If people have existing systems for managing API tokens, this might conflict.

Second, it's unenforceable and not required for interoperability, so we it's not our job to tell people what to do.

ekr:
FWIW, I agree with @chris-wood: this isn't necessary for interop testing, where authentication doesn't matter that much (as an example, we did much of our TLS 1.3 interop testing with self-signed certs). Obviously, if people want to deploy a system with real data, then some authentication is required, but enabling people to do that does not need to be a priority for the WG at this point; they can just make their own arrangements.

draft-ietf-ppm-dap.md
tgeoghegan: Thanks for the reference, this is really useful! However I think this paragraph in [section 8](
https://datatracker.ietf.org/doc/html/rfc3205#section-8) supports using an HTTP error code:

>    A layered application should use appropriate HTTP error codes to
>      report errors resulting from information in the HTTP request-line
>      and *header fields associated with the request*.  This request
>      information is part of the HTTP protocol and errors which are
>      associated with that information should therefore be reported
>      using HTTP protocol mechanisms.

In this specific case, the error results from information in the `PPM[/DAP]-Auth-Token` HTTP header and so indicating an error in HTTP seems appropriate. However the point may be moot if we end up deciding not to dictate specific auth mechanisms at this stage or ever.

ekr: Agreed. Good point.

draft-ietf-ppm-dap.md
cjpatton: Good points.

draft-ietf-ppm-dap.md
cjpatton: Oof, it would be useful to add an error code column to the error table. Will punt on this for now and simply add this to the text.

cjpatton (commented):
After conferring offline, we decided to merge this PR afterall. The branch has been rebased.

tgeoghegan (approved):

PR #243, opened 2022-05-11
Replace "ppm" with "dap" and align context strings with next draft

cjpatton:
This change is basically a search-and-replace for /ppm/dap/. However, I have also re-aligned the context strings to match the next draft that will be cut (-01).

I think the change is mostly correct, my main concern is the "URN Sub-namespace for PPM (urn:ietf:params:dap)" section. 🚲 🏠 ⏰: Does it make sense for all drafts in the PPM WG to share the namespace, or do we want to define one specifically for DAP?

### DIFF
--- a/draft-ietf-ppm-dap.md
+++ b/draft-ietf-ppm-dap.md
@@ -374,7 +374,8 @@ MAY return status code 405 (Method Not Allowed).
 When the server responds with an error status, it SHOULD provide additional
 information using a problem document {{!RFC7807}}. To facilitate automatic
 response to errors, this document defines the following standard tokens for use
-in the "type" field (within the DAP URN namespace "urn:ietf:params:ppm:error:"):
+in the "type" field (within the DAP URN namespace
+"urn:ietf:params:ppm:dap:error:"):
 
 | Type                    | Description                                                                                  |
 |:------------------------|:---------------------------------------------------------------------------------------------|
@@ -402,7 +403,7 @@ and filename safe alphabet with no padding defined in sections 5 and 3.2 of
 In the remainder of this document, we use the tokens in the table above to refer
 to error types, rather than the full URNs. For example, an "error of type
 'unrecognizedMessage'" refers to an error document with "type" value
-"urn:ietf:params:ppm:error:unrecognizedMessage".
+"urn:ietf:params:ppm:dap:error:unrecognizedMessage".
 
 This document uses the verbs "abort" and "alert with `[some error message]`" to
 describe how protocol participants react to various error conditions.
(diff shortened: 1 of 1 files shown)


draft-ietf-ppm-dap.md
```diff
@@ -1812,16 +1812,16 @@ the input shares and defeat privacy.
...
-- AggregateInitResp {{collect-flow}}: "message/ppm-aggregate-init-resp"
-- AggregateContinueReq {{collect-flow}}: "message/ppm-aggregate-continue-req"
-- AggregateContinueResp {{collect-flow}}: "message/ppm-aggregate-continue-resp"
-- AggregateShareReq {{collect-flow}}: "message/ppm-aggregate-share-req"
-- AggregateShareResp {{collect-flow}}: "message/ppm-aggregate-share-resp"
-- CollectReq {{collect-flow}}: "message/ppm-collect-req"
-- CollectResp {{collect-flow}}: "message/ppm-collect-resp"
+- HpkeConfig {{hpke-config}}: "application/dap-hpke-config"
+- Report {{upload-request}}: "message/dap-report"
+- AggregateInitReq {{collect-flow}}: "message/dap-aggregate-init-req"
```
chris-wood, line 1817: I assume this will conflict now that it's AggregateInitializeReq/Resp?

chris-wood (approved):
LGTM pending the conflict comment.

draft-ietf-ppm-dap.md
cjpatton: Fixed after rebase.

chris-wood:
> Does it make sense for all drafts in the PPM WG to share the namespace, or do we want to define one specifically for DAP?

Separate namespaces seems better, especially if we're thinking of DAP and STAR. Names are cheap, after all.

draft-ietf-ppm-dap.md
```diff
@@ -374,7 +374,8 @@ MAY return status code 405 (Method Not Allowed).
 When the server responds with an error status, it SHOULD provide additional
 information using a problem document {{!RFC7807}}. To facilitate automatic
 response to errors, this document defines the following standard tokens for use
-in the "type" field (within the DAP URN namespace "urn:ietf:params:ppm:error:"):
+in the "type" field (within the DAP URN namespace "urn:ietf:params:dap:error:"):
+>>>>>>> 4752c36 (Replace "ppm" with "dap" and align context strings with next draft)
```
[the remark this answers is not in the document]
cjpatton: Oops, that was sloppy. Fixed.

tgeoghegan (approved):

tgeoghegan:
> Does it make sense for all drafts in the PPM WG to share the namespace, or do we want to define one specifically for DAP?

Separate namespaces, because otherwise I think we'd have to write a further document enumerating the error type URNs that both this document and other PPM drafts like STAR could reference. We should avoid spreading the specification across multiple documents more than necessary. Maybe including the WG in the URN path like `urn:ietf:params:ppm:dap:error` helps?

PR #249, opened 2022-05-12
Revise usage of HPKE encryption.

### LINKED ISSUE #221: Usage of HPKE `info` and `aad` fields (was "HPKE: do not use both application_info & AAD.")
Per https://www.rfc-editor.org/rfc/rfc9180.pdf section 8.1:
> Applications that only use the single-shot APIs described in Section 6 should use the Setup info parameter for specifying auxiliary authenticated information. Implementations which only expose single-shot APIs should not allow applications to use both Setup info and Context aad or exporter_context auxiliary information parameters.

PPM effectively uses the single-shot APIs; we should consider refactoring PPM's usage of HPKE to use only one of application_info & AAD. Following the advice of the HPKE RFC, we would want to use application_info.

This would permit slightly simpler implementations. I am not sure how much it would simplify the HPKE computations.

BranLwyd:
Specifically:
  * Specify use of the "one-shot" HPKE APIs defined in RFC9180,
    replacing equivalent use of the "multi-shot" APIs. (This change is
    intended to be a clarification, not a functional change.)
  * Move the task_id parameter from the "info" parameter to the "aad"
    parameter. This is intended to protect against
    key-commitment-related attacks[1]. (This is a functional change.)
  * Clarify that the HpkeCiphertext.enc field is an encapsulated key,
    rather than an encryption context. (This is a clarification, not a
    functional change.)

[1] See discussion in https://github.com/ietf-wg-ppm/ppm-specification/issues/221.

Closes #221.

### DIFF
--- a/draft-ietf-ppm-dap.md
+++ b/draft-ietf-ppm-dap.md
@@ -479,7 +479,7 @@ uint8 HpkeConfigId;
 /* An HPKE ciphertext. */
 struct {
   HpkeConfigId config_id;    // config ID
-  opaque enc<1..2^16-1>;     // encapsulated HPKE context
+  opaque enc<1..2^16-1>;     // encapsulated HPKE key
   opaque payload<1..2^16-1>; // ciphertext
 } HpkeCiphertext;
 ~~~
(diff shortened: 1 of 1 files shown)


draft-ietf-ppm-dap.md
```diff
@@ -624,30 +624,18 @@ This message is called the client's *report*. It contains the following fields:
...
 sequence of input shares as specified by the VDAF in use. To encrypt an input
-share, the client first generates an HPKE {{!RFC9180}} context for
-the aggregator by running
+share, the client generates an HPKE {{!RFC9180}} ciphertext and encapsulated
+key for the aggregator by running
 
 ~~~
-enc, context = SetupBaseS(pk, Report.task_id || "dap-01 input share" ||
-                              0x01 || server_role)
+enc, payload = SealBase(pk, "dap-01 input share" || 0x01 || server_role,
```
tgeoghegan, line 631: Should we take the opportunity to bump this to `dap-02 input share` since we plan to submit `draft-ietf-ppm-dap-02` soon? Same goes for `dap-02 aggregate share` elsewhere in the doc.

tgeoghegan (approved):

draft-ietf-ppm-dap.md
cjpatton: We're at -00 right now: https://datatracker.ietf.org/doc/draft-ietf-ppm-dap/

cjpatton (approved):
Perfect, great work!

draft-ietf-ppm-dap.md
BranLwyd: Leaving as `dap-01` for now since we are currently at `dap-00` and per Tim's comment we will soon be bumping versions.

PR #250, opened 2022-05-13
Lock spec to draft-irtf-cfrg-vdaf-00

cjpatton:
The current draft is compatible with vdaf-00, and there will likely be
API-breaking changes in vdaf-01. For planned changes see
https://mailarchive.ietf.org/arch/msg/cfrg/Fvd_m64V9bC4VVJm2Zsuwr441mM/.

### DIFF
--- a/draft-ietf-ppm-dap.md
+++ b/draft-ietf-ppm-dap.md
@@ -146,7 +146,7 @@ Input:
 
 Input share:
 : An aggregator's share of the output of the VDAF
-   {{?VDAF=I-D.draft-irtf-cfrg-vdaf}} sharding algorithm. This algorithm is run by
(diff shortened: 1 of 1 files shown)


cjpatton:
@divergentdave, done (nice catch).

tgeoghegan (approved):

chris-wood (approved):
No objections from me, but we could also just wait until VDAF-01 is out and then pin 🤷

cjpatton:
> No objections from me, but we could also just wait until VDAF-01 is out and then pin 🤷

The big task remaining for VDAF-01 is finishing Poplar1, which likely won't happen for several weeks. I suppose we could cut VDAF-01 w/o Poplar1 and save it for VDAF-02?

cjpatton:
I'm going to merge this now, it'll be easy enough to point to the next draft once we're ready.

PR #251, opened 2022-05-13
Revise usage of HPKE encryption (take two).

### LINKED ISSUE #221: Usage of HPKE `info` and `aad` fields (was "HPKE: do not use both application_info & AAD.")

BranLwyd:
Specifically:
  * Specify use of the "one-shot" HPKE APIs defined in RFC9180,
    replacing equivalent use of the "multi-shot" APIs. (This change is
    intended to be a clarification, not a functional change.)
  * Move the task_id parameter from the "info" parameter to the "aad"
    parameter. This is intended to protect against
    key-commitment-related attacks[1].) This is a functional change.
  * Clarify that the HpkeCiphertext.enc field is an encapsulated key,
    rather than an encryption context. (This is a clarification, not a
    functional change.)

This is a second copy of a previous PR (#249) which was merged too
hastily, and then reverted. There are no changes from that PR.

[1] See discussion in #221.

Closes #221.

### DIFF
--- a/draft-ietf-ppm-dap.md
+++ b/draft-ietf-ppm-dap.md
@@ -479,7 +479,7 @@ uint8 HpkeConfigId;
 /* An HPKE ciphertext. */
 struct {
   HpkeConfigId config_id;    // config ID
-  opaque enc<1..2^16-1>;     // encapsulated HPKE context
+  opaque enc<1..2^16-1>;     // encapsulated HPKE key
   opaque payload<1..2^16-1>; // ciphertext
 } HpkeCiphertext;
 ~~~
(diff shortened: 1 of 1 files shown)


tgeoghegan (approved):

cjpatton (approved):

chris-wood (approved):

kraouf3:
. I will
Appreciate any reviews or feedback from the group.

To summarize changes, one of the changes is an active change:
* The value of DAP work ID that was already included in "Data
There was a field of different HPKE encryption and deceptive operation
Instead went to the "Aad" field. This change will make protocol
Very interested for attackers trying to benefit from lack
Key commitment in the main schemes of AEAD. See

PR #262, opened 2022-05-19
Drop job_id from Aggregate{Initialize,Continue}Resp.

### LINKED ISSUE #261: Aggregate{Initialize,Continue}Resp should not include the aggregation job ID
`Aggregate{Initialize,Continue}Resp` both include the relevant aggregation job ID; but in both cases, the `Aggregate{Initialize,Continue}Req` includes the aggregation job ID, the response will always reflect the aggregation job ID that was included in the request, and a request is always paired with a response by means of being part of the same HTTP request/response. This looks to be unnecessary communication overhead--if that's accurate, I think we can just drop these fields.

BranLwyd:
Closes #261. (see that issue for justification of this change)

### DIFF
--- a/draft-ietf-ppm-dap.md
+++ b/draft-ietf-ppm-dap.md
@@ -880,13 +880,11 @@ struct {
 } PrepareStep;
 
 struct {
-  AggregationJobID job_id;
   PrepareStep prepare_steps<1..2^16-1>;
 } AggregateInitializeResp;
 ~~~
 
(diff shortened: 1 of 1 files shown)


cjpatton (approved):

chris-wood (approved):
Agreed that these are redundant with the current design, so I'm OK merging this as-is. Let's split out how we convey task and job parameters (via URL or application message or whatever) to a separate issue.

PR #263, opened 2022-05-19
Include task_id in AggregateContinueReq.

### LINKED ISSUE #260: Clarify the domain in which aggregation job IDs must be unique
The current DAP specification says (section 4.3.1.1):
```
Generate a fresh AggregationJobID. This ID MUST be unique within the context of the corresponding DAP task.
```
...which implies that the domain of uniqueness for aggregation job IDs is at least as large as all aggregation job IDs for the related task.

But then, the aggregate continue request definition is (section 4.3.2.1):
```
struct {
  AggregationJobID job_id;
  PrepareStep prepare_shares<1..2^16-1>;
} AggregateContinueReq;
```
... which implies that aggregation job IDs must be unique over all aggregation jobs (on any task), since the `task_id` is omitted.

Depending on intent, I suggest one of the following edits:
* Update section 4.3.1.1 to state that aggregation job IDs must be unique over all aggregation jobs (for any task the aggregator knows about).
* Update section 4.3.2.1 to include a `TaskId` in the `AggregateContinueReq` message.

BranLwyd:
See https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/issues/260 for the
reasoning behind this change. In short, this keeps aggregation job IDs
scoped to be unique-per-task rather than universally-unique.

Note this isn't the only possible solution -- I'm opinionatedly sending this PR, if the discussion on #260 leads to a different conclusion, I'll update the PR accordingly.

Should this change be communicate with the WG mailing list?

Closes #260.

### DIFF
--- a/draft-ietf-ppm-dap.md
+++ b/draft-ietf-ppm-dap.md
@@ -1013,6 +1013,7 @@ structured as follows:
 
 ~~~
 struct {
+  TaskID task_id;
   AggregationJobID job_id;
   PrepareStep prepare_shares<1..2^16-1>;
 } AggregateContinueReq;


chris-wood (approved):
Indeed -- this was a bug. The task_id should have been included, just as it is in the initialization request. Good catch!

PR #265, opened 2022-05-23
Switch all messages from the "message" to the "application" media type.

### LINKED ISSUE #264: Consider unifying the top-level media type used for DAP messages.
Currently, DAP uses the following media types (defined in section 7.1):
* `application/dap-hpke-config`
* `message/dap-report`
* `message/dap-aggregate-initialize-req`
* `message/dap-aggregate-initialize-resp`
* `message/dap-aggregate-continue-req`
* `message/dap-aggregate-continue-resp`
* `message/dap-aggregate-share-req`
* `message/dap-aggregate-share-resp`
* `message/dap-collect-req`
* `message/dap-collect-resp`

Note that the first uses the `application` top-level type, while the remainder use the `message` top-level type. For consistency, I suggest we use the same top-level type for all messages.

Which type is more appropriate? Per [RFC 2046](https://www.rfc-editor.org/rfc/rfc2046.html):
* `application` is for "either uninterpreted binary data or information to be processed by an application"
* `message` is for "an encapsulated message"; all of the motivating examples (see section 5.2) are around encaspulating one message format inside another, or fragmenting a message into multiple chunks. [RFC 6838](https://www.rfc-editor.org/rfc/rfc6838.html) provides further guidance that "multipart and message are composite types; that is, they provide a means of encapsulating zero or more objects, each one a separate media type."

Given that DAP messages are not a method of encapsulating a message of a different type, I suggest that `application` is the more appropriate top-level type to use. This is also supported by the rather large number of `application` media types vs the relatively tiny number of `message` media types in [IANA's media types list](https://www.iana.org/assignments/media-types/media-types.xhtml).

(this issue fell out of discussion on https://github.com/divviup/janus/pull/183)

BranLwyd:
Closes #264. (opinionatedly: please see discussion on that issue)

### DIFF
--- a/draft-ietf-ppm-dap.md
+++ b/draft-ietf-ppm-dap.md
@@ -1800,15 +1800,15 @@ This specification defines the following protocol messages, along with their
 corresponding media types types:
 
 - HpkeConfig {{hpke-config}}: "application/dap-hpke-config"
-- Report {{upload-request}}: "message/dap-report"
-- AggregateInitializeReq {{collect-flow}}: "message/dap-aggregate-initialize-req"
-- AggregateInitializeResp {{collect-flow}}: "message/dap-aggregate-initialize-resp"
-- AggregateContinueReq {{collect-flow}}: "message/dap-aggregate-continue-req"
-- AggregateContinueResp {{collect-flow}}: "message/dap-aggregate-continue-resp"
-- AggregateShareReq {{collect-flow}}: "message/dap-aggregate-share-req"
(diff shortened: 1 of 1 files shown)


cjpatton (approved):
Approving based on the text quoted by @BranLwyd in the linked issue. It would be great to get feedback from folks who have more experience working on HTTP-based applications.

divergentdave (approved):

chris-wood (approved):
As a non-HTTP expert, I agree with the rationale. Thanks, @BranLwyd!

PR #269, opened 2022-06-09
Pretty pictures

martinthomson:
Anyone running builds locally will need a few extra pieces of software to manage this, but the result is something like this:

image

...with the appropriate palette inversion for people using dark mode.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,6 +9,8 @@
 .tags
 .targets.mk
 /*-[0-9][0-9].xml
+/node_modules/
+/package-lock.json
 archive.json
 draft-ietf-ppm-dap.xml
 lib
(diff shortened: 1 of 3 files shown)


martinthomson:
Oh, the software in question is 'svgcheck' (installed with pip3/python) and 'aasvg' (installed with npm/nodejs).

cjpatton:
I'm curious if there are any other drafts adopting this kind of thing? Happy to help blaze a trail if not!

martinthomson:
There is a small change I can make to have this work more cleanly.  Then you only need npm installed, not the specific package.

cjpatton:
I'd like to revive this PR, I'm willing to figure out the tooling.

martinthomson:
I'll let you rebase this PR then.  That's all that needs to happen, unless you want to tag a few other pictures and tweak them so that they look OK in SVG.

Issue #272, opened 2022-06-15
Increase Ciphertext size limit


wangshan:
Current HPKE Ciphertext size is limited to 2 bytes, about 65KB, many use cases will need more than this, Prio for eg. doesn't have this limit. Can we change this to 4 bytes  so the size can be `<1..2^32-1>`?

```
struct {
  HpkeConfigId config_id;    // config ID
  opaque enc<1..2^16-1>;     // encapsulated HPKE key
  opaque payload<1..2^16-1>; // ciphertext
} HpkeCiphertext;
```

wangshan:
@chris-wood will do.
Use case wise, 65KB can only represent 4B int vector of 16K dimension, if we encode correlations in individual reports then it can easily be breached, for eg. if I want to see histogram of errors by locales, let's say I have 100 error codes and 200 locales, that's already 20K dimension in one hot encoding.

PR #279, opened 2022-06-24
Increase payload size in HpkeCiphertext to 2^32-1

wangshan:
See discussion here: https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/issues/272

### DIFF
--- a/draft-ietf-ppm-dap.md
+++ b/draft-ietf-ppm-dap.md
@@ -606,7 +606,7 @@ struct {
   TaskID task_id;
   Nonce nonce;
   Extension extensions<0..2^16-1>;
-  HpkeCiphertext encrypted_input_shares<1..2^16-1>;
+  HpkeCiphertext encrypted_input_shares<1..2^32-1>;
 } Report;
 ~~~
 
@@ -815,7 +815,7 @@ struct {
   TaskID task_id;
   AggregationJobID job_id;
   opaque agg_param<0..2^16-1>;
-  ReportShare report_shares<1..2^16-1>;
+  ReportShare report_shares<1..2^32-1>;
 } AggregateInitializeReq;
 ~~~
 
@@ -1141,7 +1141,7 @@ and a body consisting of a `CollectResp`:
 
 ~~~
 struct {
-  HpkeCiphertext encrypted_agg_shares<1..2^16-1>;
+  HpkeCiphertext encrypted_agg_shares<1..2^32-1>;
 } CollectResp;
 ~~~
 
--- a/draft-ietf-ppm-dap.md
+++ b/draft-ietf-ppm-dap.md
@@ -481,7 +481,7 @@ uint8 HpkeConfigId;
 struct {
   HpkeConfigId config_id;    // config ID
   opaque enc<1..2^16-1>;     // encapsulated HPKE key
-  opaque payload<1..2^32-1>; // ciphertext
+  opaque payload<1..2^24-1>; // ciphertext
 } HpkeCiphertext;
 ~~~
 
--- a/draft-ietf-ppm-dap.md
+++ b/draft-ietf-ppm-dap.md
@@ -481,7 +481,7 @@ uint8 HpkeConfigId;
 struct {
   HpkeConfigId config_id;    // config ID
   opaque enc<1..2^16-1>;     // encapsulated HPKE key
-  opaque payload<1..2^24-1>; // ciphertext
+  opaque payload<1..2^32-1>; // ciphertext
 } HpkeCiphertext;
 ~~~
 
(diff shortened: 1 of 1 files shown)


draft-ietf-ppm-dap.md
```diff
@@ -481,7 +481,7 @@ uint8 HpkeConfigId;
 struct {
   HpkeConfigId config_id;    // config ID
   opaque enc<1..2^16-1>;     // encapsulated HPKE key
-  opaque payload<1..2^16-1>; // ciphertext
+  opaque payload<1..2^32-1>; // ciphertext
```
cjpatton, line 484: I think a 4-byte length prefix is overkill, given that we're never going to have a 4GB `payload` here. Can we change this to a 3-byte prefix so that we don't have a dead byte on the wire?
```suggestion
  opaque payload<1..2^24-1>; // ciphertext
```

cjpatton:
Also, note that this will require bumping the length prefix in other fields as well.

draft-ietf-ppm-dap.md
wangshan: 3 bytes can only support up to 16MB, it's not impossible to need more than that in some use cases, especially if we go beyond histogram collecting. Having said that, I think that will be rare cases.

chris-wood: We're talking abut a single byte here, so I would just do 4 bytes. That would also keep implementations simpler (`uint32_t` is everywhere, but `uint24_t` is not really a thing).

wangshan:
@cjpatton any pointer to where I should look at in particular? From what I can see only the definition of HpkeCiphertext specifies length of encrypted payload, the other fields like `Report.encrypted_input_shares` meant the number of HpkeCiphertext in an array, unless I misunderstood something.

divergentdave:
Per the [TLS presentation language](https://www.rfc-editor.org/rfc/rfc8446#section-3.4), and its definition of variable-length vectors, the numbers are in angle brackets, such as in `HpkeCiphertext encrypted_input_shares<1..2^16-1>` are numbers of bytes, not numbers of vector elements. Thus, we would need to recursively increase the maximum length of each vector that contains any `HpkeCiphertext` to realize benefits.

wangshan:
@cjpatton I've updated the arrays that contain ciphertext to 2^32-1, if I understand correctly, this allows those arrays to contain up to 256 ciphertext with 2^16-1 payload size. However, the number of ReportShare in AggInit request would be smaller than 256 due to ReportShare itself contains more than just ciphertext.

chris-wood:
> I've updated the arrays that contain ciphertext to 2^32-1, if I understand correctly, this allows those arrays to contain up to 256 ciphertext with 2^16-1 payload size.

The <1..2^32-1> syntax means that the size of the contents in the vector can be up to 2^32-1 bytes in length. It places no constraints on how those bytes are interpreted.

draft-ietf-ppm-dap.md
```diff
@@ -481,7 +481,7 @@ uint8 HpkeConfigId;
 struct {
   HpkeConfigId config_id;    // config ID
   opaque enc<1..2^16-1>;     // encapsulated HPKE key
-  opaque payload<1..2^16-1>; // ciphertext
+  opaque payload<1..2^24-1>; // ciphertext
```
chris-wood, line 484:
```suggestion
  opaque payload<1..2^32-1>; // ciphertext
```

chris-wood (approved):

chris-wood (changes requested):
Oh, wait, don't we also need to increase the length for prep shares (for the VDAF) in `AggregateContinueReq` and the response?

wangshan:
> Oh, wait, don't we also need to increase the length for prep shares (for the VDAF) in `AggregateContinueReq` and the response?

Do we need to? the `AggregateContinueReq` stores PrepareStep which has no direct relationship with Ciphertext. Are we saying because `AggregateInitializeReq` can have 2^32-1, to keep the index aligned `AggregateContinueReq` must also be 2^32-1?

wangshan:
@chris-wood @cjpatton I changed all types that contain PrepStep and ReportShare to 2^32-1, also taking @chris-wood 's advice to keep 2^32 everywhere by bumping ciphertext to 2^32 as well.

wangshan:
@chris-wood @cjpatton Something I'm slightly confused about TLS: the current commit has both payload and encrypted_input_shares at 2^32-1, this is not consistent because when payload is at 2^32-1, then the payload cannot possibly fit in to encrypted_input_shares, which also has a lower bound of 1. I'm not sure if I'm interpreting the TLS representation correctly.

chris-wood (approved):

chris-wood:
Sorry for not replying, @wangshan. In general, we're pretty loose with these bounds. They're not exact. For example, the extensions slot holds up to 2^16 bytes, where each extension itself carries up to 2^16 bytes. Implementations typically just make sure things fit within the bounds of each field when creating and encoding messages. 

All that said, this change looks good. Thanks!

PR #281, opened 2022-06-27
Make report timestamps more granular

### LINKED ISSUE #274: Consider supporting coarse-grained timestamp in Nonce
Not sure if DAP enforces precision of `Nonce.Time`:

```
```

But with fine-grained timestamp, it is easier for attacker to associate a report with the user sent that report. For example, if timestamp is accurate to second, and leader is compromised, attacker can compare users that uploaded at a given second (perhaps from a routing server) and reports' timestamp, then figure out if a particular user has contributed to a task. With a coarse-grained timestamp, however, we can minimise this threat with a secure routing server. Anti-replay can still be satisfied with the UUID part of Nonce.

chris-wood:
Closes #274.

We might want to consider renaming `Nonce` to something based on what it's actually doing, which is helping with replay protection. Maybe `ReportID`?


draft-ietf-ppm-dap.md
```diff
@@ -459,11 +459,10 @@ struct {
 } Interval;
 
 /* A nonce used to uniquely identify a report in the context of a DAP task. It
-includes the time at which the report was generated and a random, 64-bit
-integer. */
+includes the timestamp of the current batch and a random 16-byte value. */
 struct {
   Time time;
```
cjpatton, line 464: What about moving `time` into `Report`? Then you could modify Report by changing the type of `nonce` to `uint8 [16]`.

chris-wood: That would be another way to address the name issue, yeah. But we can do that separately.

tgeoghegan (commented):
If we require clients to round `Nonce.time` down to the nearest `min_batch_duration` multiple, then that means the vast majority of `Nonce` values are now invalid and should be rejected by servers (either that or we'd have to specify what servers should do if `Nonce.time` isn't a multiple) (what I mean here is that for `min_batch_duration = 10`, `Nonce.time = 100` is OK, but `Nonce.time = 103` is illegal and meaningless).

What if we replaced `Nonce.time` with something like `Nonce.batch_unit`, which is an integer representing how many `min_batch_duration`-length intervals have elapsed since the epoch? i.e., if `min_batch_duration = 10` and my report is generated at second 104, then instead of `Nonce.time = 104`, I do `Nonce.batch_unit = 10`?

Besides making invalid values impossible to represent, this would extend the range of DAP nonces much farther into the future, or we could choose to save some bytes on the wire by making `Nonce.batch_unit` narrower.

chris-wood:
> If we require clients to round Nonce.time down to the nearest min_batch_duration multiple, then that means the vast majority of Nonce values are now invalid and should be rejected by servers (either that or we'd have to specify what servers should do if Nonce.time isn't a multiple). 

Wait, what? Replay is enforced on the basis of Nonce in its entirety, which currently consists of a (1) timestamp and (2) a random value.

tgeoghegan (approved):
Chris W. and I hashed this out [on the IETF Slack](https://ietf.slack.com/archives/C02P9620XT6/p1656604238052029). Chris convinced me that as written, the rounding of `Nonce.time` is optional for the client, and aggregators don't need to do anything special to handle unrounded nonces: if that happens aggregators can still use the unrounded nonce to correctly identify a bucket to put the report in, so this is something that a client should do in order to improve privacy, but nothing breaks if they fail to. With that in mind, I don't think we need new text to explain what aggregators should do with nonces that aren't rounded correctly, nor does the new text for the client need a MUST.

Issue #267, opened 2022-05-30
Specify handling of agg_param


simon-friedberger:
There is an agg_param in AggregateInitializeReq and in CollectReq and in AggregateShareReq.
IIUC CollectReq.agg_param may need to be forwarded as AggregateInitializeReq.agg_param or - depending on the VDAF - later as AggregateShareReq.agg_param.

ISTM the logic needs to be specified to implement DAP correctly.

cjpatton:
Yes, both the agg_param and batch_interval need to match. Is there anything that needs to change in order to make this so?

simon-friedberger:
Well, if the agg_param in all three (CollectReq, AggregateInitializeReq, AggregateShareReq) must be the same the leader cannot start aggregating before the collector sends the request. Which is an optimization that was requested, right?

I think there are two valid scenarios:
1. AggregateInitializeReq.agg_param can be determined by the leader, e.g. because it is empty. In this case the leader can start aggregating ahead of time, without a request. Later AggregateShareReq.agg_param will be set to CollectReq.agg_param.
2. AggregateInitializeReq.agg_param does need information from the collector. So it is set to CollectReq.agg_param. Presumably AggregateShareReq.agg_param can then be empty.

cjpatton:
> Well, if the agg_param in all three (CollectReq, AggregateInitializeReq, AggregateShareReq) must be the same the leader cannot start aggregating before the collector sends the request. Which is an optimization that was requested, right?
>
> I think there are two valid scenarios:
> 
>     1. AggregateInitializeReq.agg_param can be determined by the leader, e.g. because it is empty. In this case the leader can start aggregating ahead of time, without a request. Later AggregateShareReq.agg_param will be set to CollectReq.agg_param.
> 
>     2. AggregateInitializeReq.agg_param does need information from the collector. So it is set to CollectReq.agg_param. Presumably AggregateShareReq.agg_param can then be empty.

Right, these are the two valid scenarios. For (2.) the current use case is Prio3, where the agg_param is always "" and aggregation can begin prior to receiving a collect request. For (1.) the envisioned use case is Poplar1, where the agg parameter encodes a set of candidate prefixes. In general these cannot be known until the agg result for a previous Poplar1 evaluation has been collected.

simon-friedberger:
So we need a flag in the Task to determine which scenario applies, right?

cjpatton:
[...]

PR #282, opened 2022-06-29
Clarify the handling of `agg_param` to close https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/issues/267

simon-friedberger:


draft-ietf-ppm-dap.md
```diff
@@ -293,8 +293,8 @@ The basic unit of DAP is the "task" which represents a single measurement
 includes the following parameters:
 
 * The type of each measurement.
-* The aggregation function to compute (e.g., sum, mean, etc.) and an optional
-  aggregation parameter.
+* The aggregation function to compute (e.g., sum, mean, etc.) and optional
+  task-constant parameters for the aggregation function.
```
chris-wood, line 297: This parameter changes in Poplar, so I don't think this definition is correct.

draft-ietf-ppm-dap.md
```diff
@@ -827,8 +827,12 @@ The `nonce` and `extensions` fields of each ReportShare match that in the Report
 uploaded by the client. The `encrypted_input_share` field is the `HpkeCiphertext`
 whose index in `Report.encrypted_input_shares` is equal to the index of the aggregator
 in the task's `aggregator_endpoints` to which the AggregateInitializeReq is being sent.
-The `agg_param` field is an opaque, VDAF-specific aggregation parameter. The
-`job_id` parameter contains the leader's chosen AggregationJobID.
+The `agg_param` field is an opaque, VDAF-specific aggregation parameter which is constant per collection flow.
+Note that this is the same value as in `CollectReq` and `AggregateShareReq`.
+Bandwith can be optimized by referencing the value from `AggregateInitializeReq` in `AggregateShareReq`.
```
chris-wood, line 832: While tempting, I don't think this is good advice. The leader doesn't necessarily know that helper will know how to interpret `agg_param` in an `AggregateShareReq` as a reference. In any case, it seems like premature optimization to me, so I'd suggest removing it.

chris-wood (changes requested):
Beyond editorial and implementation recommendation issues, I think this needs to be aligned with Poplar before we land it.

draft-ietf-ppm-dap.md
simon-friedberger: That is actually the point I wanted to make here. Previously it looked like the aggregation parameter (the thing that changes in Poplar) was part of the task definition. Since it changes, that's not possible. But of course other things that do not change can be given here.

chris-wood: Yeah, the current language seems wrong too. Maybe we could say something like this?

```
* The aggregation function to compute (e.g., sum, mean, etc.) and a way to
  produce optional parameters for the aggregation function.
```

draft-ietf-ppm-dap.md
simon-friedberger: tbh I am not quite sure why we are sending it twice. Maybe elaborating on that here would be more useful.

draft-ietf-ppm-dap.md
simon-friedberger: Maybe we should just decomplicate things and remove it entirely and assume that this kind of thing is part of "the function".

cjpatton: I think removing it entirely is the right way to go.

draft-ietf-ppm-dap.md
cjpatton: The agg param is needed here because this the first time the Helper learns its value. The AggregateShareReq message isn't sent until later, once the batch is ready to be collected.

If we wanted to save bandwidth, we could consider replacing AggregateShareReq.agg_param with something like SHA-256(agg_param). However I agree that this is premature and can wait until someone has implemented Poplar1.

draft-ietf-ppm-dap.md
```diff
@@ -827,8 +826,11 @@ The `nonce` and `extensions` fields of each ReportShare match that in the Report
 uploaded by the client. The `encrypted_input_share` field is the `HpkeCiphertext`
 whose index in `Report.encrypted_input_shares` is equal to the index of the aggregator
 in the task's `aggregator_endpoints` to which the AggregateInitializeReq is being sent.
-The `agg_param` field is an opaque, VDAF-specific aggregation parameter. The
-`job_id` parameter contains the leader's chosen AggregationJobID.
+The `agg_param` field is an opaque, VDAF-specific aggregation parameter which is constant per collection flow.
```
chris-wood, line 829:
```suggestion
The `agg_param` field is an opaque, VDAF-specific aggregation parameter provided during a collection flow.
```

draft-ietf-ppm-dap.md
```diff
@@ -827,8 +826,11 @@ The `nonce` and `extensions` fields of each ReportShare match that in the Report
...
+Note that this is the same value as in `CollectReq` (see {{collect-init}} below) and `AggregateShareReq` (see {{collect-aggregate}} below).
```
chris-wood, line 830: Can we move this note down to these sections? That is, when describing `CollectReq`, we can place a reference back to this section. Likewise for `AggregateShareReq`.

chris-wood (commented):
I'd like to see some text moved down, but otherwise this is good pending one concrete suggestion.

draft-ietf-ppm-dap.md
simon-friedberger: Done.

PR #284, opened 2022-07-05
report-replayed: allow varying aggregation parameter.

### LINKED ISSUE #283: Clarify language around `report-replayed` report share error
Currently, the `report-replayed` report share error is returned "if the report has already been aggregated" ([specification reference](https://ietf-wg-ppm.github.io/draft-ietf-ppm-dap/draft-ietf-ppm-dap.html#name-input-share-validation)).

Some VDAFs, like `poplar1`, will require repeatedly aggregating the same client reports with a varying aggregation parameter. I think the specification language would effectively block this use-case currently.

I think the fix is to return a `report-replayed` error if the report has already been aggregated *with the same aggregation parameter*. Note that this change would not allow a malicious leader to potentially aggregate a different set of reports with a different aggregation parameter: the existing `batch-collected` error would be returned if a client attempted to upload another report after the first collection attempt.

(Given the small size of the proposed fix, I'll send a PR shortly, but discussion on this issue/the WG mailing list takes precedence -- I'll update the PR to follow any changes from discussion.)

BranLwyd:
Previously, report-replayed would be returned if the same client report
was aggregated more than once. However, some VDAFs (e.g. poplar1) will
require repeatedly aggregating a given client report with varying
aggregation parameters. This commit relaxes the condition on which
report-replayed will be returned to allow aggregating the same client
report multiple times, as long as each aggregation has a distinct
aggregation parameter.

Closes #283.

### DIFF
--- a/draft-ietf-ppm-dap.md
+++ b/draft-ietf-ppm-dap.md
@@ -939,12 +939,12 @@ The validation checks are as follows.
    is marked as invalid with the error `batch-collected`. This prevents
    additional reports from being aggregated after its batch has already
    been collected.
-2. Check if the report has already been aggregated. If this check fails,
-   the input share is marked as invalid with the error `report-replayed`.
-   This is the case if the report was used in a previous aggregate request
-   and is therefore a replay. An aggregator may also choose to mark an
-   input share as invalid with the  error `report-dropped` under the conditions
-   prescribed in {{anti-replay}}.
+2. Check if the report has already been aggregated with this aggregation
+   parameter. If this check fails, the input share is marked as invalid with
+   the error `report-replayed`. This is the case if the report was used in a
+   previous aggregate request and is therefore a replay. An aggregator may also
+   choose to mark an input share as invalid with the  error `report-dropped`
+   under the conditions prescribed in {{anti-replay}}.
 
 If both checks succeed, the input share is not marked as invalid.
 


draft-ietf-ppm-dap.md
```diff
@@ -939,12 +939,12 @@ The validation checks are as follows.
    is marked as invalid with the error `batch-collected`. This prevents
    additional reports from being aggregated after its batch has already
    been collected.
-2. Check if the report has already been aggregated. If this check fails,
-   the input share is marked as invalid with the error `report-replayed`.
-   This is the case if the report was used in a previous aggregate request
-   and is therefore a replay. An aggregator may also choose to mark an
-   input share as invalid with the  error `report-dropped` under the conditions
-   prescribed in {{anti-replay}}.
+2. Check if the report has already been aggregated with this aggregation
```
BranLwyd, line 942: Note: the reference to an "aggregation parameter" in this section is new. However, I think it's OK: this section is referenced from the Leader Initialization & the Helper Initialization sections. The Leader Initialization section already assumes an aggregation parameter, and the Helper Initialization section explicitly receives an aggregation parameter from the leader's request. So I think this reference is well-formed.

divergentdave:
There is a separate check, under the section "Validating Batch Parameters", that each report is used in at most `max_batch_lifetime` aggregate shares. This should be set to 1 for Prio3 and VDAFs like you describe. If `max_batch_lifetime` for a task is 1, then the present change will make no difference, but if `max_batch_lifetime` for a task is greater than one, we need this change in order to allow that second batch.

chris-wood (approved):
Thanks @BranLwyd -- this change LGTM.

PR #286, opened 2022-07-08
Align draft with draft-irtf-cfrg-vdaf-01

### LINKED ISSUE #266: Align spec with vdaf-01
The next VDAF draft is out. It includes some minor changes that we'll need to adopt here.
https://datatracker.ietf.org/doc/draft-irtf-cfrg-vdaf/01/

cjpatton:
Closes #266.


draft-ietf-ppm-dap.md
```diff
@@ -962,24 +960,36 @@ Each aggregator runs this procedure for a given input share with corresponding n
 as follows:
 
 ~~~
-prep_state = VDAF.prep_init(vdaf_verify_param, agg_param, nonce, input_share)
+prep_state = VDAF.prep_init(vdaf_verify_key,
+                            server_role - 0x02,
```
tgeoghegan, line 964: I think the `- 0x02` is a bit misleading here since we explicitly spell out that it's `0x00` and `0x01` for leader and helper, respectively, and don't make it clear that `server_role` is a value from the `Role` enum.

Why does this differ from the role values used elsewhere at all? The description of [`Vdaf.prep_init`](https://github.com/cfrg/draft-irtf-cfrg-vdaf/blob/main/draft-irtf-cfrg-vdaf.md#preparation-sec-vdaf-prepare) only seems to care that the `agg_id` be unique, not that it start at 0.

tgeoghegan (changes requested):

draft-ietf-ppm-dap.md
```diff
@@ -517,10 +517,8 @@ of the aggregators is configured with following parameters:
 
 * `collector_config`: The {{!HPKE=RFC9180}} configuration of the collector (described in
   {{hpke-config}}); see {{compliance}} for information about the HPKE configuration algorithms.
-* `vdaf_verify_param`: The aggregator's VDAF verification parameter output by
-  the setup algorithm computed jointly by the aggregators before the start of the
-  DAP protocol {{?VDAF=I-D.draft-irtf-cfrg-vdaf-00}}). [OPEN ISSUE: This is yet
-  to be specified. See issue#161.]
+* `vdaf_verify_key`: The VDAF verification key shared by the aggregators. This
+  key is used in the aggregation sub-protocol ({{aggregate-flow}}).
```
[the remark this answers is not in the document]
cjpatton: Good call, done.

draft-ietf-ppm-dap.md
cjpatton: hmmm yeah, I was trying to avoid introducing `agg_id` here, as doing so seemed overly complex. Ideally we would not have to subtract `0x02`, as noted in the OPEN ISSUe below.

tgeoghegan: Overall I think we should just use the values `0x02` and `0x03` here, but so long as there's an OPEN ISSUE about this I'm happy. For now my vote would be to delete the ` - 0x02`, since I think the description of `server_role` below makes this explicit, and right now it could be read as requiring `0x00 - 0x02` for the leader and `0x01 - 0x02` for the helper.

cjpatton: That would make it incompatible with VDAF-01, unfortunately. Indexing needs to start at 0.

BranLwyd: As currently written, this PR does suggest that this parameter take `0x00 - 0x02 = -0x02` for leader and `0x01 - 0x02 = -0x01` for helper (assuming signed math in both cases). I think we need to get rid of either the subtraction of `0x02` in the highlighted line of the code block, or modify the text in the following paragraph to suggest using `0x02` for leader and `0x03` for helper, in order to ultimately map this parameter to either `0x00` or `0x01` as intended.

(FWIW, I support modifying the DAP spec to make leader & helper `0x00` & `0x01` directly, but that's probably a change larger than the scope of this PR.)

tgeoghegan: For some historical context: the reason that the DAP `Role` enum has `Leader = 0x02, Helper = 0x03` is in order to allow for additional helpers whose `Role` would naturally be `0x04`, `0x05`, etc. If we go to `Leader = 0x00, Helper = 0x01`, then we are kinda-sorta committing DAP to never having more than two aggregators, or at least to having this peculiar discontinuity in aggregator role values.

Anyway, for this PR, let's strike the `- 0x02` for clarity.

cjpatton: Ohhh yeah, sorry. The parenthetical was wrong, which is what is misleading. I've replaced `server_role - 0x02` with `agg_id`, which should at least be unambiguous. It would be nice to change `Role` to match, but we can do that later.

BranLwyd (approved):

chris-wood (approved):

PR #287, opened 2022-07-19
Add `report_count` to `CollectResp`

divergentdave:
This adds a field to collect responses with the number of reports that contributed to the aggregation. Collectors will likely want this count to aid in the interpretation of aggregate results. (see #163) There is also a change in draft-irtf-cfrg-vdaf-02 that opens the door for VDAFs that would require the report count during unsharding. (see cfrg/draft-irtf-cfrg-vdaf#95)

Note that while this field is not part of the adjacent `HpkeCiphertext`s, it will be protected by to the confidentiality of HTTPS and the leader-collector bearer token authentication.

### DIFF
--- a/draft-ietf-ppm-dap.md
+++ b/draft-ietf-ppm-dap.md
@@ -1159,11 +1159,13 @@ and a body consisting of a `CollectResp`:
 
 ~~~
 struct {
+  uint64 report_count;
   HpkeCiphertext encrypted_agg_shares<1..2^32-1>;
 } CollectResp;
 ~~~
 
-The `encrypted_agg_shares` field is the vector of encrypted aggregate shares.
+* `report_count` is the number of reports included in the aggregation.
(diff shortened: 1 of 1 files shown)


divergentdave:
I'm open to pushing it into the ciphertext, but my thinking was that a tampered report_count would only affect robustness, not privacy, and thus we only need to arrive at the correct report_count when all aggregators follow the protocol.

cjpatton (approved):
I agree with David, a binding between `report_count` and the ciphertext is not essential. If we wanted this for defense-in-depth, an alternative to sticking it in the ciphertext is to stick it in the AAD. This seems slightly easier to implement to me.

chris-wood (approved):

chris-wood:
@divergentdave can you please rebase?

cjpatton (approved):

tgeoghegan (approved):
OK, seems reasonable to me.

Issue #288, opened 2022-07-21
Polling collect URI changes the Leader's state


cjpatton:
When the Collector issues a collect request to the Leader, the Leader responds with a URI the Collector can poll later on to get the (encrypted) aggregate result. This is done via an HTTP GET to the collect URI returned by the Leader.

My understanding is that the Leader is free to remove the aggregate results from storage thereafter. (See https://github.com/cloudflare/daphne/issues/53 for discussion.) The spec doesn't require this request to be authenticated, so it's possible for someone to cause the Leader to delete this state before the Collector has a chance to fetch it.

This issue can be solved by:
1. requiring the Leader to store results for some period of time;
2. making the collect URI hard to predict (e.g., by sticking a random number in the URI string); or
2. requiring authentication for this endpoint.

(2.) seems easiest to implement, but may be tricky to spell. Any thoughts on this?

tgeoghegan:
> My understanding is that the Leader is free to remove the aggregate results from storage thereafter.

That's not what [section 4.4.1](https://ietf-wg-ppm.github.io/draft-ietf-ppm-dap/draft-ietf-ppm-dap.html#name-collection-initialization) says:

> The leader MUST retain a collect job's results until the collector sends an HTTP DELETE request to the collect job URI, in which case the leader responds with HTTP status 204 No Content.
>
> [OPEN ISSUE: Allow the leader to drop aggregate shares after some reasonable amount of time has passed, but it's not clear how to specify that. ACME doesn't bother to say anything at all about this when describing how subscribers should fetch certificates: https://datatracker.ietf.org/doc/html/rfc8555#section-7.4.2]

That text is consistent with [@chris-wood's idea](https://github.com/cloudflare/daphne/issues/53#issuecomment-1191499038). So I argue the current situation is (1). We do have to solve the problem of how long a leader can wait before tossing collect results -- or maybe we don't, since ACME doesn't bother to. But we could use this issue to discuss that question.

re: (2): I think we should avoid as much as possible imposing requirements on the construction of URIs and relative paths[^1]. An implementation should feel free to generate unpredictable collect job URIs (Janus randomly generates a v4 UUID) but I don't think DAP should mandate it, just so we don't have to write the text to spell it out.

[...]

PR #292, opened 2022-07-26
Allow leader to remove collect results without an explicit DELETE.

BranLwyd:
The previous wording required the leader to store collect results until
the collector sent an explicit DELETE request to the collect job's URI.
Now, the relevant text is reworded to allow the leader to remove collect
results for other reasons (for example, a leader can now garbage-collect
old results after a suitable period of time).

This fell out of discussion on #288.

### DIFF
--- a/draft-ietf-ppm-dap.md
+++ b/draft-ietf-ppm-dap.md
@@ -1164,7 +1164,7 @@ If obtaining aggregate shares fails, then the leader responds to subsequent HTTP
 GET requests to the collect job URI with an HTTP error status and a problem
 document as described in {{errors}}.
 
-The collector may send an HTTP DELETE request to the collect job URI, to which
+The collector can send an HTTP DELETE request to the collect job URI, to which
 the leader MUST respond with HTTP status 204 No Content. The leader MAY respond
 with HTTP status 204 No Content for requests to a collect job URI which has not
 received a DELETE request, for example if the results have been deleted due to
--- a/draft-ietf-ppm-dap.md
+++ b/draft-ietf-ppm-dap.md
@@ -1164,14 +1164,9 @@ If obtaining aggregate shares fails, then the leader responds to subsequent HTTP
 GET requests to the collect job URI with an HTTP error status and a problem
 document as described in {{errors}}.
 
-The leader MUST retain a collect job's results until the collector sends an HTTP
-DELETE request to the collect job URI, in which case the leader responds with
-HTTP status 204 No Content.
-
-[OPEN ISSUE: Allow the leader to drop aggregate shares after some reasonable
-amount of time has passed, but it's not clear how to specify that. ACME doesn't
-bother to say anything at all about this when describing how subscribers should
-fetch certificates: https://datatracker.ietf.org/doc/html/rfc8555#section-7.4.2]
+The leader MUST remove a collect job's results when the collector sends an HTTP
+DELETE request to the collect job URI. The leader responds with HTTP status 204
+No Content for requests to a collect job URI whose results have been removed.
 
 [OPEN ISSUE: Describe how intra-protocol errors yield collect errors (see
 issue#57). For example, how does a leader respond to a collect request if the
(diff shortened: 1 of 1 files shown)


draft-ietf-ppm-dap.md
```diff
@@ -1164,14 +1164,9 @@ If obtaining aggregate shares fails, then the leader responds to subsequent HTTP
...
-DELETE request to the collect job URI, in which case the leader responds with
-HTTP status 204 No Content.
-
-[OPEN ISSUE: Allow the leader to drop aggregate shares after some reasonable
-amount of time has passed, but it's not clear how to specify that. ACME doesn't
-bother to say anything at all about this when describing how subscribers should
-fetch certificates: https://datatracker.ietf.org/doc/html/rfc8555#section-7.4.2]
+The leader MUST remove a collect job's results when the collector sends an HTTP
+DELETE request to the collect job URI. The leader responds with HTTP status 204
+No Content for requests to a collect job URI whose results have been removed.
```
tgeoghegan, line 1169: I think this MUST should be rewritten to describe externally observable behavior. For instance:

```suggestion
The collector may send an HTTP DELETE request to the collect job URI, to which
the leader MUST respond with HTTP status 204 No Content. The leader MUST
respond to subsequent requests to the collect job URI with HTTP status 204 No
Content.
```

Note that this also specifies how the leader should respond to the DELETE, which is missing in the existing text. My intuition is that there's no way for other protocol participants to know whether or not the leader's storage still has some aggregate share. The only thing they can see is how the leader responds to a request.

Besides that: this text makes it clear what a leader MUST do if it gets an explicit DELETE, but I think we also need a MAY sentence making it clear that the leader can start respond to the collect URI with HTTP 204 _before_ a collector sends DELETE.

tgeoghegan (changes requested):

draft-ietf-ppm-dap.md
BranLwyd: OK -- I had left the `MAY` off since it's IMO implied by the updated text, but explicit is usually better than implicit.

draft-ietf-ppm-dap.md
```diff
@@ -1164,14 +1164,11 @@ If obtaining aggregate shares fails, then the leader responds to subsequent HTTP
...
-DELETE request to the collect job URI, in which case the leader responds with
-HTTP status 204 No Content.
-
-[OPEN ISSUE: Allow the leader to drop aggregate shares after some reasonable
-amount of time has passed, but it's not clear how to specify that. ACME doesn't
-bother to say anything at all about this when describing how subscribers should
-fetch certificates: https://datatracker.ietf.org/doc/html/rfc8555#section-7.4.2]
+The collector may send an HTTP DELETE request to the collect job URI, to which
+the leader MUST respond with HTTP status 204 No Content. The leader MAY respond
+with HTTP status 204 No Content for requests to a collect job URI which has not
```
simon-friedberger, line 1169: I think it would be nice to add a little "(e.g. because the data has been deleted due to age or lack of space)". It's a little confusing if you don't have that on your mind.

tgeoghegan (approved):
Approved as this change achieves the intended mechanical purpose, though I agree with Simon's point about clarity, and others more experienced in the RFC writing style might comment further.

BranLwyd:
OK -- I left the motivating example out as I didn't want to be too prescriptive to implementations. I would also be interested in comments from folks more experienced with RFC writing; I've added the motivating example for now.

chris-wood (approved):

draft-ietf-ppm-dap.md
```diff
@@ -1164,15 +1164,13 @@ If obtaining aggregate shares fails, then the leader responds to subsequent HTTP
 GET requests to the collect job URI with an HTTP error status and a problem
 document as described in {{errors}}.
 
-The leader MUST retain a collect job's results until the collector sends an HTTP
-DELETE request to the collect job URI, in which case the leader responds with
+The collector may send an HTTP DELETE request to the collect job URI, to which
```
chris-wood, line 1167:
```suggestion
The collector can send an HTTP DELETE request to the collect job URI, to which
```

Issue #180, opened 2022-01-15
Aggregators required to store nonce sets indefinitely


cjpatton:
Since #169, aggregators are required to store report nonces for batch intervals that have not yet been collected. This is necessary in order to prevent replay attacks. The problem is that nothing guarantees that a batch interval will be collected in a timely manner: In theory, it's possible that aggregators will have to store nonce sets indefinitely. To prevent this, there needs to be a way for the aggregators to agree on when to stop accepting reports for a given batch interval.

The strawman solution is to allow each aggregator to, at its own discretion, ignore a report if it can't confirm if the report was replayed or not. The problem is that this potentially leads to unnecessary data loss.

cjpatton:
I propose resolving this with #338.

PR #297, opened 2022-07-30
Add support for fixed-size tasks

### LINKED ISSUE #273: Collect without interval
If the collector is not interested in time and interval, but simply want to collect aggregation in a batch that meets `min_batch_size`. Then can the protocol support a 'batch-based collection' instead of current interval-based?

Consider a Prio usecase, when `max_batch_lifetime == 1`, collector will only need to collect the aggregation so far, with a batch size B >= `min_batch_size`. This can be orchestrated by the leader, which can implement a counting service to track batch size for a task, once it reaches `min_batch_size`, leader sends AggregateShareReq to collect helper's `aggregate_share` and return to collector.

This requires a new id to tie agg-flow with agg-share-flow. For example, in addition to `agg_job_id`, leader can send a unique `batch_id` in every AggregateReq. At collect time, leader use the same `batch_id` to collect `output_share` in helper (helper can still proactively aggregate output_shares to `aggregate_share`, since there are no more batch windows, helper can store `aggregate_share` by `agg_job_id`, or accumulate all aggregation jobs' `output_share` to one `aggregate_share`, and store it with `batch_id`). Illustrated as following:

```
|<------------------------ batch_id1 ----------------->| <== AggregationShareReq
|    agg_job_id1     |   agg_job_id2   |   agg_job_id3 | <== AggregateReq
T0                                                    Tm <== Time
```
Here [T0, Tm] is the time takes to aggregate `min_batch_size` number of reports, it has no relationship with `min_batch_duration` or collector interval.

As this issue pointed out: https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/issues/195, to avoid privacy leak, each batch window must contain at least `min_batch_size` reports, otherwise attacker can find ways to slice intervals to break privacy guarantee. But if the protocol does require each batch window meets `min_batch_size`, then the collect interval itself is no longer useful, since the duration that takes to meet `min_batch_size` is the smallest duration that can be queried. Therefore, it seems to make sense to base collection entirely on batch size, not interval.

cjpatton:
Closes #273.

Generalizes "batch_interval" to "query" and enumerate the different query
types and constraints that apply to each. The existing collect request
semantics is subsumed by the `time-interval` query type.

Adds a query type, `fixed-size`, that is designed to support use cases
in which the sample size needs to be tightly controlled. The main
feature of this query type is that the Aggregators enforce limits on the
size of each batch. We given them a little wiggle room in order to make
coordination of aggregation jobs a bit easier.

One notable feature of this change is that each tasks supports exactly
one query type. This allows us to punt on the coomplexity of having to
composing different query types per task. For example: If a report can
be batched into a chunk or a batch interval, then it's not clear that
it's possible for Aggregators to aggregate reports ahead of a collect
request.

### DIFF
--- a/draft-ietf-ppm-dap.md
+++ b/draft-ietf-ppm-dap.md
@@ -386,8 +386,8 @@ in the "type" field (within the DAP URN namespace
 | reportTooLate              | Report could not be processed because it arrived too late. |
 | reportTooEarly             | Report could not be processed because its timestamp is too far in the future. |
 | batchInvalid               | A collect or aggregate-share request was made with invalid batch parameters. |
-| insufficientBatchSize      | There are not enough reports in the batch interval to satisfy the task's minimum batch size. |
-| batchLifetimeExceeded      | The batch lifetime has been exceeded for one or more reports included in the batch interval. |
+| invalidBatchSize           | There are an invalid number of reports in the batch. |
+| batchLifetimeExceeded      | The batch lifetime has been exceeded for one or more reports included in the batch. |
 | batchMismatch              | Aggregators disagree on the report shares that were aggregated in a batch. |
 | unauthorizedRequest        | Authentication of an HTTP request failed (see {{https-sender-auth}}). |
 
@@ -458,12 +458,8 @@ struct {
   Duration duration;
 } Interval;
 
-/* A nonce used to uniquely identify a report in the context of a DAP task. It
-includes the timestamp of the current batch and a random 16-byte value. */
-struct {
-  Time time;
-  uint8 rand[16];
-} Nonce;
+/* A nonce used to uniquely identify a report in the context of a DAP task. */
+Nonce uint8[16];
 
 /* The various roles in the DAP protocol. */
 enum {
@@ -484,6 +480,124 @@ struct {
 } HpkeCiphertext;
 ~~~
 
+## Queries {#query}
+
+Aggregated results are computed based on sets of report, called batches. The
+Collector influences which reports are used in a batch via a "query." The
+Aggregators use this query to carry out the aggregation flow and produce
+aggregate shares encrypted to the Collector.
+
+This document defines the following query types:
+
+~~~
+enum {
+   reserved(0), // Reserved for testing purposes
+   time-interval(1),
+   fixed-size(2),
+   (65535)
+} QueryType;
+~~~
+
+The time-interval query type is described in {{time-interval-query}}; the
+fixed-size query type is described in {{fixed-size-query}}. Future
+specifications can introduce new query types as needed. A query includes
+parameters used by the Aggregators to select a batch of reports specific to the
+given query type. A query is defined as follows:
+
+~~~
+enum {
+    next-batch(0),
+    by-batch-id(1),
+} FixedSizeQueryType;
+
+opaque BatchId[32];
+
+struct {
+    FixedSizeQueryType fixed_size_query_type;
+    select (fixed_size_query_type) {
+        next-batch:  Empty;
+        by-batch-id: BatchId batch_id;
+    }
+} FixedSizeQuery;
+
+struct {
+    select (query_type) { // determined by query configuration
+        case time-interval: Interval batch_interval;
+        case fixed-size: FixedSizeQuery fixed_size_query;
+    }
+} Query;
+~~~
+
+The parameters pertaining to each query type are described in one of the
+subsections below. The query is issued in-band as part of the collect
+sub-protocol ({{collect-flow}}). Its content is determined by the "query type",
+which in turn is encoded by the "query configuration" configured out-of-band.
+All query types have the following configuration parameters in common:
+
+- `min_batch_size` - The smallest number of reports the batch is allowed to
+  include. In a sense, this parameter controls the degree of privacy that will
+  be obtained: The larger the minimum batch size, the higher degree of privacy.
+  However, this ultimately depends on the application and the nature of the
+  reports and aggregation function.
+
+- `time_precision` - Clients use this value to truncate their report timestamps;
+  see {{upload-flow}}. Additional semantics may apply, depending on the query
+  type. (See {{batch-validation}} for details.)
+
+The parameters pertaining to specific query types are described in the relevant
+subsection below.
+
+### Time-interval Queries {#time-interval-query}
+
+The first query type, `time-interval`, is designed to support applications in
+which reports are collected over a long period of time. The Collector specifies
+a "batch interval" that determines the time range for reports included in the
+batch: For each report in the batch, the time at which that report was generated
+(see {{upload-flow}}) must fall within the batch interval specified by the
+Collector.
+
+Typically the Collector issues queries for which the batch intervals are
+continuous, monotonically increasing, and have the same duration. For example,
+the sequence of batch intervals. For example, the sequence of batch intervals
+`(1659544000, 1000)`, `(1659545000, 1000)`, `(1659545000, 1000)`, `(1659546000,
+1000)` satisfies these conditions. (The first element of the pair denotes the
+start of the batch interval and the second denotes the duration.) Of course,
+there are cases in which Collector may need to issue queries out-of-order. For
+example, a previous batch might need to be queried again with a different
+aggregation parameter (e.g, for Poplar1). In addition, the Collector may need to
+vary the duration to adjust to changing report upload rates.
+
+### Fixed-size Queries {#fixed-size-query}
+
+The `fixed-size` query type is used to support applications in which the
+Collector needs the ability to strictly control the sample size. This is
+particularly important for controlling the amount of noise added to reports
+by Clients (or added to aggregate shares by Aggregators) in order to achieve
+differential privacy.
+
+To get the aggregate of the next batch, the Collector issues a query of type
+`next-batch`. (See FixedSizeQuery in {{query}}.) The response to this query (see
+{{collect-flow}}) includes a "batch ID", which can be used to query the same
+batch later on. This is done using the `by-batch-id` batch-query type.
+
+For this query type, the Aggregators group reports into arbitrary batches such
+that each batch has roughly the same number of reports. In addition to the
+minumum batch size common to all query types, the conifugration includes a
+"maximum batch size", `max_batch_size`, that determines maximum number of
+reports per batch.
+
+Implementation note: The goal for the Aggregators is to aggregate precisely
+`min_batch_size` reports per batch. Doing so, however, may be challenging
+for Leader deployments in which multiple, independent nodes running the
+aggregate sub-protocol (see {{aggregate-flow}}) need to be coordinated. The
+maximum batch size is intended to allow room for error. Typically the difference
+between the minimum and maximum batch size will be a small fraction of the
+target batch size for each batch.
+
+[OPEN ISSUE: It may be feasible to require a fixed batch size, i.e.,
+`min_batch_size == max_batch_size`. We should know better once we've had some
+implementation/deployment experience.]
+
 ## Task Configuration {#task-configuration}
 
 Prior to the start of execution of the protocol, each participant must agree on
@@ -502,13 +616,11 @@ number generator. Each task has the following parameters associated with it:
   leader's endpoint MUST be the first in the list. The order of the
   `encrypted_input_shares` in a `Report` (see {{upload-flow}}) MUST be the
   same as the order in which aggregators appear in this list.
+* The query configuration for this task (see {{query}}). This determines the
+  query type for batch selection and the properties that all batches for this
+  task must have.
 * `max_batch_lifetime`: The maximum number of times a batch of reports may be
-  used in collect requests.
-* `min_batch_size`: The minimum number of reports that appear in a batch.
-* `min_batch_duration`: The minimum time difference between the oldest and
-  newest report in a batch. This defines the boundaries with which the batch
-  interval of each collect request must be aligned. (See
-  {{batch-parameter-validation}}.)
+  queried by the Collector.
 * A unique identifier for the VDAF instance used for the task, including the
   type of measurement associated with the task.
 
@@ -605,46 +717,52 @@ structured as follows:
 
 ~~~
 struct {
+    Time time;
+    Nonce nonce;
+    Extension extensions<0..2^16-1>;
+} ReportMetadata;
+
+struct {
   TaskID task_id;
-  Nonce nonce;
-  Extension extensions<0..2^16-1>;
+  ReportMetadata metadata;
   HpkeCiphertext encrypted_input_shares<1..2^32-1>;
 } Report;
 ~~~
 
-This message is called the client's report. It contains the following fields:
+This message is called the Client's report. It consists of a "header" and the
+encrypted input share of each of the Aggregators. The header consists of the
+task ID and report "metadata". The metadata consists of the following fields:
 
-* `task_id` is the task ID of the task for which the report is intended.
-* `nonce` is the report nonce generated by the client. This field is used by the
-  aggregators to ensure the report appears in at most one batch. (See
-  {{anti-replay}}.)
-* `extensions` is a list of extensions to be included in the Upload flow; see
-  {{upload-extensions}}.
-* `encrypted_input_shares` contains the encrypted input shares of each of the
-  aggregators. The order in which the encrypted input shares appear MUST match
-  the order of the task's `aggregator_endpoints` (i.e., the first share should
-  be the leader's, the second share should be for the first helper, and so on).
+* A timestamp representing the time at which the report was generated.
+  Specifically, the `time` field is set to the number of seconds elapsed since
+  the start of the UNIX epoch. The client SHOULD round this value down to the nearest
+  multiple of `time_precision` in order to ensure that that the timestamp cannot
+  be used to link a report back to the Client that generated it.
 
-To generate the report, the client begins by initializing the report nonce.
-Specifically, the client first sets `Nonce.rand` to 16 random bytes output from
-a cryptographically secure random number generator. It then sets `Nonce.time`
-to the number of seconds elapsed since the start of the UNIX epoch, rounded down to the nearest multiple of `min_batch_duration`.
-This truncation is done to ensure that the report's timestamp cannot be used to
-link the report back to the originating client.
+* A nonce used by the Aggregators to ensure the report appears in at most one
+  batch. (See {{anti-replay}}.) The Client MUST generate this by generating 16
+  random bytes using a cryptographically secure random number gnerator.
 
-The client then finishes the report generation by sharding its measurement into
-a sequence of input shares as specified by the VDAF in use. To encrypt an input
-share, the client generates an {{HPKE}} ciphertext for the aggregator by running
+* A list of extensions to be included with the report. (See
+  {{upload-extensions}}.)
+
+To generate a report, the Client first shards its measurement into input shares
+as specified by the VDAF. It then encrypts each input share as follows:
 
 ~~~
-enc, payload = SealBase(pk, "dap-01 input share" || 0x01 || server_role,
-    task_id || nonce || extensions, input_share)
+enc, payload = SealBase(pk,
+    "dap-01 input share" || 0x01 || server_role,
+    task_id || metadata, input_share)
 ~~~
 
 where `pk` is the aggregator's public key; `server_role` is the Role of the
-intended recipient (`0x02` for the leader and `0x03` for the helper);
-`task_id`, `nonce`, and `extensions` are the corresponding fields of `Report`;
-and `input_share` is the aggregator's input share.
+intended recipient (`0x02` for the leader and `0x03` for the helper), `task_id`
+is the task ID, `metadata` is the report metadata, and `input_share` is the
+Aggregator's input share.
+
+The order of the encrypted input shares appear MUST match the order of the
+task's `aggregator_endpoints`. That is, the first share should be the leader's,
+the second share should be for the first helper, and so on.
 
 The leader responds to well-formed requests to `[leader]/upload` with HTTP status
 code 200 OK and an empty body. Malformed requests are handled as described in {{errors}}.
@@ -658,12 +776,11 @@ cached aggregator `HpkeConfig` and retry with a freshly generated Report. If
 this retried report does not succeed, clients MUST abort and discontinue
 retrying.
 
-The leader MUST ignore any report whose nonce contains a timestamp that falls in
-a batch interval for which it has received at least one collect request from the
-collector. (See {{collect-flow}}.) Otherwise, comparing the aggregate result to
-the previous aggregate result may result in a privacy violation. (Note that the
-helpers enforce this as well; see {{collect-flow}}.) In addition, the
-leader SHOULD abort the upload protocol and alert the client with error
+The Leader MUST ignore any report pertaining to a batch that has already been
+collected. (See {{input-share-batch-validation}} for details.) Otherwise,
+comparing the aggregate result to the previous aggregate result may result in a
+privacy violation. (Note that the Helpers enforce this as well.) In addition,
+the leader SHOULD abort the upload protocol and alert the client with error
 "reportTooLate".
 
 Leaders can buffer reports while waiting to aggregate them. The leader SHOULD
@@ -789,6 +906,7 @@ enum {
   hpke-unknown-config-id(3),
   hpke-decrypt-error(4),
   vdaf-prep-error(5),
+  batch-saturated(6),
 } ReportShareError;
 ~~~
 
@@ -814,8 +932,7 @@ structured as follows:
 
 ~~~
 struct {
-  Nonce nonce;
-  Extension extensions<0..2^16-1>;
+  ReportMetadata metadata;
   HpkeCiphertext encrypted_input_share;
 } ReportShare;
 
@@ -823,20 +940,48 @@ struct {
   TaskID task_id;
   AggregationJobID job_id;
   opaque agg_param<0..2^16-1>;
+  select (query_type) { // determined by task configuration
+    time-interval: Empty;
+    fixed-size: BatchId batch_id;
+  }
   ReportShare report_shares<1..2^32-1>;
 } AggregateInitializeReq;
 ~~~
 
-[[OPEN ISSUE: consider sending report shares separately (in parallel) to the aggregate instructions. RIght now, aggregation parameters and the corresponding report shares are sent at the same time, but this may not be strictly necessary. ]]
+[[OPEN ISSUE: Consider sending report shares separately (in parallel) to the
+aggregate instructions. Right now, aggregation parameters and the corresponding
+report shares are sent at the same time, but this may not be strictly
+necessary.]]
 
-The `nonce` and `extensions` fields of each ReportShare match that in the Report
-uploaded by the client. The `encrypted_input_share` field is the `HpkeCiphertext`
-whose index in `Report.encrypted_input_shares` is equal to the index of the aggregator
-in the task's `aggregator_endpoints` to which the AggregateInitializeReq is being sent.
-The `agg_param` field is an opaque, VDAF-specific aggregation parameter provided during a collection flow.
-The `job_id` parameter contains the leader's chosen AggregationJobID.
+This message consists of:
 
-[[OPEN ISSUE: Check that this handling of `agg_param` is appropriate when the definition of Poplar is done.]]
+* The ID of the task for which the reports will be aggregated.
+
+* The aggregation job ID.
+
+* The opaque, VDAF-specific aggregation parameter provided during the
+  collection flow (`agg_param`),
+
+  [[OPEN ISSUE: Check that this handling of `agg_param` is appropriate when the
+  definition of Poplar is done.]]
+
+* Information used by the Aggregators to determine how to aggregate each report:
+
+    * For fixed-size tasks, the Leader specifies a "batch ID" that determines
+      the batch to which each report for this aggregation job belongs.
+
+      [OPEN ISSUE: For fixed-size tasks, the Leader is in complete control over
+      which batch a report is included in. For time-interval tasks, the Client has
+      some control, since the timestamp determines which batch window it falls
+      in. Is this desirable from a privacy perspective? If not, it might be
+      simpler to drop the timestamp altogether and have the agg init request
+      specify the batch window instead.]
+
+* The sequence of report shares to aggregate. The `encrypted_input_share` field
+  of the report share is the `HpkeCiphertext` whose index in
+  `Report.encrypted_input_shares` is equal to the index of the aggregator in the
+  task's `aggregator_endpoints` to which the AggregateInitializeReq is being
+  sent.
 
 Let `[aggregator]` denote the helper's API endpoint. The leader sends a POST
 request to `[aggregator]/aggregate` with its AggregateInitializeReq message as
@@ -907,18 +1052,19 @@ abort with error "unrecognizedMessage".
 
 #### Input Share Decryption {#input-share-decryption}
 
-Each report share has a corresponding task ID, nonce, list of extensions, and encrypted
-input share. Let `task_id`, `nonce`, `extensions`, and `encrypted_input_share` denote these
-values, respectively. Given these values, an aggregator decrypts the input
-share as follows. First, the aggregator looks up the HPKE config and corresponding
-secret key indicated by `encrypted_input_share.config_id`. If not found, then it
-marks the report share as invalid with the error `hpke-unknown-config-id`.
-Otherwise, it decrypts the payload with the following procedure:
+Each report share has a corresponding task ID, report metadata (timestamp,
+nonce, and extensions), and encrypted input share. Let `task_id`, `metadata`,
+and `encrypted_input_share` denote these values, respectively. Given these
+values, an aggregator decrypts the input share as follows. First, the aggregator
+looks up the HPKE config and corresponding secret key indicated by
+`encrypted_input_share.config_id`. If not found, then it marks the report share
+as invalid with the error `hpke-unknown-config-id`. Otherwise, it decrypts the
+payload with the following procedure:
 
 ~~~
-input_share = OpenBase(encrypted_input_share.enc, sk, "dap-01 input share" ||
-    0x01 || server_role, task_id || nonce || extensions,
-    encrypted_input_share.payload)
+input_share = OpenBase(encrypted_input_share.enc, sk,
+    "dap-01 input share" || 0x01 || server_role,
+    task_id || metadata, encrypted_input_share.payload)
 ~~~
 
 where `sk` is the HPKE secret key, and `server_role` is the role of the
@@ -935,16 +1081,22 @@ The validation checks are as follows.
 
 1. Check if the report has never been aggregated but is contained by
    a batch that has been collected. If this check fails, the input share
-   is marked as invalid with the error `batch-collected`. This prevents
+   MUST be marked as invalid with the error `batch-collected`. This prevents
    additional reports from being aggregated after its batch has already
    been collected.
 2. Check if the report has already been aggregated with this aggregation
-   parameter. If this check fails, the input share is marked as invalid with
-   the error `report-replayed`. This is the case if the report was used in a
-   previous aggregate request and is therefore a replay. An aggregator may also
-   choose to mark an input share as invalid with the  error `report-dropped`
-   under the conditions prescribed in {{anti-replay}}.
-
+   parameter. If this check fails, the input share MUST be marked as invalid
+   with the error `report-replayed`. This is the case if the report was used in
+   a previous aggregate request and is therefore a replay. An aggregator may
+   also choose to mark an input share as invalid with the error
+   `report-dropped` under the conditions prescribed in {{anti-replay}}.
+3. Depending on the query type for the task, additional checks may be applicable:
+    * For fixed-size tasks, the Aggregators need to ensure that each batch is
+      roughly the same size. If the number of reports aggregated for the current
+      batch exceeds the maximum batch size (per the task configuration), the
+      Aggregator MAY mark the input share as invalid with the error
+      "batch-saturated". Note that this behavior is not strictly enforced here
+      but during the collect sub-protocol. (See {{batch-validation}}.)
 If both checks succeed, the input share is not marked as invalid.
 
 #### Input Share Preparation {#input-share-prep}
@@ -1086,12 +1238,12 @@ The helper then awaits the next message from the leader.
 
 ## Collecting Results {#collect-flow}
 
-In this phase, the collector requests aggregate shares from each aggregator
-and then locally combines them to yield a single, aggregate output. In particular,
-the collector asks the leader to collect and return the results for a given
-DAP task over a given time period. The aggregate shares are encrypted to the
-collector so that it can decrypt and combine them to yield the aggregate output.
-This entire process is composed of two interactions:
+In this phase, the Collector requests aggregate shares from each aggregator and
+then locally combines them to yield a single aggregate result. In particular,
+the Collector issues a query to the Leader ({{query}}), which the Aggregators
+use to select a batch of reports to aggregate. Each emits an aggregate share
+encrypted to the Collector so that it can decrypt and combine them to yield the
+aggregate result. This entire process is composed of two interactions:
 
 1. Collect request and response between the collector and leader, specified
    in {{collect-init}}
@@ -1115,7 +1267,7 @@ attacks.]
 ~~~
 struct {
   TaskID task_id;
-  Interval batch_interval;
+  Query query;
   opaque agg_param<0..2^16-1>; // VDAF aggregation parameter
 } CollectReq;
 ~~~
@@ -1123,27 +1275,27 @@ struct {
 The named parameters are:
 
 * `task_id`, the DAP task ID.
-* `batch_interval`, the request's batch interval.
+* `query`, the Collector's query.
 * `agg_param`, an aggregation parameter for the VDAF being executed.
   This is the same value as in `AggregateInitializeReq` (see {{leader-init}}).
 
 Depending on the VDAF scheme and how the leader is configured, the leader and
-helper may already have prepared all the reports falling within `batch_interval`
-and be ready to return the aggregate shares right away, but this cannot be
+helper may already have prepared a sufficient number of reports satisfying the
+query and be ready to return the aggregate shares right away, but this cannot be
 guaranteed. In fact, for some VDAFs, it is not be possible to begin preparing
 inputs until the collector provides the aggregation parameter in the
 `CollectReq`. For these reasons, collect requests are handled asynchronously.
 
 Upon receipt of a `CollectReq`, the leader begins by checking that the request
 meets the requirements of the batch parameters using the procedure in
-{{batch-parameter-validation}}. If so, it immediately sends the collector a
+{{batch-validation}}. If so, it immediately sends the collector a
 response with HTTP status 303 See Other and a Location header containing a URI
 identifying the collect job that can be polled by the collector, called the
 "collect job URI".
 
-The leader then begins working with the helper to prepare the shares falling
-into `CollectReq.batch_interval` (or continues this process, depending on the
-VDAF) as described in {{aggregate-flow}}.
+The leader then begins working with the helper to prepare the shares satisfying
+the query (or continues this process, depending on the VDAF) as described in
+{{aggregate-flow}}.
 
 After receiving the response to its CollectReq, the collector makes an HTTP GET
 request to the collect job URI to check on the status of the collect job and
@@ -1159,15 +1311,24 @@ and a body consisting of a `CollectResp`:
 
 ~~~
 struct {
+  select (query_type) { // determined by task configuration
+    time-interval: Empty;
+    fixed-size: BatchId batch_id;
+  }
   uint64 report_count;
   HpkeCiphertext encrypted_agg_shares<1..2^32-1>;
 } CollectResp;
 ~~~
 
-* `report_count` is the number of reports included in the aggregation.
-* `encrypted_agg_shares` is the vector of encrypted aggregate shares.
-They MUST appear in the same order as the aggregator endpoints list of the task
-parameters.
+This structure includes the following:
+
+* information used to bind the aggregate result to the query. For fixed-size
+  tasks, this includes the batch ID assigned to the batch by the Leader.
+
+* The number of reports included in the batch.
+
+* The vector of encrypted aggregate shares. They MUST appear in the same order
+  as the aggregator endpoints list of the task parameters.
 
 If obtaining aggregate shares fails, then the leader responds to subsequent HTTP
 GET requests to the collect job URI with an HTTP error status and a problem
@@ -1188,10 +1349,10 @@ helper drops out?]
 
 The leader obtains each helper's encrypted aggregate share in order to respond
 to the collector's collect response. To do this, the leader first computes a
-checksum over the set of output shares included in the batch window identified
-by the collect request. The checksum is computed by taking the SHA256 hash of
-each nonce from the client reports included in the aggregation, then combining
-the hash values with a bitwise-XOR operation.
+checksum over the set of output shares included in the batch. The checksum is
+computed by taking the SHA256 hash of each nonce from the client reports
+included in the aggregation, then combining the hash values with a bitwise-XOR
+operation.
 
 Then, for each aggregator endpoint `[aggregator]` in the parameters associated
 with `CollectReq.task_id` (see {{collect-flow}}) except its own, the leader sends
@@ -1199,31 +1360,49 @@ a POST request to `[aggregator]/aggregate_share` with the following message:
 
 ~~~
 struct {
+  select (query_type) { // determined by task configuration
+    time-interval: Interval batch_interval;
+    fixed-size: BatchId batch_id;
+  }
+} BatchSelector;
+
+struct {
   TaskID task_id;
-  Interval batch_interval;
+  BatchSelector batch_selector;
   opaque agg_param<0..2^16-1>;
   uint64 report_count;
   opaque checksum[32];
 } AggregateShareReq;
 ~~~
 
-* `task_id` is the task ID associated with the DAP parameters.
-* `batch_interval` is the batch interval of the request.
-* `agg_param`, an aggregation parameter for the VDAF being executed.
-  This is the same value as in `AggregateInitializeReq` (see {{leader-init}})
-  and in `CollectReq` (see {{collect-init}}).
-* `report_count` is the number of reports included in the aggregation.
-* `checksum` is the checksum computed over the set of client reports.
+The message contains the following parameters:
+
+* The task ID.
+
+* The "batch seletor", which encodes parameters used to determine the batch
+  being aggregated. The value depends on the query type for the task:
+
+    * For time-interval tasks, the request specifies the batch interval.
+
+    * For fixed-size tasks, the request specifies the batch ID.
+
+* The opaque aggregation parameter for the VDAF beging executed. This value MUST
+  match the same value in the the `AggregateInitializeReq` message sent in at
+  least one run of the aggregate sub-protocol. (See {{leader-init}}). and in
+  `CollectReq` (see {{collect-init}}).
+
+* The number number of reports included in the batch.
+
+* The batch checksum.
 
 This request MUST be authenticated as described in {{https-sender-auth}}. To
 handle the leader's request, the helper first ensures that the request meets the
 requirements for batch parameters following the procedure in
-{{batch-parameter-validation}}.
+{{batch-validation}}.
 
-Next, it computes a checksum based on its view of the output shares included in
-the batch window, and checks that the `report_count` and `checksum` included in
-the request match its computed values. If not, then it MUST abort with error
-"batchMismatch".
+Next, it computes a checksum based on the reports that satisfy the query, and
+checks that the `report_count` and `checksum` included in the request match its
+computed values. If not, then it MUST abort with error "batchMismatch".
 
 Next, it computes the aggregate share `agg_share` corresponding to the set of
 output shares, denoted `out_shares`, for the batch interval, as follows:
@@ -1235,7 +1414,7 @@ agg_share = VDAF.out_shares_to_agg_share(agg_param, out_shares)
 Note that for most VDAFs, it is possible to aggregate output shares as they
 arrive rather than wait until the batch is collected. To do so however, it is
 necessary to enforce the batch parameters as described in
-{{batch-parameter-validation}} so that the aggregator knows which aggregate
+{{batch-validation}} so that the aggregator knows which aggregate
 share to update.
 
 The helper then encrypts `agg_share` under the collector's HPKE public
@@ -1260,10 +1439,10 @@ computed above and `encrypted_aggregate_share.ciphertext` is the ciphertext
 After receiving the helper's response, the leader uses the HpkeCiphertext to
 respond to a collect request (see {{collect-flow}}).
 
-After issuing an aggregate-share request for a given batch interval, it is an
-error for the leader to issue any more aggregate or aggregate-init requests for
-additional reports in the batch interval. These reports will be rejected by helpers as
-described {{agg-init}}.
+After issuing an aggregate-share request for a given query, it is an error for
+the leader to issue any more aggregation jobs for additional reports that
+satisfy the query. These reports will be rejected by helpers as described
+{{agg-init}}.
 
 Before completing the collect request, the leader also computes its own aggregate
 share `agg_share` by aggregating all of the prepared output shares that fall within
@@ -1293,7 +1472,7 @@ as follows:
 
 ~~~
 enc, payload = SealBase(pk, "dap-01 aggregate share" || server_role || 0x00,
-  AggregateShareReq.task_id || AggregateShareReq.batch_interval, agg_share)
+  AggregateShareReq.task_id || AggregateShareReq.batch_selector, agg_share)
 ~~~
 
 where `pk` is the HPKE public key encoded by the collector's HPKE key,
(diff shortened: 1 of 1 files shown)


draft-ietf-ppm-dap.md
```diff
@@ -1305,45 +1509,70 @@ given batch interval, denoted `batch_interval`, decryption works as follows:
...
-  alert the peer with "batchLifetimeExceeded".
-
-Finally, the aggregator checks that the batch interval defined by the collect
-request satisfies one of the conditions:
-
-1. The batch interval does not overlap with the batch interval of any prior
-   completed collect requests.
-1. The batch interval, including its start and duration values, match a prior
-   completed collect request.
+Frist the aggregator checks that the request respects any batch "boundaries"
```
cjpatton, line 1533: Maybe "boundary" is the wrong word? For time-series tasks, these are  the batch windows. For fixed-chunk tasks, these are the chunks.

draft-ietf-ppm-dap.md
```diff
@@ -484,6 +480,133 @@ struct {
...
+
+For this query type, the Aggregators batch measuremnet into arbitrary "chunks"
+such that each chunk has roughly the same size. The configuration includes
+parameters `min_batch_size` and `max_batch_size` that determine, respectively,
+the minumum and maximum number of measurements per chunk.
+
+To get the aggregate of the next chunk, the Colector issues a "chunk query" of
+type `next-chunk`. (See ChunkQuery in {{query}}.) The response to this query
+((see {{collect-flow}})) includes a "chunk ID", which can be used to query the
+same chunk later on. This is done using the `by-chunk-id` chunk query type.
```
cjpatton, line 600: TODO: Deal with the case where a client queries an unknown chunk.

draft-ietf-ppm-dap.md
```diff
@@ -484,6 +480,133 @@ struct {
...
+
+### Fixed-size Chunks
+
+The `fixed-chunk` query type is used to support applications in which the
+Collector needs the ability to strictly control the sample size. This is
+particularly important for controlling the amount of noise added to measurements
+by Clients (or added to aggregate shares by Aggregators) in order to achieve
+differential privacy.
+
+For this query type, the Aggregators batch measuremnet into arbitrary "chunks"
```
tgeoghegan, line 592:
```suggestion
For this query type, the Aggregators batch measurements into arbitrary "chunks"
```

draft-ietf-ppm-dap.md
```diff
@@ -484,6 +480,133 @@ struct {
...
+such that each chunk has roughly the same size. The configuration includes
+parameters `min_batch_size` and `max_batch_size` that determine, respectively,
+the minumum and maximum number of measurements per chunk.
+
+To get the aggregate of the next chunk, the Colector issues a "chunk query" of
+type `next-chunk`. (See ChunkQuery in {{query}}.) The response to this query
+((see {{collect-flow}})) includes a "chunk ID", which can be used to query the
```
tgeoghegan, line 599:
```suggestion
(see {{collect-flow}}) includes a "chunk ID", which can be used to query the
```

draft-ietf-ppm-dap.md
```diff
@@ -605,46 +726,63 @@ structured as follows:
...
+
+* A nonce used by the Aggregators to ensure the report appears in at most one
+  batch. (See {{anti-replay}}.) The Client MUST generate this by generating 16
+  random bytes using a cryptographically secure random number gnerator.
+
+* A list of extensions to be included with the report. (See
+  {{upload-extensions}}.)
+
+To generate a report, the Client first shards its measurement into input shares
+as specifieed by the VDAF. It then encrypts each input share as follows:
```
tgeoghegan, line 771:
```suggestion
as specified by the VDAF. It then encrypts each input share as follows:
```

draft-ietf-ppm-dap.md
```diff
@@ -935,16 +1106,22 @@ The validation checks are as follows.
...
-   parameter. If this check fails, the input share is marked as invalid with
-   the error `report-replayed`. This is the case if the report was used in a
-   previous aggregate request and is therefore a replay. An aggregator may also
-   choose to mark an input share as invalid with the  error `report-dropped`
-   under the conditions prescribed in {{anti-replay}}.
-
+   parameter. If this check fails, the input share MUST be marked as invalid
+   with the error `report-replayed`. This is the case if the report was used in
+   a previous aggregate request and is therefore a replay. An aggregator may
+   also choose to mark an input share as invalid with the  error
```
tgeoghegan, line 1116:
```suggestion
   also choose to mark an input share as invalid with the error
```

draft-ietf-ppm-dap.md
```diff
@@ -935,16 +1106,22 @@ The validation checks are as follows.
...
+   `report-dropped` under the conditions prescribed in {{anti-replay}}.
+3. Depending on the query type for the task, additional checks may be applicable:
+    * For fixed-chunk tasks, the Aggregators need to ensure that each batch is
+      roughly the same size. If the number of reports aggregated for the current
+      chunk exceeds the maximum batch size (per the task configuration), the
+      Aggregator MAY mark the input share as invalid with the error
+      "batch-saturated". Note that this behasvior is not strictly enforced here
```
tgeoghegan, line 1123:
```suggestion
      "batch-saturated". Note that this behavior is not strictly enforced here
```

draft-ietf-ppm-dap.md
```diff
@@ -1086,12 +1263,12 @@ The helper then awaits the next message from the leader.
...
 ## Collecting Results {#collect-flow}
 
-In this phase, the collector requests aggregate shares from each aggregator
-and then locally combines them to yield a single, aggregate output. In particular,
-the collector asks the leader to collect and return the results for a given
-DAP task over a given time period. The aggregate shares are encrypted to the
-collector so that it can decrypt and combine them to yield the aggregate output.
-This entire process is composed of two interactions:
+In this phase, the Collector requests aggregate shares from each aggregator and
+then locally combines them to yield a single, aggregate result. In particular,
```
tgeoghegan, line 1267:
```suggestion
then locally combines them to yield a single aggregate result. In particular,
```

draft-ietf-ppm-dap.md
tgeoghegan, line 1533:
```suggestion
First the aggregator checks that the request respects any batch "boundaries"
```

tgeoghegan (changes requested):
First pass for spelling and a couple editorial gripes, still reviewing for content

draft-ietf-ppm-dap.md
```diff
@@ -1188,42 +1374,60 @@ helper drops out?]
...
+The message contains the following parameters:
+
+* The task ID.
+
+* The "batch seletor", which encodes parameters used to determine the batch
+  being aggregated. The value depends on the query type for the task:
+
+    * For time-series tasks, the request specifies the batch interval.
+
+    * For fixed-chunk tasks, the requiest specifies the chunk ID.
```
tgeoghegan, line 1412:
```suggestion
    * For fixed-chunk tasks, the request specifies the chunk ID.
```

tgeoghegan (changes requested):
I think this change accomplishes what it intends. However, besides specific comments left inline, I have an orthogonal, meta-thought here: can we design this so that the helper is agnostic to what kind of query the task uses?

Before this PR, we only supported batch interval queries. The leader's job was to assign reports to aggregation jobs and then communicate that mapping to helpers. The helper would then have to keep track of:

 - which reports belong to which aggregation jobs;
 - which reports belong to which batch intervals.

With the new fixed chunk queries, the helper has to keep track of:

 - which reports belong to which aggregation jobs (as before);
 - which reports belong to which chunks OR which aggregation jobs belong to which chunks.

The second mapping in both cases is new code and maybe new database tables that the helper has to deal with. Maybe we can spell this out so that for new query types, only the leader and collector need to know about them.

Ultimately, the only thing a helper is enforcing is that the number of reports included in an aggregate share fall within a range. `min_batch_size <= count < infinity` in the batch interval case, or `min_batch_size < count < max_batch_size` in the fixed chunk case.

So instead of:

```
struct {
  TaskID task_id;
  BatchSelector batch_selector;
  opaque agg_param<0..2^16-1>;
  uint64 report_count;
  opaque checksum[32];
} AggregateShareReq;
```

...what if we had:

```
struct {
  TaskID task_id;
  AggregationJobId job_ids<0..2^?-1>;
  opaque agg_param<0..2^16-1>;
  uint64 report_count;
  opaque checksum[32];
} AggregateShareReq;
```

To service that request, the helper looks up the set of reports associated with each `AggregationJobId` (recall that the helper already has to keep track of this mapping), then verifies that the total number of reports described satisfies the range requirement, and releases the aggregate share if so. There are no options based on query flavour, and the helper has a single codepath for handling aggregate share requests.

So long as new query types' results can ultimately be described by the leader in terms of a list of aggregation job IDs, then helpers should be able to support them without code changes.

The downsides or questions w.r.t. this idea are:

P1 - The current design of the batch interval query case allows aggregators to "eagerly" aggregate into `min_batch_duration`-sized buckets ahead of a collect request providing the actual `batch_interval`. If we make everything aggregation job ID based, then the helper can't eagerly aggregate across aggregation jobs. The leader could choose to construct aggregation jobs in some pathological manner that makes the helper extremely inefficient but I think that might already have been the case.

P2 - In the batch interval query case, the helper can no longer enforce `min_batch_duration`. But was that ever needed for robustness or privacy, or was that just to enable to aggregators to eagerly accumulate report shares ahead of a collect request? If an aggregate share contains enough reports to satisfy the task's minimum, does it matter if they all arrived in the same instant?

P3 - The helper can no longer verify that parameters in an `AggregateShareReq` have anything to do with an authentic collector `CollectReq`. If `AggregateShareReq` is written in terms of a list of `AggregationJobId`, then the leader gets total control over what reports are included. This was already the case in the fixed chunk case, but it does seem like a new power in the batch interval case.

That's a pitch for expressing both batch interval and fixed chunk queries to the helper in terms of a list of aggregation job IDs. But I think we could also express fixed chunk queries in terms of batch intervals. Supposing that report nonces contained time regardless of query type, then when the leader constructs an `AggregateShareReq`, it could construct a batch interval from the chunk's lowest and highest timestamps and then send that to the helper.

draft-ietf-ppm-dap.md
```diff
@@ -1115,35 +1292,35 @@ attacks.]
 ~~~
 struct {
   TaskID task_id;
-  Interval batch_interval;
+  Query query;
```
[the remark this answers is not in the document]
cjpatton: > In the fixed chunk model, however, we grant the leader total power to assign reports to chunks (modulo the max and min sizes the task defines), so we should carefully analyze what powers that gives the leader.

Nice catch. There should already be an OPEN ISSUE somewhere in this PR about this.

> For instance, suppose the leader can see the uploading client's IP and do geolocation. Maybe it constructs one chunk that only contains reports from clients in country A and another chunk that only contains reports from country B in an effort to skew either chunk's aggregation somehow?

Agreed, this is a risk worth considering. I imagine there are other things the Leader can do like this already, but you're right that the intention of including the timestamp in the report was to allow the Client to control (at least in part), which batch its report goes in.

It's worth pointing out that replay protection requirements haven't changed: Each Aggregator still needs to ensure that a single report doesn't get replayed in multiple chunks.

> My intuition on how this feature is used is that you want to enable the leader to create a chunk every `min_chunk_size` reports. So maybe we want to enable the client to verify that the chunks defined by the leader are made up of non-overlapping intervals? That would require putting the `Time` component back in a report `Nonce`, but this would use code the helper needs to have anyway to enforce that batch interval queries don't overlap.

Ohhhh that's an interesting idea! Your intuition is right I think. It's a little stricter than what this PR does now. (And thus harder to implement?). The Leader can still mount a similar "attack" by only processing reports from the Client population it's interested in. However there is a cost: It'll end up losing the data for all other Clients.

Feedback on feasibility of the implementation would be useful.

cjpatton:
Thanks @tgeoghegan for the comments! Based on your feedback and feedback in the PPM slack, I will definitely be adding the timestamp back into the report struct for fixed-chunk tasks.

draft-ietf-ppm-dap.md
```diff
@@ -484,6 +480,133 @@ struct {
...
+
+To get the aggregate of the next chunk, the Colector issues a "chunk query" of
+type `next-chunk`. (See ChunkQuery in {{query}}.) The response to this query
+((see {{collect-flow}})) includes a "chunk ID", which can be used to query the
+same chunk later on. This is done using the `by-chunk-id` chunk query type.
+
+Implementation note: The goal for the Aggregators is to aggregate precisely
+`min_batch_size` measurements per chunk. Doing so, however, is challenging for
+Leader deployments in which multiple, independent nodes running the aggregate
+sub-protocol (see {{aggregate-flow}}) need to be coordinated. The maximum batch
```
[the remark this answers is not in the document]
BranLwyd: I think creating an aggregation job and attaching it to a given chunk requires knowing [an overestimate of] the sum of the number of reports successfully aggregated + the number of "outstanding" reports in in-progress aggregation jobs. Otherwise, creating an aggregation job may cause the chunk to grow past `max_batch_size`, effectively losing the chunk.

If you can update this sum "transactionally" with creation of the aggregation job, you can always hit `min_batch_size` exactly.

If you can't, I think avoiding chunk overflow requires having enough "leeway" (in `max_batch_size - min_batch_size`) such that every worker creating a max-size job at once will not move a chunk that has not yet aggregated `min_batch_size` reports to have aggregated more than `max_batch_size` reports, since every worker might simultaneously see the same information and independently decide to schedule an aggregation job. Another way of putting it is: I think increasing the "precision" of the size of your chunks would require reducing parallelism in the number of workers simultaneously aggregating into the chunk; equivalently, increasing worker parallelism to aggregate a task more quickly will lead to decreased "precision" in the size of the chunks that are generated.

Is that sort of what you're thinking (especially in the analysis in the previous paragraph)? I'm also curious if hitting `min_batch_size` exactly provides advantages over just knowing the number of reports in the chunk -- it would be relatively easy to communicate the report count to the collector, and if doing so would let us stop worrying about `max_batch_size` that might be the easier path.

draft-ietf-ppm-dap.md
```diff
@@ -484,6 +480,133 @@ struct {
...
+  will be obtained: The larger the minimum batch size, the higher degree of
+  privacy. However, this ultimately depends on the application and the nature of
+  the measurements and aggregation function.
+
+- `min_batch_duration` - The duration of the smallest permitted batch interval.
+  Additional restrictions apply; see {{batch-validation}}.
+
+The Collector is free to choose any sqeuence batch intervals it wishes, subject
+to the restrctions prescribed in {{batch-validation}}. However, it is
+RECOMMENDED that the sequence of batch intervals requested by the Collector all
```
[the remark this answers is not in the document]
BranLwyd: One thing useful to call out (but probably not add to the spec text) is that collectors may want to send a collect request before reports for the "query" have finished arriving. For example, in a time-series task, the collector may want to send a collect request for a batch interval that has not yet expired, with the expectation that the created collect job will not be resolved until the interval has completed.

Chris, you pointed out that this is nice in the sense that it means the collector & the aggregators don't need to agree on the clock. Makes sense to me.

I think this is even more important for fixed-chunk tasks: the next chunk being "ready for collection" is based on the rate of report arrival, which the collector can't see. So I think we _need_ to support the "aggressive collector" use-case to support fixed-chunk tasks.

All that said, I'm agnostic on whether this kind of detail should make it into the spec text. :) I think mentioning that the the collector can choose whatever intervals it wishes; I'm less sure of the value of `RECOMMEND`ing a particular sequence of intervals.

draft-ietf-ppm-dap.md
```diff
@@ -605,46 +727,63 @@ structured as follows:
...
 where `pk` is the aggregator's public key; `server_role` is the Role of the
-intended recipient (`0x02` for the leader and `0x03` for the helper);
-`task_id`, `nonce`, and `extensions` are the corresponding fields of `Report`;
-and `input_share` is the aggregator's input share.
+intended recipient (`0x02` for the leader and `0x03` for the helper), `header`
+is the report header, and `input_share` is the Aggregator's input share.
+
+The order of the encrypted input shares appear MUST match the order of the
+task's `aggregator_endpoints`. That is, the first share should be the leader's,
+the second share should be for the first helper, and so on.
```
BranLwyd, line 786: I think L800 (a few lines down from here) needs to change. It currently reads:

> The leader MUST ignore any report whose nonce contains a timestamp that falls in
a batch interval for which it has received at least one collect request from the
collector.

Nonces no longer have a timestamp. Depending on the task configuration, reports might no longer have a timestamp at all.

More importantly, this would block an "aggressive collector" which wants to send collect requests before reports have finished arriving for a given batch, with the expectation that the collect job will stay live until the batch is "over". Supporting the aggressive collector is a nice-to-have for time-series tasks (since it breaks the need for a collector to have a clock that matches the aggregators', and removes a case where a collector can trigger data loss accidentally), and I think a requirement for fixed-chunk tasks (since batches are determined by the rate of report arrival, and the collector does not have any insight into this, so it can't know when to make another collect request).

draft-ietf-ppm-dap.md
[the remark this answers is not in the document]
tgeoghegan: My bias is for simplicity in the protocol text, which I think argues for requiring fixed size chunks, and my intuition is that this should be achievable. My intuition (I'm using "intuition" as a weasel word here so that I don't actually have to support my arguments) is that the coordination you describe should be achievable in a datastore with [serializability](https://en.wikipedia.org/wiki/Serializability), which I think DAP implementations require anyway in order to do things like anti-replay.

However I think I want to get more implementation experience with the fixed chunk mode before I argue this point further, so let's go ahead with what's here now.

Finally, we already have #163 to wire up the report count to the collector, orthogonally to what's going on in this PR.

draft-ietf-ppm-dap.md
```diff
@@ -484,6 +480,134 @@ struct {
...
+            uint64 max_batch_size;
+    }
+} QueryConfig;
+~~~
+
+### Time-series Telemetry
+
+The first query type, `time-series`, is designed to support applications in
+which some process is observed over a long period of time. The Collector
+specifes a "batch interval" that determines the time range for measurements
```
tgeoghegan, line 550:
```suggestion
specifies a "batch interval" that determines the time range for measurements
```

draft-ietf-ppm-dap.md
```diff
@@ -604,47 +726,60 @@ Clients upload reports by using an HTTP POST to `[leader]/upload`, where
...
+This message is called the Client's report. It consists of a "header" and the
+encrypted input share of each of the Aggregators. The header consists of:
+
+* The task ID for which the report is intended.
+
+* A timestamp representing the time at which the report was generated.
+  Specifically, the `time` field is set to the number of seconds elapsed since
+  the start of the UNIX epoch.
+
+    * For time-series tasks, the client MUST round this value to the nearest
```
tgeoghegan, line 751: This wasn't a MUST previously, and we discussed why at the time in [#281](https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/pull/281#pullrequestreview-1025110801).

draft-ietf-ppm-dap.md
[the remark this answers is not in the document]
hanschen1996: I think this concern is valid, but making sure each chunk has non-overlapping intervals can cause us to lose some late-arriving reports. For example, assuming max_batch_lifetime is 1 for simplicity, and I have a task that can quickly fulfill min chunk size in a short period of time (e.g. 1 minute). If leader has constructed a chunk that meets minimum chunk size, the min and max time of all the reports in this chunk defines the interval for this chunk. Now for the next chunk, leader is only allowed to collect reports that have timestamps later than the max timestamp of the previous chunk. If there are late arriving reports, leader is forced to drop all of them even though they are valid, so the availability of the leader service may take a hit in this case.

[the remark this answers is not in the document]
tgeoghegan: re: Brandon's points:

(1): This was already a case we had to handle in the batch interval query, and I think we would handle it the same way here: if a report arrives after the collect job for its chunk has been serviced, it gets ignored.

(2): Ooh, this is interesting. I think this can be handled, though. Let's say a collector comes along and asks for the result of aggregating chunk ID 4, and then immediately asks for chunk ID 5. The leader will generate a couple of collect job URIs and send those back to the collector -- at this point there's no aggregator commitment to which reports go in which chunk. The leader might then identify a set of candidate reports for chunk 4 and another set for chunk 5. It then carves each set of reports into some number of aggregation jobs and executes them with the helper.

Now, suppose enough reports are rejected from the candidate set assigned to chunk 4 that `min_batch_size` isn't satisfied. The leader will want to then steal some aggregation jobs from chunk 5 to get chunk 4 over the minimum.

OK, but what if the aggregation jobs for chunk 5 finished first and the aggregate result was delivered to the collector? In that case, yes, we are stuck and can never satisfy the collect request for chunk 4. But we could assert that the leader can't deliver the results for the collect job for chunk 5 until it has delivered the results for chunk 4. Note that the constraint here is not on when or in what order the aggregators can do the work of preparation. The constraint is on when the results can be delivered, because it's only at that point that the aggregators commit to assigning a set of reports to a chunk.

So I think this means that aggregators could still be running all the aggregate jobs in parallel. They just have to require that before you can deliver the results for collect job _n_, the collect jobs for the previous _n-1_ chunks must have been completed.

The problem with this is that we want to allow collectors to skip ahead in the stream of chunks. I suppose this could be accomplished by having the collector send DELETE requests to collect job URIs for those chunks it wants to skip?

draft-ietf-ppm-dap.md
```diff
@@ -814,29 +950,60 @@ structured as follows:
...
-whose index in `Report.encrypted_input_shares` is equal to the index of the aggregator
-in the task's `aggregator_endpoints` to which the AggregateInitializeReq is being sent.
-The `agg_param` field is an opaque, VDAF-specific aggregation parameter provided during a collection flow.
-The `job_id` parameter contains the leader's chosen AggregationJobID.
+* The ID of the task for which the reports will be aggregated.
 
-[[OPEN ISSUE: Check that this handling of `agg_param` is appropriate when the definition of Poplar is done.]]
+* The aggregation job ID.
+
+* The opaque, VDAF-specifict aggregation parameter provided during the
```
junyechen1996, line 980: VDAF-specifict -> VDAF-specific

draft-ietf-ppm-dap.md
```diff
@@ -907,18 +1074,19 @@ abort with error "unrecognizedMessage".
...
 #### Input Share Decryption {#input-share-decryption}
 
-Each report share has a corresponding task ID, nonce, list of extensions, and encrypted
-input share. Let `task_id`, `nonce`, `extensions`, and `encrypted_input_share` denote these
-values, respectively. Given these values, an aggregator decrypts the input
-share as follows. First, the aggregator looks up the HPKE config and corresponding
-secret key indicated by `encrypted_input_share.config_id`. If not found, then it
-marks the report share as invalid with the error `hpke-unknown-config-id`.
-Otherwise, it decrypts the payload with the following procedure:
+Each report share has a corresponding task ID, report info (nonce, batch
```
junyechen1996, line 1077: report info should be nonce, time, and extensions?

draft-ietf-ppm-dap.md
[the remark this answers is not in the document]
junyechen1996: @tgeoghegan Regarding bullet point 1, what if each report rounds down its timestamp to the nearest n-hour window (a coarse-grained timestamp to protect client privacy, let's say 4 hours for example)? Leader can create a chunk with just enough reports that have timestamp 00:00, doesn't that mean leader will be forced to drop all future reports that mark its timestamp at 00:00?

draft-ietf-ppm-dap.md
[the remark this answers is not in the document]
cjpatton: TODO: Ready for text

draft-ietf-ppm-dap.md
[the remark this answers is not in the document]
cjpatton: @BranLwyd as discussed at IETF, the main reason to target a given batch size is that in some statistical applications you need to be able to control for the sample size. This would be important, for example, when using local differential privacy.

I'd like to start resolving threads where possible. The question left open here is whether requiring `min_batch_size == max_batch_size` is feasible. @tgeoghegan are you alright with resolving this with an OPEN ISSUE in the text?

draft-ietf-ppm-dap.md
```diff
@@ -484,6 +480,134 @@ struct {
...
+        case fixed-chunk: ChunkQuery chunk_query;
+    }
+} Query;
+~~~
+
+The query is issued in-band as part of the collect sub-protocol
+({{collect-flow}}). Its content is determined by the "query type", which in turn
+is encoded by the "query configuration" configured out-of-band:
+
+~~~
```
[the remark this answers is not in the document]
cjpatton: Ack.

draft-ietf-ppm-dap.md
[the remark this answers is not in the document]
tgeoghegan: Yes, I'm comfortable revisiting this after we get some experience implementing the fixed chunk mode. However I would urge you to highlight this particular question in the thread you started on the ppm-ietf list just in case someone there has an opinion.

draft-ietf-ppm-dap.md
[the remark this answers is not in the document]
cjpatton: Dropped the RECOMMENDATION. I'm not sure how to spell the Aggregator behavior, so I left as an open issue. WE can deal with this later.

draft-ietf-ppm-dap.md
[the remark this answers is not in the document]
cjpatton: Done.

draft-ietf-ppm-dap.md
cjpatton: Haha yup good catch!

draft-ietf-ppm-dap.md
cjpatton: Good catch. I think the following solves both issues. This also matches the Helper's behavior in {{input-share-batch-validation}}.

> The Leader MUST ignore any report pertaining to a batch that has already been collected.

For time-interval tasks the intended interpretation is "the Leader rejects the report if the timestamp falls in a batch interval that has been collected". For fixed-size tasks the intended interpretation is "The Leader MUST NOT include the report in a previously collected batch."

draft-ietf-ppm-dap.md
cjpatton: Oh good, I thought the intention was MUST. How about SHOULD?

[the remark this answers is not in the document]
cjpatton: Right, I don't know what to do about this awkwardness. One way to resolve this is to find a reason not to include the timestamp for fixed-size tasks. (The argument would be that preventing overlapping batch intervals for this task type has marginal benefit or is not feasible.) There is an OPEN ISSUE in the text in case we want to punt.

draft-ietf-ppm-dap.md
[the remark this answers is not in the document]
cjpatton: Thanks, everyone, for taking time to help with this. I think we're now at a point where we need to take a step back and think through requirements.

First, in response to feedback from @tgeoghegan and @BranLwyd, not that we have renamed "time-series" to "time-interval" and "fixed-chunk" to "fized-size". Instead of "chunk" we'll just say "batch".

A quick summary to see if we're on the same page. @tgeoghegan points out a desirable property of time-interval tasks: Given the sequence of queries, a Clients know exactly *which* batch their report would be included in, since we have the following properties:
- (1.) a report can belong to exactly one batch
- (2.) batches are non-overlapping
- (3.) a query determines a unique batch 

For "fixed-size" tasks, we currently guarantee something strictly weaker: a Client is know their report is included in at most one batch, but it does not know which one. This is because we have (1.) and (2.) but not (3.).

We don't seem to have a workable solution to this yet:
- (a.) We could have the Client include the batch ID in its report. This is the direct analogue of including the timestamp in the report. However this is a non-starter because it shifts coordination of the batch selection to Clients.
- (b.) We could require that batches don't overlap in time, just as we do for time-interval tasks. @BranLwyd and @junyechen1996 have pointed out various reasons why this is tricky.

These difficulties beg the following essential question: **How important is property (3.)?** Without it, the Leader is capable of sorting reports into batches as it pleases. For example:
* The Leader can batch reports by User-Agent, location, and so on. This is is still possible with (3.), but this would at least force the Leader to "throw away" reports from clients that are not in the target group, but pertain to the query.
* The Leader can do a kind of weak stuffing attack. Stuffing attacks ares still possible with (3.), but again, it would at least force the Leader to throw away any reports that pertain to the query.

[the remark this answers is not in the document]
cjpatton: My take: This threat falls into a big bucket of Sybil-like attacks that I would like to see a more systematic solution for. (Yes Sybil attacks are hard, but they're at least tractable.) Thus I don't think it's worth expending too much effort trying to get (3.). My vote would be to not require it for every query type (and fixed-size queries in particular), but make sure whatever we spell doesn't prevent it.

draft-ietf-ppm-dap.md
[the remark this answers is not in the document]
BranLwyd: Thinking about this a little more, I suppose the task doesn't need a new explicit parameter -- clients can freely round their submitted timestamps to any desired resolution already.

I suppose this will be left as an OPEN ISSUE right now, but one way to resolve would be to say that clients can round their timestamps, and additionally mandate that clients of a time-series task must round to at most the `min_batch_duration` to keep their reports in the correct buckets.

draft-ietf-ppm-dap.md
```diff
@@ -484,6 +480,130 @@ struct {
...
+includes parameters `min_batch_size` and `max_batch_size` that determine,
+respectively, the minimum and maximum number of measurements per batch.
+
+To get the aggregate of the next batch, the Collector issues a query of type
+`next-batch`. (See FixedSizeQuery in {{query}}.) The response to this query (see
+{{collect-flow}}) includes a "batch ID", which can be used to query the same
+batch later on. This is done using the `by-batch-id` batch-query type.
+
+Implementation note: The goal for the Aggregators is to aggregate precisely
+`min_batch_size` measurements per batch. Doing so, however, umay be challenging
```
BranLwyd, line 596:
```suggestion
`min_batch_size` measurements per batch. Doing so, however, may be challenging
```

(typo)

BranLwyd (approved):
Thanks for your patience. :) This LGTM.

I would definitely like if we could find a way to somehow unify the different query types (lots of places in the spec now basically bifurcate struct content/validation logic/etc), but IIUC this is a somewhat exploratory change to discover the different query types that need to be supported, so merging this LGTM.

cjpatton:
> I would definitely like if we could find a way to somehow unify the different query types (lots of places in the spec now basically bifurcate struct content/validation logic/etc), but IIUC this is a somewhat exploratory change to discover the different query types that need to be supported, so merging this LGTM.

I think we should aim to merge what we intend to implement. I agree that that it would be great to make the aggregation flow as agnostic as possible to to the query type. I think we've done about as good a job as we can here, but I'm open to alternative ideas. Let's keep discussing!

BranLwyd:
My current thinking (pretty high-level, may be sketchy, review appreciated):

* I think something like property (3) is very desirable to have. We've been discussing achieving this property by requiring that distinct batches be non-overlapping in time. I think one way to make it feasible to implement that idea is to change how client reports, aggregation jobs, and batches are bound together:
  * Currently, we bind a client report to a batch when the aggregation job is created via an `AggregateInitializeReq`. But there's no reason this conceptually needs to be true: we could instead choose to bind client reports to batches at or after the end of aggregation. We also require that all reports in the same aggregation job be bound to the same batch, but again, there is no reason this conceptually needs to be true: we could instead allow different client reports in the same aggregation job to be bound to different batches. We could implement these conceptual changes by adding an additional step to the aggregate flow, or perhaps even better by having the `AggregateShareReq` include enough information to identify the aggregated reports that are bound to the batch. I think the latter option would be better since, if we commit to batches having non-overlapping time intervals in the `fixed-size` task type, the appropriate set of reports could be identified by a batch interval in both the `time-interval` and `fixed-size` task types.
  * These changes would make it much easier to group reports into batches while still meeting the `{min,max}_batch_size` & non-overlapping-in-time batch constraints for a few reasons. First, since the binding happens after aggregation concludes, reports that fail the aggregation flow can no longer unexpectedly "remove" reports from a batch -- we only add successfully-aggregated reports to batches. Even more importantly, by allowing different reports in the same aggregation job to be bound to different batches, we can effectively split (successfully aggregated) reports in an aggregation job to more than one batch -- so an aggregation job that is "too big" to fit in the currently-open batch can still provide enough reports to finish the currently-open batch, with the remaining reports going into a following batch. (In fact, I think with these changes we could probably always hit `min_batch_size` exactly -- the leader would just select an interval end that selects the correct number of reports. We probably want some way to break timestamp ties -- details likely a bit messy.)
  * This doesn't necessarily _uniquely_ determine batches if we allow leeway in the batch size. But it's close, and like I said, I think we actually don't need the leeway anymore in which case the batches are uniquely determined.
  * `BatchId` no longer needs to exist as a concept, since batches are now always identified by an interval.

* Once we have (3) for the `fixed-size` case, I think we can unify both the `fixed-size` and the `time-interval` cases to nearly the same logic -- they are both effectively aggregating reports over intervals of time, the largest difference is how the intervals of time are decided.
  * Specifically, the `time-interval` case would determine the batches as `[0, T)`, `[T, 2T)`, ... just as today (where `T` is the `min_batch_duration`). The `fixed-size` case would generate intervals based on the rate of client reports arriving -- the interval boundaries would be selected to create batches of the appropriate size.
  * We still need the ability added in this PR for the collector to request the "next" batch ID. It's a nice-to-have for the `time-interval` case, but a requirement for the `fixed-size` case since in this case the collector has no insight into what the batch intervals might be.
  * The ability for the collector to collect multiple contiguous batches is still required for the `time-interval` case, since we might need to do so to hit `min_batch_size`. We could implement it for `fixed-size` if we wished, by mandating that the generated batches be contiguous (i.e. no gaps between batches).
  * The expected collector behavior for `time-interval` wouldn't change: occasionally use the collect flow to collect the next interval of interest. (With these changes, the collector could safely request an interval that is not yet complete; or, if it wanted to, ask to be *told* what the "next" batch interval is, though I don't think this would be required.)
  * The expected collector behavior for `fixed-size` would be very similar to what is suggested in this PR: occasionally use the collect flow to collect the "next" batch interval, using the "tell-me-the-next-batch" functionality.
  * I think we'd want to keep the change in this PR to allow reports to arrive after a collect request arrives (but not after the collection completes), since otherwise the "tell-me-the-next-batch" functionality could not be used without potentially cutting a batch short.

Like I said, this is pretty sketchy. I'm probably missing some things, review would be appreciated.

simon-friedberger:
>       * Currently, we bind a client report to a batch when the aggregation job is created via an `AggregateInitializeReq`. But there's no reason this conceptually needs to be true: we could instead choose to bind client reports to batches at or after the end of aggregation. We also require that all reports in the same aggregation job be bound to the same batch, but again, there is no reason this conceptually needs to be true: we could instead allow different client reports in the same aggregation job to be bound to different batches. We could implement these conceptual changes by adding an additional step to the aggregate flow, or perhaps even better by having the `AggregateShareReq` include enough information to identify the aggregated reports that are bound to the batch. I think the latter option would be better since, if we commit to batches having non-overlapping time intervals in the `fixed-size` task type, a batch would be identified by a batch interval in both the `time-interval` and `fixed-size` task types.

IIUC you're saying we can do the report validation first and only assign reports to batches for the actual aggregation. I completely agree with that. If that's not what you're saying, could you please elaborate?

>       we can effectively split (successfully aggregated) reports in an aggregation job to more than one batch 

I think your wording here is a bit confusing. Once the jobs are aggregated they are already in a batch because that's what we're aggregating over. I think I know what you mean but I'm not sure.

simon-friedberger:
Meta: Can we move this discussion to an issue? The order gets confusing and parts are lost in PRs.

branlwyd:
> > ```
> >   * Currently, we bind a client report to a batch when the aggregation job is created via an `AggregateInitializeReq`. But there's no reason this conceptually needs to be true: we could instead choose to bind client reports to batches at or after the end of aggregation. We also require that all reports in the same aggregation job be bound to the same batch, but again, there is no reason this conceptually needs to be true: we could instead allow different client reports in the same aggregation job to be bound to different batches. We could implement these conceptual changes by adding an additional step to the aggregate flow, or perhaps even better by having the `AggregateShareReq` include enough information to identify the aggregated reports that are bound to the batch. I think the latter option would be better since, if we commit to batches having non-overlapping time intervals in the `fixed-size` task type, a batch would be identified by a batch interval in both the `time-interval` and `fixed-size` task types.
> > ```
> 
> IIUC you're saying we can do the report validation first and only assign reports to batches for the actual aggregation. I completely agree with that. If that's not what you're saying, could you please elaborate?

Yes, that's correct. With how the PR is currently worded, in the `fixed-size` case reports are assigned to batches when aggregation begins; my suggestion is to change that to assign reports to batches implicitly based on their timestamp.

> > ```
> >   we can effectively split (successfully aggregated) reports in an aggregation job to more than one batch 
> > ```
> 
> I think your wording here is a bit confusing. Once the jobs are aggregated they are already in a batch because that's what we're aggregating over. I think I know what you mean but I'm not sure.

The aggregation flow is done via a grouping of reports called an "aggregation job". Collection is done via a (different!) grouping of reports called a "batch", I suppose -- this "batch" concept is explicit for the `fixed-size` case which introduces an explicit batch ID, and was I believe implicit for the existing `time-interval` case (which identified batches by "batch interval").

With how this PR is currently worded, all of the reports in a given aggregation job must fall into the same batch in the `fixed-size` case, since reports are bound to a batch during `AggregateInitializeReq`. My suggestion is to change this to bind reports to batches implicitly based on their timestamp, and identify batches by a batch interval in all cases, rather than using a "batch ID" in the `fixed-size` case.

> Meta: Can we move this discussion to an issue? The order gets confusing and parts are lost in PRs.

I'd agree with this, is #273 a good place for further discussion?

junyechen1996:
I think unifying the types in `AggregateShareReq` regardless of task query type is great, but whether we use interval or list of aggregation job IDs in `AggregateShareReq` is inconvenient for one or other task query type: I 
- If we use interval in `AggregateShareReq`: This can already support interval-based query type today, but this is hard to support fixed-size batch query, especially with coarse-grained timestamps. Regarding @branlwyd's suggestions to "assign reports to batches implicitly based on their timestamp", if the clients choose to round down their timestamps to a `n`-hour window (i.e `min_batch_duration = n hours`), we could possibly have multiple batches with minimum batch size in each n-hour interval. Then we cannot guarantee that the interval in `AggregateShareReq` can uniquely identify a batch. We either need to distinguish batches within an interval (but DAP prevents overlapping intervals in many places), or just throw away reports that belong to an interval after the minimum batch size has been met in that interval, which is a waste of client measurements IMO.
- If we use list of aggregation job IDs in `AggregateShareReq`: this is relatively easy for the fixed-size batch query, because leader just needs to create aggregation jobs with the appropriate size based on the number of reports in the open batch/chunk. However, this is hard for the interval-based query, for leader particularly. Leader needs to assign reports, with the same unit batch interval start time, to the same aggregation job. However, in today's DAP, leader and helper pre-aggregate output shares implicitly based on their timestamps, so there is no restriction on how leader assigns reports to aggregation jobs.
- A separate note is that we can use the batch selector (interval or batch ID) to uniquely identify a batch, so that when leader wants to retry an `AggregateShareReq` in case of any temporary network error, the batch selector serves as the key to retrieve the aggregate share from helper side. Whereas using something like list of aggregation job IDs can be ambiguous, whether leader wants to retry a previous request, or wants to collect the same list of aggregation job IDs again.

My take is we can see if future task query types need to be supported, and evaluate again whether unifying the types in `AggregateShareReq` is worth doing? Other than that, I think this PR looks good to me.

junyechen1996 (approved):

draft-ietf-ppm-dap.md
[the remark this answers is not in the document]
wangshan: I don't think property (3) is critical for this feature, since leader can do similar things no matter what collection type is used, with pretty trivial effort. If the leader/collector also authors clients, then it can design a task that apply a malicious filter from clients, I don't think this is the type of threat DAP can address on its own.

draft-ietf-ppm-dap.md
```diff
@@ -814,29 +945,56 @@ structured as follows:
...
+
+* The opaque, VDAF-specific aggregation parameter provided during the
+  collection flow (`agg_param`),
+
+  [[OPEN ISSUE: Check that this handling of `agg_param` is appropriate when the
+  definition of Poplar is done.]]
+
+* Information used by the Aggregators to determine how to aggregate each report:
+
+    * For fixed-size tasks, the Leader specifies a "batch ID" that determines
```
wangshan, line 983: Is it worth mentioning that the binding of batch_id and agg_job_ids is permanent? so collect query can be idempotent: query of the same batch_id should return the same result, if max_batch_lifetime > 0.

draft-ietf-ppm-dap.md
```diff
@@ -604,47 +722,60 @@ Clients upload reports by using an HTTP POST to `[leader]/upload`, where
...
+This message is called the Client's report. It consists of a "header" and the
+encrypted input share of each of the Aggregators. The header consists of:
+
+* The task ID for which the report is intended.
+
+* A timestamp representing the time at which the report was generated.
+  Specifically, the `time` field is set to the number of seconds elapsed since
+  the start of the UNIX epoch.
+
+    * For time-interval tasks, the client SHOULD round this value to the nearest
```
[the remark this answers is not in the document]

draft-ietf-ppm-dap.md
```diff
@@ -604,47 +722,60 @@ Clients upload reports by using an HTTP POST to `[leader]/upload`, where
...
+    Time time;
+    Nonce nonce;
+    Extension extensions<0..2^16-1>;
+} ReportInfo;
+
 struct {
   TaskID task_id;
-  Nonce nonce;
-  Extension extensions<0..2^16-1>;
+  ReportInfo info;
```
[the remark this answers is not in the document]

draft-ietf-ppm-dap.md
cjpatton: Maybe, but not here. This would go in the collect sub-protocol I'd think.

draft-ietf-ppm-dap.md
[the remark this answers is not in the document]
cjpatton: @wangshan made the same suggestion. What I decided to do was replace `min_batch_duration` with `time_precision`, which is used by the Clients to truncate their report and, in the case time-series tasks, to define the min batch duration.

cjpatton:
FYI, I have squashed this branch into a single commit.

draft-ietf-ppm-dap.md
```diff
@@ -484,6 +480,132 @@ struct {
...
+particularly important for controlling the amount of noise added to measurements
+by Clients (or added to aggregate shares by Aggregators) in order to achieve
+differential privacy.
+
+For this query type, the Aggregators batch measurements into arbitrary batches
+such that each batch has roughly the same number of reports. The configuration
+includes parameters `min_batch_size` and `max_batch_size` that determine,
+respectively, the minimum and maximum number of measurements per batch.
+
+To get the aggregate of the next batch, the Collector issues a query of type
```
wangshan, line 592: Do we want to describe the behaviour when collector calls with `next-batch` before the last `batch-id` has been returned? I'm not sure there is a use case for collecting fixed sized batches in parallel. One of the reasons to have fixed size collection is to close a batch asap so any iterative tasks can move on to the next iteration, if there are two opening `batch-id`s for one task, then I imagine the aggregators would distribute reports to these two batches and make the collection slower for one batch? @cjpatton

draft-ietf-ppm-dap.md
```diff
@@ -1305,50 +1493,89 @@ given batch interval, denoted `batch_interval`, decryption works as follows:
...
+
+Next, the Aggregator checks that batch contains a valid number of reports. Let
+`X` denote the set of reports in the batch:
+
+* For time-interval tasks, the query configuration specifies the minimum batch
+  size, `min_batch_size`. The Aggregator checks that `len(X) >= min_batch_size`.
+
+* For fixed-size tasks, the query configuration specifies the minimum batch
+  size, `min_batch_size`, and maximum batch size. The Aggregator checks that
+  `len(X) >= min_batch_size` and `len(X) < max_batch_size`.
```
wangshan, line 1549: nit: `max_batch_size` is the only upper bound that's exclusive, should we make it `<=` ? @cjpatton

draft-ietf-ppm-dap.md
cjpatton: > One of the reasons to have fixed size collection is to close a batch asap so any iterative tasks can move on to the next iteration, ...

Is there anything in the spec that precludes this? 

> if there are two opening `batch-id`s for one task, then I imagine the aggregators would distribute reports to these two batches and make the collection slower for one batch?

It seems like this is a matter for the implementation. I agree this behavior is permitted, but the Leader could also fill the first batch first, then the second.

draft-ietf-ppm-dap.md

wangshan (approved):
Looks good to me. I have some opinions about the proposal by @branlwyd and @tgeoghegan , I'll continue in https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/issues/273. This PR looks ready to be merged for me.

draft-ietf-ppm-dap.md
```diff
@@ -484,6 +480,132 @@ struct {
...
+
+Implementation note: The goal for the Aggregators is to aggregate precisely
+`min_batch_size` measurements per batch. Doing so, however, may be challenging
+for Leader deployments in which multiple, independent nodes running the
+aggregate sub-protocol (see {{aggregate-flow}}) need to be coordinated. The
+maximum batch size is intended to allow room for error. Typically the difference
+between the minimum and maximum batch size will be a small fraction of the
+target batch size for each batch.
+
+[OPEN ISSUE: It may be feasilble to require a fixed batch size, i.e.,
```
tgeoghegan, line 605:
```suggestion
[OPEN ISSUE: It may be feasible to require a fixed batch size, i.e.,
```

draft-ietf-ppm-dap.md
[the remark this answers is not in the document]
tgeoghegan: +1 to not defining this as a struct, for the reasons Bran lays out but also to be consistent with how we discuss task parameters.

draft-ietf-ppm-dap.md
```diff
@@ -604,47 +724,54 @@ Clients upload reports by using an HTTP POST to `[leader]/upload`, where
...
-a sequence of input shares as specified by the VDAF in use. To encrypt an input
-share, the client generates an {{HPKE}} ciphertext for the aggregator by running
+This message is called the Client's report. It consists of a "header" and the
+encrypted input share of each of the Aggregators. The header consists of:
+
+* The task ID for which the report is intended.
+
+* A timestamp representing the time at which the report was generated.
+  Specifically, the `time` field is set to the number of seconds elapsed since
+  the start of the UNIX epoch. The client SHOULD round this value to the nearest
```
tgeoghegan, line 747:
```suggestion
  the start of the UNIX epoch. The client SHOULD round this value down to the nearest
```
The existing text had "rounded down", so let's be consistent with that.

tgeoghegan (approved):
A couple more minor changes, but given the open questions we have agreed on, I think this is OK to merge.

draft-ietf-ppm-dap.md
```diff
@@ -484,6 +480,132 @@ struct {
 } HpkeCiphertext;
 ~~~
 
+## Queries {#query}
+
+DAP is designed to give the Collector control over which measurements are
+included in a batch for aggregation. However, since reports are uploaded to the
+Leader, the Collector cannot choose the batch directly. Instead, it issues a "query" used
+by the Aggregators to guide batch selection.
```
chris-wood, line 488:
```suggestion
Aggregated results are computed based on sets of report, called batches. The Collector
influences which reports are used in a batch via a "query." The Aggregators use this query
to carry out the aggregation flow and produce aggregate shares encrypted to the Collector.
```

draft-ietf-ppm-dap.md
```diff
@@ -484,6 +480,132 @@ struct {
...
+~~~
+enum {
+   reserved(0), // Reserved for testing purposes
+   time-interval(1),
+   fixed-size(2),
+   (65535)
+} QueryType;
+~~~
+
+Additional query types can be defined as needed. A query includes parameters
```
chris-wood, line 501:
```suggestion
Future specifications can introduce new query types as needed. A query includes parameters
```

draft-ietf-ppm-dap.md
```diff
@@ -484,6 +480,132 @@ struct {
...
+used by the Aggregators to select a given batch of measurements specific to the
```
chris-wood, line 502:
```suggestion
used by the Aggregators to select a batch of reports specific to the
```

draft-ietf-ppm-dap.md
```diff
@@ -484,6 +480,132 @@ struct {
...
+given query type:
+
+~~~
+enum {
+    next-batch(0),
+    by-batch-id(1),
+} FixedSizeQueryType;
```
chris-wood, line 509: Why not split these out into two separate query types, rather than a single query type with different parameters?

draft-ietf-ppm-dap.md
```diff
@@ -484,6 +480,132 @@ struct {
...
+    QueryType query_type;
+    Duration time_precision;
+    uint64 min_batch_size;
+    select (query_type) {
+        case time-interval:
+            Empty;
+        case fixed-size:
+            uint64 max_batch_size;
+    }
+} QueryConfig;
```
chris-wood, line 544: We need to define what all these fields mean. For example, what is the `time_precision` field?

draft-ietf-ppm-dap.md
```diff
@@ -484,6 +480,132 @@ struct {
...
+~~~
+
+### Time-interval Queries
+
+The first query type, `time-interval`, is designed to support applications in
+which some process is observed over a long period of time. The Collector
```
chris-wood, line 550:
```suggestion
which measurements are collected over a long period of time. The Collector
```

draft-ietf-ppm-dap.md
```diff
@@ -484,6 +480,132 @@ struct {
...
+specifies a "batch interval" that determines the time range for measurements
```
chris-wood, line 551:
```suggestion
specifies a "batch interval" that determines the time range for reports
```

chris-wood: s/measurement/report generally everywhere we are actually operating on reports.

draft-ietf-ppm-dap.md
```diff
@@ -484,6 +480,132 @@ struct {
...
+
+The configuration of this query type specifies two paramters:
+
+- `min_batch_size` - Is the smallest number of measurements the batch is allowed
+  to include. In a sense, this parameter controls the degree of privacy that
+  will be obtained: The larger the minimum batch size, the higher degree of
+  privacy. However, this ultimately depends on the application and the nature of
+  the measurements and aggregation function.
+
+- `time_precision` - The duration of the smallest permitted batch interval.
```
chris-wood, line 564:
```suggestion
- `min_batch_interval` - The duration of the smallest permitted batch interval.
```

chris-wood: Since `time_precision` is confusing to me.

draft-ietf-ppm-dap.md
```diff
@@ -484,6 +480,132 @@ struct {
...
+
+### Fixed-size Queries
+
+The `fixed-size` query type is used to support applications in which the
+Collector needs the ability to strictly control the sample size. This is
+particularly important for controlling the amount of noise added to measurements
+by Clients (or added to aggregate shares by Aggregators) in order to achieve
+differential privacy.
+
+For this query type, the Aggregators batch measurements into arbitrary batches
```
chris-wood, line 587:
```suggestion
For this query type, the Aggregators group reports into arbitrary batches
```

draft-ietf-ppm-dap.md
```diff
@@ -484,6 +480,132 @@ struct {
...
+such that each batch has roughly the same number of reports. The configuration
+includes parameters `min_batch_size` and `max_batch_size` that determine,
+respectively, the minimum and maximum number of measurements per batch.
+
+To get the aggregate of the next batch, the Collector issues a query of type
+`next-batch`. (See FixedSizeQuery in {{query}}.) The response to this query (see
+{{collect-flow}}) includes a "batch ID", which can be used to query the same
+batch later on. This is done using the `by-batch-id` batch-query type.
```
chris-wood, line 595: What's the use case for querying by-batch-id again in the future? Poplar?

draft-ietf-ppm-dap.md
```diff
@@ -484,6 +480,132 @@ struct {
...
+The first query type, `time-interval`, is designed to support applications in
+which some process is observed over a long period of time. The Collector
+specifies a "batch interval" that determines the time range for measurements
+included in the batch: For each measurement in the batch, the time at which that
+measurement was generated (see {{upload-flow}}) must fall within the batch
+interval specified by the Collector.
+
+The configuration of this query type specifies two paramters:
+
+- `min_batch_size` - Is the smallest number of measurements the batch is allowed
```
chris-wood, line 558: Why is this a query-specific parameter and not something that applies to _all_ types of queries?

draft-ietf-ppm-dap.md
```diff
@@ -604,47 +724,54 @@ Clients upload reports by using an HTTP POST to `[leader]/upload`, where
...
-Specifically, the client first sets `Nonce.rand` to 16 random bytes output from
-a cryptographically secure random number generator. It then sets `Nonce.time`
-to the number of seconds elapsed since the start of the UNIX epoch, rounded down to the nearest multiple of `min_batch_duration`.
-This truncation is done to ensure that the report's timestamp cannot be used to
-link the report back to the originating client.
-
-The client then finishes the report generation by sharding its measurement into
-a sequence of input shares as specified by the VDAF in use. To encrypt an input
-share, the client generates an {{HPKE}} ciphertext for the aggregator by running
+This message is called the Client's report. It consists of a "header" and the
```
chris-wood, line 740: If we're going to refer to the header, why not actually introduce a header struct?

```
struct {
    TaskID task_id;
    Time time;
    Nonce nonce;
    Extension extensions<0..2^16-1>;
} ReportHeader;
```

chris-wood: Also, this change seems unrelated to the query type change. Is it possible to apply it in a separate PR?

draft-ietf-ppm-dap.md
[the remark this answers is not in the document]
chris-wood: Let's use the open issue to track resolution to this. I'm not convinced that we want to abandon (3) just yet.

draft-ietf-ppm-dap.md
```diff
@@ -1305,50 +1493,89 @@ given batch interval, denoted `batch_interval`, decryption works as follows:
...
+First the aggregator checks that the request respects any batch "boundaries"
+determined by the query configuration:
+
+* For time-interval tasks, the query configuration specifies the "minimum batch
+  duration", which is determined by `time_precision`. For the `batch_interval`
+  included with the query, the Aggregator checks that both
+  `batch_interval.start` and `batch_interval.duration` are divisible by
+  `time_precision` and that `batch_interval.duration >= time_precision`.
+
+* For fixed-size tasks, the batch is assocaited with a "batch ID" selected by
```
chris-wood, line 1526:
```suggestion
* For fixed-size tasks, the batch is associated with a "batch ID" selected by
```

draft-ietf-ppm-dap.md
```diff
@@ -1305,50 +1493,89 @@ given batch interval, denoted `batch_interval`, decryption works as follows:
...
+  the Leader. Thus the Aggregator needs to check that the query is associated
+  with a known batch ID:
+
+    * If the query contained in the CollectReq has type `by-batch-id`, the
+      Leader checks that the provided batch ID corresponds to a batch ID it
+      returned in a previous CollectResp for the task.
```
chris-wood, line 1532: Why would the Collector not be able to opportunistically query by batch ID? What if the IDs were monotonically increasing and the Collector just wanted to iterate over the batches?

draft-ietf-ppm-dap.md
```diff
@@ -1305,50 +1493,89 @@ given batch interval, denoted `batch_interval`, decryption works as follows:
...
+its query:
+
+* For time-interval tasks, the batch selector is the batch interval specified in
+  the query.
 
-### Validating Batch Parameters {#batch-parameter-validation}
+* For fixed-size tasks, the batch selector is the batch ID assigned sent in the
+  response.
+
+### Batch Validation {#batch-validation}
```
chris-wood, line 1511: Can we refactor this such that there are subsections for each type of query? I know that there's some shared logic, but the mixed time/fixed-size validation steps is a bit hard to parse, especially if I'm only implementing one of them.

draft-ietf-ppm-dap.md
```diff
@@ -1305,50 +1493,89 @@ given batch interval, denoted `batch_interval`, decryption works as follows:
...
+
+* For time-interval tasks, the Aggregator checks that the batch interval does not
+  overlap with the batch interval of any previous query. If this batch interval
+  check fails, then the Aggregator MAY abort with error "batchOverlap".
 
 [[OPEN ISSUE: #195 tracks how we might relax this constraint to allow for more
 collect query flexibility. As of now, this is quite rigid and doesn't give the
 collector much room for mistakes.]]
 
+[OPEN ISSUE: We might need to permit the Aggregator to drop a request as needed, if
```
chris-wood, line 1575: Definitely. Why is this even an open issue?

chris-wood (changes requested):
This is good. I have some requests for editorial changes, as well as some protocol questions, but it's pretty close to being ready.

draft-ietf-ppm-dap.md
cjpatton: I've added a line to punt to the relevant subsection. Will this suffice?
> The parameters pertaining to each query configuration are described in the relevant the subsection below.

draft-ietf-ppm-dap.md
cjpatton: Yup!

draft-ietf-ppm-dap.md
cjpatton: We kicked this around a bit. It seems like this could get tricky. For example, there would need to be a mechanism to ensure that that queries are issued in order: Imagine the collector picks batch id = 10,000, then batch id = 1.

On balance, explicit batch IDs picked by the leader seemed simpler.

draft-ietf-ppm-dap.md
cjpatton: Good idea, I will work on this now.

draft-ietf-ppm-dap.md
cjpatton: It does, I just didn't spell it that way. I'll take another crack.

draft-ietf-ppm-dap.md
cjpatton: Yup, this appears to be redundant: see https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/issues/180. Thanks @tgeoghegan!

draft-ietf-ppm-dap.md
```diff
@@ -484,6 +480,138 @@ struct {
...
+    next-batch(0),
+    by-batch-id(1),
+} FixedSizeQueryType;
+
+opaque BatchId[32];
+
+struct {
+    FixedSizeQueryType fixed_size_query_type;
+    select (fixed_size_query_type) {
+        next-batch:  Empty;
```
branlwyd, line 518: What is the idempotency story for `next-batch` requests? I think achieving idempotency requires the interaction to look something like this:

1. Collector issues `next-batch` collect request to Leader, receives collect job URI. Collector notes that the collect job URI needs to be collected to its own durable storage.
2. Collector polls collect job URI until the Leader gives a completed collect response, records the aggregation value & that the collect job is complete to its own durable storage.

If that's accurate, I think that we should spell out that `next-batch` returns the same collect job until the Collector actually collects the job--I think this isn't spell out currently, and if we don't have those semantics, repeated `next-batch` requests might cause batches to be dropped if recording the collect job URI to the Collector's durable storage fails.

Also, if it's acceptable to allow batch IDs to be enumerable (e.g. a counter, as suggested by others), I think a simpler idempotent design would be to have the Collector track which batch ID is next itself, and change `next-batch` to something like `max-batch` to allow Collectors to "start up" without knowing what the next batch is. This approach would simplify the Collector's state updates.

draft-ietf-ppm-dap.md
cjpatton: Done, please have a look.

draft-ietf-ppm-dap.md
cjpatton: I've lifted the parameters that apply to all query types to supersection.

draft-ietf-ppm-dap.md
cjpatton: > What is the idempotency story for `next-batch` requests? I think achieving idempotency requires the interaction to look something like this:
> 
>     1. Collector issues `next-batch` collect request to Leader, receives collect job URI. Collector notes that the collect job URI needs to be collected to its own durable storage.
> 
>     2. Collector polls collect job URI until the Leader gives a completed collect response, records the aggregation value & that the collect job is complete to its own durable storage.

This is accurate.

> If that's accurate, I think that we should spell out that `next-batch` returns the same collect job until the Collector actually collects the job--I think this isn't spelled out currently, and if we don't have those semantics, repeated `next-batch` requests might cause batches to be dropped if recording the collect job URI to the Collector's durable storage fails.

Each `next-batch` query would correspond to a different batch, and thus a different collect job. If the Collector queues up a bunch of collect jobs, then each job would still correspond to a different batch, right?


> Also, if it's acceptable to allow batch IDs to be enumerable (e.g. a counter, as suggested by others), I think a simpler idempotent design would be to have the Collector track which batch ID is next itself, and change `next-batch` to something like `max-batch` to allow Collectors to "start up" without knowing what the next batch is. This approach would simplify the Collector's state updates.

draft-ietf-ppm-dap.md
chris-wood: I agree that, in the current formulation, the subtypes make sense. However, I don't really see the need for the by-batch-id use case. Why would a collector want to repeatedly query the same batch? Poplar?

draft-ietf-ppm-dap.md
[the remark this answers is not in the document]
chris-wood: I agree with Brandon here. Can we describe a configuration the same way we do a task configuration?

draft-ietf-ppm-dap.md
chris-wood: That's fine, yeah.

draft-ietf-ppm-dap.md
chris-wood: OK, well, if we didn't have Poplar, would we still need this? I'm growing less and less convinced that Poplar will be actually usable, so I'm not sure we want to add complexity for the sake of something that will not get used.

draft-ietf-ppm-dap.md
chris-wood: (Saw the response below -- we can resolve this)

draft-ietf-ppm-dap.md
chris-wood: That's a fine way to spell the header, but isn't the task ID also included in each report? In any case, we refer to this thing called a "header" but then don't actually define what it is. Let's either drop the word "header", or clarify that header == task ID + metadata. 

(And in the future, we need to stop landing unrelated changes in PRs like this!)

draft-ietf-ppm-dap.md
```diff
@@ -604,47 +731,54 @@ Clients upload reports by using an HTTP POST to `[leader]/upload`, where
...
-a cryptographically secure random number generator. It then sets `Nonce.time`
-to the number of seconds elapsed since the start of the UNIX epoch, rounded down to the nearest multiple of `min_batch_duration`.
-This truncation is done to ensure that the report's timestamp cannot be used to
-link the report back to the originating client.
-
-The client then finishes the report generation by sharding its measurement into
-a sequence of input shares as specified by the VDAF in use. To encrypt an input
-share, the client generates an {{HPKE}} ciphertext for the aggregator by running
+This message is called the Client's report. It consists of a "header" and the
+encrypted input share of each of the Aggregators. The header consists of:
```
chris-wood, line 748: Here let's say that header consists of (1) task ID and (2) metadata, and then list what the metadata consists of. That way the text is consistent with the struct.

draft-ietf-ppm-dap.md
chris-wood: Looks good!

draft-ietf-ppm-dap.md
```diff
@@ -1305,50 +1500,97 @@ given batch interval, denoted `batch_interval`, decryption works as follows:
...
+
+##### Boundary Check
+
+For fixed-size tasks, the batch boundaries are defined by "batch IDs" selected
+by the Leader. Thus the Aggregator needs to check that the query is associated
+with a known batch ID:
+
+* If the query contained in the CollectReq has type `by-batch-id`, the
+  Leader checks that the provided batch ID corresponds to a batch ID it
+  returned in a previous CollectResp for the task.
```
chris-wood, line 1581: I don't think this requirement is necessary. The Collector should be able to query for batches by ID opportunistically, I think.

chris-wood (changes requested):
Only 1 blocking comment on the constraint that Collectors can't opportunistically query for batch IDs.

draft-ietf-ppm-dap.md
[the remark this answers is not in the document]

draft-ietf-ppm-dap.md
cjpatton: The task ID is included in each `Report`, but is not sent in each `ReportShare` sent during the agg flow. OTOH the metadata is unique to each `ReportShare`, which is why I've split metadata and task ID.

Agreed on minimizing PRs. Like I said, it's an artifact of a previous version of this PR. Still, I think it's a useful change. I'm happy to pull it out into a separate PR.

draft-ietf-ppm-dap.md
cjpatton: Good idea. Done!

draft-ietf-ppm-dap.md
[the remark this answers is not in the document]
cjpatton: I think that approach would drop (3.) for time-interval tasks, which at this point I prefer not to do :) Perhaps there is a version of this that'll work.

draft-ietf-ppm-dap.md
cjpatton: I don't see a way to enable this without creating lots of corner cases. I think we need two things in order to pursue this:
1. What's the motivation? Is there a reason why the current scheme is not workable for Collectors? What's the advantage of knowing the batch ID in advance, beyond not having to store it?
2. A way to address corner cases like: What does the Collector do if the Leader requests batch 9999, then batch 1? If the Leader is aggregating opportunistically (like it may do for Prio), it might have to wait to starting filling batch 9999 until 1, ..., 9998 have been filled.

This point has already been discussed at some length on the slack. I'm happy to keep discussing, but could we do so in an issue so we have a chance to level set?

cjpatton (commented):
> Only 1 blocking comment on the constraint that Collectors can't opportunistically query for batch IDs.

I think this point needs a bit more discussion. Can we do so in a new issue?

chris-wood:
> > Only 1 blocking comment on the constraint that Collectors can't opportunistically query for batch IDs.
>
> I think this point needs a bit more discussion. Can we do so in a new issue?

I don't really see how this introduces edge cases. If the batch ID doesn't exist yet, the leader can just discard it. The behavior is pretty straightforward.

I don't see where this has been discussed on slack. Please drop an OPEN ISSUE so we can track relaxing this constraint.

cjpatton:
> I don't see where this has been discussed on slack. Please drop an OPEN ISSUE so we can track relaxing this constraint.

Done. Filed https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/issues/301.

chris-wood (approved):
Nice work! Ship it!

cjpatton:
Squashed and updated commit message.

draft-ietf-ppm-dap.md
branlwyd: (sorry, this comment is a little late to the party)

> Each next-batch query would correspond to a different

I see -- in that case, I think there is potential for data-loss if the Collector crashes at an inopportune time. Consider the following:

1. The Leader has a collect job ready to be retrieved, call it `X`.
2. The Collector issues a `next-batch` request, receives a collect job corresponding to `X`.
3. Before storing the collect job ID to durable storage or polling for completion, the Collector (or some relevant subcomponent) crashes & loses its state.

In this case, I think the Collector has lost batch `X` permanently: it no longer knows the relevant collect job ID, and there is no way to interact with the Leader to retrieve it since further `next-batch` requests will result in different collect jobs being returned.

I think the best way to fix this might be based on the idea of making the batch IDs enumerable as suggested by #301. I'll add a comment to that bug.

PR #300, opened 2022-08-17
[retracted] Merge "fixed-interval" & "fixed-size" task types (mostly).

branlwyd:
# PR is retracted, please don't review.

([Reasoning for retracting PR](https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/pull/300#issuecomment-1218456857))

Almost all functionality & message wire formats are the same for the two
task types again. The largest point of divergence is in the batch
validation implemented by the leader & helper: this validation is used
to enforce the desired properties on each batch (having a fixed interval
of time or including a fixed number of reports, respectively).

See [motivating discussion](https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/issues/273#issuecomment-1213552532) in #273.

### DIFF
--- a/draft-ietf-ppm-dap.md
+++ b/draft-ietf-ppm-dap.md
@@ -386,8 +386,8 @@ in the "type" field (within the DAP URN namespace
 | reportTooLate              | Report could not be processed because it arrived too late. |
 | reportTooEarly             | Report could not be processed because its timestamp is too far in the future. |
 | batchInvalid               | A collect or aggregate-share request was made with invalid batch parameters. |
-| insufficientBatchSize      | There are not enough reports in the batch interval to satisfy the task's minimum batch size. |
-| batchLifetimeExceeded      | The batch lifetime has been exceeded for one or more reports included in the batch interval. |
+| invalidBatchSize           | There are an invalid number of reports in the batch. |
+| batchLifetimeExceeded      | The batch lifetime has been exceeded for one or more reports included in the batch. |
 | batchMismatch              | Aggregators disagree on the report shares that were aggregated in a batch. |
 | unauthorizedRequest        | Authentication of an HTTP request failed (see {{https-sender-auth}}). |
 
@@ -458,12 +458,8 @@ struct {
   Duration duration;
 } Interval;
 
-/* A nonce used to uniquely identify a report in the context of a DAP task. It
-includes the timestamp of the current batch and a random 16-byte value. */
-struct {
-  Time time;
-  uint8 rand[16];
-} Nonce;
+/* A nonce used to uniquely identify a report in the context of a DAP task. */
+Nonce uint8[16];
 
 /* The various roles in the DAP protocol. */
 enum {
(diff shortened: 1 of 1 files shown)


draft-ietf-ppm-dap.md
```diff
@@ -624,11 +501,20 @@ number generator. Each task has the following parameters associated with it:
   leader's endpoint MUST be the first in the list. The order of the
   `encrypted_input_shares` in a `Report` (see {{upload-flow}}) MUST be the
   same as the order in which aggregators appear in this list.
-* `query_config`: The query configuration (`QueryConfig`) for this task. This
-  determines the query type for batch selection and the properties that all
-  batches for this task must have.
+* `batch_generation_method`: The batch generation method determines how batches
+  of reports are generated & validated by the leader & helper for this task. It
```
cjpatton, line 505: Too informal for an RFC
```suggestion
  of reports are generated and validated by the leader and helper for this task. It
```

draft-ietf-ppm-dap.md
```diff
@@ -1276,23 +1137,28 @@ attacks.]
...
 The named parameters are:
 
 * `task_id`, the DAP task ID.
-* `query`, the Collector's query.
+* `batch`, the batch to collect.
 * `agg_param`, an aggregation parameter for the VDAF being executed.
   This is the same value as in `AggregateInitializeReq` (see {{leader-init}}).
 
+If the Collector's request includes a `batch` with both `start` and `end`
+having a zero `time` & and all-zero `nonce`, the Leader will choose a batch
```
cjpatton, line 1153:
```suggestion
having a zero `time` and all-zero `nonce`, the Leader will choose a batch
```

draft-ietf-ppm-dap.md
```diff
@@ -1481,92 +1332,56 @@ as follows:
...
-      AggregateShareReq corresponds to a batch ID used in a previous
-      AggregateInitializeReq for the task.
-
-If the boundary check fails, then the Aggregator MUST abort with error
-"batchInvalid".
+First the aggregator checks that the request respects any batch "boundaries":
+for the `batch` included in the query, the Aggregator checks that
+`batch.end > batch.start`. For fixed-interval tasks, the Aggregator
+additionally checks that both `batch.start.time` and `batch.end.time` are
+divisible by `min_batch_duration`, and that
```
cjpatton, line 1364: Do you mean time_precision here, or something else?
```suggestion
divisible by `time_precision`, and that
```

draft-ietf-ppm-dap.md
```diff
@@ -1481,92 +1332,56 @@ as follows:
...
-
-* For fixed-size tasks, the query configuration specifies the minimum batch
-  size, `min_batch_size`, and maximum batch size. The Aggregator checks that
-  `len(X) >= min_batch_size` and `len(X) <= max_batch_size`.
-
-If the batch size check fails, then the Aggregator MUST abort with error
-"invalidBatchSize".
+`N` denote the number of reports in the batch. All tasks must satisfy that
+`N >= min_batch_size`. Fixed-size tasks must satisfy the stricter condition
+that `N = min_batch_size`. If the batch size check fails, then the Aggregator
```
cjpatton, line 1371: I'm fairly sure we'll still need a bit of flexibility.
```suggestion
that `N <= max_batch_size`. If the batch size check fails, then the Aggregator
```

draft-ietf-ppm-dap.md
```diff
@@ -447,19 +447,22 @@ We start with some basic type definitions used in other messages.
...
+  uint8 rand[16];
+} Nonce;
 
-/* A nonce used to uniquely identify a report in the context of a DAP task. */
-Nonce uint8[16];
+/* A batch is defined by a range of nonces, from start (inclusive) to end (exclusive). */
+struct {
+  Nonce start;
+  Nonce end;
+} Batch;
```
cjpatton, line 465: (Just to check that I understand:) A batch `batch1` is "valid" if:
1. (Total ordering) `batch1.start.time < batch1.end.time` or `batch1.start.time == batch1.end.time` and `batch1.start.nonce < batch1.end.nonce`
2. (Non-overlapping) There is no prior batch `batch0` that intersects with `batch1`, i.e., there is at least one nonce contained in both batches
3. (Boundary aligned) `batch1.start.time` and `batch1.start.end` are both equal to `0 (mod time_precision)`

Anything I'm missing?

branlwyd:
Based on https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/issues/273#issuecomment-1218133868 & the goal to have `fixed-size` batches be collected as soon as they are available, I am retracting this PR: it introduces a delay in collection which may be as long as the timestamp rounding factor, which per the linked comment may be on the order of hours.

Bringing this back to draft for now in case some of the text is useful, but please hold off on reviewing.

cjpatton:
Thanks for working on this Brandon, it was useful to see this idea spelled out. A good high level goal should be to minimize the differences between various DAP "modes", so we should revisit in the future.

Issue #278, opened 2022-06-23
Resource oriented HTTP API (was: Conveying task and job IDs)


chris-wood:
Issue #261 touched on how task and job parameters are conveyed between aggregators across request and response messages. This issue tracks the more general question of how we convey these parameters. They could be conveyed in application messages as they are now, in URLs as query parameters, or even headers. 

@mnot, if you have cycles, I'd be very interested to hear what you think.

_Originally posted by @BranLwyd in https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/issues/261#issuecomment-1132242994_

mnot:
What's the deployment model? In other words, if I wanted to deploy this protocol, would I:

a) spin up a new, dedicated server and treat it like a black box
b) create new resources on an potentially existing server that are specific to this protocol
c) apply the protocol to potentially existing resources on a server

?

chris-wood:
I think (a) is going to be the most common deployment model here.

cjpatton:
Based on my experience implementing the current draft, I think it would be great to move the task ID to either a header or the query string of the URL. The latter might be preferable since it's more consistent with `/hpke_config`. I'm agnostic about the aggregation-job ID.

The problem solved by moving the task ID to either the URL or a header is the following: Before parsing a message it's useful to first make sure that the message is authentic: https://github.com/cloudflare/daphne/blob/main/daphne/src/roles.rs#L421

However to authenticate a DAP message it's necessary to at least parse the task ID in order to look up the bearer token: https://github.com/cloudflare/daphne/blob/main/daphne/src/auth.rs#L84-L87

mnot:
Will it ever be necessary to send these requests in an unmodified browser (either by directly entering the Urls into the location bar, or using forms or javascript)?

If not, it's largely a matter of stylistic preference. 

URL parameters are visible in the location bar (and in logs); their semantics are determined by the resource you're interacting with. So, if you use them you should be approaching the protocol as an exercise in resource modelling -- i.e., "this resource takes those parameters, and does that".

The body is similar to URL parameters in terms of scope of semantics, but needs to be identified with a media type. Also it's not visible, and not available on some request methods, and its processing is tied to the request method to some degree (most loosely with POST).

[...]

Issue #248, opened 2022-05-12
No negotiation for HPKE Configs


ekr:
The server only supplies one config, so what happens if the server would support X22519 and P-256? We can fix this by having the server supply multiple configs.

cjpatton:
This is probably a good idea. We could just change the payload of the response to `GET /hpke_config` to a sequence of `HpkeConfig` messages rather than one.

cjpatton:
As mentioned in https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/pull/303#discussion_r954175414, it might be useful to allow the client to advertise which algorithms it supports. This would increase the likelihood of successfully negotiating an a HPKE suite that both parties support.

cjpatton:
@ekr should we try to tackle this in DAP-03?

PR #303, opened 2022-08-24
Make "task_id" optional in "/hpke_config" endpoint

### LINKED ISSUE #289: Allow hpke_config to be queried without task_id
In [HPKE Configuration Request](https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/blob/main/draft-ietf-ppm-dap.md#hpke-configuration-request-hpke-config), the `[aggregator]/hpke_config` endpoint will be called with a `task_id`, unrecognised `task_id` will result in a 404. 
This requires clients (or any entity wishes to call `[aggregator]/hpke_config`) to know a task id in advance. This is unnecessary for deployments that want to share HPKE config across multiple tasks. Can we consider make `task_id` optional in this endpoint, when queried without `task_id`, aggregator returns either one or a list of `hpke_config`.

junyechen1996:
Closes #289.

### DIFF
--- a/draft-ietf-ppm-dap.md
+++ b/draft-ietf-ppm-dap.md
@@ -390,6 +390,7 @@ in the "type" field (within the DAP URN namespace
 | batchLifetimeExceeded      | The batch lifetime has been exceeded for one or more reports included in the batch. |
 | batchMismatch              | Aggregators disagree on the report shares that were aggregated in a batch. |
 | unauthorizedRequest        | Authentication of an HTTP request failed (see {{https-sender-auth}}). |
+| missingTaskID              | HPKE configuration was requested without specifying a task ID. |
 
 This list is not exhaustive. The server MAY return errors set to a URI other
 than those defined above. Servers MUST NOT use the DAP URN namespace for errors
@@ -654,14 +655,23 @@ configuration of each aggregator. See {{compliance}} for information on HPKE
 algorithm choices.
 
 Clients retrieve the HPKE configuration from each aggregator by
-sending an HTTP GET request to `[aggregator]/hpke_config?task_id=[task-id]`, where
-`[aggregator]` is the aggregator's endpoint URL, obtained from the task
-parameters, and `[task-id]` is the task ID obtained from the task parameters,
+sending an HTTP GET request to `[aggregator]/hpke_config`, where `[aggregator]`
+is the aggregator's endpoint URL, obtained from the task parameters.
+Clients MAY specify a query parameter `task_id` when sending an
+HTTP GET request to `[aggregator]/hpke_config?task_id=[task-id]`,
+where `[task-id]` is the task ID obtained from the task parameters,
 encoded in Base 64 with URL and filename safe alphabet with no padding, as
 specified in sections 5 and 3.2 of {{!RFC4648}}. If the aggregator does not
 recognize the task ID, then it responds with HTTP status code 404 Not Found and
-an error of type `unrecognizedTask`. The aggregator responds to well-formed
-requests with HTTP status code 200 OK and an `HpkeConfig` value:
+an error of type `unrecognizedTask`.
+
+An aggregator is free to use different HPKE configurations for each task with
+which it's configured. If the task ID is missing from a client's request,
+the aggregator MAY abort with an error of type `missingTaskID`, in which case
+the client SHOULD retry the request with a well-formed task ID included.
+
+An aggregator responds to well-formed requests with HTTP status code 200 OK
+and an `HpkeConfig` value:
 
 [TODO: Allow aggregators to return HTTP status code 403 Forbidden in deployments
 that use authentication to avoid leaking information about which tasks exist.]


wangshan:
looks good, @chris-wood is the size of hpke_configs <1..2^16-1> needs expanding, given that it can contain more than one configs?

draft-ietf-ppm-dap.md
```diff
@@ -654,19 +654,29 @@ configuration of each aggregator. See {{compliance}} for information on HPKE
...
+where `[task-id]` is the task ID obtained from the task parameters,
 encoded in Base 64 with URL and filename safe alphabet with no padding, as
 specified in sections 5 and 3.2 of {{!RFC4648}}. If the aggregator does not
 recognize the task ID, then it responds with HTTP status code 404 Not Found and
-an error of type `unrecognizedTask`. The aggregator responds to well-formed
-requests with HTTP status code 200 OK and an `HpkeConfig` value:
+an error of type `unrecognizedTask`. If a deployment enforces different HPKE
+configurations for different tasks, and a client doesn't specify `task_id` in the
+HTTP GET request, then the aggregator responds with HTTP status code 400 Bad Request
+and an error of type `unrecognizedMessage`. The aggregator responds to well-formed
```
tgeoghegan, line 669: `unrecognizedMessage` and HTTP 400 seem inappropriate. The problem isn't an incorrectly encoded message, it's that this particular aggregator chooses not to implement this functionality. If a client does `GET /hpke_config` and the server doesn't support it, it should get some indication that it should now try `GET /hpke_config?task_id=[task_id]`.

[HTTP 422](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/422) seems pretty appropriate for this case ("the server understands the content type of the request entity, and the syntax of the request entity is correct, but it was unable to process the contained instructions"). I think we should use that status code and define a new problem document type (in section 3.1 errors) for this specific case.

tgeoghegan (changes requested):

draft-ietf-ppm-dap.md
```diff
@@ -654,19 +654,29 @@ configuration of each aggregator. See {{compliance}} for information on HPKE
 algorithm choices.
 
 Clients retrieve the HPKE configuration from each aggregator by
-sending an HTTP GET request to `[aggregator]/hpke_config?task_id=[task-id]`, where
+sending an HTTP GET request to `[aggregator]/hpke_config`, where
 `[aggregator]` is the aggregator's endpoint URL, obtained from the task
-parameters, and `[task-id]` is the task ID obtained from the task parameters,
+parameters. A deployment MAY choose to use different HPKE configurations
+for different tasks, so clients can optionally specify a query parameter `task_id`
```
cjpatton, line 660:
```suggestion
parameters. A deployment may choose to use different HPKE configurations
for different tasks, so clients MAY specify a query parameter `task_id`
```

draft-ietf-ppm-dap.md
cjpatton: +1 to new abort type (e.g., "missingTaskId"), but @tgeoghegan, we currently we return the same status code for all abort cases (400). It might be a good idea to let the status code differ for different abort types, but we should leave that out of this change I think.

draft-ietf-ppm-dap.md
```diff
@@ -654,19 +654,29 @@ configuration of each aggregator. See {{compliance}} for information on HPKE
...
+and an error of type `unrecognizedMessage`. The aggregator responds to well-formed
+requests with HTTP status code 200 OK and a vector of `HpkeConfig` value:
 
 [TODO: Allow aggregators to return HTTP status code 403 Forbidden in deployments
 that use authentication to avoid leaking information about which tasks exist.]
 
 ~~~
+struct {
+  HpkeConfig hpke_configs<1..2^16-1>;
+} HpkeConfigResp;
```
cjpatton, line 678: I don't think this adequately addresses #248:
1. You have defined a new struct for a sequence of HPKE configs, but you have not changed the return type (see line 670 above).
2. This hasn't been discussed in the issue yet, but merely allowing the server to return multiple HPKE configs doesn't totally solve the problem. What is needed, I think, is a mechanism for allowing the client to indicate which cipher suites its supports.

In any case, I think we should push this change to a future PR. Best practices for spec review (as well as code review!) is to keep changes minimal.

cjpatton (changes requested):

draft-ietf-ppm-dap.md
tgeoghegan: That's fair enough. The problem document type will make this explicit enough, and [RFC 9205](https://httpwg.org/specs/rfc9205.html#using-http-status-codes) recommends using the most general possible HTTP status anyway (in this case, 400).

junyechen1996: Sounds good, I added a new `missingTaskId` problem type under {{Errors}}.

draft-ietf-ppm-dap.md
junyechen1996: Fixed now.

draft-ietf-ppm-dap.md
junyechen1996: Sounds good, I reverted the changes on returning a vector of `HpkeConfig`. Previously I changed line 670 to indicate the endpoint would return a vector of `HpkeConfig` values, maybe I didn't explicitly say the return type would change to `HpkeConfigResp`.

My initial thinking was the endpoint would return all `HpkeConfig` that it supports, and clients can use any of them with a cipher suite that it recognizes. But I'm fine with addressing this in a separate PR if that's clearer.

cjpatton: In practice that might decrease the likelihood of failing to negotiate an HPKE suite, but I think we can do a bit better. In any case, this can/should be handled separately.

draft-ietf-ppm-dap.md
```diff
@@ -390,6 +390,7 @@ in the "type" field (within the DAP URN namespace
 | batchLifetimeExceeded      | The batch lifetime has been exceeded for one or more reports included in the batch. |
 | batchMismatch              | Aggregators disagree on the report shares that were aggregated in a batch. |
 | unauthorizedRequest        | Authentication of an HTTP request failed (see {{https-sender-auth}}). |
+| missingTaskId              | `/hpke_config` endpoint received a request without `task_id` parameter (see {{hpke-config}}). |
```
cjpatton, line 393: While precise, this description of error is hard to parse without knowing the protocol context. What's going on here, intuitively?

cjpatton: Also, @tgeoghegan: would you prefer `missingTaskId` or `missingTaskID`? I think the latter might be more consistent.

draft-ietf-ppm-dap.md
```diff
@@ -654,13 +655,19 @@ configuration of each aggregator. See {{compliance}} for information on HPKE
...
+for different tasks, so clients MAY specify a query parameter `task_id`
+when sending an HTTP GET request to `[aggregator]/hpke_config?task_id=[task-id]`,
+where `[task-id]` is the task ID obtained from the task parameters,
 encoded in Base 64 with URL and filename safe alphabet with no padding, as
 specified in sections 5 and 3.2 of {{!RFC4648}}. If the aggregator does not
 recognize the task ID, then it responds with HTTP status code 404 Not Found and
-an error of type `unrecognizedTask`. The aggregator responds to well-formed
+an error of type `unrecognizedTask`. If a deployment enforces different HPKE
+configurations for different tasks, and a client doesn't specify `task_id` in the
+HTTP GET request, then the aggregator responds with HTTP status code 400 Bad Request
```
cjpatton, line 669:
```suggestion
HTTP GET request, then the aggregator MUST respond with HTTP status code 400 Bad Request
```

cjpatton: This could be stated a bit simpler as:
> then the Aggregator MUST abort with error `missingTaskId`.

cjpatton (changes requested):
Looking good. One high level question I have: How does a client decide whether to add a query parameter or not? Perhaps we should try to spell out this discovery process:
* The Client first sends GET /hpke_config.
* The Aggregator MAY abort with "missingTaskID", in which case the client retries with `task_id`.
* The Client is of course free to skip the first request if it knows it's going to need to provide a `task_id`.

draft-ietf-ppm-dap.md
junyechen1996: > While precise, this description of error is hard to parse without knowing the protocol context. What's going on here, intuitively?

Should I add on to the above text with "if a deployment enforces different HPKE configurations for different tasks"?

Regarding TaskId vs TaskID, I think the type definition is `TaskId` at line 608, but there are struct fields where `TaskID` is used, so it seems like there is some inconsistency here. I was trying to follow the original type definition.

cjpatton: > Should I add on to the above text with "if a deployment enforces different HPKE configurations for different tasks"?

I would simplify it. How about:
```
HPKE config requested without specifying the task ID
```

> Regarding TaskId vs TaskID, I think the type definition is `TaskId` at line 608, but there are struct fields where `TaskID` is used, so it seems like there is some inconsistency here. I was trying to follow the original type definition.

Ah OK, nice catch. Let's correct the inconsistency here. It seems like the preference is likely to be "ID".

junyechen1996:
> Looking good. One high level question I have: How does a client decide whether to add a query parameter or not? Perhaps we should try to spell out this discovery process:
> 
> * The Client first sends GET /hpke_config.
> * The Aggregator MAY abort with "missingTaskID", in which case the client retries with `task_id`.
> * The Client is of course free to skip the first request if it knows it's going to need to provide a `task_id`.

Do you mean I should have some texts specifying the behavior when a client doesn't know if a deployment enforces different HPKE configurations for different tasks? Some thoughts I have here:
- As a first step, after this sentence: "then the aggregator responds with HTTP status code 400 Bad Request and an error of type `missingTaskID`.", I can add: "The client can then retry the request with a well-formed Task ID".
- With the above sentence, I think it's fairly clear that clients should retry the request with a task ID if the problem type is `missing TaskID`. I can try to spell out the discovery process as well, but I don't know if it makes the protocol text unnecessarily long.
- If we need to spell this process out, regarding the "MAY" in your second bullet point, should I avoid using it, because I think it is a deployment-specific behavior? I think I can just say "If the aggregator aborts with `missingTaskID`, client can retry the request by specifying `task_id`.

cjpatton:
> Do you mean I should have some texts specifying the behavior when a client doesn't know if a deployment enforces  different HPKE configurations for different tasks? Some thoughts I have here:

Yes :)

> - As a first step, in the existing protocol text, after this sentence: "then the aggregator responds with HTTP status code 400 Bad Request and an error of type `missingTaskID`", I can add: "The client can then retry the request by specifying a well-formed Task ID as the `task_id` parameter".
> - With the above sentence, I think it's fairly clear that clients should retry the request with a task ID if the problem type is `missingTaskID`. I can try to spell out the discovery process as well, but I don't know if it makes the protocol text unnecessarily long.
> - If we need to spell this process out, regarding the "MAY" in your second bullet point, should I avoid using it, because I think it is a deployment-specific behavior? I think I can just say "If the aggregator aborts with `missingTaskID`, client can retry the request by specifying `task_id`.

I think spelling it out is a good idea. It shouldn't be too much text, but it may require a bit of refactoring. "MAY" is intended to capture deployment-specific behavior, so it's appropriate here.

draft-ietf-ppm-dap.md
junyechen1996: I added the MUST keyword, but still kept the HTTP status code, to be more explicit.

junyechen1996:
> I think spelling it out is a good idea. It shouldn't be too much text, but it may require a bit of refactoring. "MAY" is intended to capture deployment-specific behavior, so it's appropriate here.

Sounds good. I split the texts into three paragraphs: the first paragraph introduces the `/hpke_config` endpoint. The second paragraph talks about a deployment MAY enforce different HPKE configurations for different tasks, and different cases of aggregators' aborting. The third paragraph spells out client's discovery process of whether `task_id` is required. Let me know if it makes sense to you.

draft-ietf-ppm-dap.md
tgeoghegan: [RFC 7807](https://www.rfc-editor.org/rfc/rfc7807.html#section-4) doesn't have any obvious guidance I can see, but if we look at [ACME aka RFC 8555](https://www.rfc-editor.org/rfc/rfc8555.html#section-6.7)'s errors, it has types like `badCSR`, `caa` and `dns`, which I think would lead us to using `missingTaskId`, and that would be consistent with the struct definitions elsewhere. But I'm happy with either; the ambiguous case problems weren't introduced by this change and can always be cleaned up later in the standardization process.

draft-ietf-ppm-dap.md
```diff
@@ -654,14 +655,27 @@ configuration of each aggregator. See {{compliance}} for information on HPKE
...
 
 Clients retrieve the HPKE configuration from each aggregator by
-sending an HTTP GET request to `[aggregator]/hpke_config?task_id=[task-id]`, where
-`[aggregator]` is the aggregator's endpoint URL, obtained from the task
-parameters, and `[task-id]` is the task ID obtained from the task parameters,
+sending an HTTP GET request to `[aggregator]/hpke_config`, where `[aggregator]`
+is the aggregator's endpoint URL, obtained from the task parameters.
+
+A deployment MAY enforce different HPKE configurations for different tasks.
+In this case, clients MUST specify a query parameter `task_id` when
```
tgeoghegan, line 662: I think MUST isn't right in this case, because it's not a violation of the protocol for the client to not send a `task_id` query parameter when the aggregator lacks a "global" HPKE config. Same goes for the one one line 669, because an aggregator is always free to respond to a request without the `task_id` query parameter.

tgeoghegan (changes requested):

draft-ietf-ppm-dap.md
junyechen1996: I think this paragraph tries to characterize the behaviors of an aggregator, when the deployment needs to enforce different HPKE configurations for different tasks. In this case, clients are required to send `task_id`, and aggregators are required to abort if no `task_id` is included in the request. Should I be clearer about the assumptions in this paragraph, or should I relax the "MUST" keyword, perhaps to "SHOULD"?

draft-ietf-ppm-dap.md
cjpatton: Works for me

draft-ietf-ppm-dap.md
```diff
@@ -654,14 +655,27 @@ configuration of each aggregator. See {{compliance}} for information on HPKE
...
+HTTP GET request, then the aggregator MUST abort with HTTP status code 400 Bad Request
+and an error of type `missingTaskID`.
+
+When a client doesn't know whether it needs to provide the `task_id` parameter, it can
+first send an HTTP GET request to `[aggregator]/hpke_config`. The aggregator MAY abort with
+an error of type `missingTaskID`, in which case the client can retry the request with a
+well-formed `task_id`. Clients are free to skip the first HTTP request, if they know
+they need to provide a `task_id`.
+
+The aggregator responds to well-formed requests with HTTP status code 200 OK and an `HpkeConfig` value:
```
cjpatton, line 678: nit: Wrap lines at 80 characters

draft-ietf-ppm-dap.md
cjpatton: That's fine, but if later on we decide to change the status code, we'll have to update here as well. That's why it's better to specify something once.

draft-ietf-ppm-dap.md
cjpatton: I agree with Tim here. The problem is that the Client may not know a priori that the enforces HPKE key separation per task. I think a different tack is needed.

In the paragraph above where you specify the GET request, you should also add: "The Client MAY include the task ID with the query parameters. [Spell out what the path looks like.]"

In this paragraph, you should start off with: "The Aggregator is free to use different HPKE configurations for each task with which it's configured. If the task ID is missing from the Client's request, the Aggregator MAY abort with error "missingTaskID", in which case the Client SHOULD retry the request with the task ID included."

This accomplishes two things:
* It provides a way to obtain an HPKE config, regardless of how the Aggregator is configured.
* It permits the optional, deployment-specific beahvior.

draft-ietf-ppm-dap.md
```diff
@@ -654,14 +655,27 @@ configuration of each aggregator. See {{compliance}} for information on HPKE
...
-requests with HTTP status code 200 OK and an `HpkeConfig` value:
+an error of type `unrecognizedTask`. If a client doesn't specify `task_id` in the
+HTTP GET request, then the aggregator MUST abort with HTTP status code 400 Bad Request
+and an error of type `missingTaskID`.
+
+When a client doesn't know whether it needs to provide the `task_id` parameter, it can
+first send an HTTP GET request to `[aggregator]/hpke_config`. The aggregator MAY abort with
+an error of type `missingTaskID`, in which case the client can retry the request with a
+well-formed `task_id`. Clients are free to skip the first HTTP request, if they know
+they need to provide a `task_id`.
```
cjpatton, line 676: I would fold this into the previous paragraph, as described above.

draft-ietf-ppm-dap.md
junyechen1996: Fixed now.

draft-ietf-ppm-dap.md
junyechen1996: @tgeoghegan @cjpatton Please take a look again at the revised wording based on Chris's suggestions.

draft-ietf-ppm-dap.md
```diff
@@ -654,14 +655,23 @@ configuration of each aggregator. See {{compliance}} for information on HPKE
...
-`[aggregator]` is the aggregator's endpoint URL, obtained from the task
-parameters, and `[task-id]` is the task ID obtained from the task parameters,
+sending an HTTP GET request to `[aggregator]/hpke_config`, where `[aggregator]`
+is the aggregator's endpoint URL, obtained from the task parameters.
+Clients MAY specify a query parameter `task_id` when sending an
+HTTP GET request to `[aggregator]/hpke_config?task_id=[task-id]`,
+where `[task-id]` is the task ID obtained from the task parameters,
 encoded in Base 64 with URL and filename safe alphabet with no padding, as
 specified in sections 5 and 3.2 of {{!RFC4648}}. If the aggregator does not
 recognize the task ID, then it responds with HTTP status code 404 Not Found and
```
cjpatton, line 665: (Orthogonal to this PR, so we can clean up later if needed.) @tgeoghegan is 404 correct here you think?

cjpatton (approved):
LGTM! One minor question for reviewers, which we can punt on.

draft-ietf-ppm-dap.md
cjpatton: Looks good to me.

junyechen1996:
> LGTM! One minor question for reviewers, which we can punt on.

Thanks! I just squashed commits into one.

draft-ietf-ppm-dap.md
tgeoghegan: My reading of [RFC 9205's discussion of HTTP statuses](https://www.rfc-editor.org/rfc/rfc9205#name-using-http-status-codes) is that we should lean toward HTTP 400 and communicate more specific issues with the problem document, but [I made a note in #278](https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/issues/278#issuecomment-1226237695) that we should take a look at all this, so I don't think we need to iterate on that in this PR.

tgeoghegan (approved):
This works, thank you for iterating on it with us!

tgeoghegan:
These changes were approved about a week ago and there have been no comments since, so I am merging.

PR #304, opened 2022-08-24
Adding max_task_lifetime to task parameters

### LINKED ISSUE #291: Consider adding max_task_lifetime to force retire a task
The current spec doesn't explain how to retire or delete a task. For some use cases, it's useful to have a hard task deadline, when expired, the task and all its aggregations must be made inactive and eventually deleted. This can be a new Duration type task parameter called `max_task_lifetime`, or `max_task_duration` to avoid similar named but int typed `max_batch_lifetime`.

Other than the practical benefit, for clients that uploads repeatedly and periodically, this can add privacy benefit to limit the number of uploads from the same client to the same task (related to https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/issues/210).

PS. I feel like "lifetime" is more appropriate for a deadline than a max allowed count.

wangshan:
Closes #291.

### DIFF
--- a/draft-ietf-ppm-dap.md
+++ b/draft-ietf-ppm-dap.md
@@ -622,6 +622,9 @@ number generator. Each task has the following parameters associated with it:
   task must have.
 * `max_batch_lifetime`: The maximum number of times a batch of reports may be
   queried by the Collector.
+* `task_expiration`: The time up to which clients are expected to upload to this
+  task. The task is considered completed after this time. Aggregators MAY
+  reject reports that have timestamps later than `task_expiration`.
 * A unique identifier for the VDAF instance used for the task, including the
   type of measurement associated with the task.
 
@@ -784,9 +787,11 @@ retrying.
 The Leader MUST ignore any report pertaining to a batch that has already been
 collected. (See {{input-share-batch-validation}} for details.) Otherwise,
 comparing the aggregate result to the previous aggregate result may result in a
-privacy violation. (Note that the Helpers enforce this as well.) In addition,
-the leader SHOULD abort the upload protocol and alert the client with error
-"reportTooLate".
+privacy violation. (Note that the Helpers enforce this as well.) The Leader MAY
+ignore any reports whose timestamp is past the task's `task_expiration`. When it does so,
+the leader SHOULD abort the upload protocol and alert the client with
+error "reportTooLate". Client MAY choose to opt out of the task if its own
+clock has passed `task_expiration`.
 
 Leaders can buffer reports while waiting to aggregate them. The leader SHOULD
 NOT accept reports whose timestamps are too far in the future. Implementors MAY
@@ -940,6 +945,7 @@ enum {
   hpke-decrypt-error(4),
   vdaf-prep-error(5),
   batch-saturated(6),
+  task-expired(7),
   (255)
 } ReportShareError;
 ~~~
@@ -1134,7 +1140,10 @@ The validation checks are as follows.
       Aggregator MAY mark the input share as invalid with the error
       "batch-saturated". Note that this behavior is not strictly enforced here
       but during the collect sub-protocol. (See {{batch-validation}}.)
-If both checks succeed, the input share is not marked as invalid.
+      If both checks succeed, the input share is not marked as invalid.
+4. Check if the report's timestamp has passed its task's `task_expiration`
+   time, if so the Aggregator MAY mark the input share as invalid with the error
+   "task-expired".
 
 #### Input Share Preparation {#input-share-prep}
 
@@ -1800,16 +1809,22 @@ designed to allow implementations to reduce operational costs in certain cases.
 
 ### Reducing storage requirements
 
-In general, the aggregators are required to keep state for all valid reports for
-as long as collect requests can be made for them. In particular, the aggregators
-must store a batch as long as the batch has not been queried more than
-`max_batch_lifetime` times. However, it is not always necessary to store the
-reports themselves. For schemes like Prio in which the input-validation protocol
-is only run once per report, each aggregator only needs to store its
+In general, the aggregators are required to keep state for tasks and all valid
+reports for as long as collect requests can be made for them. In particular,
+aggregators must store a batch as long as the batch has not been queried more
+than `max_batch_lifetime` times. However, it is not always necessary to store
+the reports themselves. For schemes like Prio3 {{!VDAF}} in which the input-validation
+protocol is only run once per report, each aggregator only needs to store its
 aggregate share for each possible batch interval, along with the number of times
 the aggregate share was used in a batch. This is due to the requirement that the
 batch interval respect the boundaries defined by the DAP parameters. (See
 {{batch-validation}}.)
+
+Furthermore, the aggregators must store data related to a task as long as the
+current time has not passed this task's `task_expiration`. Aggregator MAY delete
+the task and all data pertaining to this task after `task_expiration`.
+Implementors SHOULD provide for some leeway so the collector can collect the batch
+after some delay.
 
 # Compliance Requirements {#compliance}
 


draft-ietf-ppm-dap.md
```diff
@@ -621,6 +621,8 @@ number generator. Each task has the following parameters associated with it:
   task must have.
 * `max_batch_lifetime`: The maximum number of times a batch of reports may be
   queried by the Collector.
+* `max_task_lifetime`: The maximum timestamp of a task, after which the aggregators
```
cjpatton, line 624: "maximum timestamp of a task" is a bit ill-defined. Can you be more precise about what this means? Is it that reports with timestamps exceeding the max_task_lifetime are rejected? Or is it that requests are aborted if they arrive after max_task_lifetime?

cjpatton (changes requested):
We will need to be more precise about client/aggregator/collector behavior. Specifically, the text needs to prescribe optional, or even mandatory behavior. For example:  "Aggregators MAY (or MUST?) reject reports whose timestamps exceed `max_task_lifetime`."

wangshan:
@cjpatton I have made the definition more precise, in the process changing the meaning of this field a bit, please have another pass.

draft-ietf-ppm-dap.md
```diff
@@ -1712,8 +1727,8 @@ designed to allow implementations to reduce operational costs in certain cases.
 
 ### Reducing storage requirements
 
-In general, the aggregators are required to keep state for all valid reports for
-as long as collect requests can be made for them. In particular, the aggregators
+In general, the aggregators are required to keep state for tasks and all valid reports for as long
```
cjpatton, line 1730: nit: Wrap lines at 80 characters.

cjpatton (changes requested):
This is looking pretty good to me. One thing we might be missing is report rejection during the aggregation flow. In particular, do we need to spell out how the Leader/Helper rejects reports with time stamps that exceed the max batch lifetime?

draft-ietf-ppm-dap.md
```diff
@@ -1723,6 +1738,12 @@ the aggregate share was used in a batch. This is due to the requirement that the
 batch interval respect the boundaries defined by the DAP parameters. (See
 {{batch-validation}}.)
 
+Furthermore, the aggregators must store data related to a task as long as current epoch
+time has not passed this task's `max_task_lifetime`. Aggregator MAY delete the task and all
+data pertaining to this task after `max_task_lifetime`. Implementors SHOULD provide for some
+leeway so collector can collect the batch after some delay.
```
tgeoghegan, line 1744: We already define type `Time` in terms of the Unix epoch, so no need to refer to that here.
```suggestion
Furthermore, the aggregators must store data related to a task as long as the current time
has not passed this task's `max_task_lifetime`. Aggregator MAY delete the task and all
data pertaining to this task after `max_task_lifetime`. Implementors SHOULD provide for some
leeway so collector can collect the batch after some delay.
```

draft-ietf-ppm-dap.md
```diff
@@ -666,6 +669,9 @@ requests with HTTP status code 200 OK and an `HpkeConfig` value:
 [TODO: Allow aggregators to return HTTP status code 403 Forbidden in deployments
 that use authentication to avoid leaking information about which tasks exist.]
 
+[OPEN ISSUE: Decide whether aggregators should return 403 or 404 if `max_task_lifetime`
+for the given `task-id` has passed.]
```
tgeoghegan, line 673: Let's not add new language referencing HTTP error codes more specific than 400 or 500 since [RFC 9205 cautions against that](https://www.rfc-editor.org/rfc/rfc9205#name-using-http-status-codes). In any case, I don't think we should require that aggregators stop advertising HPKE config for some task ID at any moment in time. The reason is that HPKE configs are likely to be cached for a long time (elsewhere the document explicitly suggests caching them for days), meaning it'd be difficult for implementations to meaningfully guarantee that they'll stop serving a task's HPKE config at the precise instant of task expiration. I don't see any harm in an aggregator continuing to serve expired HPKE configs because the subsequent upload requests to an expired task will reliably fail anyway.

draft-ietf-ppm-dap.md
```diff
@@ -789,6 +795,11 @@ provide for some small leeway, usually no more than a few minutes, to account
 for clock skew. If the leader rejects a report for this reason, it SHOULD abort
 the upload protocol and alert the client with error "reportTooEarly".
 
+The Leader SHOULD ignore any reports pertaining to a task that have timestamps pass
+`max_task_lifetime`. Leader SHOULD abort the upload protocol and alert the client
+with error "reportTooLate". Client MAY choose to opt out of the task if its own clock
+has passed `max_task_lifetime`.
```
tgeoghegan, line 801: Should the leader make this determination based on the timestamp in the report, or based on the leader's perception of the current time when it receives the report?

Put another way: suppose a task has `max_batch_lifetime` = `t_1`. The leader receives a report with timestamp `t_0 < t_1`. The leader checks its own clock and determines that the current time is `t_2 > t_1`. Should the report be rejected because it was too late by the time the leader saw it, or accepted because its timestamp indicates the measurement was taken before the end of the task?

I think it's got to be the former (because otherwise clients could send back-dated reports long after a task has expired), in which case this should be rewritten to discuss time of report reception rather than the timestamp.

draft-ietf-ppm-dap.md
```diff
@@ -621,6 +621,9 @@ number generator. Each task has the following parameters associated with it:
   task must have.
 * `max_batch_lifetime`: The maximum number of times a batch of reports may be
   queried by the Collector.
+* `max_task_lifetime`: The timestamp up to which clients are allowed to upload
```
tgeoghegan, line 624: I think defining this as a timestamp (i.e., the encoding of an instant in time that might appear in a message) is wrong. The `max_task_lifetime` is an instant in time.

`lifetime` also seems like the wrong word. For one thing, it collides awkwardly with `max_batch_lifetime` (which refers to a totally different quantity). For another, "lifetime" implies to me a span of time, or a duration, but this is a single instant. Would `task_expiration` be better?

draft-ietf-ppm-dap.md
```diff
@@ -1501,11 +1512,15 @@ its query:
...
 
-Before an Aggregator responds to a CollectReq or AggregateShareReq, it must
+Before an Aggregator responds to a CollectReq or AggregateShareReq, it MUST
 first check that the request does not violate the parameters associated with the
 DAP task. It does so as described here.
 
-First the Aggregator checks that the batch respects any "boundaries" determined
+First the Aggregator checks that the task still exists, if the task identified by
+`task_id` no longer exists (for e.g. deleted after `max_task_lifetime` passed),
+then the Aggregator MUST abort with error "unrecognizedTask".
```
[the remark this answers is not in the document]
tgeoghegan: I don't think we need this paragraph, but for a different reason: it should be OK to do aggregations on a task after the aggregators have stopped accepting client reports for it. Let's say a task has `max_task_lifetime` `t` and it takes `d` seconds to run the collect protocol. So far as clients know, it is OK for them to continue submitting reports right up until instant `t`. As written, the collector must initiate the collect protocol at time `t - d`. So that means any reports that arrive during the interval `(t - d, t)` won't be included in the aggregation.

The other problem is that there's no good way in the protocol for the collector to figure out what `d` is and thus figure out how late it can issue a collect request without risking losing an aggregation because the task expired.

In fact, that seems to be implied by the text added in "Reducing storage requirements".

tgeoghegan (changes requested):
There's an ambiguity here: is `task.max_batch_lifetime` an instant in time, or a timestamp (i.e., an encoding of an instant in time that might appear in a protocol message)?

draft-ietf-ppm-dap.md
wangshan: I agree `task_expiration` is a better name if it's defined as a timestamp.

What's the main concern of encoding an instant in time in the protocol message (sorry if this is  obvious). Defining it as a duration will make it very hard for client to check this, since a client doesn't know the starting time of a task.

draft-ietf-ppm-dap.md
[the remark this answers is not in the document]
wangshan: I'll revert the change.

@tgeoghegan The current description indeed doesn't say explicitly how soon a task should be retired after expiry, IMO it should be left to the deployment. Collector should know `t` and try to collect before `t` expires.

draft-ietf-ppm-dap.md
wangshan: @tgeoghegan I think it should be based on a timestamp in the report. Assuming we want to give client a chance to check this, it can't be a duration since client may "wake up" at very different times.
Even for aggregators, if we determine by arrival time, then we could lose reports due to any latency caused on route to leader. The threat you mentioned is real, but attacker would have to hijack a large number of clients to make this attack effective.

wangshan:
> In particular, do we need to spell out how the Leader/Helper rejects reports with time stamps that exceed the max batch lifetime?

do you mean we should specify whether expired reports should be result in upload error or be dropped silently?

draft-ietf-ppm-dap.md
wangshan: having said that, I realised that given the `time_precision` is used to blur client timestamp, if we really want client to check this expiry time, then we should round (up?) `task_expiry` with `time_precision` too, otherwise if time_expiry is close to the rounded timestamp, people could figure out a more accurate timestamp of client.

draft-ietf-ppm-dap.md
divergentdave: I think the distinction is that `task_expiration` should be defined as an abstract instant in time, not a timestamp representing that instant. Since the expiration is going to be shared out of bounds, and won't appear in encoded form anywhere in the protocol, it would be cleanest to define it as a point in time, and not define a timestamp encoding for it that the protocol doesn't use.

draft-ietf-ppm-dap.md
tgeoghegan: I think you're right that it makes sense to trust the timestamp in the report.

Can you expand on why evaluating `task_expiration` could leak information about the report? Which actor (the client, the aggregator, a network observer) could learn something about the true time a measurement was taken based on what observable behavior of which actor?

tgeoghegan (commented):
I think the text in its current iteration accomplishes what it should, which is that clients may upload reports right up until `task_expiration`, and collectors can initiate the collect flow _after_ `task_expiration`. I left a few more comments about word choices and some typos but I think this is OK to go forward.

draft-ietf-ppm-dap.md
wangshan: @tgeoghegan for e.g. if the precision is in 1hr, and 
1. one client has rounded it's true timestamp to 14:00:00
2. the aggregators know the `task_expiration` is effectively 14:00:01 UTC
3. the aggregators know clients will check their timestamp < `task_expiration` before sending.

Then the aggregator could know that the client has a true timestamp between 14:00:00 and 14:00:01.

draft-ietf-ppm-dap.md
```diff
@@ -1712,17 +1717,23 @@ designed to allow implementations to reduce operational costs in certain cases.
...
+protocol is only run once per report, each aggregator only needs to store its
 aggregate share for each possible batch interval, along with the number of times
 the aggregate share was used in a batch. This is due to the requirement that the
 batch interval respect the boundaries defined by the DAP parameters. (See
 {{batch-validation}}.)
 
+Furthermore, the aggregators must store data related to a task as long as the
+current time has not passed this task's `task_expiration`. Aggregator MAY delete
+the task and all data pertaining to this task after `task_expiration`.
+Implementors SHOULD provide for some leeway so collector can collect the batch
```
cjpatton, line 1734:
```suggestion
Implementors SHOULD provide for some leeway so the collector can collect the batch
```

wangshan:
> Suppose the Helper receives an aggregation job for a task that has not yet "expired" but contains at least one report whose timestamp is passed the expiration date. What should it do?
> 
> I think it would be appropriate to add a `ReportShareError` variant for this case.

good point, let me add that

tgeoghegan (approved):

draft-ietf-ppm-dap.md
```diff
@@ -621,6 +621,9 @@ number generator. Each task has the following parameters associated with it:
   task must have.
 * `max_batch_lifetime`: The maximum number of times a batch of reports may be
   queried by the Collector.
+* `task_expiration`: The time up to which clients are allowed to upload to this
```
cjpatton, line 624:
```suggestion
* `task_expiration`: The time up to which clients are expected to upload to this
```

cjpatton (approved):
A couple editorial nits, otherwise I'm good landing this.

cjpatton:
@wangshan please squash this branch and fix the build issue so that it's ready to merge :)

wangshan:
@cjpatton squashed and ready to merge

draft-ietf-ppm-dap.md
cjpatton: Can we resolve this convo?

draft-ietf-ppm-dap.md
```diff
@@ -779,16 +782,21 @@ retrying.
...
 collected. (See {{input-share-batch-validation}} for details.) Otherwise,
 comparing the aggregate result to the previous aggregate result may result in a
-privacy violation. (Note that the Helpers enforce this as well.) In addition,
-the leader SHOULD abort the upload protocol and alert the client with error
-"reportTooLate".
+privacy violation. (Note that the Helpers enforce this as well.) The Leader MAY
+ignore any reports pertaining to a task that have timestamps pass
+`task_expiration`. In addition, the leader SHOULD abort the upload protocol
+and alert the client with error "reportTooLate". Client MAY choose to opt out
+of the task if its own clock has passed `task_expiration`
```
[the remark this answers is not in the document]
cjpatton: Can we resolve this?

draft-ietf-ppm-dap.md
tgeoghegan: I see what you mean @wangshan. I think requiring that `task_expiration` be aligned with `time_precision` would help. Would you please file an issue about that so we can deal with it in another PR?

wangshan: will do @tgeoghegan, let me first resolve this conversation and rebase this PR for merging

chris-wood (approved):

Issue #293, opened 2022-07-28
Revisit requirements and recommendations for request and response authentication


tgeoghegan:
At IETF 114, we discussed how [draft-01](https://datatracker.ietf.org/doc/draft-ietf-ppm-dap/01/) contains inconsistent guidance and/or requirements on authentication. Some details can be found in [this slide deck](https://datatracker.ietf.org/meeting/114/materials/slides-114-ppm-whats-new-for-dap/), especially slide 7.

DAP should figure out what the security requirements are for each of the channels between the protocol participants, and then decide what if any prescriptions to make.

tgeoghegan:
In my presentation at IETF 114, I argued that DAP should say as little as possible about authentication, in order to allow organizations deploying DAP to use whatever authn mechanism they already support for their HTTP APIs (e.g., maybe a vendor already operates a bunch of other APIs using OAuth 2 and so using that for DAP makes it easy for them to fit it into their existing permissions model).

@ekr suggested that we should reach out to the HTTP WG for guidance here (which might prompt the HTTP WG to develop some guidance here).

One clear precedent to look at here is ACME, which requires the use of JWT for authentication of requests. We should talk to ACME operators in the wild (like Let's Encrypt) to see what they think.

PR #305, opened 2022-08-24
Revisit request authentication

tgeoghegan:
Remove the requirement that implementations use `DAP-Auth-Token` with a
fixed, pre-negotiated secret and instead allow any client auth scheme
that is composable with HTTP (which includes `DAP-Auth-Token`).

Relevant to #293

### DIFF
--- a/draft-ietf-ppm-dap.md
+++ b/draft-ietf-ppm-dap.md
@@ -368,17 +368,18 @@ DAP is made up of several sub-protocols in which different subsets of the
 protocol's participants interact with each other.
 
 In those cases where a channel between two participants is tunneled through
-another protocol participant, DAP mandates the use of authenticated encryption
+another protocol participant, DAP mandates the use of public-key encryption
 using {{!HPKE=RFC9180}} to ensure that only the intended recipient can see a
 message in the clear.
 
 In other cases, DAP requires HTTPS client authentication. Any authentication
 scheme that is composable with HTTP is allowed. For example, {{!OAuth2=RFC6749}}
 credentials are presented in an Authorization HTTP header, which can be added to
-any DAP protocol message. This allows organizations deploying DAP to use
-existing well-known HTTP authentication mechanisms that they already support.
-Discovering what authentication mechanisms are supported by a DAP participant
-is outside of this document's scope.
+any DAP protocol message, or TLS client certificates are another viable
+solution. This allows organizations deploying DAP to use existing well-known
+HTTP authentication mechanisms that they already support. Discovering what
+authentication mechanisms are supported by a DAP participant is outside of this
+document's scope.
 
 ## Errors
 
--- a/draft-ietf-ppm-dap.md
+++ b/draft-ietf-ppm-dap.md
@@ -358,10 +358,112 @@ client could not report 10^6s or -20s.
 
 # Message Transport
 
-Communications between DAP entities are carried over HTTPS {{!RFC2818}}. HTTPS
-provides server authentication and confidentiality. When client authenticaiton
-is also required, the client uses the mechanism described in
-{{https-sender-auth}}.
+Communications between DAP participants are carried over HTTPS {{!RFC2818}}.
+HTTPS provides server authentication and confidentiality. Use of HTTPS is
+REQUIRED.
+
+## Request Authentication {#request-authentication}
+
+DAP is made up of several sub-protocols in which different subsets of the
+protocol's participants interact with each other. In this section we enumerate
+those interactions and their security requirements.
+
+In those cases where a channel between two participants is tunneled through
+another protocol participant, DAP mandates the use of authenticated encryption
+using {{!HPKE=RFC9180}} to ensure that only the intended recipient can see a
+message in the clear.
+
+In other cases, DAP requires client authentication. Any authentication scheme
+that is composable with HTTP is allowed. For example, {{!OAuth2=RFC6749}}
+credentials are presented in an Authorization HTTP header, which can be added to
+any DAP protocol message. This allows organizations deploying DAP to use
+existing well-known HTTP authentication mechanisms that they already supports.
+Discovering what authentication mechanisms are supported by a DAP participant
+is outside of this document's scope.
+
+### Client to aggregator
+
+Clients upload reports to aggregators, as detailed in {{upload-flow}}. The
+contents of each input share must be kept confidential from everyone but the
+client and the aggregator it is being sent to and clients must be able to
+authenticate the aggregator they upload to.
+
+The HTTPS channel provides confidentiality between the client and the leader,
+but this is not sufficient since the helper's report shares are relayed through
+the leader. Confidentiality of report shares is achieved by encrypting each
+report share to a public key held by the respective aggregator. Clients fetch
+the public keys from each aggregator over HTTPS, allowing them to authenticate
+the server.
+
+Aggregators MAY require clients to authenticate when uploading reports. This is
+an effective mitigation against Sybil attacks in deployments where it is
+practical for each client to have an identity provisioned (e.g., a user logged
+into an online service or a hardware device programmed with an identity). If it
+is used, client authentication MUST use a scheme that meets the requirements in
+{{request-authentication}}.
+
+In some deployments, it will not be practical to require clients to authenticate
+(e.g., a widely distributed application that does not require its users to login
+to any service), so client authentication is not mandatory in DAP. In the
+absence of client authentication, implementations SHOULD deploy some mitigation
+against Sybil attacks.
+
+### Inter-aggregator
+
+The leader and helper aggregators communicate to prepare input shares, as
+detailed in {{aggregate-flow}}, and to aggregate output shares into aggregate
+shares, as detailed in {{collect-aggregate}}. These messages must be
+confidential and mutually authenticated.
+
+The aggregate sub-protocol is driven by the leader acting as an HTTPS client,
+making requests to the helper's HTTPS server. HTTPS provides confidentiality and
+authenticates the helper to the leader.
+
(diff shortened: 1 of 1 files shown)


draft-ietf-ppm-dap.md
```diff
@@ -687,8 +761,6 @@ cipher suites (a la OHTTP/ECH).]
 The client MUST abort if any of the following happen for any HPKE config
 request:
 
-* the client and aggregator failed to establish a secure,
-  aggregator-authenticated channel;
```
tgeoghegan, line 691: I deleted this because it's implicit by virtue of using HTTPS: you won't even be able to establish a connection if you can't verify server identity, so we don't need more protocol/application layer language to repeat features offered by the transport.

draft-ietf-ppm-dap.md
```diff
@@ -358,10 +358,112 @@ client could not report 10^6s or -20s.
 
 # Message Transport
 
-Communications between DAP entities are carried over HTTPS {{!RFC2818}}. HTTPS
-provides server authentication and confidentiality. When client authenticaiton
-is also required, the client uses the mechanism described in
-{{https-sender-auth}}.
+Communications between DAP participants are carried over HTTPS {{!RFC2818}}.
+HTTPS provides server authentication and confidentiality. Use of HTTPS is
+REQUIRED.
```
tgeoghegan, line 363: I think there's good arguments against relying on a transport-level concept of identity to authenticate protocol-level concepts, but [this is good enough for ACME](https://www.rfc-editor.org/rfc/rfc8555.html#section-6) so perhaps we can lean on that precedent.

cjpatton: Ultimately this question would be answered by a proof of (in)security in a formal model. I'm perfectly happy to make this a requirement for now.

One nit: Does RFC2818 cover HTTP/3? I'm wondering if @martinthomson can help us spell this requirement :)

draft-ietf-ppm-dap.md
```diff
@@ -358,10 +358,27 @@ client could not report 10^6s or -20s.
...
+
+## Request Authentication {#request-authentication}
```
cjpatton, line 365: ?

```suggestion
## HTTP Request Authentication {#request-authentication}
```

draft-ietf-ppm-dap.md
```diff
@@ -358,10 +358,27 @@ client could not report 10^6s or -20s.
...
+
+DAP is made up of several sub-protocols in which different subsets of the
+protocol's participants interact with each other.
+
+In those cases where a channel between two participants is tunneled through
+another protocol participant, DAP mandates the use of authenticated encryption
+using {{!HPKE=RFC9180}} to ensure that only the intended recipient can see a
+message in the clear.
+
+In other cases, DAP requires client authentication. Any authentication scheme
```
cjpatton, line 375: We should try to avoid overloading "Client" wherever possible.

```suggestion
In other cases, DAP requires sender authentication. Any authentication scheme
```

draft-ietf-ppm-dap.md
```diff
@@ -813,6 +800,31 @@ encoded value of the following form:
 "extension_type" indicates the type of extension, and "extension_data" contains
 information specific to the extension.
 
+### Upload Message Security
+
+The contents of each input share must be kept confidential from everyone but the
+client and the aggregator it is being sent to and clients must be able to
```
cjpatton, line 806:
```suggestion
client and the aggregator it is being sent to. In addition, clients must be able to
```

draft-ietf-ppm-dap.md
```diff
@@ -813,6 +800,31 @@ encoded value of the following form:
...
+practical for each client to have an identity provisioned (e.g., a user logged
+into an online service or a hardware device programmed with an identity). If it
+is used, client authentication MUST use a scheme that meets the requirements in
+{{request-authentication}}.
+
+In some deployments, it will not be practical to require clients to authenticate
+(e.g., a widely distributed application that does not require its users to login
+to any service), so client authentication is not mandatory in DAP. In the
+absence of client authentication, implementations SHOULD deploy some mitigation
+against Sybil attacks.
```
cjpatton, line 826: As it stands, this problem is unsolved. I don't think it's fair to nudge implementors until we have a better sense of what alternatives are possible. It might be worth pointing to an issue regarding Sybil attacks (we have at least one somewhere).

```suggestion
to any service), so client authentication is not mandatory in DAP.
```

draft-ietf-ppm-dap.md
```diff
@@ -1499,6 +1518,45 @@ its query:
...
+leader to:
+
+* Send collect parameters to the helper that do not reflect the parameters
+  chosen by the collector
+* Discard the aggregate share computed by the helper and then fabricate
+  aggregate shares that combine into an arbitrary aggregate result
+
+These are both attacks on soundness, which we already assume to hold only if
+both aggregators are honest, which puts these malicious-leader attacks out of
+scope.
```
cjpatton, line 1553: Point to security considerations here.

draft-ietf-ppm-dap.md
```diff
@@ -1499,6 +1518,45 @@ its query:
...
+
+[[OPEN ISSUE: Should we have authentication in either direction between the
+helper and the collector? #155]]
+[[OPEN ISSUE: DAP and/or VDAF should spell out what "soundness" and "privacy"
+mean]]
```
cjpatton, line 1558: VDAF does already. Does DAP not as well?

draft-ietf-ppm-dap.md
```diff
@@ -358,10 +358,27 @@ client could not report 10^6s or -20s.
...
+In those cases where a channel between two participants is tunneled through
+another protocol participant, DAP mandates the use of authenticated encryption
+using {{!HPKE=RFC9180}} to ensure that only the intended recipient can see a
+message in the clear.
+
+In other cases, DAP requires client authentication. Any authentication scheme
+that is composable with HTTP is allowed. For example, {{!OAuth2=RFC6749}}
+credentials are presented in an Authorization HTTP header, which can be added to
+any DAP protocol message. This allows organizations deploying DAP to use
+existing well-known HTTP authentication mechanisms that they already supports.
```
branlwyd, line 379:
```suggestion
existing well-known HTTP authentication mechanisms that they already support.
```

(typo)

draft-ietf-ppm-dap.md
martinthomson: You can start to refer to RFC 9110 for that stuff now.  That definitely covers HTTP/3.

(Technically, RFC2818 did cover HTTP/3, but only circuitously.  Also, RFC 2818 is a little old in a few important ways.)

draft-ietf-ppm-dap.md
tgeoghegan: Oh, whoops, I was searching for "soundness" in VDAF, but it discusses "robustness". And in security considerations we discuss "privacy" and "correctness", but the doc has references to "soundness" and "robustness", both in protocol text and the odd OPEN ISSUE block. I filed #306 to clean up this language in DAP and align it with VDAF.

draft-ietf-ppm-dap.md
tgeoghegan: I put in a reference to {{sec-considerations}} and use the word "correctness", which is currently the term that Security Considerations uses. That's wrong, but it'll get cleaned up in #306.

It also occurs to me that maybe everything in these "{Upload, Aggregate, Collect} Message Security" sections that isn't a MUST or a MAY should be punted to security considerations, since it's discussion, context and justification rather than prescription.

cjpatton: I think it's OK to point to security considerations as needed in the normative text, especially when it clarifies certain design choices to the reader. That said, mnimizing where possible is a good idea. In particular I'm not sure I would call these sections ".* Message Security".

draft-ietf-ppm-dap.md
```diff
@@ -358,10 +358,27 @@ client could not report 10^6s or -20s.
...
+HTTPS provides server authentication and confidentiality. Use of HTTPS is
+REQUIRED.
+
+## HTTPS Request Authentication {#request-authentication}
+
+DAP is made up of several sub-protocols in which different subsets of the
+protocol's participants interact with each other.
+
+In those cases where a channel between two participants is tunneled through
+another protocol participant, DAP mandates the use of authenticated encryption
```
cjpatton, line 371: Bump: We're not using HPKE for authenticated encryption here. The properties we need are confidentiality and non-malleability.
```suggestion
another protocol participant, DAP mandates the use of public-key encryption
```

draft-ietf-ppm-dap.md
```diff
@@ -358,10 +358,27 @@ client could not report 10^6s or -20s.
...
+using {{!HPKE=RFC9180}} to ensure that only the intended recipient can see a
+message in the clear.
+
+In other cases, DAP requires HTTPS client authentication. Any authentication
+scheme that is composable with HTTP is allowed. For example, {{!OAuth2=RFC6749}}
+credentials are presented in an Authorization HTTP header, which can be added to
+any DAP protocol message. This allows organizations deploying DAP to use
+existing well-known HTTP authentication mechanisms that they already support.
+Discovering what authentication mechanisms are supported by a DAP participant
+is outside of this document's scope.
```
cjpatton, line 381: It's worth mentioning mutual TLS here as a viable option.

cjpatton (approved):
I think it would be useful to minimizing security considerations language in the Message Security sections, but I don't have a clear idea how. I think this is good enough for now, hence the approval, but let's see if we can get more feedback before merging.

branlwyd (approved):

tgeoghegan:
I had to do a rebase to fix a trivial conflict with #303. These changes were approved about a week ago and there's been no further comment, so I am merging.

PR #307, opened 2022-08-25
Move draft-irtf-cfrg-vdaf-01 to 03

### LINKED ISSUE #296: Align with latest VDAF draft
This will be VDAF-02 or 03, depending on when we need to cut DAP-02.

cjpatton:
Closes #296.

Align with the latest VDAF draft. This requires two minor protocol
changes:

* 02 added a value called the "public_share" to the VDAF syntax. This is
  to support Poplar1 (and likely other schemes that need a similar
  extractability property). To accommodate this, the public_share field
  has been added to the Report and ReportShare structures and appended to
  the AAD for input share encryption. (This may end up being useful for
  privacy.)

* 02 added the report count to the aggregate result computation. This is
  already passed to the Collector in the CollectResp, so we just need to
  update the aggregate result computation in the draft to match.


divergentdave (approved):

draft-ietf-ppm-dap.md
```diff
@@ -725,11 +725,13 @@ struct {
 struct {
   TaskID task_id;
   ReportMetadata metadata;
+  opaque public_share<0..2^32-1>;
```
tgeoghegan, line 728: AFAICT, `public_share` always accompanies `ReportMetadata`, so it'd be nice to tuck the former into the latter, because then the public share would get included in message AAD and appear in all the places it needs to "for free". My only reservation is that calling the public share "metadata" feels like a misnomer.

tgeoghegan (approved):

draft-ietf-ppm-dap.md
```diff
@@ -146,7 +146,7 @@ Input:
 
 Input share:
 : An aggregator's share of the output of the VDAF
-   {{?VDAF=I-D.draft-irtf-cfrg-vdaf-01}} sharding algorithm. This algorithm is run by
+   {{?VDAF=I-D.draft-irtf-cfrg-vdaf-03}} sharding algorithm. This algorithm is run by
```
tgeoghegan, line 149: While we're in here touching these lines: I think VDAF qualifies as a normative reference rather than an informative one based on [the IETF's criteria](https://www.ietf.org/about/groups/iesg/statements/normative-informative-references/). This and all the other VDAF references should be updated to use `!`.
```suggestion
   {{!VDAF=I-D.draft-irtf-cfrg-vdaf-03}} sharding algorithm. This algorithm is run by
```

tgeoghegan (changes requested):

draft-ietf-ppm-dap.md
cjpatton: Yes, you're right. We could rename to "header" maybe, but @chris-wood rejected this name in an earlier PR. Maybe it makes sense to rename if we include public share?

draft-ietf-ppm-dap.md
cjpatton: Done!

tgeoghegan (approved):

cjpatton:
Rebased and pushed a minor update: The public share in Report and ReportStore now both have the same length prefix.

cjpatton:
Rebased.

Issue #301, opened 2022-08-18
Consider making batch IDs predictable


cjpatton:
For fixed-size tasks, the Leader picks a batch ID and passes it to the Collector with the CollectResp. This is so that the batch can be queried later on with a different aggregation parameter (as required for Poplar1). Alternatively, we could arrange for batch IDs to be "predictable" so that the Collector doesn't have to store batch IDs.

_Originally posted by @chris-wood in https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/issues/297#issuecomment-1219697601_

branlwyd:
With the current `next-batch` semantics, there is the possibility that a Collector crashing at an inopportune time might lead to a batch being lost permanently.

If batch IDs were predictable, this would be fixable in a fairly straightforward manner (from a [PR comment](https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/pull/297#discussion_r948460390)):

> Also, if it's acceptable to allow batch IDs to be enumerable (e.g. a counter, as suggested by others), I think a simpler idempotent design would be to have the Collector track which batch ID is next itself, and change next-batch to something like max-batch to allow Collectors to "start up" without knowing what the next batch is.

cjpatton:
Here would be my suggested change for this: 
* Change type of `BatchID` from `uint8[32]` to `uint64`. Maybe call it `BatchNum`, since "ID" is always a 32-byte string elsewhere.
* Have the Collector specify the `BatchID` in the collect request.
* Allow the Leader to reject `BatchID`s that are too far into the future.

Anything else to consider?

I'll aim to have a PR up by end of week.

branlwyd:
You may wish to change `next-batch` to something like `current-batch` which causes the Leader to return the batch ID that is "current." (Defined as something like "the next batch that will be completed.") This would allow a new Collector to start up with recent data, without needing out-of-band knowledge of the current batch ID.

OTOH, I'm not sure if this `current-batch` functionality is necessary for any currently-planned deployment's needs; it's also possible to say that a Collector should simply start at the first Batch ID & iterate from there to find the latest Batch ID. In that case, the `next-batch` functionality could be dropped.

cjpatton:
Partly solved by #304, we'll continue on in #342, which is more specific.

PR #308, opened 2022-08-26
Fixed-size tasks: remove `next-batch` functionality.

branlwyd:
This PR does not specify how the Collector knows what batch IDs to query for.
This will be addressed in a follow-on PR.

### DIFF
--- a/draft-ietf-ppm-dap.md
+++ b/draft-ietf-ppm-dap.md
@@ -321,7 +321,6 @@ measurement for the collector. Depending on the VDAF, it may be possible to
 incrementally process each report as it comes in, or may be necessary to wait
 until the entire batch of reports is received.
 
-
 ## Validating Inputs {#validating-inputs}
 
 An essential task of any data collection pipeline is ensuring that the data
@@ -357,7 +356,6 @@ instance, the user might have spent 30s on a task but the client might report
 60s. This is a problem with any measurement system and DAP does not attempt to
 address it; it merely ensures that the data is within acceptable limits, so the
 client could not report 10^6s or -20s.
-
 
 # Message Transport
 
@@ -506,28 +504,14 @@ A query includes parameters used by the Aggregators to select a batch of reports
 specific to the given query type. A query is defined as follows:
 
 ~~~
-enum {
-    next-batch(0),
-    by-batch-id(1),
-    (255)
-} FixedSizeQueryType;
-
-opaque BatchId[32];
-
-struct {
-    FixedSizeQueryType fixed_size_query_type;
-    select (FixedSizeQuery.fixed_size_query_type) {
-        case next-batch:  Empty;
-        case by-batch-id: BatchId batch_id;
-    };
-} FixedSizeQuery;
+opaque BatchID[32];
 
 struct {
     QueryType query_type;
     select (Query.query_type) {
         case time-interval: Interval batch_interval;
-        case fixed-size: FixedSizeQuery fixed_size_query;
-    };
+        case fixed-size: BatchID batch_id;
+    }
 } Query;
 ~~~
 
@@ -578,28 +562,31 @@ particularly important for controlling the amount of noise added to reports
 by Clients (or added to aggregate shares by Aggregators) in order to achieve
 differential privacy.
 
-To get the aggregate of the next batch, the Collector issues a query of type
-`next-batch`. (See FixedSizeQuery in {{query}}.) The response to this query (see
-{{collect-flow}}) includes a "batch ID", which can be used to query the same
-batch later on. This is done using the `by-batch-id` batch-query type.
-
 For this query type, the Aggregators group reports into arbitrary batches such
-that each batch has roughly the same number of reports. In addition to the
-minumum batch size common to all query types, the conifugration includes a
-"maximum batch size", `max_batch_size`, that determines maximum number of
-reports per batch.
+that each batch has roughly the same number of reports. These batches are
+identified by opaque "batch IDs", allocated in an arbitrary fashion by the
+Leader. To get the aggregate of a batch, the Collector issues a query
+specifying the batch ID of interest (see {{query}}).
+
+In addition to the minimum batch size common to all query types, the
+configuration includes a "maximum batch size", `max_batch_size`, that
+determines maximum number of reports per batch.
 
 Implementation note: The goal for the Aggregators is to aggregate precisely
 `min_batch_size` reports per batch. Doing so, however, may be challenging
 for Leader deployments in which multiple, independent nodes running the
 aggregate sub-protocol (see {{aggregate-flow}}) need to be coordinated. The
-maximum batch size is intended to allow room for error. Typically the difference
-between the minimum and maximum batch size will be a small fraction of the
-target batch size for each batch.
+maximum batch size is intended to allow room for error. Typically the
+difference between the minimum and maximum batch size will be a small fraction
+of the target batch size for each batch.
 
 [OPEN ISSUE: It may be feasible to require a fixed batch size, i.e.,
 `min_batch_size == max_batch_size`. We should know better once we've had some
 implementation/deployment experience.]
+
+[OPEN ISSUE: It may be desirable to allow Collectors to query for a current/
+recent batch ID. How important this is will be determined by deployment
+experience.]
 
 ## Task Configuration {#task-configuration}
 
@@ -988,7 +975,7 @@ struct {
   QueryType query_type;
   select (AggregateInitializeReq.query_type) {
     case time-interval: Empty;
-    case fixed-size: BatchId batch_id;
+    case fixed-size: BatchID batch_id;
   };
   ReportShare report_shares<1..2^32-1>;
 } AggregateInitializeReq;
@@ -1382,7 +1369,7 @@ struct {
   QueryType query_type;
   select (CollectResp.query_type) {
     case time-interval: Empty;
-    case fixed-size: BatchId batch_id;
+    case fixed-size: BatchID batch_id;
   };
   uint64 report_count;
   HpkeCiphertext encrypted_agg_shares<1..2^32-1>;
@@ -1435,7 +1422,7 @@ struct {
   QueryType query_type;
   select (BatchSelector.query_type) {
     case time-interval: Interval batch_interval;
-    case fixed-size: BatchId batch_id;
+    case fixed-size: BatchID batch_id;
   };
 } BatchSelector;
 
@@ -1452,7 +1439,7 @@ The message contains the following parameters:
 
 * The task ID.
 
-* The "batch seletor", which encodes parameters used to determine the batch
+* The "batch selector", which encodes parameters used to determine the batch
   being aggregated. The value depends on the query type for the task:
 
     * For time-interval tasks, the request specifies the batch interval.
@@ -1671,18 +1658,12 @@ reports in the batch.
 
 ##### Boundary Check
 
-For fixed-size tasks, the batch boundaries are defined by "batch IDs" selected
-by the Leader. Thus the Aggregator needs to check that the query is associated
-with a known batch ID:
+For fixed-size tasks, the batch boundaries are defined by opaque batch IDs.
+Thus the Aggregator needs to check that the query is associated with a known
+batch ID:
 
-* If the query contained in the CollectReq has type `by-batch-id`, the
-  Leader checks that the provided batch ID corresponds to a batch ID it
-  returned in a previous CollectResp for the task.
-
-  [OPEN ISSUE: Consider making batch IDs predictable. See issue #301.]
-
-* The Helper checks that the batch ID provided by the leader in its
-  AggregateShareReq corresponds to a batch ID used in a previous
+* For an AggregateShareReq, the Helper checks that the batch ID provided by the
+  Leader in its corresponds to a batch ID used in a previous
   AggregateInitializeReq for the task.
 
 ##### Size Check


draft-ietf-ppm-dap.md
```diff
@@ -575,24 +573,29 @@ particularly important for controlling the amount of noise added to reports
...
-"maximum batch size", `max_batch_size`, that determines maximum number of
-reports per batch.
+that each batch has roughly the same number of reports. These batches are
+identified by "batch IDs", which are numerical identifiers allocated
+consecutively starting from 0.
+
+To get the aggregate of a batch, the Collector issues a query of type
+`by-batch-id`, specifying the batch ID of interest in the query. (See
+`FixedSizeQuery` in {{query}}.) The Collector may not know which batch ID it is
+interested in (for example, in startup scenarios); in this case, it can also
```
branlwyd, line 584: The `current-batch` functionality may not be ideal:

* It assumes that a Collector might want to start up against a task that has already collected some batches, and will want to start from "recent" data. If this is not the case (for example, if we assume that a Collector can always safely start from batch ID 0, or are OK saying that communicating a "recent" batch ID can be done out-of-band), we could drop this functionality entirely.
* Even if we keep this functionality, it would be nicer to separate "give me a recent batch ID" from "start collecting a batch ID". One way to do this would be to expose an endpoint that forwards to a recent batch ID; this would be easier to specify with #278, at which point the batch ID might be specified as part of the URL.

draft-ietf-ppm-dap.md
```diff
@@ -506,17 +504,17 @@ given query type. A query is defined as follows:
...
 ~~~
 enum {
-    next-batch(0),
-    by-batch-id(1),
+    by-batch-id(0),
+    current-batch(1),
 } FixedSizeQueryType;
 
-opaque BatchId[32];
+uint64 BatchId;
```
cjpatton, line 511: Since other "IDs" have type `opaque [32]`, I wonder if it makes sense to give this a different name. What about `BatchNum`?

draft-ietf-ppm-dap.md
```diff
@@ -1556,19 +1559,17 @@ reports in the batch.
...
+"batch IDs". Thus the Aggregator needs to check that the query is associated
 with a known batch ID:
 
-* If the query contained in the CollectReq has type `by-batch-id`, the
-  Leader checks that the provided batch ID corresponds to a batch ID it
-  returned in a previous CollectResp for the task.
-
-  [OPEN ISSUE: Consider making batch IDs predictable. See issue #301.]
+* If the query contained in the CollectReq has type `by-batch-id`, the Leader
+  MAY fail if the batch ID is considered out of range (too small or too large).
```
cjpatton, line 1567:
```suggestion
  MAY abort with error "XXX" if the batch ID is considered out of range (too small or too large).
```

draft-ietf-ppm-dap.md
```diff
@@ -1556,19 +1559,17 @@ reports in the batch.
...
 
-* The Helper checks that the batch ID provided by the leader in its
+* The Helper checks that the batch ID provided by the Leader in its
   AggregateShareReq corresponds to a batch ID used in a previous
-  AggregateInitializeReq for the task.
+  AggregateInitializeReq for the task. The Helper MAY fail if the batch ID is
```
cjpatton, line 1571:
```suggestion
  AggregateInitializeReq for the task. The Helper MAY abort with error "XXX" if the batch ID is
```

draft-ietf-ppm-dap.md
cjpatton: Yeah, this is a bit hairy. The protocol seems like it would be a lot simpler if all the Collector could do is specify which batch ID it wants.

Do you thin this functionality is needed right now? If not, maybe better to punt on and keep DAP-02 simple.

branlwyd: I think an eventual "final"/"production-quality" design would definitely want some provision to allow Collectors to discover a recent batch ID from the Aggregators.

However, you're probably right that this isn't needed for initial testing: I suspect interop-test-level deployments can get away with the Collector always starting at batch ID 0, and if that doesn't work out, out-of-band communications would allow configuration of the Collector to start from the correct batch ID.

I went ahead and tore out the `current-batch` functionality for now.

draft-ietf-ppm-dap.md
branlwyd: I like calling it an "ID" because that is an accurate description of this type's semantics -- it's a primary identifier for a batch. The difference in types doesn't bother me so much; IMO, it's fairly natural for IDs for different kinds of entities to have different types, depending on the needs for the ID.

OTOH, we might rename some messages such that a "batch ID" is either an interval (for `time-interval` tasks) or an opaque numeric value (for `fixed-size` tasks), i.e.:

```
struct BatchID {
  select (query_type) {
    case time-interval: Interval batch_interval;
    case fixed-size: BatchID batch_id;
  }
}
```

This would be "cleaner" (IMO) -- this new BatchID struct would effectively replace the Query -- but I'm not sure it's worth taking the time to do right now.

draft-ietf-ppm-dap.md
branlwyd: In the Batch Validation section, these boundary conditions are specified as causing a failure of type "batchInvalid". I reworded this from "failing" to "checking", since this section should be describing a set of checks.

draft-ietf-ppm-dap.md
branlwyd: Same as above: in the Batch Validation section, these boundary conditions are specified as causing a failure of type "batchInvalid". I reworded this from "failing" to "checking", since this section should be describing a set of checks.

draft-ietf-ppm-dap.md
cjpatton: Good point about my suggestion. The second idea I like quite a lot as well, though I think we'd have to rename the inner "batch_id" :) If the change is not overly invasive, I say go for it and let's see what it looks like! Just be ready to revert if needed.

draft-ietf-ppm-dap.md
cjpatton: Yup, you're right. I was a bit careless there. Text looks good after revision.

draft-ietf-ppm-dap.md
cjpatton: Sounds good. Most likely we'll be able to spell this feature when we actually need it.

draft-ietf-ppm-dap.md
cjpatton: Actually, thinking about it more, it sounds like what you want is to rename "Query" to "BatchID"?

cjpatton (approved):
LGTM. One potential downside is that the Leader can no longer choose batch IDs at random, since it needs to fill batch IDs consecutively. I wonder if this could potentially create a processing bottleneck for the Leader.

draft-ietf-ppm-dap.md
```diff
@@ -575,29 +560,36 @@ particularly important for controlling the amount of noise added to reports
...
+that each batch has roughly the same number of reports. These batches are
+identified by "batch IDs", which are numerical identifiers allocated
+consecutively starting from 0.
+
+To get the aggregate of a batch, the Collector issues a query specifying the
+batch ID of interest (see {{query}}). The Collector will typically request each
+batch in turn in consecutive order; it is expected that the batch ID to start
+from will be configured out-of-band from the DAP protocol.
+
+In addition to the minumum batch size common to all query types, the
```
tgeoghegan, line 573:
```suggestion
In addition to the minimum batch size common to all query types, the
```

draft-ietf-ppm-dap.md
```diff
@@ -575,29 +560,36 @@ particularly important for controlling the amount of noise added to reports
...
-batch later on. This is done using the `by-batch-id` batch-query type.
-
 For this query type, the Aggregators group reports into arbitrary batches such
-that each batch has roughly the same number of reports. In addition to the
-minumum batch size common to all query types, the conifugration includes a
-"maximum batch size", `max_batch_size`, that determines maximum number of
-reports per batch.
+that each batch has roughly the same number of reports. These batches are
+identified by "batch IDs", which are numerical identifiers allocated
+consecutively starting from 0.
```
tgeoghegan, line 566: Should we say "monotonic" to allow discontinuities in batch IDs? Does anything bad happen if there are discontinuities in batch IDs? I assume we already have text instructing aggregators what to do if they don't recognize a batch ID.

tgeoghegan (approved):

draft-ietf-ppm-dap.md
branlwyd: I think we don't want discontinuities: the idea of this approach is that the Collector can send requests for batch 0, followed by batch 1, followed by batch 2, and so on -- i.e. that the Collector can predict the sequence of batch identifiers. While a failed aggregation or technical issues might lead to an uncollectable batch, which (hopefully) is communicated well enough for the Collector to realize it won't be able to collect that batch, allowing the Leader to skip arbitrary batch IDs would currently be painful[1]. For example, if the Leader decided to skip `10^6` batch IDs, I suppose the Collector would have to walk through `10^6` collect requests getting an "unknown batch" before finding the next "good" batch.


[1] I think the best way to reduce this pain, if we decide we need to, would be to have some way for the Collector to ask the Leader which batch IDs are usable/acceptable. That's what I was trying to get at with the now-removed `current-batch` functionality, to allow Collectors to startup against Aggregators that have been running for awhile without the need for out-of-band communication. Like with that functionality, IMO we should see whether we need to implement something more complicated before we specify it.

branlwyd: > LGTM. One potential downside is that the Leader can no longer choose batch IDs at random, since it needs to fill batch IDs consecutively. I wonder if this could potentially create a processing bottleneck for the Leader. (I don't know the answer so I'm hoping other reviewers will take the time to think this through!)

I'm not _too_ worried about this over-and-above the current design: I think there is already effectively a "serialization point" required when the Leader assigns the next aggregation job to a batch -- at that point, it needs to know how many outstanding aggregation jobs are assigned to batches, and how many reports are in those aggregation jobs, in order to decide the batch ID to use for its next aggregation job. I think it needs to know this since, if this knowledge is unavailable or imprecise, it might assign too many reports to a single batch, leading to those reports being uncollectable.

Once you have a serialization point at point-of-generation-of-aggregation-job, choosing a new batch ID would be something like the pseudo-SQL `SELECT MAX(batch_id) + 1 FROM batches WHERE batch_full = false`.

That said -- maybe the existing serialization point will be too much of a performance hit, with or without generating a consecutive batch ID? I guess implementation experience will tell. I'm not too worried since this serialization point wouldn't affect the aggregation protocol, which is likely to be the costliest part of the overall protocol, but it's hard to be certain without actually implementing.

(also, did I miss something with the above? maybe there's some kinds of imprecision in knowledge of assignment of client reports/aggregation jobs to batches that would allow a serialization point to be safely skipped?)

branlwyd:
> LGTM. One potential downside is that the Leader can no longer choose batch IDs at random, since it needs to fill batch IDs consecutively. I wonder if this could potentially create a processing bottleneck for the Leader. (I don't know the answer so I'm hoping other reviewers will take the time to think this through!)

I'm not _too_ worried about this over-and-above the current design: I think there is already effectively a "serialization point" required when the Leader assigns the next aggregation job to a batch -- at that point, it needs to know how many outstanding aggregation jobs are assigned to batches, and how many reports are in those aggregation jobs, in order to decide the batch ID to use for its next aggregation job. I think it needs to know this since, if this knowledge is unavailable or imprecise, it might assign too many reports to a single batch, leading to those reports being uncollectable.

Once you have a serialization point at point-of-generation-of-aggregation-job, choosing a new batch ID would be something like the pseudo-SQL `SELECT MAX(batch_id) + 1 FROM batches WHERE batch_full = false`.

That said -- maybe the existing serialization point will be too much of a performance hit, with or without generating a consecutive batch ID? I guess implementation experience will tell. I'm not too worried since this serialization point wouldn't affect the aggregation protocol, which is likely to be the costliest part of the overall protocol, but it's hard to be certain without actually implementing.

(also, did I miss something with the above? maybe there's some kinds of imprecision in knowledge of assignment of client reports/aggregation jobs to batches that would allow a serialization point to be safely skipped?)

draft-ietf-ppm-dap.md
tgeoghegan: OK, then do we need a MUST describing how the leader should assign batch IDs?

branlwyd: Sure -- the existing text in the "Fixed-size Queries" section defined them this way, but it's probably valuable to reiterate this point in the "Leader Initialization" section since that section defines the specific protocol steps. I did so in terms of a `MUST`.

draft-ietf-ppm-dap.md
branlwyd: Thinking about this some more, I think I don't want to rename Query to BatchID: while this would work well for `fixed-size`, I think the interval included in `time-interval` isn't quite an ID: specifically, `time-interval` tasks allow queries that span multiple `min_batch_duration` intervals, so for `time-interval` the query is actually describing a range of batches rather than a single batch. IMO, a batch ID would identify a single `min_batch_duration` interval for `time-interval`.

(of course, I could keep Query and separately define BatchID, but at that point the suggested change is no longer a simplification...)

cjpatton:
> (also, did I miss something with the above? maybe there's some kinds of imprecision in knowledge of assignment of client reports/aggregation jobs to batches that would allow a serialization point to be safely skipped?)

I'm not sure myself, I haven't fully thought this through. I'm also happy to wait for implementation experience to tell us if we need to change course here.

MichaelScaria:
In #301:

> With the current next-batch semantics, there is the possibility that a Collector crashing at an inopportune time might lead to a batch being lost permanently.


Would this happen when the Leader commits that it sent the Collector the next batch and the Collector crashes before it commits that it received the batch?

MichaelScaria:
I'm a little concerned with requiring that the Leader assigns batch ids consecutively. If there are multiple Leaders instances running (for scale purposes), it would require that we share some type of counter across the instances. Additionally, it's not clear to me what happens if we lose this state.

I'm also just catching up on this thread, so apologies in advance if I'm missing any details. Looking forward to hearing back!

draft-ietf-ppm-dap.md
```diff
@@ -968,7 +960,8 @@ This message consists of:
 * Information used by the Aggregators to determine how to aggregate each report:
 
     * For fixed-size tasks, the Leader specifies a "batch ID" that determines
-      the batch to which each report for this aggregation job belongs.
+      the batch to which each report for this aggregation job belongs. The
+      Leader MUST allocate these IDs consecutively starting from 0.
```
wangshan, line 964: this contradicts with previous text: "it is expected that the batch ID to start
 from will be configured out-of-band from the DAP protocol." at line #571

wangshan:
Today we already require collector to store some states, for e.g. the task ID, and the "collect job URI" returned by leader, the (random) batchID is just an addition to that.
I guess I don't see the added value of having consecutive numbers.

MichaelScaria:
For this PR, having the Collector to indicate the desired batch ID (as proposed), but not requiring the Leader to consecutively order the batches? That allows implementations to have the flexibility to:

  a) Make the Leader generate random batch ids and explicitly hand it off to the collector.
  
  or
  
   b) Make the Leader consecutively generate batch ids and allow the collector to predict the batch id.
   
     In a separate issue/PR we can discuss how a Leader could implement the “next batch” functionality. Currently, it isn’t clear how the Leader and Collector reconcile the batch id if state is lost for whatever reason. An analogy I have in my mind would be how Kafka maintains offsets for consumer groups. Similarly, the Leader could maintain the “batch offset” the collector last committed receiving.

branlwyd:
> In #301:
> 
> > With the current next-batch semantics, there is the possibility that a Collector crashing at an inopportune time might lead to a batch being lost permanently.
> 
> Would this happen when the Leader commits that it sent the Collector the next batch and the Collector crashes before it commits that it received the batch?

Correct, that is the situation I was concerned with. To be fully explicit, the problematic scenario with the old `next-batch` semantics was:

1. The Collector issues a `next-batch` request to the Leader. The Leader returns the next batch ID and updates its durable storage to note that this batch has been returned to the Collector.
2. Before the Collector stores the batch ID back to its own durable storage, it crashes.

At this point, there would be no way for the Collector to interact with the Leader to retrieve this batch ID again, so the batch would effectively be lost.

Putting it another way, I think the core of this problem is that the old `next-batch` semantics required an (inherently-fallible) durable state update from both the Leader & the Collector; these updates were not transactional, so the Leader's update could succeed while the Collector's update could fail; and if this did occur, there was no way for the Collector to interact with the Leader to retrieve the same batch ID again.

cjpatton:
Thanks @MichaelScaria, I'm with you on the requirements. However I just noticed a potential problem here: If the Leader doesn't get to choose the batch ID, then it doesn't get to proactively aggregate reports as they come in: It has to wait until the batch ID is assigned by the Collector.

branlwyd:
> I'm a little concerned with requiring that the Leader assigns batch ids consecutively. If there are multiple Leaders instances running (for scale purposes), it would require that we share some type of counter across the instances.

In this scenario, do the multiple Leader instances share state or not? ("state" in this case being received client reports, aggregation jobs, etc)

If the Leader instances do share state, I think the argument in https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/pull/308#issuecomment-1230705299 applies: assigning batch IDs without risking adding too many client reports to a given batch already implies serialization of creation of aggregation jobs. Once that serialization is accepted, the counter will not affect performance over-and-above the existing performance cost of serializing the rest of the aggregation-job-creation operation. (That said, I'd be interested in review of the linked argument; I may well be missing some clever implementation strategy.)

If the Leader instances don't share state, I think the two Leaders are effectively separate DAP deployments and are free to use a per-Leader counter.


> Additionally, it's not clear to me what happens if we lose this state.

This is a good point -- I don't think the DAP specification covers data-loss-recovery scenarios currently. For this case in particular, if only the counter is lost, out-of-band communication would allow recovery. I think that is probably acceptable for an initial pass. (If more data is lost, e.g. client reports/aggregation jobs, recovery may be much more difficult--but that's not new with this PR & IMO would best be covered in separate work.)

draft-ietf-ppm-dap.md
branlwyd: Ah, good point. The intent of the text you reference on line 571 is that a Collector that starts after the Aggregators have been running may not be able to start from 0 -- while the Aggregators generated batch IDs starting form 0, by the time the Collector starts, batch ID 0 will have already been garbage collected & no longer be available for collection. Given that, the Collector needs some way to know where to start from.

That said, this text is confusing & likely unnecessary at this point. I deleted the relevant text around line 571.

wangshan:
> 1. The Collector issues a next-batch request to the Leader. The Leader returns the next batch ID and updates its durable storage to note that this batch has been returned to the Collector.
> 2. Before the Collector stores the batch ID back to its own durable storage, it crashes.

Can we not ask the collector to re-send `next-batch` request in this case?  

> Upon receipt of a CollectReq, the leader begins by checking that the request meets the requirements of the batch parameters using the procedure in {{batch-validation}}. If so, it immediately sends the collector a response with HTTP status 303 See Other and a Location header containing a URI identifying the collect job that can be polled by the collector, called the "collect job URI". 

Should we include the batchID in the "collect job URI", so if the collector crashes before saving the batchID, it will have to re-send CollectReq with `next-batch` since the collect job URI will be lost too?

branlwyd:
> Today we already require collector to store some states, for e.g. the task ID, and the "collect job URI" returned by leader, the (random) batchID is just an addition to that. I guess I don't see the added value of having consecutive numbers.

The reason for this change are:
* This allows the Collector to opportunistically query for batches. (the discussion around this point starts at https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/pull/297#discussion_r949117056)
* The semantics of `next-batch` could lead to data loss by design (see https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/pull/308#issuecomment-1233352095 for a full explanation). Allowing the Collector to predict batch IDs avoids this problem. (but it is not the only possible solution to this problem)

branlwyd:
> > 1. The Collector issues a next-batch request to the Leader. The Leader returns the next batch ID and updates its durable storage to note that this batch has been returned to the Collector.
> > 2. Before the Collector stores the batch ID back to its own durable storage, it crashes.
> 
> Can we not ask the collector to re-send `next-batch` request in this case?

The semantics of `next-batch` were defined such that sending another `next-batch` request would receive a different batch. There was no way for the Leader to know that it should return the same batch twice for two different `next-batch` requests. See https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/pull/297#discussion_r948460390 for original discussion around these semantics.


> > Upon receipt of a CollectReq, the leader begins by checking that the request meets the requirements of the batch parameters using the procedure in {{batch-validation}}. If so, it immediately sends the collector a response with HTTP status 303 See Other and a Location header containing a URI identifying the collect job that can be polled by the collector, called the "collect job URI".
> 
> Should we include the batchID in the "collect job URI", so if the collector crashes before saving the batchID, it will have to re-send CollectReq with `next-batch` since the collect job URI will be lost too? @branlwyd @cjpatton @MichaelScaria

I think in the medium-term, including batch ID in the collect job URI is likely a good idea. I think this work would be part of (or perhaps an eventual extension of) https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/issues/278.

You're correct that the data-loss scenario under discussion would lose the collect job URI, as well as the batch ID. However, I'm not certain how including the batch ID in the collect job URI helps:

* When using `next-batch` (i.e. the approach before this PR), a second `next-batch` request will receive a different batch, so including the batch ID in the collect job URI does not help to avoid data loss.
* When using the approach in this PR, if the Collector loses the collect job URI, it can issue a second collect request with the same batch ID to receive it again.

branlwyd:
> For this PR, having the Collector to indicate the desired batch ID (as proposed), but not requiring the Leader to consecutively order the batches?

That would work, but we'd need to specify how the Collector "discovers" batches. That is, I think we'd need to specify the equivalent of "next-batch" functionality in this PR or at least before DAP-02 is published.


> Similarly, the Leader could maintain the “batch offset” the collector last committed receiving.

A commit-based approach would definitely work; my only complaint is that it may be more complex than is needed.

wangshan:
@branlwyd thank you for pointing out previous comments, definitely helped explaining the ideas. But I'm curious why the "current-batch" solution was removed, it feels like the right solution to this problem: the "current" batch is maintained by leader, if collector restarts without the batchID, it can ask leader with a "current-batch" CollectReq. 

Alternatively, we can let collector specify a randomised batchID (what @MichaelScaria said), in this case the "current" batch is maintained by collector, there's no need for batchID discovery, collector could persist the batchID before sending CollectReq. But as @cjpatton pointed out, since the batchID is used in AggregateInitializeReq, the aggregators will have to wait for a CollectReq before aggregating. Unless we change AggregateInitializeReq to use a list of agg job IDs instead of a batchID and associate the list with batchID when CollectReq is received.

MichaelScaria:
> That is, I think we'd need to specify the equivalent of "next-batch" functionality in this PR or at least before DAP-02 is published.

Any reason we need "next-batch" functionality before draft 02? If we remove that feature and assume that the Leader and Collector have a way to share the batchID (for example, encoding it in the collect URI), then fixed size query isn't regressing from a data loss perspective.

Just to clear up any confusion, I wrote earlier: 

> For this PR, having the Collector to indicate the desired batch ID (as proposed), but not requiring the Leader to consecutively order the batches? 

To clarify, Collector would set the batchID on CollectReq that the Leader had previously generated.

branlwyd:
> I'm curious why the "current-batch" solution was removed, it feels like the right solution to this problem: the "current" batch is maintained by leader, if collector restarts without the batchID, it can ask leader with a "current-batch" CollectReq.

The only reason it was removed was we believed we didn't need it in an initial implementation (https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/pull/308#discussion_r956491618).

I'd be OK restoring it if desirable -- @cjpatton any thoughts?

cjpatton:
> > I'm curious why the "current-batch" solution was removed, it feels like the right solution to this problem: the "current" batch is maintained by leader, if collector restarts without the batchID, it can ask leader with a "current-batch" CollectReq.
> 
> The only reason it was removed was we believed we didn't need it in an initial implementation ([#308 (comment)](https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/pull/308#discussion_r956491618)).
> 
> I'd be OK restoring it if desirable -- @cjpatton any thoughts?

I"m not sure restoring solves the problem at hand. It might be helpful to take a step back and organize our requirement for this change. I believe these are as follows (let me know if I'm missing anything):
1. Our main goal is to prevent data loss in case the Collector forgets the batch ID (due to @branlwyd).
2. Restrictions on ordering of batch IDs is untenable (due to @MichaelScaria). In particular, it would be useful if the Leader could choose batch IDs at random.

The decision tree is as follows:
* If folks generally agree on these requirements, then our goal is to extend this PR so that it solves for (2.), without regressing on (1.).
   * Someone observed previously that forgetting the collect URI also results in data loss. Thus one way we can solve for these requirements without *increasing* the risk of data loss is to require the collect URI to encode the batch ID. This appears to be the best suggestion right now.
   * Perhaps we can de-couple the batch ID used by the Aggregators for aggregation from the collect flow. In other words, the Leader maintains in its a map from batch IDs use during aggregation and "batch indices" queried by the collector. For example, when the Collector queries "batch index == 0", the Leader looks for the batch that's closest to being finished and associated the corresponding batch ID with that index. This would allow the Collector to "predict" batches without storing them (thereby solving (1.)) without constraining the Leader.
* If folks don't agree on these requirements, then we should move the discussion back to the issue and figure out what we need.

wangshan:
I think 1. is the main requirement, 2. is only an issue if we choose to go the predictable ID route. @cjpatton 

I think the decision tree should be following:

If we want to maintain the "current batch id" in leader, then a clean solution could be using an idempotent `current-batch` query, and add batch ID to collect job URI. Collector only sends CollectReq with `current-batch`
* if it saves the returned collect job URI successfully, then it also knows the batchID, and from that moment on it can pull from collect job URI. 
* If it failed to save the returned collect job URI, it can send the same `current-batch` again.

Having a predicable ID is another solution but I would imagine the complexity of maintaining such count in leader is no less than maintaining the current batch ID.

If we want to maintain the "current batch id" in collector, then we just need the collector to come up with an(random) ID and always use that in CollectReq, we can still add batchID to collect job URI so collector only need to maintain one state. 
* If we want to go this route, we must resolve what batchID to send in `AggregateInitailiseReq`, we could, for e.g. not send any batchID in `AggregateInitailiseReq`, but send the list of agg job IDs in `AggregateShareReq`, but that has the problem of how long the list of agg job ID can become, pointed out by someone from the mailing list when the issue was open.

cjpatton:
Ok, it sounds like we might be converging on the following changes to this PR:
* Let the Leader pick the batch ID
* (Re-)introduce `current-batch` query, in response to which the Collector returns the batch ID for the latest batch being processed.
* Require the collect URI to encode the batch ID. (Probably we would just spell out the structure of the URI explicitly.)

Please 👍 this comment if you're happy with this :)

cjpatton:
I'll just note that, for the third point, we might be able to remove the batch ID from the CollectResp.

chris-wood:
Let's please keep the semantics of how batch IDs are chosen outside of this PR, and instead do the simple thing ((3) in @cjpatton's last comment) here. There are lots of ways in which implementations might choose to pick the batch IDs. I don't think we need to be prescriptive here.

wangshan:
@chris-wood (3) is desirable, I think we should do it regardless of how batchID is chosen, but as @branlwyd has pointed out above: https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/pull/308#issuecomment-1233450503, it won't solve the requirement 1, if we keep current `next-batch` query.

chris-wood:
@wangshan that's fine, but this is a _separable change_. I am suggesting we make this change separately. This PR is doing too much.

cjpatton (changes requested):
Unapproving, pending changes discussed in slack.

draft-ietf-ppm-dap.md
```diff
@@ -968,7 +956,8 @@ This message consists of:
 * Information used by the Aggregators to determine how to aggregate each report:
 
     * For fixed-size tasks, the Leader specifies a "batch ID" that determines
-      the batch to which each report for this aggregation job belongs.
+      the batch to which each report for this aggregation job belongs. The
```
branlwyd, line 959: I'm not totally sure "The Leader MAY allocate batch IDs in any order." is necessary -- we could remove this sentence and leave this implicit. Thoughts?

branlwyd (commented):
OK, this is ready for review w/ the changes suggested from Slack discussion.

draft-ietf-ppm-dap.md
cjpatton: I agree it's unnecessary.

draft-ietf-ppm-dap.md
```diff
@@ -1556,18 +1545,15 @@ reports in the batch.
...
-* If the query contained in the CollectReq has type `by-batch-id`, the
-  Leader checks that the provided batch ID corresponds to a batch ID it
-  returned in a previous CollectResp for the task.
+For fixed-size tasks, the batch boundaries are defined by numeric batch IDs.
+Thus the Aggregator needs to check that the query is associated with a known
+batch ID:
 
-  [OPEN ISSUE: Consider making batch IDs predictable. See issue #301.]
+* For a CollectReq, the Leader checks that the provided batch ID corresponds to
+  a batch ID it returned in a previous CollectResp for the task.
```
cjpatton, line 1553: Is this required? For this PR we're assuming the mechanism by which batch IDs are assigned is implementation-specific. (This will be addressed in the follow-up PR.)

draft-ietf-ppm-dap.md
```diff
@@ -505,25 +503,12 @@ parameters used by the Aggregators to select a batch of reports specific to the
...
-opaque BatchId[32];
-
-struct {
-    FixedSizeQueryType fixed_size_query_type;
-    select (fixed_size_query_type) {
-        next-batch:  Empty;
-        by-batch-id: BatchId batch_id;
-    }
-} FixedSizeQuery;
+uint64 BatchID;
```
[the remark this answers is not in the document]
branlwyd: 👍🏻 I restored it to 32 bytes as originally written; I suppose this matches the size of other "opaque identifiers for which we want to enable random generation with negligible collision chance". (I'd advocate for changing all of these identifiers to 16 bytes rather than 32, but I also think that's a separable change, and so for now I think maintaining consistency is more valuable.)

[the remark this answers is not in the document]
cjpatton: I'd be all for 16 bytes, but yes this change is separable :) One thing we need to keep in mind is whether we need IDs to be the hash of things, in which case 32 bytes might be better.

draft-ietf-ppm-dap.md
branlwyd: Removed.

draft-ietf-ppm-dap.md
branlwyd: Good point, I think it's not required in this PR. Removed.

cjpatton (approved):

wangshan (approved):

chris-wood:
@branlwyd sorry :( I made this conflict. Could you please resolve.

branlwyd:
> @branlwyd sorry :( I made this conflict. Could you please resolve.

No worries -- rebased again, GH reports no conflicts. This should be ready for review again.

PR #313, opened 2022-09-02
Fixed-size tasks: specify `current-batch` to allow Collector to discover batch IDs.

branlwyd:
This is implemented via a new `current-batch` query type for fixed-size
tasks, which allows the Leader to select a batch to return to the
Collector.

This is stacked on https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/pull/308. (Unfortunately, I don't know how to use that PR as a base without merging into
a branch on my personal fork of the DAP specification repository, which is
probably not desirable.)

### DIFF
--- a/draft-ietf-ppm-dap.md
+++ b/draft-ietf-ppm-dap.md
@@ -119,7 +119,7 @@ seen in the clear by any server.
 - Define a new task configuration parameter, called the "query type", that
   allows tasks to partition reports into batches in different ways. In the
   current draft, the Collector specifies a "query", which the Aggregators use to
-  guide selection of the batch. Two query types are defined: the "time-interval"
+  guide selection of the batch. Two query types are defined: the "time_interval"
   type captures the semantics of draft 01; and the "fixed_size" type allows the
   Leader to partition the reports arbitrarily, subject to the constraint that
   each batch is roughly the same size. (\*)
@@ -543,11 +543,24 @@ specific to the given query type. A query is defined as follows:
 ~~~
 opaque BatchID[32];
 
+enum {
+  by_batch_id(0),
+  current_batch(1),
+} FixedSizeQueryType;
+
+struct {
+  FixedSizeQueryType query_type;
+  select (query_type) {
+    by_batch_id: BatchID batch_id;
+    current_batch: Empty;
+  }
+} FixedSizeQuery;
+
 struct {
   QueryType query_type;
   select (Query.query_type) {
     case time_interval: Interval batch_interval;
-    case fixed_size: BatchID batch_id;
+    case fixed_size: FixedSizeQuery fixed_size_query;
   }
 } Query;
 ~~~
@@ -602,12 +615,17 @@ differential privacy.
 For this query type, the Aggregators group reports into arbitrary batches such
 that each batch has roughly the same number of reports. These batches are
 identified by opaque "batch IDs", allocated in an arbitrary fashion by the
-Leader. To get the aggregate of a batch, the Collector issues a query specifying
-the batch ID of interest (see {{query}}).
+Leader.
+
+To get the aggregate of a batch, the Collector issues a query specifying the
+batch ID of interest (see {{query}}). The Collector may not know which batch ID
+it is interested in; in this case, it can also issue a query of type
+`current_batch`, which allows the Leader to select a recent batch to aggregate.
+The leader SHOULD select a batch which has not yet began collection.
 
 In addition to the minimum batch size common to all query types, the
-configuration includes a "maximum batch size", `max_batch_size`, that determines
-maximum number of reports per batch.
+configuration includes a "maximum batch size", `max_batch_size`, that
+determines maximum number of reports per batch.
 
 Implementation note: The goal for the Aggregators is to aggregate precisely
 `min_batch_size` reports per batch. Doing so, however, may be challenging for
@@ -620,10 +638,6 @@ for each batch.
 [OPEN ISSUE: It may be feasible to require a fixed batch size, i.e.,
 `min_batch_size == max_batch_size`. We should know better once we've had some
 implementation/deployment experience.]
-
-[OPEN ISSUE: It may be desirable to allow Collectors to query for a current/
-recent batch ID. How important this is will be determined by deployment
-experience.]
 
 ## Task Configuration {#task-configuration}
 
@@ -1730,13 +1744,17 @@ reports in the batch.
 
 ##### Boundary Check
 
-For fixed_size tasks, the batch boundaries are defined by opaque batch IDs. Thus
-the Aggregator needs to check that the query is associated with a known batch
-ID:
+For fixed_size tasks, the batch boundaries are defined by opaque batch IDs.
+Thus the Aggregator needs to check that the query is associated with a known
+batch ID:
+
+* For a CollectReq containing a query of type `by_batch_id`, the Leader checks
+  that the provided batch ID corresponds to a batch ID it returned in a
+  previous CollectResp for the task.
 
 * For an AggregateShareReq, the Helper checks that the batch ID provided by the
-  Leader in its corresponds to a batch ID used in a previous
-  AggregateInitializeReq for the task.
+  Leader corresponds to a batch ID used in a previous AggregateInitializeReq
+  for the task.
 
 ##### Size Check
 


branlwyd:
@MichaelScaria & @wangshan: GitHub's UI will not allow me to add you as reviewers, but I would appreciate your review.

draft-ietf-ppm-dap.md
```diff
@@ -505,18 +503,18 @@ parameters used by the Aggregators to select a batch of reports specific to the
...
-opaque BatchId[32];
-
 struct {
-    FixedSizeQueryType fixed_size_query_type;
-    select (fixed_size_query_type) {
-        next-batch:  Empty;
+    FixedSizeQueryType query_type;
+    select (query_type) {
         by-batch-id: BatchId batch_id;
+        current-batch: Empty;
```
MichaelScaria, line 517: Won't the line above this need to be `BatchID`?

draft-ietf-ppm-dap.md
```diff
@@ -942,7 +944,7 @@ struct {
   opaque agg_param<0..2^16-1>;
   select (query_type) { // determined by task configuration
     time-interval: Empty;
-    fixed-size: BatchId batch_id;
+    fixed-size: BatchID batch_id;
```
MichaelScaria, line 947: Do we need to also make an update to `AggregateContinueReq`?

draft-ietf-ppm-dap.md
branlwyd: Indeed -- updated.

draft-ietf-ppm-dap.md
branlwyd: What update are you thinking of?

On the assumption it is something along the lines of "include the batch ID in the `AggregateContinueReq` as well", I think we don't need an update to `AggregateContinueReq`: once the mapping of aggregation job to batch ID is determined by the `AggregateInitializeReq`, it is sufficient to specify only the aggregation job ID in further communications.

(apologies if I'm misinterpreting your comment!)

branlwyd:
(I also picked up changes based on review comments for #308)

draft-ietf-ppm-dap.md
MichaelScaria: Nope, you assumed right! Would the Helper need to have some sort of reverse map to recover the batch ID from the job ID?

MichaelScaria: Thinking aloud here - 

On init, Helper gets batch and job IDs. 
On cont, Helper gets job ID.
On collect, Helper gets batchID.

So we would need batchID -> jobID and vice versa right?

branlwyd: > Nope, you assumed right! Would the Helper need to have some sort of reverse map to recover the batch ID from the job ID?

Indeed: the Helper will need to track this as an additional piece of state. Presumably this reverse map would be an index (or equivalent) in a Helper implementation.

Thinking about whether this state is needed, I think it is: even if we decided that `AggregateContinueReq` should include a `BatchID` as well, I think we would want to enforce that the `BatchID` matches for all `Aggregate{Initialize,Continue}Req` messages for the same aggregation job ID, which would imply the helper storing the state there as well. (maybe there is a larger change possible?)


>Thinking aloud here -
>
>On init, Helper gets batch and job IDs.
>On cont, Helper gets job ID.
>On collect, Helper gets batchID.
>
>So we would need jobID -> batchID (for cont) and batchID -> jobID (for collect) right?

That's correct. Effectively, a batch is made up of one or more aggregation jobs. Informally, aggregation jobs are the unit of batching between the Aggregators for the majority of the aggregation process (i.e. the Aggregators evaluating the VDAF up to the point of recovering their respective output shares). Batches are the unit of collection by the Collector.

One reason that the concept of "batch" is separate from "aggregation job" is that it may be practically infeasible to group together enough reports at a time to meet the batch requirements (e.g. `min_batch_size`). Also, in many scenarios Aggregators can begin the aggregation process long before the collection process; this is aided by allowing aggregation jobs to be created even before e.g. enough reports have arrived to fill a batch. That is, the concept of aggregation jobs allows aggregating already-received reports for a batch concurrently with additional reports arriving.

wangshan: why do we need a reverse map of jobID-> batchID?

draft-ietf-ppm-dap.md
```diff
@@ -575,24 +573,28 @@ particularly important for controlling the amount of noise added to reports
...
-minumum batch size common to all query types, the conifugration includes a
-"maximum batch size", `max_batch_size`, that determines maximum number of
-reports per batch.
+that each batch has roughly the same number of reports. These batches are
+identified by opaque "batch IDs", allocated in an arbitrary fashion by the
+Leader.
+
+To get the aggregate of a batch, the Collector issues a query
+specifying the batch ID of interest (see `FixedSizeQuery` in {{query}}). The
+Collector may not know which batch ID it is interested in; in this case, it can
```
wangshan, line 583: If the batchID is allocated by Leader in an arbitrary fashion, then wouldn't the collect HAVE TO issue a current-batch query first? Or are we deliberately not being specific about the way current-batch is used, so some OOB agreement can be used to "predict" the batchID?

draft-ietf-ppm-dap.md
```diff
@@ -1556,18 +1558,16 @@ reports in the batch.
...
 
-  [OPEN ISSUE: Consider making batch IDs predictable. See issue #301.]
+* For a CollectReq containing a query of type `by-batch-id`, the Leader checks
+  that the provided batch ID corresponds to a batch ID it returned in a
+  previous CollectResp for the task.
 
-* The Helper checks that the batch ID provided by the leader in its
-  AggregateShareReq corresponds to a batch ID used in a previous
+* For an AggregateShareReq, the Helper checks that the batch ID provided by the
+  Leader in its corresponds to a batch ID used in a previous
```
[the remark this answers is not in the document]
branlwyd: Removed, thanks.

draft-ietf-ppm-dap.md
branlwyd: Typically yes, I expect that the Collector would use `current-batch` first (outside of OOB agreement you mention), and then possibly use one or more `by-batch-id` queries for that batch, depending on the VDAF.

I was writing this text as an explanation of available functionality/expectations for the `fixed-size` query type, rather than a step-by-step guide to how the query types are expected to be used. I can change if folks like?

draft-ietf-ppm-dap.md
divergentdave: I think the main reason we need a map from aggregation job IDs to collect batches is for enforcing collect-time privacy requirements (the aggregators will need to examine all collect batches that each aggregation job was previously a part of, given the candidate list of aggregation jobs for the current collect batch).

The reverse map may also be used during handling of AggregateContinueRequest, to look up state relating to the collect batch and incrementally incorporate output shares into an aggregate share, but that's an implementation decision. Aggregators could just as well store output shares indexed by aggregation job, and then only fetch output shares and combine them all at once when handling a CollectRequest or AggregateShareRequest.

MichaelScaria: @branlwyd Wouldn't the simplest thing be to send both batchID and jobID on init and continue? On continue, if those two parameters don't uniquely resolve to a set of transitions, then we send back some error. On aggregate share, the Leader only sends the batchID, then the Helper finds all jobs for that batchID and can enforce the collect-time privacy requirements on the output shares for those jobs.

From an implementation perspective, you could use any distributed key value store that additionally supports a prefix query (S3 and Redis should be good candidates for this).

Ideally we can find a solution that uses a single data structure.

branlwyd: IMO, your suggestion is actually more complicated, in the sense that it sends more data over the wire & creates an additional possible error case (i.e. aggregate continue req contains a mismatched aggregation job ID/batch ID) for implementations to potentially mishandle. (Resending the data does ensure that the two aggregators don't have different conceptions of the aggregation job<->batch mapping; but this would be a pretty nasty bug, and would be caught at collection time even without resending the data.)

For that reason, I'd prefer to keep the current text as written.

===

re: single data structure

I'd be amenable to a single data structure somehow merging the concept of an aggregation job & a batch; however, I don't believe the suggestion to send both batch ID & aggregation job ID moves us towards that goal. More generally, I think we'd need to somehow overcome/sidestep the practical issues I mention upthread from this comment:

> One reason that the concept of "batch" is separate from "aggregation job" is that it may be practically infeasible to group together enough reports at a time to meet the batch requirements (e.g. min_batch_size). Also, in many scenarios Aggregators can begin the aggregation process before the collection process can start; this is aided by allowing aggregation jobs to be created even before e.g. enough reports have arrived to fill a batch. That is, the concept of aggregation jobs allows aggregating already-received reports for a batch concurrently with additional reports arriving.

(To add a few more words to this: I like the ability to "unhitch" the privacy parameters, like `min_batch_size`, from the technical parameters, such as "how many client reports will we batch into a single unit of aggregation" -- I wouldn't be too happy if we ended up with a design which required trading off privacy to reach desired performance/reliability properties. And I consider the ability to commonly begin the aggregation process before the batch is full to be an important performance property of the current design.)

FWIW, I think implementations may practically already be in the world where only one data structure is required: it is possible to arrange an implementation such that "aggregation jobs" are a real/reified data structure, and "batches" exist only as a collection of aggregation jobs -- the mapping being stored with the aggregation job data structure. (In this implementation strategy, collection information is also attached to aggregation jobs, which may be surprising but also assists in ensuring that the same aggregation job is not aggregated/collected too many times.) Since in this strategy batches are represented entirely as an indexed identifier associated with an aggregation job, in this implementation strategy batches are implicit.

MichaelScaria: Sure, the tradeoff seems to be either we send an extra field (batchID) and maintain one mapping `(batchID, jobID) -> state` on the Helper, or we only send jobID in `AggregateContinueReq` and maintain two mappings `batchID -> jobID` and `jobID -> state`. 

I was wrong earlier when I thought we would need a reverse map, so I no longer feel as strongly in either direction.

That all said, this conversation is probably out of scope for this PR and I don't want to hold this up. Happy to continue the discussion elsewhere if needed.

As for what I meant by single data structure, I was talking about the number of mappings. But I 💯 agree with the last paragraph, we were actually implementing it that way while the fixed-size proposal was iterated on. Now that Init requires the batchID to be generated, we can no longer do that I think. @branlwyd

branlwyd: OK -- I'm happy to make a change if there's consensus (or if additional discussion provides even stronger justification for making this change), but will leave the text as-is for now.

===

One question (this might be moot if we are OK going with the current direction, but might influence how we iterate on this approach in the future):

> But I 100 agree with the last paragraph, we were actually implementing it that way while the fixed-size proposal was iterated on. Now that Init requires the batchID to be generated, we can no longer do that I think.

Could you elaborate on this point? I'm not sure how associating batch ID at time-of-aggregate-job-initialization blocks this implementation strategy. Also, I'm not sure what the suggested alternative is: even with the text as written at `HEAD`, i.e. before PRs #308 & #313, the text was written such aggregation jobs were associated with a batch at time of aggregation job initialization. (to my memory, #297 was always written that way--although I may be misremembering, the PR branch was squashed before merging so I've lost some of the history)

I'd like to enable implementations which represent batches "implicitly", so if I'm missing something here I'd be interested in fixing things up. (but as long as you're OK with not holding up this PR, I do think it might be valuable to continue discussion elsewhere)

cjpatton (approved):
Very well done (though I'm sure reviewers played no small part). I'm happy to land this and consider the issue closed.

cjpatton:
Thanks all for reviews and thanks to @branlwyd for creating the PR. Based on conversations I've had, it doesn't seem like we quite have consensus on this. Unless it's a blocker for anyone, I would like to suggest we punt to DAP-03.

The main problem this PR is intended to solve is interop between the Leader and Collector: As of #308, the communicating batch IDs from Leader to Collector happens out-of-band, i.e., is deployment specific: The Leader could choose predictable batch IDs, or it could choose them at random and communicate them, somehow, to the Collector. It's also possible to implement the main idea of this PR by exposing an HTTP endpoint for the "current batch".

TL;DR: My suggestion is to wait for deployment experience to decide if we need to spell anything out here. Any objections?

wangshan:
I think this PR looks fine on its own. My main question/comment is it looks like the main difference between existing `next-batch` query and this PR is that the former requires collector to call next-batch first, then `by-batch-id`, where this PR says there is `current-batch` and `by-batch-id` but how they are used by collector is unspecified. If the main purpose for the PR is to define inter-op between leader and collector, I don't think it quite achieved that goal.

So I support punting this to DAP-03.

branlwyd:
> My main question/comment is it looks like the main difference between existing `next-batch` query and this PR is that the former requires collector to call next-batch first, then `by-batch-id`, where this PR says there is `current-batch` and `by-batch-id` but how they are used by collector is unspecified.

For what it's worth/for clarity: the intended use of `current-batch` was effectively the same as `next-batch` -- the Collector would call `current-batch` (to discover a batch), then `by-batch-id` (if necessary).

The reason `current-batch` is preferable to `next-batch` is `next-batch` requires durable state updates from both the Leader & Collector. These updates were not transactional, so the Leader's update could fail while the Collector's update could succeed. In the case that the Leader's update succeeded but the Collector's update failed, the returned batch would effectively be lost -- there would be no way for the Collector to interact with the Leader to re-retrieve the batch. That is, `current-batch` is intended as a bugfix on `next-batch`, but the intended usage is otherwise the same.

cjpatton:
I've filed https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/issues/342. Let's consider landing this in DAP-03. @branlwyd if you want you can rebase this now, or we can close this PR and re-open when we're ready to revisit.

branlwyd:
I've rebased -- I will rebase again if/when folks are ready to consider this PR again, or on request.

@wangshan I'd like to alleviate the confusion over the expected usage of this approach; I'd like your opinion. I think the most relevant text in the PR is:

> To get the aggregate of a batch, the Collector issues a query specifying the
batch ID of interest (see {{query}}). The Collector may not know which batch ID
it is interested in; in this case, it can also issue a query of type
`current-batch`, which allows the Leader to select a recent batch to aggregate.

Is there anything you would suggest to make the intended usage clearer? Would it be helpful if `current-batch` was mentioned before mentioning that the Collector can query by batch ID? (reviewing this, I think I should mention also `by-batch-id` by name)

cjpatton:
> If leader decides the batch-id, how can collector know the batch-id without calling current-batch first?

It doesn't: The CollrectResp will carry the batch ID, which is how the Collector learns this for the first time.

branlwyd:
I have rebased this PR on current main; it should be ready for review once again.

cjpatton (changes requested):
Revoking approval while until I can do another pass.

draft-ietf-ppm-dap.md
```diff
@@ -543,11 +543,24 @@ specific to the given query type. A query is defined as follows:
 ~~~
 opaque BatchID[32];
 
+enum {
+  by-batch-id(0),
```
wangshan, line 547: the current draft unified enum cases using underscore.

wangshan:
@cjpatton @branlwyd I see, so there are two interactions if batch-id is chosen by leader:
1A. collector sends CollectReq with `current-batch` or 1B. collector sends CollectReq with `by-batch-id`
2. leader responses with a 303 + "collect job URI"
3. collector pulls "collect job URI", leader responds 200 + CollectResp which includes batch-id

if crash happens between 2 and 3, collector will go to step 1A, instead of relying on collect job URI; if crash happens after 3, collector will go to step 1B. I think it might worth calling out collect job URI is transactional and should not be relied on as an identifier, but I guess this doesn't affect the inter-op around batch-id and should be addressed separately. 

I'm happy to approve this as long as the enum value format is corrected.

draft-ietf-ppm-dap.md
cjpatton: (here and below)

draft-ietf-ppm-dap.md
```diff
@@ -602,12 +615,16 @@ differential privacy.
...
 that each batch has roughly the same number of reports. These batches are
 identified by opaque "batch IDs", allocated in an arbitrary fashion by the
-Leader. To get the aggregate of a batch, the Collector issues a query specifying
-the batch ID of interest (see {{query}}).
+Leader.
+
+To get the aggregate of a batch, the Collector issues a query specifying the
+batch ID of interest (see {{query}}). The Collector may not know which batch ID
+it is interested in; in this case, it can also issue a query of type
+`current-batch`, which allows the Leader to select a recent batch to aggregate.
```
cjpatton, line 623: We might add a SHOULD here to guide the Leader towards returning the oldest, not-yet-collected batch.

cjpatton (commented):
> [Shan] if crash happens between 2 and 3, collector will go to step 1A, instead of relying on collect job URI; if crash happens after 3, collector will go to step 1B. I think it might worth calling out collect job URI is transactional and should not be relied on as an identifier, but I guess this doesn't affect the inter-op around batch-id and should be addressed separately.

+1 here. Might as well address it in this PR, I think.

tgeoghegan (approved):
I think this change works, in that it enables a collector that has no idea what batch IDs exist for a task to bootstrap itself into collecting a sequence of aggregate results. We should take this change because without it, fixed-batch DAP doesn't work without some out of band arrangement between leader and collector on where batch IDs start. I'm guessing that DAP-02 implementations will have no choice but to implement this or something that is functionally equivalent, so we might as well pull it into the protocol text.

However, I think that `current-batch` obscures what we're really specifying here, which is an ordered list of batches with a cursor maintained by the leader. My suspicion, which I'm exploring in #367, is that we want the leader to expose a resource for a list of known batches that the collector can traverse. But I don't want to complicate this PR, which I think is well-focused and achieves a meaningful improvement, with that conversation, so to re-iterate, I think we should take this change.

draft-ietf-ppm-dap.md
tgeoghegan: I'm not sure this notion can be meaningful. The leader necessarily keeps track of the query count against a batch for privacy reasons, but unfortunately I don't think that means it can correctly maintain the position of the `current-batch` cursor. Suppose the leader services a collect request and delivers an aggregate result to the collector, but the collector crashes while receiving the message. The leader now might consider the batch in question collected and thus advance the cursor to the next batch. But when the collector restarts and queries `current-batch`, it really should get the batch that the leader just advanced past. Without an explicit ack from the collector, the leader has to assume the result has not been delivered.

The more I think about it, the more I think that this is the semantics of a message queue like AWS SQS or GCP PubSub, both of which will redeliver messages to consumers until a message is explicitly acked.

Anyway, the best the leader can do given what's in this PR is to advance the cursor regardless of the message being received by the collector. This means that some batches could be lost, but this proposal is still workable in the common case where the collector doesn't crash.

branlwyd: I'm happy to add a SHOULD if folks think it wise.

FWIW, I did not intend to imply list-like semantics for `current-batch` -- I attempted to be careful not to imply anything about the ordering of batches returned by `current-batch`. A few plausible implementation strategies:

* A queue, as discussed upthread. (and yep, I think `current-batch` would allow a fairly nice implementation based on GCP Pub/Sub, AWS SQS, or similar technologies)
* Randomly choose an acceptable batch with each call to `current-batch`.
* For a sharded implementation: choose a random shard, and ask it for a current batch. (each shard might have its own queue, select randomly, or use some other strategy)

Thoughts? I purposefully left the implementation strategy ambiguous (on the idea that it's often wise to be less prescriptive rather than more prescriptive when it comes to implementation strategies), but if everyone thinks a queue is most appropriate then we can certainly suggest that. A queue or queue-like implementation is what I'd go with, but I didn't want to block implementations from experimenting with different selection methods.

===

> Anyway, the best the leader can do given what's in this PR is to advance the cursor regardless of the message being received by the collector. This means that some batches could be lost, but this proposal is still workable in the common case where the collector doesn't crash.

Can you elaborate? With `current-batch`, the batch can continue to be returned until the collector actually attempts collection for the first time. This means that an ill-timed crash can be recovered from by polling `current-batch` again -- the same batch will eventually be returned again. The expectation is that the collector will durably store the collect URL before beginning to poll.

Or, to put it another way:
* If the crash happens before the collector durably stores the collect URI, the collector will eventually receive the same batch back from `current-batch` again -- no data loss.
* If the crash happens after the collector durably stores the collect URI, the collector has the collect URI in its durable storage & can begin/continue polling -- no data loss.

cjpatton: Fair points, both :) We probably SHOULDn't then.

branlwyd:
> @cjpatton @branlwyd I see, so there are two interactions if batch-id is chosen by leader: 1A. collector sends CollectReq with `current-batch` or 1B. collector sends CollectReq with `by-batch-id` 2. leader responses with a 303 + "collect job URI" 3. collector pulls "collect job URI", leader responds 200 + CollectResp which includes batch-id
> 
> if crash happens between 2 and 3, collector will go to step 1A, instead of relying on collect job URI; if crash happens after 3, collector will go to step 1B. I think it might worth calling out collect job URI is transactional and should not be relied on as an identifier, but I guess this doesn't affect the inter-op around batch-id and should be addressed separately.

Yep, correct, except that the crash happening before the collector stores the collect job URI would cause it to go back to 1A; the crash happening after the collector stores the collect job URI would cause it to continue polling. (`by-batch-id` is used for collecting the same batch a second time, for VDAFs where that is relevant such as `poplar1`)

draft-ietf-ppm-dap.md
cjpatton: @branlwyd I think this hasn't actually been resolved yet? Basically the ask is: change "-" to "_" in enum variants

branlwyd:
> However, I think that `current-batch` obscures what we're really specifying here, which is an ordered list of batches with a cursor maintained by the leader. My suspicion, which I'm exploring in #367, is that we want the leader to expose a resource for a list of known batches that the collector can traverse. But I don't want to complicate this PR, which I think is well-focused and achieves a meaningful improvement, with that conversation, so to re-iterate, I think we should take this change.

Only thing I'd add here is that there's no requirement for ordering -- the aggregator is free to choose nearly any strategy to choose a batch, there is no need for a linear order. Agreed to discuss the rest on #367.

draft-ietf-ppm-dap.md
branlwyd: Looking again: I added text that the Leader `SHOULD` return a batch that has not yet began collection, as the expected semantics there were IMO not clear enough. I left off the ordering part for now, but I'm happy to add it if folks think we should enforce queue-like behavior.

cjpatton: LGTM

MichaelScaria:
Could the collector miss out on batches if there isn't a requirement to order?

draft-ietf-ppm-dap.md
tgeoghegan: I'm wary of responding to a resolved thread in a PR that I approved already but I want to note one more thing in this context.

branlwyd:
>Can you elaborate? With current-batch, the batch can continue to be returned until the collector actually attempts collection for the first time. This means that an ill-timed crash can be recovered from by polling current-batch again -- the same batch will eventually be returned again. The expectation is that the collector will durably store the collect URL before beginning to poll.

IIUC, you're saying that the collector doing `POST /collect` with `current-batch` is akin to a subscriber pulling a message, and then the first `GET {collect-uri}` is akin to acking the message, in that it indicates to the leader that the collector has received the collect URI and thus the leader can then respond to `current-batch` with a different collect URI? I think that makes sense, and I see how this is fault tolerant, but now it seems like `GET {collect-uri}` has a side effect of changing how the server responds to `POST /collect`, which seems a bit off. So I wonder if we should change it to `POST {collect-uri}`. But I don't think we should make that change here, as we need to consider more broadly what abstraction we want for delivering aggregate shares from aggregators to collectors.

branlwyd:
> Could the collector miss out on batches if there isn't a requirement to order?

That depends on the implementation's strategy to return batches (as well as how quickly uncollected batches are GC'ed & how aggressively the Collector polls for new batches to collect).

(Strawman example showing how batches might end up missed: consider an aggregator which chooses a random uncollected batch each time. If the Collector happens to only requests a `current-batch` when there are at least two batches ready for collection, simple bad luck could lead the Aggregator to never randomly select a given batch to be returned to the Collector. This batch would eventually be GC'ed & therefore lost.)

Since this is avoidable by implementations, I think it's OK -- that is, I'd view such an issue as an implementation bug and/or operational configuration problem, rather than a problem with the specification. I'd also note that every example I can think of here requires the Collector to collect batches at a rate slower than they are being generated -- otherwise, any given batch will eventually be returned to the Collector because it will eventually be the only available batch to return. [I'm curious if anyone can think of a scenario that doesn't require a slow Collector.] But I think we should expect to lose batches if the Collector is collecting them slower than they are being generated, since IMO it's fairly unreasonable to expect the Aggregator to build up an arbitrarily-sized backlog of batches ready for collection -- that is, I think it is important for Aggregators to be able to eventually GC batches even if they have not yet been collected.

This leads me to wonder if we might eventually have some hint to the Collector that it should collect more frequently or risk losing some batches, but I think that's beyond the scope of this PR. It might help deployment scenarios where the rate of batch generation is unpredictably variable. Likely we shouldn't think too hard about this unless/until implementation experience shows that it might be helpful.

draft-ietf-ppm-dap.md
branlwyd: > IIUC, you're saying that the collector doing POST /collect with current-batch is akin to a subscriber pulling a message, and then the first GET {collect-uri} is akin to acking the message, in that it indicates to the leader that the collector has received the collect URI and thus the leader can then respond to current-batch with a different collect URI?

Correct, though I would note that the Leader is free to return a different batch even before the Collector "acks" it by beginning collection. (This would be handy to support use-cases where the Collector wants to collect potentially many batches in parallel -- no need to synchronize the "receive a `current-batch` for collection" logic with the "poll a collect URI" logic, they can run more-or-less independently.) Or, to put it another way: "acking" a batch removes it from the set of batches that can be returned from `current-batch`; there is no requirement for `current-batch` to continuously return the same batch until it is "acked".


> I think that makes sense, and I see how this is fault tolerant, but now it seems like GET {collect-uri} has a side effect of changing how the server responds to POST /collect, which seems a bit off. So I wonder if we should change it to POST {collect-uri}. But I don't think we should make that change here, as we need to consider more broadly what abstraction we want for delivering aggregate shares from aggregators to collectors.

Oh, that's a good point -- this is indeed a state change via a `GET`. While I think it is technically-speaking safe/non-problematic in this particular instance, it's worthy of being cleaned up to follow best practices. The most direct way would be to switch to `POST`, as you suggest.

tgeoghegan:
What are the semantics of `current-batch` for VDAFs where the aggregation parameter varies? Concretely:

Let's suppose a leader is running a chunky DAP task. It's continuously receiving reports, and assigning them to batch IDs as it goes. At some point in time, it has accumulated enough reports to fill three batches, which it assigns batch IDs 1, 2 and 3.

A collector comes along and makes a collect req with some aggregation parameter `a_1` and a `current-batch` query. I think it makes sense that the leader would run aggregation jobs for batch ID 1 with agg param `a_1`. Next, the collector sends another collect req with the same aggregation parameter and `current-batch`. Having already served up results for batch ID 1, the leader decides that batch ID 2 is now current, and runs aggregation jobs over it with `a_1`.

Now, suppose the collector sends a collect req with `current-batch` and aggregation parameter `a_2`. What batch does the leader prepare?

Does it prepare batch 3, on the premise that it's now current? Or should the change in aggregation parameter reset the cursor position back to batch 1? Or do we expect the leader to maintain a different cursor in the batch stream for each aggregation parameter?

I understand that `by-batch-id` is the mechanism by which collectors are intended to query a batch multiple times, but it's still possible for a collector to do nothing but `current-batch` with varying agg params. I suppose any collector that does this is accepting that it's at the mercy of whatever leader implementation it's talking to?

PR #326, opened 2022-09-09
confid_id vs id for HpkeConfigId in HpkeConfig

seanturner:
Not sure about this one.


tgeoghegan:
I prefer `id` in spots like this because if you imagine referring to it as `HpkeConfig.config_id`, then `config` is repeated unnecessarily, whereas `HpkeConfig.id` is concise and I hope unambiguous. However the names of these fields doesn't ever go on the wire so i don't object to changing them if others think this is more clear.

cjpatton:
I agree with Tim, I think we should leave as-is.

PR #327, opened 2022-09-09
Fixing FixedSizeQuerry presentation syntax

seanturner:
I think this is right based on RFC 8446 always have a prefix like "FixedSizeQuery." when the choice comes after a type.


cjpatton (commented):
No objection, but I was under the impression we were supposed to skip this.

wangshan:
I can't find explicitly rules regarding this in RFC8446 section3, but I think this is the right change even if just to keep it consistent with the other occurrences of such variant, which is `select (PrepareStep.prepare_step_result)`

cjpatton (approved):

PR #333, opened 2022-09-09
Clarify handling of replays in "Upload Requests"

### LINKED ISSUE #191: reconsider empty body requirement for status 200 responses
Section 4.2.2:

"Clients SHOULD NOT upload the same measurement value in more than one report if the leader responds with status 200 and an empty body.”"

Is a non-empty body understood to signal an error condition based on some other context?

What if the client got status 200 from the leader but a non-empty body? Should the client retry as this text suggests?

I don't know enough about HTTP semantics to know for sure, but it seems like specifying an empty body here is unnecessary. It seems fine to specify that the leader should *send* an empty body for 200, but maybe be less strict on the consumption side and drop the empty body requirement?

tgeoghegan:
We didn't have any text that actually instructed leaders to reject uploads deemed to be replays based on their nonce. The text in {{anti-replay}} says leaders should do this, but then refers to the uploads section, which doesn't contain any text about checking nonces or what the leader should do.

We now define a new problem document type "reportRejected", which replaces the previous "reportTooLate", previously used in the case where the leader detects a report pertaining to an already-collected batch. The rationale for using a less specific error type for these two cases is twofold:

  - In these cases, we don't expect the client to do anything but give up on the report, so nothing more specific than a generic rejection is needed;
  - If we did provide more specific error codes, we'd be giving attackers an oracle for what nonces have already been uploaded or what reports have already been collected.

While I was in this text, I also:

  - resolved #191.
  - clarified that the client should do if it gets "reportTooEarly". If the client shouldn't do anything special in that situation, then I argue the error should be the generic "reportRejected".
  - downgraded a MUST to a SHOULD in the case where the client gets "outdatedConfig" twice. There's no way the protocol can meaningfully enforce that MUST, and if the client continues to upload reports referencing an outdated or non-existent HPKE config, the leader can just continue to ignore the reports.

Resolves #191

### DIFF
--- a/draft-ietf-ppm-dap.md
+++ b/draft-ietf-ppm-dap.md
@@ -442,7 +442,7 @@ in the "type" field (within the DAP URN namespace
 | unrecognizedTask           | An endpoint received a message with an unknown task ID. |
 | unrecognizedAggregationJob | An endpoint received a message with an unknown aggregation job ID. |
 | outdatedConfig             | The message was generated using an outdated configuration. |
-| reportTooLate              | Report could not be processed because it arrived too late. |
+| reportRejected             | Report could not be processed for an unspecified reason. |
 | reportTooEarly             | Report could not be processed because its timestamp is too far in the future. |
 | batchInvalid               | A collect or aggregate-share request was made with invalid batch parameters. |
 | invalidBatchSize           | There are an invalid number of reports in the batch. |
@@ -867,23 +867,31 @@ with error of type 'outdatedConfig'. If the leader supports multiple HPKE
 configurations, it can use trial decryption with each configuration to determine
 if requests match a known HPKE configuration. When clients receive an 'outdatedConfig'
 error, they SHOULD invalidate any cached aggregator `HpkeConfigList` and retry with
-a freshly generated Report. If this retried report does not succeed, clients MUST
-abort and discontinue retrying.
+a freshly generated Report. If this retried report does not succeed, clients
+SHOULD abort and discontinue retrying.
+
+If a report's ID matches that of a previously uploaded report, the leader
+MUST ignore it and the leader MAY alert the client with error `reportRejected`.
+See the implementation note in {{input-share-validation}}.
 
 The Leader MUST ignore any report pertaining to a batch that has already been
-collected. (See {{input-share-validation}} for details.) Otherwise,
-comparing the aggregate result to the previous aggregate result may result in a
-privacy violation. (Note that the Helpers enforce this as well.) The Leader MAY
-ignore any reports whose timestamp is past the task's `task_expiration`. When it
-does so, the leader SHOULD abort the upload protocol and alert the client with
-error "reportTooLate". Client MAY choose to opt out of the task if its own clock
-has passed `task_expiration`.
+collected (see {{input-share-validation}} for details). Otherwise, comparing the
+aggregate result to the previous aggregate result may result in a privacy
+violation. Note that this is enforced by all Aggegators, not just the leader. In
+addition, the leader MAY abort the upload protocol and alert the client with
+error `reportRejected`.
+
+The Leader MAY ignore any reports whose timestamp is past the task's
+`task_expiration`. When it does so, the leader SHOULD abort the upload protocol
+and alert the client with error `reportRejected`. Client MAY choose to opt out
+of the task if its own clock has passed `task_expiration`.
 
 Leaders can buffer reports while waiting to aggregate them. The leader SHOULD
 NOT accept reports whose timestamps are too far in the future. Implementors MAY
 provide for some small leeway, usually no more than a few minutes, to account
 for clock skew. If the leader rejects a report for this reason, it SHOULD abort
-     the upload protocol and alert the client with error "reportTooEarly".
+the upload protocol and alert the client with error `reportTooEarly`. In this
+situation, clients MAY re-upload the report later on.
 
 If the Report contains an unrecognized extension, or of two extensions have the
 same ExtensionType, then the Leader MAY abort the upload request with error


tgeoghegan:
I don't think we should hold up draft-02 for this, unless it takes long enough for the fixed-task stuff to land anyway.

divergentdave (changes requested):
The term "replay" is overloaded between two contexts, clients uploading reports to the leader and the leader sending report shares to the helper. I think this change is mixing up the two.

In the context of `/upload` requests, leaders need to at least ignore reports with repeated nonces for the following reasons.

- To protect the privacy of misbehaving clients that choose nonces with insufficient entropy
- To ensure the rest of the protocol can be executed consistently while indexing reports, report shares, preparation message shares, etc. by report nonce
- To make `/upload` an idempotent endpoint, so that clients can safely retry uploads in the event of network-level issues

At the time of `/upload`, the "anti-replay" logic we require is just ensuring that the leader stores and uses at most one report with any given nonce. (Whether this is resolved during the upload request or after, so long as aggregation only ever uses one report with a given nonce) There is also a requirement that subsequent uploads that fall within previously completed collect requests be ignored in future collect requests, but this seems subjectively more distant from the umbrella of "anti-replay". (In fact, it could just as well be enforced during aggregation or collection, by not including it in aggregation jobs, validating AggregateShareReqs against past requests, etc.)

This differs from the anti-replay requirements in the context of the collect flow. In that case aggregators need to account for how many aggregations a report (or its nonce) has been included in, and enforce the `max_batch_lifetime` condition before servicing either a CollectReq or AggregateShareReq. This is the subject of section 4.5.7.

When we add text covering what the leader should do when it receives a new report with a nonce it has seen previously, I suggest that the leader should ignore the report and return 200 OK, on the theory that this will occur most commonly when retrying HTTP requests. If we do define an error type for this situation, we should not require that the leader reply with it upon repeated nonces in uploads, because that would require increased coordination between concurrent upload requests.

> If we did provide more specific error codes, we'd be giving attackers an oracle for what nonces have already been uploaded or what reports have already been collected.

I think combining the two error conditions in this manner is only a minor impediment for an attacker, as they would instead have an oracle for the disjunction of the two error conditions. Instead, the unpredictability of the random component of the nonce (from an honest client) is what will prevent an attacker from gaining any advantage from `/upload` error codes. With that said, I support combining these different conditions into `reportRejected` for the first reason, because we don't expect the client to make any remedial change either way.

cjpatton (approved):
Modulo changing the {{anti-replay}} reference to {{input-share-batch-validation}}, I think this change is in good enough shape to land. I think changing "reportTooLate" to "reportRejected" is strictly and improvement, since it better reflects the feedback the Leader gets from the Helper during the aggregation flow. We might consider adding more signaling, but we should do so in a later PR.

I want to remind folks of one of our design goals for the upload flow. Checking whether a report would be early rejected (due to decryption failure, repeated nonce, collected batch, etc.) requires interacting with backend storage. Thus the Leader is permitted to perform these checks offline, just before running an aggregation job. That way this (potentially expensive) work is not on the hot path with the client. Of course, the Leader is free to provide this feedback if it wishes. (In fact the spec even says it SHOULD.)

cjpatton (changes requested):
Merge conflicts

draft-ietf-ppm-dap.md
```diff
@@ -767,30 +767,35 @@ The order of the encrypted input shares appear MUST match the order of the
...
-retrying.
+HTTP status code 400 Bad Request and an error of type 'outdatedConfig'. Clients
+SHOULD invalidate any cached aggregator `HpkeConfig` and retry with a freshly
+generated Report. If this retried report does not succeed, clients SHOULD abort
+and discontinue retrying.
 
-The Leader MUST ignore any report pertaining to a batch that has already been
+If the leader determines that a report is a replay based on the conditions in
+{{anti-replay}}, the report MUST be ignored and the leaader SHOULD alert the
+client with error "reportRejected".
```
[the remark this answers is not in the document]
tgeoghegan: Reviewing Chris' #382 reminded me to revisit this change. David's point about a circular reference is a good one. What I want to do is repurpose {{anti-replay}} to discuss the criteria for determining if a report is a replay, but then leave it up to either {{upload-flow}} or {{early-input-share-validation}} to specify what should be done when a replay is detected. So I'll do that after #382 lands, since that PR is changing the text of {{anti-replay}}.

chris-wood:
@tgeoghegan can you please rebase this?

tgeoghegan:
I've updated this change. The section on handling uploads no longer references {{input-share-validation}}, per @divergentdave's argument that leader checks during upload and aggregator checks during aggregation are not the same thing. Instead we plainly state in the text of "Upload Requests" that leaders MUST ignore replays and MAY alert the client.

divergentdave (approved):

draft-ietf-ppm-dap.md
```diff
@@ -867,23 +867,29 @@ with error of type 'outdatedConfig'. If the leader supports multiple HPKE
 configurations, it can use trial decryption with each configuration to determine
 if requests match a known HPKE configuration. When clients receive an 'outdatedConfig'
 error, they SHOULD invalidate any cached aggregator `HpkeConfigList` and retry with
-a freshly generated Report. If this retried report does not succeed, clients MUST
-abort and discontinue retrying.
+a freshly generated Report. If this retried report does not succeed, clients
+SHOULD abort and discontinue retrying.
+
+If a report's metadata matches that of a previously uploaded report, the leader
```
cjpatton, line 873: I think this check is too stringent. For anti-replay, we require tracking the report IDs at a minimum.

draft-ietf-ppm-dap.md
```diff
@@ -867,23 +867,29 @@ with error of type 'outdatedConfig'. If the leader supports multiple HPKE
...
-collected. (See {{input-share-validation}} for details.) Otherwise,
-comparing the aggregate result to the previous aggregate result may result in a
-privacy violation. (Note that the Helpers enforce this as well.) The Leader MAY
-ignore any reports whose timestamp is past the task's `task_expiration`. When it
-does so, the leader SHOULD abort the upload protocol and alert the client with
-error "reportTooLate". Client MAY choose to opt out of the task if its own clock
-has passed `task_expiration`.
+collected (see {{input-share-validation}} for details). Otherwise, comparing the
+aggregate result to the previous aggregate result may result in a privacy
+violation. Note that the Helpers enforce this as well. In addition, the leader
```
cjpatton, line 879: In case we end up hard-coding two aggregators in the future.
```suggestion
violation. Note that this is enforced by all Aggregators, not just the Leader. In addition, the leader
```

draft-ietf-ppm-dap.md
tgeoghegan: I changed the wording to discuss report ID. I think this needs the same nuance about sharding as the corresponding item in {{input-share-validation}} but I'm not sure how to effectively reference that here. We could just copy paste in the "implementation note".

cjpatton: Or just point to that section?

draft-ietf-ppm-dap.md
```diff
@@ -867,23 +867,32 @@ with error of type 'outdatedConfig'. If the leader supports multiple HPKE
...
 error, they SHOULD invalidate any cached aggregator `HpkeConfigList` and retry with
-a freshly generated Report. If this retried report does not succeed, clients MUST
-abort and discontinue retrying.
+a freshly generated Report. If this retried report does not succeed, clients
+SHOULD abort and discontinue retrying.
+
+If a report's ID matches that of a previously uploaded report, the leader
+MUST ignore it and the leader MAY alert the client with error `reportRejected`.
+[TODO: should we reference the implementation note from
+{{input-share-validation}} here?]
```
cjpatton, line 876: yes :)

cjpatton (approved):

PR #335, opened 2022-09-12
Align agg_id with role.

### LINKED ISSUE #328: s4.4.1.5: agg_id
Is there any reason that agg_id values can't match the rest of the document? In agg_id, leader is 0x00 and helper is 0x01, but everywhere else it's 0x01 and 0x02 for leader and helper, respectively.

simon-friedberger:
Resolve #328


chris-wood (approved):

draft-ietf-ppm-dap.md
```diff
@@ -1161,7 +1161,8 @@ out = VDAF.prep_next(prep_state, None)
 ~~~
 
 `vdaf_verify_key` is the VDAF verification key shared by the aggregators;
-`agg_id` is the aggregator ID (`0x00` for the Leader and `0x01` for the helper);
+`agg_id` is the aggregator ID, equal to the role in {{protocol-definition}}
+(`0x02` for the Leader and `0x03` for the helper);
```
divergentdave, line 1165: This would cause problems given the current text of draft-irtf-cfrg-vdaf. There, `agg_id` is expected to be less than the total number of shares. The pseudocode also hinges some decisions on whether `agg_id == 0`. To simplify this, we either need to subtract two here, renumber the Role enum above, or coordinate some larger rearrangement with a future VDAF draft.

cjpatton, line 1165: This is a breaking change for VDAF. We could either:
1. Update the agg ID semantics in VDAF and cut a draft
2. Align server role with agg ID

Option (2.) might look like this:

```suggestion
`agg_id` is the aggregator ID, equal to `server_role - 0x02`, where `server_role` is the aggregators' role in {{protocol-definition}};
```
Alternatively, we can re-assign the server roles to align with the VDAF draft.

cjpatton (changes requested):

draft-ietf-ppm-dap.md
cjpatton: ha, just saw this :)

chris-wood: It seems bad that an arbitrary change in DAP breaks VDAF. Let's do whatever is needed to decouple these things so that the aggregator ID can be freely assigned.

cjpatton: Duplicates @divergentdave's comment.

cjpatton: In general the computations of each of the aggregator's aren't identical. For example, for Prio3 the first aggregator's (agg_id==0) is larger than the other's. On the other hand, for Poplar1 they're all roughly the same size. In general there is no reason to think that there will always be a notion of a "leader share".

Do you have thoughts about how to clean this up? David suggested one fix. An alternative would be to re-label the aggregators here.

simon-friedberger: Oh, indeed it is being used basically as an array index to identify each aggregator's data.

I have no strong opinion.
This was based on the desire to expand to more aggregators later. We could also use 0 = leader, 1 = helper (and 2, 3, .. if necessary) 254 = collector, 255 = client and change role accordingly?

Or just leave everything as is, though I was equally surprised as Sean about the discrepancy.

cjpatton: I like you're suggestion for relabeling Simon, let's go with that and see what it looks like. Note that you'll have to propagate the changes to wherever we HPKE encrypt or decrypt something.

simon-friedberger: ISTM @chris-wood is advocating for just leaving it as is. We can but we somehow need to specify how to convert the arbitrary IDs from the DAP spec to continues numbers with leader=0 for the VDAF spec.

tgeoghegan:
I don't have a helpful suggestion here, but this reminds me of this issue in VDAF that I wrote some time ago: https://github.com/cfrg/draft-irtf-cfrg-vdaf/issues/40

I haven't thought it all the way through, but if VDAF had its own notion of Leader and Helper, then DAP and VDAF could independently maintain numberings of roles and I think we'd achieve the property Chris W. wants.

cjpatton:
> I don't have a helpful suggestion here, but this reminds me of this issue in VDAF that I wrote some time ago: [cfrg/draft-irtf-cfrg-vdaf#40](https://github.com/cfrg/draft-irtf-cfrg-vdaf/issues/40)
> 
> I haven't thought it all the way through, but if VDAF had its own notion of Leader and Helper, then DAP and VDAF could independently maintain numberings of roles and I think we'd achieve the property Chris W. wants.

This works for SHARES==2, but for SHARES>2 it breaks down because the agg_id is used for domain separation in the Fiat-Shamir step for Prio3. We can do whatever we want in DAP, but in Prio3, agg_id is going to have to be distinct for each aggregator.

cjpatton:
Thanks for looking at this @simon-friedberger! Because this breaks compatibility with VDAF, I'm going to decline this PR. There is still plenty of time to address #328 before we cut DAP-02, so feel free to send a PR.

PR #336, opened 2022-09-12
Add report time to `struct PrepareStep`

tgeoghegan:
PR #297 (which added fixed chunk tasks) broke the report nonce apart into a timestamp and a nonce. This was not reflected in the `PrepareStep` structures in an `AggregateInitResp` or `AggregateContinueReq`.

### DIFF
--- a/draft-ietf-ppm-dap.md
+++ b/draft-ietf-ppm-dap.md
@@ -1054,6 +1054,7 @@ enum {
 } PrepareStepResult;
 
 struct {
+  Time time;
   Nonce nonce;
   PrepareStepResult prepare_step_result;
   select (PrepareStep.prepare_step_result) {


tgeoghegan:
Elsewhere we use `ReportMetadata` to get a single handle to a report's nonce and timestamp, but since that structure includes the optional extensions, it feels incorrect to include it in `PrepareStep`.

cjpatton (commented):
Good catch, this might be a regression. I didn't end up needing this in my implementation. Is this useful for Janus?

Unless we have a compelling reason to add the times tamp back, I'd suggest we axe it. To me it's conceptually simpler to rely strictly on the nonce for indexing the reports.

tgeoghegan:
> Unless we have a compelling reason to add the times tamp back, I'd suggest we axe it. To me it's conceptually simpler to rely strictly on the nonce for indexing the reports.

That's an interesting proposal. Certainly a 16 byte random nonce ought to be sufficient to uniquely identify a nonce. Let me see how this plays out as I wire up some of this in Janus. In the meantime, I'll make this PR a draft.

tgeoghegan:
Indeed, I was able to implement everything without having the report time in the PrepareStep, because a nonce is enough to uniquely identify a report. Closing this -- it's easy enough to make a new PR if we do decide we need this.

PR #338, opened 2022-09-13
Allow per-report artificats to be dropped before task expiration

### LINKED ISSUE #180: Aggregators required to store nonce sets indefinitely

cjpatton:
Closes #180.

Clarify that `report-dropped` is used when it's impossible to determine if an input share is valid. One situation we need this for is when an Aggregator drops some anti-replay state in order to reduce storage costs. This should result in data loss in the worst case, not a loss of privacy.

### DIFF
--- a/draft-ietf-ppm-dap.md
+++ b/draft-ietf-ppm-dap.md
@@ -776,7 +776,7 @@ this retried report does not succeed, clients MUST abort and discontinue
 retrying.
 
 The Leader MUST ignore any report pertaining to a batch that has already been
-collected. (See {{input-share-batch-validation}} for details.) Otherwise,
+collected. (See {{early-input-share-validation}} for details.) Otherwise,
 comparing the aggregate result to the previous aggregate result may result in a
 privacy violation. (Note that the Helpers enforce this as well.) The Leader MAY
 ignore any reports whose timestamp is past the task's `task_expiration`. When it does so,
@@ -952,7 +952,7 @@ shares as follows:
    corresponding DAP task. It is RECOMMENDED that this be set to a random string output
    by a cryptographically secure pseudorandom number generator.
 1. Decrypt the input share for each report share as described in {{input-share-decryption}}.
-1. Check that the resulting input share is valid as described in {{input-share-batch-validation}}.
+1. Check that the resulting input share is valid as described in {{early-input-share-validation}}.
 1. Initialize VDAF preparation as described in {{input-share-prep}}.
 
 If any step yields an invalid report share, the leader removes the report share from
@@ -1037,10 +1037,10 @@ error "unrecognizedMessage". If this check succeeds, the helper then attempts to
 input share in `AggregateInitializeReq.report_shares` as follows:
 
 1. Decrypt the input share for each report share as described in {{input-share-decryption}}.
-1. Check that the resulting input share is valid as described in {{input-share-batch-validation}}.
+1. Check that the resulting input share is valid as described in {{early-input-share-validation}}.
 1. Initialize VDAF preparation and initial outputs as described in {{input-share-prep}}.
 
-[[OPEN ISSUE: consider moving the helper nonce check into #input-share-batch-validation]]
+[[OPEN ISSUE: consider moving the helper nonce check into #early-input-share-validation]]
 
 Once the helper has processed each valid report share in `AggregateInitializeReq.report_shares`, the
 helper then creates an AggregateInitializeResp message to complete its initialization. This message is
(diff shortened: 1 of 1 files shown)


draft-ietf-ppm-dap.md
```diff
@@ -1804,12 +1809,18 @@ In general, the aggregators are required to keep state for all valid reports for
...
-{{batch-validation}}.)
+reports themselves. For schemes like Prio3 in which reports are aggregated once,
+each aggregator only needs to store its aggregate share for each batch.
+
+However, Aggregators are also required to implement several per-report checks
+that require retaining a number of data artificts. For example, to detect replay
+attacks, it is necessary for each Aggregator to retain the set of nonces of
+reports that have been aggregated for the task so far. Depending on the task
+lifetime and report upload rate, this can result in high storage costs. To
+elliviate this burden, DAP allows Aggregators to drop this state as needed, so
```
branlwyd, line 1820:
```suggestion
alleviate this burden, DAP allows Aggregators to drop this state as needed, so
```

branlwyd (approved):

draft-ietf-ppm-dap.md
```diff
@@ -1109,32 +1109,37 @@ aggregator (`0x02` for the leader and `0x03` for the helper). If decryption
...
+   a previous aggregate request and is therefore a replay.
+1. Depending on the query type for the task, additional checks may be applicable:
     * For fixed-size tasks, the Aggregators need to ensure that each batch is
       roughly the same size. If the number of reports aggregated for the current
       batch exceeds the maximum batch size (per the task configuration), the
       Aggregator MAY mark the input share as invalid with the error
       "batch-saturated". Note that this behavior is not strictly enforced here
       but during the collect sub-protocol. (See {{batch-validation}}.)
-If both checks succeed, the input share is not marked as invalid.
+1. Finally, if an Aggregator cannot determine if an input share is valid. it
```
tgeoghegan, line 1136:
```suggestion
1. Finally, if an Aggregator cannot determine if an input share is valid, it
```

draft-ietf-ppm-dap.md
```diff
@@ -1804,12 +1809,18 @@ In general, the aggregators are required to keep state for all valid reports for
...
+reports themselves. For schemes like Prio3 in which reports are aggregated once,
+each aggregator only needs to store its aggregate share for each batch.
+
+However, Aggregators are also required to implement several per-report checks
+that require retaining a number of data artificts. For example, to detect replay
+attacks, it is necessary for each Aggregator to retain the set of nonces of
+reports that have been aggregated for the task so far. Depending on the task
+lifetime and report upload rate, this can result in high storage costs. To
+elliviate this burden, DAP allows Aggregators to drop this state as needed, so
+long as reports are droppped properly as described in
```
tgeoghegan, line 1821:
```suggestion
long as reports are dropped properly as described in
```

draft-ietf-ppm-dap.md
```diff
@@ -1109,32 +1109,37 @@ aggregator (`0x02` for the leader and `0x03` for the helper). If decryption
...
       roughly the same size. If the number of reports aggregated for the current
       batch exceeds the maximum batch size (per the task configuration), the
       Aggregator MAY mark the input share as invalid with the error
       "batch-saturated". Note that this behavior is not strictly enforced here
       but during the collect sub-protocol. (See {{batch-validation}}.)
-If both checks succeed, the input share is not marked as invalid.
+1. Finally, if an Aggregator cannot determine if an input share is valid. it
+   MUST mark the input share as invalid with error `report-dropped`. This
+   situation arises if, for example, the Aggregator has evicted from long-term
+   storage the state required to perform the check. (See
```
tgeoghegan, line 1139: This would also happen if for instance a helper received an `AggregateContinueReq` that contained a `PrepareStep` referencing a report that the helper has never seen before, right? If so, maybe worth noting as much in the "Helper Continuation" section that discusses handling of `AggregateContinueReq`.

tgeoghegan (approved):

draft-ietf-ppm-dap.md
cjpatton: That's not really the situation I'm thinking of. I suspect what I'm imagining will be much more common: Suppose a task runs for year. An Aggregator may choose to evict all report metadata that is older than a week; this text is meant to cover the situation where a report arrives over a week late.

PR #341, opened 2022-09-14
Ensure all protocol messages are self-describing

### LINKED ISSUE #340: Determining query type before parsing a request
The query type is determined by the task config, which in turn is determined by the task ID. The task ID is encoded by the message, so it is necessary to parse the task ID before the Aggregator knows how to parse it.

PR https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/pull/297 introduced the following problem: If an Aggregator implements multiple query types ("time-interval" and "fixed-size", say), then a message's format depends on the query type for the task. This means it is necessary to parse the task ID before the rest of the message.

This is perfectly implementable, but it has two downsides:
1. This likely breaks abstraction boundaries in current implementations and will require refactoring. (For example, [here is how reports are handled in Daphne](https://github.com/cloudflare/daphne/blob/main/daphne/src/roles.rs#L251-L255). Notice that the message is parsed in one step.)
2. Going forward, we'll need to make sure that the task ID always appears first in any DAP message so that it can be easily parsed. This is already true today, but we might want to add some text to Operational Considerations that spells this requirement out so that we minimize the chance of regression.

Here's my question for folks: Do we want to try and do something about this? I can think of a couple of solutions:
1. Add the query type byte to messages whose format depends on this. One downside is that it creates a new error case for the Aggregator to check, namely, that the query type indicated in the message matches the query type of the task. (In fact, we tried to avoid this in #297.)
2. Move the task ID out of the message payload and into the URL query string, or perhaps the HTTP header. The only downside to this is that it complicates request authentication for deployments that want to sign the payload and bind it to the task config.  (This may be desirable from a security perspective.) One possible upshot is that it might put us in a better position to address #278. (@tgeoghegan please weigh in here!)

cjpatton:
Closes #340.

Currently parsing AggreateInitializeReq, AggregateshareReq, CollectReq, and CollectResp requires the query type associated with the task. The query type is determined by the task, but it is necessary to parse the task ID from the message before the remainder can be parsed.

This change ensures that parsing of these messages does not require the query type.

### DIFF
--- a/draft-ietf-ppm-dap.md
+++ b/draft-ietf-ppm-dap.md
@@ -978,7 +978,7 @@ struct {
   AggregationJobID job_id;
   opaque agg_param<0..2^16-1>;
   QueryType query_type;
-  select (AggreagteInitializeReq.query_type) {
+  select (AggregateInitializeReq.query_type) {
     case time-interval: Empty;
     case fixed-size: BatchId batch_id;
   };
@@ -1382,7 +1382,7 @@ This structure includes the following:
 
 * Information used to bind the aggregate result to the query. For fixed-size
   tasks, this includes the batch ID assigned to the batch by the Leader. The
-  indicatd query type MUST match the task's query type.
+  indicated query type MUST match the task's query type.
 
   [OPEN ISSUE: What should the Collector do if the query type doesn't match?]
 
@@ -1422,7 +1422,7 @@ a POST request to `[aggregator]/aggregate_share` with the following message:
 ~~~
 struct {
   QueryType query_type;
-  select (BachSelector.query_type) {
+  select (BatchSelector.query_type) {
     case time-interval: Interval batch_interval;
     case fixed-size: BatchId batch_id;
   };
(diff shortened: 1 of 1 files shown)


draft-ietf-ppm-dap.md
```diff
@@ -975,7 +977,8 @@ struct {
   TaskID task_id;
   AggregationJobID job_id;
   opaque agg_param<0..2^16-1>;
-  select (query_type) { /* determined by task configuration */
+  QueryType query_type;
+  select (AggreagteInitializeReq.query_type) {
```
tgeoghegan, line 981:
```suggestion
  select (AggregateInitializeReq.query_type) {
```

draft-ietf-ppm-dap.md
```diff
@@ -1372,8 +1380,11 @@ struct {
 
 This structure includes the following:
 
-* information used to bind the aggregate result to the query. For fixed-size
-  tasks, this includes the batch ID assigned to the batch by the Leader.
+* Information used to bind the aggregate result to the query. For fixed-size
+  tasks, this includes the batch ID assigned to the batch by the Leader. The
+  indicatd query type MUST match the task's query type.
```
tgeoghegan, line 1385:
```suggestion
  indicated query type MUST match the task's query type.
```

draft-ietf-ppm-dap.md
```diff
@@ -1410,7 +1421,8 @@ a POST request to `[aggregator]/aggregate_share` with the following message:
 
 ~~~
 struct {
-  select (query_type) { /* determined by task configuration */
+  QueryType query_type;
+  select (BachSelector.query_type) {
```
tgeoghegan, line 1425:
```suggestion
  select (BatchSelector.query_type) {
```

tgeoghegan (approved):
This accomplishes what it sets out to, modulo some typos.

chris-wood (approved):

PR #343, opened 2022-09-15
Editorial: Make paragraph formatting consistent

cjpatton:
* Justify text after bullets consistenly
* Make sure all lines break at 80 characters


tgeoghegan (approved):
LGTM, modulo conflicts. It'd be nice if we had a linter that blocked PRs until their formatting was right, since changes like this taint git history.

PR #352, opened 2022-09-20
Editorial: Define PartialBatchSelector

cjpatton:
The BatchSelector is the value agreed upon by the Aggregators during the collect flow. It is used to guide selection of a batch of reports for computing the aggregate result.

Conceptually, a PartialBatchSelector is "promoted" to a BatchSelector once the Collector's query is determined. The PartialBatchSelector is used during the aggregation flow to assign reports to batches. For example, for fixed size queries, the BatchID is selected during the Leader during the aggregation flow.

This is not a functional change, but I ended up wanting this abstraction in my implementation, so I thought I'd upstream it.

cc/ @wangshan.

### DIFF
--- a/draft-ietf-ppm-dap.md
+++ b/draft-ietf-ppm-dap.md
@@ -1014,14 +1014,18 @@ struct {
 } ReportShare;
 
 struct {
-  TaskID task_id;
-  AggregationJobID job_id;
-  opaque agg_param<0..2^16-1>;
   QueryType query_type;
-  select (AggregateInitializeReq.query_type) {
+  select (PartialBatchSelector.query_type) {
     case time_interval: Empty;
     case fixed_size: BatchID batch_id;
   };
+} PartialBatchSelector;
+
+struct {
+  TaskID task_id;
+  AggregationJobID job_id;
+  opaque agg_param<0..2^16-1>;
+  PartialBatchSelector part_batch_selector;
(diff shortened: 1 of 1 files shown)


wangshan:
I'm not sure the "promoted" behaviour is clear, it feels like just for reducing duplication? I'd also prefer to call it something closer to its true purpose.

cjpatton:
> I'm not sure the "promoted" behavior is clear, it feels like just for reducing duplication? I'd also prefer to call it something closer to its true purpose.

Deduplication is the point, yes -- this change is editorial.

Let me try another stab at explaining: A batch is determined by a "batch selector". In turn, the batch selector is determined by the Collector's query. But, there are parts of the protocol where we the batch needs to be "partially determined" prior to the Collector's query. (In particular, the Leader needs to be able to pick the batch ID before the Collector makes its query.)

branlwyd (approved):
I think that this change would slightly simplify some of Janus' implementation, so LGTM.

MichaelScaria:
LGTM. Talked to Chris P. offline about how `Query` and `BatchSelector` have the same structure, but learned that `Query` might change in the future with the current-batch PR.

MichaelScaria (approved):

wangshan (approved):

chris-wood (approved):

PR #353, opened 2022-09-20
Editorial: nit: Be consistent about tabs in pseudocode

cjpatton:


wangshan:
should we consider to be consistent with 4 spaces instead of 2?

cjpatton:
> should we consider to be consistent with 4 spaces instead of 2?

The majority of tabs are 2, so that's what I went with.

tgeoghegan (approved):

PR #356, opened 2022-09-21
Move and rename "Nonce" to "ReportID".

### LINKED ISSUE #355: Rename `Nonce` to `ReportID`
Currently, the DAP specification has a number of opaque values used to identify some DAP entity:
* `Nonce` (identifying a client report)
* `TaskID` (identifying a task)
* `AggregationJobID` (identifying an aggregation job)
* `BatchID` (identifying a batch)

`Nonce`'s name differs from the rest. I think this is because, historically, the nonce also included a time component. Now that the time component has been split out, I think we could rename `Nonce` to `ReportID` for clarity of purpose.

Thoughts?

(and I probably don't need to say this, but there is no reason to push this into DAP-02 -- it could easily land in DAP-03 instead.)

branlwyd:
Historically, the Nonce included a time component. Now that the time component is split out & Nonce is effectively an opaque identifier, ReportID is clearer & more closely matches similar identifiers (TaskID, AggregationJobID, BatchID).

Closes #355.

### DIFF
--- a/draft-ietf-ppm-dap.md
+++ b/draft-ietf-ppm-dap.md
@@ -1131,11 +1131,11 @@ The helper's response to the leader is an HTTP status code 200 OK whose body is
 the AggregateInitializeResp and media type is
 "message/dap-aggregate-initialize-resp".
 
-Upon receipt of a helper's AggregateInitializeResp message, the leader checks that the
-sequence of PrepareStep messages corresponds to the ReportShare sequence of the
-AggregateInitializeReq. If any message appears out of order, is missing, has an
-unrecognized report ID, or if two messages have the same report ID, then the leader
-MUST abort with error "unrecognizedMessage".
+Upon receipt of a helper's AggregateInitializeResp message, the leader checks
+that the sequence of PrepareStep messages corresponds to the ReportShare
+sequence of the AggregateInitializeReq. If any message appears out of order, is
+missing, has an unrecognized report ID, or if two messages have the same report
+ID, then the leader MUST abort with error "unrecognizedMessage".
 
 [[OPEN ISSUE: the leader behavior here is sort of bizarre -- to whom does it
 abort?]]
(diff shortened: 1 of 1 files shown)


branlwyd:
Note this contains a single wire-format change: `ReportMetadata`'s fields are reordered so that `ReportID` comes first, since it is the "primary" identifier. I am happy to revert this part of the PR if folks would prefer a purely-editorial change.

branlwyd:
Note this PR includes a single wire-format change: `ReportMetadata` had its fields reordered so that the `ReportID` came first, since the ID is the primary identifier. I am happy to revert this part of the PR if folks would prefer a purely-editorial PR.

MichaelScaria:
LGTM. Any thoughts on replacing `time` and `report_id` with a Nonce so then it has both?

MichaelScaria (approved):

branlwyd:
> LGTM. Any thoughts on replacing `time` and `report_id` with a Nonce so then it has both?

I'm pretty happy leaving `time` & `report_id` as separate fields -- for practical purposes, IMO `ReportMetadata` binds them together closely enough (along with the report extensions).

Conceptually, DAP is changing such that the primary identifier for a report is its opaque ID (with this PR, called `ReportID`). Previously, the primary identifier for a report was `(report timestamp, opaque report ID)`. The technical aspects of this change were made by cjpatton in his changes to introduce `fixed-size` tasks. After that point, the nonce used for cryptographic operations no longer includes the timestamp -- it's only the opaque report ID.

Thus, if we do decide to introduce a struct that is effectively `(timestamp, opaque report ID)`, I'd prefer if it was called something other than `Nonce` -- while we do expect that this value is a "value to be used only once", it's no longer used as the nonce in any cryptographic operations. And I'd argue there's no better justification to call the report ID (or `(timestamp, report ID)`) a nonce than, say, the aggregation job ID/batch ID/task ID.

cjpatton (approved):
Yeah I think this is a big win :) good idea @branlwyd

Issue #361, opened 2022-09-26
Specify unrecognized task ID error explicitly in all APIs


wangshan:
The current protocol text doesn't specify what participants should do when a task ID is not recognized, there are two places unrecognized task are mentioned:

1. In the [Errors](https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/blob/main/draft-ietf-ppm-dap.md#errors) table, it has: 

> unrecognizedTask | An endpoint received a message with an unknown task ID.

2. [`hpke_config` request](https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/blob/main/draft-ietf-ppm-dap.md#hpke-configuration-request-hpke-config) mentions "404 Not Found and an error of type unrecognizedTask", which seems to conflict with most implementation's choice of error code 400 (@cjpatton )

The text should specify error 400 with type `unrecognizedTask` is used in every API that receives unrecognized task ID.

cjpatton:
This would be useful I think. Would you mind sending a PR, @wangshan?

wangshan:
@cjpatton will do, create this issue to remind myself.

wangshan:
PR #363 merged

PR #363, opened 2022-09-27
Specify unrecognized task ID error explicitly in all APIs

wangshan:
Addresses issue https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/issues/361

### DIFF
--- a/draft-ietf-ppm-dap.md
+++ b/draft-ietf-ppm-dap.md
@@ -690,7 +690,7 @@ specify a query parameter `task_id` when sending an HTTP GET request to
 obtained from the task parameters, encoded in Base 64 with URL and filename safe
 alphabet with no padding, as specified in sections 5 and 3.2 of {{!RFC4648}}. If
 the aggregator does not recognize the task ID, then it responds with HTTP status
-code 404 Not Found and an error of type `unrecognizedTask`.
+code 400 Bad Request and an error of type `unrecognizedTask`.
 
 An aggregator is free to use different HPKE configurations for each task with
 which it is configured. If the task ID is missing from a client's request, the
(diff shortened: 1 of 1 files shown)


cjpatton (commented):
A couple high level comments:
1. The status code is implied by the definition of "abort" in {{errors}}, so this doesn't need to be mentioned.
2. It's always best to use the keyword "MUST" when prescribing behavior that is security critical.

Based on this, I think the best verbiage is:

> MUST abort with error "unrecognizedTask"

tgeoghegan:
>     1. The status code is implied by the definition of "abort" in {{errors}}, so this doesn't need to be mentioned.
I don't think that's true. [3.2 says](https://ietf-wg-ppm.github.io/draft-ietf-ppm-dap/draft-ietf-ppm-dap.html#section-3.2):
>DAP servers can return responses with an HTTP error response code (4XX or 5XX). For example, if the client submits a request using a method not allowed in this document, then the server MAY return HTTP status code 405 Method Not Allowed.

So maybe we do need to specify that the status should be 400 (client error) and not 500 (server error). FWIW, [ACME](https://datatracker.ietf.org/doc/html/rfc8555) has several places where it dictates HTTP 400 inline in the text, so I think it's OK for us to do that. 

>     2. It's always best to use the keyword "MUST" when prescribing behavior that is security critical.

I agree we should be consistent about using MUST throughout the document. Curiously, ACME has some examples where the 400 error response is a MUST, but the 200 success response has no all-caps applied to it. [From ACME 7.6](https://datatracker.ietf.org/doc/html/rfc8555#section-7.6):

> If the revocation succeeds, the server responds with status code 200
>   (OK).  If the revocation fails, the server returns an error.  For
>   example, if the certificate has already been revoked, the server
>   returns an error response with status code 400 (Bad Request) and type
>   "urn:ietf:params:acme:error:alreadyRevoked".

wangshan:
@tgeoghegan @cjpatton thanks for the comments, I've specified 400 in the errors section and shortened the words elsewhere.

cjpatton (approved):

Issue #342, opened 2022-09-15
Fixed-size queries: Discovery of batch IDs is unspecified


cjpatton:
For tasks that support fixed-size queries, the Collector indicates in its `CollectReq` the ID of the batch it wishes to aggregate. At the same time, the Leader is responsible for assigning ID, and it is free to chose any `BatchID` it wishes. (It may, for example, choose them at random.) This is to allow the Leader to opportunistically aggregate reports as they arrive, rather than wait for a request from the Collector.

The current draft doesn't specify a means by which the Collector learns these batch IDs. This is instead left up to the deployment, which presents interop issues for Collectors and Leaders that are organizationally separated.

We explored a couple possible solutions:
1. Require the batch IDs to be predicable (#301), for example, by requiring the Leader to use a counter for picking the next batch ID. The concern here is that this potentially complicates implementation of the Leader.
2. "next-batch" functionality: Initially (#297) we had spelled out a method by which the Leader could request the "next batch" and the Leader would respond with a `BatchID`. This solves the problem, but creates another: It requires the Collector to remember batch IDs that it needs to query again.
3. "current-batch" functionality (#308): Like "next-batch", except that it adds fault tolerance in case the Collector loses a recently used batch ID.

bhalleycf:
[...]

PR #367, opened 2022-10-13
DO NOT MERGE sketch of resource oriented DAP API

tgeoghegan:
This is a sketch of a resource oriented API for DAP. It's presented
separately from the main body of protocol text because before taking on
the work of integrating it in, I want to begin a discussion about this
proposal, leading up to a presentation at IETF 115.

The objectives are to refactor the DAP API into a resource-oriented API
more in line with conventional HTTP semantics and REST. In particular, I
am trying to pay close attention to idempotence and ensuring that
protocol participants can recover gracefully from network messages being
lost.

Designing the API around resources identified by a URI also simplifies
message handling: servers no longer need to parse the first part of a
message to get a task ID to then figure out what the rest of the message
contains, because now things like a task ID are in a resource's path.

We also should be able to write less protocol text, because if all of
our resources are boring HTTP nouns, then the usual semantics for things
like HTTP 202 Accepted, 3xx redirects or 404 Not Found apply.

I do _not_ introduce an ACME style API directory, because I'm not
convinced that it is appropriate to introduce an extra level of
indirection for DAP API requests, which are going to be more performance
sensitive than ACME.

I also deliberately avoid ACME's POST-as-GET requests, because our
experience at Let's Encrypt has been that those make it hard to cache
responses, because CDNs don't expect POST results to be cacheable.

I've put this up on GitHub as a PR so that it's out in public and folks
can leave comments inline reasonably gracefully, although I don't intend
to ever merge this.

Relevant to #278

### DIFF
--- a/api-resources.md
+++ b/api-resources.md
@@ -2,7 +2,6 @@
 
 DAP is structured as an HTTP-based application with the following resources:
 
-- (TODO) Tasks
 - HPKE configurations
 - Reports
 - Aggregation jobs
@@ -22,10 +21,6 @@ We use HTTP `PUT` requests for idempotent creation of resources. Server response
 Finally, a quick reminder of some HTTP semantics: `GET` and `PUT` are idempotent (i.e., the same request may be made multiple times without changing state in the server more than once, and the response will be the same each time). `POST` is not idempotent. `GET` requests may not have bodies.
 
 We care a lot about idempotence because we need to account for cases where a client makes a request but never sees the response, and thus needs a way to recover from this state of uncertainty.
-
-### Tasks
-
-TODO: do we want a task resource that we can GET and PUT in the protocol? Does this collide with https://wangshan.github.io/draft-wang-ppm-dap-taskprov/draft-wang-ppm-dap-taskprov.html or https://github.com/divergentdave/draft-dcook-ppm-dap-interop-test-design?
 
 ### HPKE configurations
 
--- a/api-resources.md
+++ b/api-resources.md
@@ -79,7 +79,7 @@ struct {
 
 ##### PUT
 
-Idempotent upload of a report to an aggregator. The request body is a `struct Report`.
+Idempotent upload of a report to an aggregator. The request body is a `struct Report`. Once uploaded, a report is immutable, meaning that subsequent `PUT` requests to a particular report resource that vary fields of `struct Report` MUST be rejected by the leader.
 
 This is analogous to DAP-02's `POST /upload`.
 
@@ -154,6 +154,8 @@ struct {
 ```
 
 If successful, the helper's response is a `struct AggregateJob` where `prepare_steps` contains the helper's first-round VDAF prepare messages for the report shares in `AggregateJobPutReq`, or errors if `vdaf.prepare` failed for any report.
+
+Once created, an aggregation job resource cannot be mutated with `PUT` requests. Subsequent `PUT` requests to an aggregation job MUST be rejected by helpers.
 
 This is analogous to DAP-02's `POST /aggregate` with an `AggregateInitReq`.
 
@@ -232,6 +234,8 @@ struct {
 
 If successful, the response from the helper is a `struct AggregateShare`. That same share may later be obtained by a GET request on the resource.
 
+Once it is created, an aggregate share cannot be mutated with a further `PUT` request. Subsequent `PUT` requests MUST be rejected by the helper.
+
 This is analogous to DAP-02's `POST /aggregate_share`.
 
 ###### A note on `BatchSelector.fixed_size`
@@ -294,6 +298,8 @@ struct {
 ```
 
 If successful, the response from the helper is a `struct Collection`. That same share may later be obtained by a GET request on the resource.
+
+Once created, a collection cannot be mutated with a further `PUT` request. Leaders MUST reject subsequent `PUT` requests to a collection resource.
 
 This is analogous to DAP-02's `POST /collect`.
 
--- a/api-resources.md
+++ b/api-resources.md
@@ -24,19 +24,13 @@ We care a lot about idempotence because we need to account for cases where an HT
 
 ### HPKE configurations
 
+This proposal deliberately does not address any of the open questions or issues around multiple HPKE configurations or negotiation (see #248) and preserves the existing `hpke_config` endpoint unchanged, in the interest of containing the scope of this proposal.
+
 #### Path
 
 `/hpke_config[?task_id=task-id]`
 
 The `task_id` query parameter is optional, as described in https://www.ietf.org/archive/id/draft-ietf-ppm-dap-02.html#section-4.3.1.
-
-##### Alternatives
-
-An HPKE config has an ID, so we could use a path `hpke_configs/hpke-config-id`. However, a client should never need to get any HPKE config except for the "current" one, so does it make sense to expand the API surface with a path param for HPKE config ID?
-
-`GET /hpke_config` could redirect to `/hpke_config/current-hpke-config-id`, which would make this mostly transparent to clients, assuming they handle 3xx redirects properly.
-
-Instead of providing the task ID as a query parameter, we could also include it in the path, like `/tasks/{task-id}/hpke_config`, on the premise that the task ID is part of the resource's permanent identity. However, it's awkward to solve the global HPKE config case with this formulation. We could include a special task ID "global" so that clients could query `/tasks/global/hpke_config`.
 
 #### Representation
 
--- a/api-resources.md
+++ b/api-resources.md
@@ -77,10 +77,6 @@ struct {
 
 #### Required HTTP methods
 
-##### GET/HEAD
-
-The response body is a `struct Report` representing the report.
-
 ##### PUT
 
 Idempotent upload of a report to an aggregator. The request body is a `struct Report`.
--- a/api-resources.md
+++ b/api-resources.md
@@ -50,7 +50,7 @@ struct {
 
 ##### GET/HEAD
 
-The response body is a `struct HpkeConfig` representing the `hpke_config`.
+The response body is a `struct HpkeConfig` representing the server's current `hpke_config`.
 
 ### Reports
 
@@ -125,7 +125,7 @@ See "POST", below, for explanation and justification of the `round` field.
 
 ##### GET/HEAD
 
-The response body is a `struct AggregateJob` representing the current state of the reports in the job.
+The response body is a `struct AggregateJob` representing the current state of the reports in the job. The server mays return an error if no aggregation job is known to it for the aggregation job ID, or if the aggregation job results have been discarded, whether in response to a `DELETE` request on the resource, the task having expired, or any other garbage collection policy the server implements.
 
 ##### PUT
 
@@ -207,7 +207,7 @@ struct {
 
 ##### GET/HEAD
 
-If the aggregate share is available, the result is a `struct AggregateShare`. Otherwise the server can return something like HTTP 202 Accepted to indicate it's not ready yet, or HTTP 404 if no such aggregate share is known to the helper.
+If the aggregate share is available, the result is a `struct AggregateShare`. Otherwise the server can return something like HTTP 202 Accepted to indicate it's not ready yet, or HTTP 404 if no such aggregate share is known to the helper. The server may also return an error if the aggregate share was discarded in response to a `DELETE` request on the resource, the task expiring or some other garbage collection policy implemented by the server.
 
 ##### PUT
 
@@ -272,7 +272,7 @@ struct {
 
 ##### GET/HEAD
 
-If the collection is available, the response body is a `struct Collection`. Otherwise the server can return something like HTTP 202 Accepted to indicate it's not ready yet, or HTTP 404 if no such aggregate share request is known to the server.
+If the collection is available, the response body is a `struct Collection`. Otherwise the server can return something like HTTP 202 Accepted to indicate it's not ready yet, or HTTP 404 if no such aggregate share request is known to the server. The server may also return an error if the collection was discarded due to a `DELETE` request on the resource, the task having expired, or some other garbage collection policy implemented by the server.
 
 This is analogous to DAP-02's `GET` on a collect job URI.
 
--- a/api-resources.md
+++ b/api-resources.md
@@ -125,7 +125,7 @@ See "POST", below, for explanation and justification of the `round` field.
 
 ##### GET/HEAD
 
-The response body is a `struct AggregateJob` representing the current state of the reports in the job. The server mays return an error if no aggregation job is known to it for the aggregation job ID, or if the aggregation job results have been discarded, whether in response to a `DELETE` request on the resource, the task having expired, or any other garbage collection policy the server implements.
+The response body is a `struct AggregateJob` representing the current state of the reports in the job. The server may return an error if no aggregation job is known to it for the aggregation job ID, or if the aggregation job results have been discarded, whether in response to a `DELETE` request on the resource, the task having expired, or any other garbage collection policy the server implements.
 
 ##### PUT
 
@@ -330,7 +330,7 @@ struct {
 
 If the collection is available, the response body is a `struct Collection`. Otherwise the server can return something like HTTP 202 Accepted to indicate it's not ready yet, or HTTP 404 if no such collection is known to the server. The server may also return an error if the collection was discarded due to a `DELETE` request on the resource, the task having expired, or some other garbage collection policy implemented by the server.
 
-Even after receiving a `struct Collection`, the collector MAY continue to send `POST` requests to a collect URI and the respones should still be the `struct Collection` (if it is still available and hasn't been garbage collected).
+Even after receiving a `struct Collection`, the collector MAY continue to send `POST` requests to a collect URI and the responses should still be the `struct Collection` (if it is still available and hasn't been garbage collected).
 
 See "Viewing fixed-size batches as a message queue" for discussion.
 
--- a/api-resources.md
+++ b/api-resources.md
@@ -5,9 +5,9 @@ DAP is structured as an HTTP-based application with the following resources:
 - (TODO) Tasks
 - HPKE configurations
 - Reports
-- Aggregation jobs (and lists thereof)
-- Aggregate shares (and lists thereof)
-- Collections (and lists thereof)
+- Aggregation jobs
+- Aggregate shares
+- Collections
 
 A resource's path is resolved relative to a server's base URL to construct a resource URI. Deployments may host resource paths arbitrarily deep relative to their domain. Paths are generally structured as `/resource-type/{resource-id}`. Anywhere `{resource-id}` (e.g., `{task-id}` or `{report-id}`) occurs in a URI is to be understood as the URL-safe, unpadded base64 representation of the resource's identifier, which itself is usually 16 random bytes (some are 32, but https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/issues/349 will normalize them to 16).
 
--- a/api-resources.md
+++ b/api-resources.md
@@ -9,9 +9,9 @@ DAP is structured as an HTTP-based application with the following resources:
 - Aggregate shares (and lists thereof)
 - Collections (and lists thereof)
 
-A resource's path is resolved relative to a server's base URL to construct a resource URI. Deployments may host resource paths arbitrarily deep relative to their domain. Paths are generally structured as `/resource-type/resource-id`. Anywhere `resource-id` (e.g., `task-id` or `report-id`) occurs in a URI is to be understood as the URL-safe, unpadded base64 representation of the resource's identifier, which itself is usually 16 random bytes (some are 32, but https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/issues/349 will normalize them to 16).
+A resource's path is resolved relative to a server's base URL to construct a resource URI. Deployments may host resource paths arbitrarily deep relative to their domain. Paths are generally structured as `/resource-type/{resource-id}`. Anywhere `{resource-id}` (e.g., `{task-id}` or `{report-id}`) occurs in a URI is to be understood as the URL-safe, unpadded base64 representation of the resource's identifier, which itself is usually 16 random bytes (some are 32, but https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/issues/349 will normalize them to 16).
 
-Some resources are owned by another resource, and so their URI will contain two identifiers. For instance, a report belongs to a task, so the URI is `[base]/tasks/task-id/reports/report-id`.
+Some resources are owned by another resource, and so their URI will contain two identifiers. For instance, a report belongs to a task, so the URI is `[base]/tasks/{task-id}/reports/{report-id}`.
 
 Generally, if a resource supports the `GET` method, it also supports `HEAD`. The response to `HEAD` is exactly the same as `GET`, but the response contains no body.
 
@@ -41,7 +41,7 @@ An HPKE config has an ID, so we could use a path `hpke_configs/hpke-config-id`. 
 
 `GET /hpke_config` could redirect to `/hpke_config/current-hpke-config-id`, which would make this mostly transparent to clients, assuming they handle 3xx redirects properly.
 
-Instead of providing the task ID as a query parameter, we could also include it in the path, like `/tasks/task-id/hpke_config`, on the premise that the task ID is part of the resource's permanent identity. However, it's awkward to solve the global HPKE config case with this formulation. We could include a special task ID "global" so that clients could query `/tasks/global/hpke_config`.
+Instead of providing the task ID as a query parameter, we could also include it in the path, like `/tasks/{task-id}/hpke_config`, on the premise that the task ID is part of the resource's permanent identity. However, it's awkward to solve the global HPKE config case with this formulation. We could include a special task ID "global" so that clients could query `/tasks/global/hpke_config`.
 
 #### Representation
 
@@ -67,7 +67,7 @@ The response body is a `struct HpkeConfig` representing the `hpke_config`.
 
 #### Path
 
-`/tasks/task-id/reports/report-id`
+`/tasks/{task-id}/reports/{report-id}`
 
 `report-id` uniquely identifies a report in a task. It is chosen by the client when uploading a report.
 
@@ -104,7 +104,7 @@ Only the helper supports this resource.
 
 #### Path
 
-`/tasks/task-id/aggregation_jobs/aggregation-job-id`
+`/tasks/{task-id}/aggregation_jobs/{aggregation-job-id}`
 
 `aggregation-job-id` uniquely identifies an aggregation job. It is chosen by the leader when it dispatches a job to the helper.
 
@@ -180,17 +180,17 @@ The idea here is that all the values in `struct AggregateJobPutReq` never need t
 
 POST to an aggregate job is how the leader steps an aggregate job. The request body is a `struct AggregateJob` where `prepare_steps` contains the leader's current-round prepare messages, and `round` is the number of the current round. The response is a `struct AggregateJob` where `prepare_steps` contains the helper's next-round prepare messages. The response's `round` field will be one more than the request's `round`.
 
-We use a POST here because this request is _not_ idempotent. Suppose the leader sends `POST /tasks/task-id/aggregate_jobs/aggregation-job-id` where the body has `round = 1`, and that this is successfully handled by the helper. Besides generating the response, this will have the side effect of advancing the state in the helper to round 2. If the leader were to resend the `POST /tasks/task-id/aggregate_jobs/aggregation-job-id` with `round = 1`, then the request should be rejected by the helper, because it has advanced to the next round.
+We use a POST here because this request is _not_ idempotent. Suppose the leader sends `POST /tasks/{task-id}/aggregate_jobs/{aggregation-job-id}` where the body has `round = 1`, and that this is successfully handled by the helper. Besides generating the response, this will have the side effect of advancing the state in the helper to round 2. If the leader were to resend the `POST /tasks/{task-id}/aggregate_jobs/{aggregation-job-id}` with `round = 1`, then the request should be rejected by the helper, because it has advanced to the next round.
 
 This is analogous to DAP-02's `POST /aggregate` with an `AggregateContinueReq`.
 
 ###### Context for the `round` field
 
-The `round` field is needed so that the leader can recover from a response being lost. Suppose a 2-round VDAF is being executed, and that the leader does `PUT /tasks/task-id/aggregate_jobs/aggregation-job-id` and that request succeeds, so the helper responds with its first-round prepare messages. Then, the leader does `POST /tasks/task-id/aggregate_jobs/aggregation-job-id` with the first-round broadcast prepare message, but the helper's response gets lost during network transit.
+The `round` field is needed so that the leader can recover from a response being lost. Suppose a 2-round VDAF is being executed, and that the leader does `PUT /tasks/{task-id}/aggregate_jobs/{aggregation-job-id}` and that request succeeds, so the helper responds with its first-round prepare messages. Then, the leader does `POST /tasks/{task-id}/aggregate_jobs/{aggregation-job-id}` with the first-round broadcast prepare message, but the helper's response gets lost during network transit.
 
-If the helper received the `POST` and advanced its state to the second round, then the leader should do `GET /tasks/task-id/aggregate_jobs/aggregation-job-id` so it can compute the second round broadcast message and then do `POST /tasks/task-id/aggregate_jobs/aggregation-job-id` to have the helper move to the finished state. If the helper did not receive the `POST`, then the leader should re-send the first-round broadcast prepare message.
+If the helper received the `POST` and advanced its state to the second round, then the leader should do `GET /tasks/{task-id}/aggregate_jobs/{aggregation-job-id}` so it can compute the second round broadcast message and then do `POST /tasks/{task-id}/aggregate_jobs/{aggregation-job-id}` to have the helper move to the finished state. If the helper did not receive the `POST`, then the leader should re-send the first-round broadcast prepare message.
 
-But in DAP-02, the leader has no way to know what happened in the helper, leaving it unable to recover from this state. Adding a `round` field to the `AggregateJob` message enables the leader to reliably and idempotently find out what state the helper is in by doing `GET /tasks/task-id/aggregate_jobs/aggregation-job-id`, and then taking the appropriate next step.
+But in DAP-02, the leader has no way to know what happened in the helper, leaving it unable to recover from this state. Adding a `round` field to the `AggregateJob` message enables the leader to reliably and idempotently find out what state the helper is in by doing `GET /tasks/{task-id}/aggregate_jobs/{aggregation-job-id}`, and then taking the appropriate next step.
 
 ##### DELETE
 
@@ -198,38 +198,13 @@ Instructs the helper to abandon the aggregate job and allows it to discard all s
 
 Requiring this would solve https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/issues/241
 
-### Aggregation jobs
-
-Only the helper supports this resource.
-
-#### Path
-
-`/tasks/task-id/aggregation_jobs`
-
-#### Representation
-
-A paginated list of aggregation jobs for the task.
-
-```
-struct {
-  u64 page;
-  AggregateJob aggregate_jobs<0..2^16-1>
-} AggregationJobList
-```
-
-#### Required HTTP methods
-
-##### GET/HEAD
-
-The request takes an optional query parameter for the page. The response body is a `struct AggregationJobList`.
-
 ### An aggregate share
 
 Only the helper supports this resource.
 
 #### Path
 
-`/tasks/task-id/aggregate_shares/aggregate-share-id`
+`/tasks/{task-id}/aggregate_shares/{aggregate-share-id}`
 
 `aggregate-share-id` uniquely identifies an aggregate share. It is chosen by the leader when it requests an aggregate share by the helper.
 
@@ -278,38 +253,11 @@ This is analogous to DAP-02's `POST /aggregate_share`.
 
 In DAP-02, a `BatchSelector` with `query_type = fixed_size` contains a `BatchID`. In this API, the batch ID hoisted up into the resource URI. This implies that `time_interval`-type queries now also have a batch ID, chosen by the leader. While aggregate shares will often (always?) be 1:1 with collections, the IDs do not have to match.
 
-Note also that `AggregateShareReq` should use a `Query` (https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/issues/366).
-
 TODO: timg to read more about chunky DAP and think about how it intersects with this proposal, especially https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/issues/342
 
 ##### DELETE
 
 Instructs the helper to abandon the aggregate share and allows the helper to discard all state related to it. This is akin to DELETE on a collection for the leader.
-
-### Aggregate shares
-
-Only the helper supports this resource.
-
-#### Path
-
-`/tasks/task-id/aggregate_shares`
-
-#### Representation
-
-A paginated list of aggregation jobs for the task.
-
-```
-struct {
-  u64 page;
-  AggregateShare aggregate_shares<0..2^16-1>
-} AggregateShareList
-```
-
-#### Required HTTP methods
-
-##### GET/HEAD
-
-The request takes an optional query parameter for the page. The response body is a `struct AggregateShareList`.
 
 ### A collection
 
@@ -317,7 +265,7 @@ Only the leader supports this resource.
 
 #### Path
 
-`/tasks/task-id/collections/collection-id`
+`/tasks/{task-id}/collections/{collection-id}`
 
 `collection-id` uniquely identifies a collection. It is chosen by the collector when it POSTs to the resource.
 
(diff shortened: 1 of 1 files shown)


api-resources.md
```diff
@@ -0,0 +1,414 @@
...
+
+### A collection
+
+Only the leader supports this resource.
+
+#### Path
+
+`/tasks/task-id/collections/collection-id`
+
+`collection-id` uniquely identifies a collection. It is chosen by the collector when it POSTs to the resource.
```
divergentdave, line 322:
```suggestion
`collection-id` uniquely identifies a collection. It is chosen by the collector when it PUTs to the resource.
```

api-resources.md
```diff
@@ -0,0 +1,414 @@
...
+
+#### Representation
+
+A paginated list of aggregation jobs for the task.
+
+```
+struct {
+  u64 page;
+  AggregateJob aggregate_jobs<0..2^16-1>
+} AggregationJobList
```
divergentdave, line 217: I think the list responses will require some additional metadata to indicate when to stop requesting pages.

tgeoghegan: Yeah, I should have put a TODO there, because the pagination stuff isn't completely thought through. In fact we should follow the lead of ACME here and use link relations. [RFC 8288](https://datatracker.ietf.org/doc/html/rfc8288) appears to be the relevant reference.

However, it's possible we can get away without providing the endpoints for listing aggregation jobs, aggregate shares and collections if we can instead work out a way for protocol participants to query for exactly the aggregation job, aggregate share or collection they care about, perhaps by putting things like aggregation parameters, batch selectors or queries into query parameters.

api-resources.md
```diff
@@ -0,0 +1,414 @@
...
+
+We use HTTP `PUT` requests for idempotent creation of resources. Server responses to `PUT` requests follow [RFC 9110 section 9.3.4](https://httpwg.org/specs/rfc9110.html#rfc.section.9.3.4).
+
+Finally, a quick reminder of some HTTP semantics: `GET` and `PUT` are idempotent (i.e., the same request may be made multiple times without changing state in the server more than once, and the response will be the same each time). `POST` is not idempotent. `GET` requests may not have bodies.
+
+We care a lot about idempotence because we need to account for cases where a client makes a request but never sees the response, and thus needs a way to recover from this state of uncertainty.
+
+### Tasks
+
+TODO: do we want a task resource that we can GET and PUT in the protocol? Does this collide with https://wangshan.github.io/draft-wang-ppm-dap-taskprov/draft-wang-ppm-dap-taskprov.html or https://github.com/divergentdave/draft-dcook-ppm-dap-interop-test-design?
```
wangshan, line 28: I think it can work with taskprov, which specifies how to provision a task in-band, it could end up calling the `PUT tasks/<task-id>` API. But so far the out-of-band task provision is not specified, so we may want to flesh that out (if we want to flesh that out) first.

api-resources.md
```diff
@@ -0,0 +1,414 @@
...
+  PartialBatchSelector part_batch_selector;
+  ReportShare report_shares<1..2^32-1>;
+} AggregateJobPutReq;
+```
+
+If successful, the helper's response is a `struct AggregateJob` where `prepare_steps` contains the helper's first-round VDAF prepare messages for the report shares in `AggregateJobPutReq`, or errors if `vdaf.prepare` failed for any report.
+
+This is analogous to DAP-02's `POST /aggregate` with an `AggregateInitReq`.
+
+This request is idempotent, so if a helper receives multiple PUT requests for some aggregation job and the members of `AggregateJobPutReq` don't change, the helper can return 200 OK.
```
wangshan, line 175: I don't think this is idempotent if aggregator proactively aggregates reports. If two identical AggregateJobPutReq is received before and after a batch is collected, then the later one will end up in error.

wangshan: even without collect flow, anti-reply will make this non-idempotent too:

https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/blob/main/draft-ietf-ppm-dap.md#early-input-share-validation-early-input-share-validation

api-resources.md
```diff
@@ -0,0 +1,414 @@
...
+
+Only the helper supports this resource.
+
+#### Path
+
+`/tasks/task-id/aggregate_shares`
+
+#### Representation
+
+A paginated list of aggregation jobs for the task.
```
branlwyd, line 299: Why do we need a paginated list of aggregate shares?

I think this can be dropped from the specification (& left up to implementations, if they find it handy).

api-resources.md
branlwyd, line 322: Currently DAP doesn't specify collection IDs. Could that be done as part of this work? I think a `CollectionID` as a 16-byte opaque identifier would fit will with similar identifiers in the spec. I think explicitly defining collections would as a side effect make some of the validation logic easier to specify/follow.

(BatchID doesn't work, unfortunately: a collection is logically identified by a batch ID along with an aggregation parameter. A more radical change might be to specify something like `/tasks/{task-id}/batches/{batch-id}/collections/{agg-param}` to collect. This would be very clean in terms of API layout & allowing our validation to be written in terms of explicitly-defined entities, but I don't think we want to commit to putting an aggregation parameter in the URI path -- we don't know how large it is, or that it will fit well in the URL-safe character set.)

api-resources.md
```diff
@@ -0,0 +1,414 @@
...
+
+##### DELETE
+
+Instructs the leader to abandon the collection and allows the leader to discard all state related to it. This is akin to DELETE on an aggregate share for the helper (indeed, the leader might send a DELETE to the helper's aggregate share resource after its corresponding collection is deleted).
+
+This is analogous to DAP-02's `DELETE` on a collect job URI.
+
+#### Alternatives: `collection-id`, `PUT` vs. `POST` and idempotence?
+
+This proposal introduces the notion of a `collection-id`, which is chosen by the collector when it sends `CollectReq` to helper. In DAP-02, a collection is already uniquely identified by the combination of the aggregation parameter and the query, meaning we can get by without `collection-id`. Let's sketch out what the API would look like so we can weigh pros and cons.
```
branlwyd, line 379: > a collection is already uniquely identified by the combination of the aggregation parameter and the query

nit: this is true today but won't be true once #342 is resolved; at that point, there will be some mechanism for the Collector's `Query` to say something like "I don't know the specific batch I want, just give me one that's ready (if any)" -- i.e. the Collector's `Query` will not directly identify a single specific batch.

I guess the more accurate way of saying this would be that a collection is identified by an aggregation parameter and a batch identifier (literally `BatchID` for fixed-size, a time interval for time-interval).

api-resources.md
```diff
@@ -0,0 +1,414 @@
...
+
+`POST /tasks/task-id/collections`
+
+The body is a `CollectReq`, defined as above. The request method is `POST` instead of `PUT` because I don't think this request can be idempotent. Let's say a collector makes a time interval query over some interval _i_ and it is received by the leader at time _t1_. At _t1_, the leader has _n1_ reports that fall into the interval _i_. Then, suppose at least one more report that falls into _i_ arrives and the leader and helper prepare it. Then, at _t2 > t1_, the collector sends a query with the same interval _i_ again.
+
+Should the leader serve up the results it computed at time _t1_? Or should it make a new collection, consuming another unit of max batch query, that includes the reports that arrived and were prepared between _t1_ and _t2_?
```
branlwyd, line 385: I'm confused, wouldn't the leader ignore the reports that fall into interval _i_ after the first collection request at _t1_? 

From DAP-02: "The Leader MUST ignore any report pertaining to a batch that has already been collected."

Multiple collections for a single batch are to allow aggregation parameter to vary; I don't think the current wording of DAP-02 allows for collecting "the same batch but with a few more reports that have arrived since the last collection" to avoid potentially breaking privacy of those new reports.

api-resources.md
```diff
@@ -0,0 +1,414 @@
...
+
+Only the leader supports this resource.
+
+#### Path
+
+`/tasks/task-id/collections`
+
+#### Representation
+
+A paginated list of collections for the task.
```
branlwyd, line 401: Why do we need a paginated list of collection jobs?

I think this can be dropped from the specification (& left up to implementations, if they find it handy).

api-resources.md
branlwyd, line 299:
```suggestion
A paginated list of aggregate shares for the task.
```

api-resources.md
tgeoghegan: It's idempotent in the sense that repeated `PUT /tasks/{task-id}/aggregation_jobs/{aggregation-job-id}` requests won't change the state, the key point being that it's safe for the leader to resend that request to the helper multiple times.

I also think that collection of a batch doesn't need to affect how the helper responds to this request. Let me sketch out a hypothetical:

1. Leader sends `PUT /tasks/{task-id}/aggregation_jobs/agg-job-1` (here, `agg-job-1` is some specific aggregation job identifier).
2. Helper replies with `struct AggregateJob` containing first-round prepare message
3. Leader sends `POST /tasks/{task-id}/aggregation_jobs/agg-job-1` with first-round broadcast prepare message
4. Helper prepares the reports, consuming batch query count for the relevant reports, and replies with `struct AggregateJob` containing a sequence of `PrepareStep` that are all in state `finished`.

At this point, if the leader sends `PUT /tasks/{task-id}/aggregation_jobs/agg-job-1` again (for whatever reason) _and if all the members of the `struct AggregateJobPutReq` are unchanged_, then the helper shouldn't re-evaluate anti-replay, because it can check its own storage, notice it's already prepared all those inputs, and return the same response it did in step (4). This doesn't violate anti-replay or privacy budgets because it's the exact same set of reports under the same aggregation parameter.

However, if the leader sends `PUT /tasks/{task-id}/aggregation_jobs/agg-job-1` _but varies any of the fields of the body_, then the helper should reject it. So I think we should specify that the aggregation job resource may not be mutated once created, except to advance its state through a POST request.

Further, if the leader sends `PUT /tasks/{task-id}/aggregation_jobs/agg-job-2` and the reports in that request's body overlap with the ones prepared by `agg-job-1`, then the helper has to apply anti-replay and either consume query limit or refuse to service the request, as appropriate.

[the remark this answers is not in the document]
tgeoghegan: I think you are raising the same question as Shan did [here](https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/pull/367/files#r997208345), so I would direct you to my response there. But you bring up an additional good point about mutating an aggregation job with a PUT, which I think we should explicitly forbid.

api-resources.md
```diff
@@ -0,0 +1,414 @@
...
+
+Only the helper supports this resource.
+
+#### Path
+
+`/tasks/task-id/aggregation_jobs`
+
+#### Representation
+
+A paginated list of aggregation jobs for the task.
```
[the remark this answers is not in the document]
tgeoghegan: Given that aggregation job IDs are assigned by the leader, I believe you are right, because the leader should be able to maintain an authoritative list of agg jobs itself instead of relying on the helper.

api-resources.md
```diff
@@ -0,0 +1,414 @@
...
+
+### An aggregate share
+
+Only the helper supports this resource.
+
+#### Path
+
+`/tasks/task-id/aggregate_shares/aggregate-share-id`
+
+`aggregate-share-id` uniquely identifies an aggregate share. It is chosen by the leader when it requests an aggregate share by the helper.
```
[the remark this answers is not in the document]
tgeoghegan: Yes, aggregate share ID is new, and introduced for consistency with other resources, and for the same fundamental reason that I introduce `collection-id`. The helper's aggregate share resource is strongly analogous to the leader's collection resource, so a lot of the same considerations apply, including the question of whether or not an aggregate share or collection is uniquely identified by the (aggregation parameter, query) tuple. Anyway, I'll discuss that question further in the collection-id portions of the doc [here](https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/pull/367/files#r997427041).

api-resources.md
```diff
@@ -0,0 +1,414 @@
...
+} AggregateShareReq;
+```
+
+If successful, the response from the helper is a `struct AggregateShare`. That same share may later be obtained by a GET request on the resource.
+
+This is analogous to DAP-02's `POST /aggregate_share`.
+
+###### A note on `BatchSelector.fixed_size`
+
+In DAP-02, a `BatchSelector` with `query_type = fixed_size` contains a `BatchID`. In this API, the batch ID hoisted up into the resource URI. This implies that `time_interval`-type queries now also have a batch ID, chosen by the leader. While aggregate shares will often (always?) be 1:1 with collections, the IDs do not have to match.
```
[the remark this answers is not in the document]
tgeoghegan: I think this again is the question of what uniquely identifies an aggregate share or collection, which I'd like to consolidate [here](https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/pull/367/files#r997427041).

api-resources.md
tgeoghegan: Agreed, I will delete this resource.

api-resources.md
tgeoghegan: You are correct, I forgot about that requirement in DAP-02. I need to revisit this text after thinking harder about `current-batch` and fixed batch queries in general.

api-resources.md
tgeoghegan: Deleted with the other list resources

api-resources.md
tgeoghegan: A recurring set of questions here is (1) whether aggregate shares and collections need an opaque ID to uniquely identify them beyond the (aggregate parameter, query) tuple and (2) who should choose that ID if it is needed. I'm still ruminating on this; more thoughts to follow in this thread.

api-resources.md
wangshan: @tgeoghegan what I meant was after the task has been provisioned following task_prov extension, the leader/helper can call a _PUT to /tasks/task-id_ to store the task object. The /tasks endpoint should not be exposed to client.

api-resources.md
wangshan: @tgeoghegan makes sense, I think it's safe privacy wise since the agg flow is for preparation. My other concern is time related states, if the 2nd `PUT /tasks/{task-id}/aggregation_jobs/agg-job-1` come at a time the task_id has expired, would helper responds differently? If not then we are potentially asking helper to store output share or aggregate share forever.

api-resources.md
```diff
@@ -0,0 +1,337 @@
...
+
+#### Path
+
+`/hpke_config[?task_id=task-id]`
+
+The `task_id` query parameter is optional, as described in https://www.ietf.org/archive/id/draft-ietf-ppm-dap-02.html#section-4.3.1.
+
+##### Alternatives
+
+An HPKE config has an ID, so we could use a path `hpke_configs/hpke-config-id`. However, a client should never need to get any HPKE config except for the "current" one, so does it make sense to expand the API surface with a path param for HPKE config ID?
```
cjpatton, line 40: As ChrisW would no doubt ask at this juncture: What are your requirements here? Why would we need multiple HPKE configs?

One reason was previously pointed out by @ekr: https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/issues/248. TL;DR: The Client and Aggregator may not initially agree on what HPKE ciphersuite to use. But we may not need to expose multiple configs simultaneously, if the Client can indicate in its request which suites it supports.

api-resources.md
```diff
@@ -0,0 +1,337 @@
...
+  opaque public_share<0..2^32-1>;
+  HpkeCiphertext encrypted_input_shares<1..2^32-1>;
+} Report;
+```
+
+#### Required HTTP methods
+
+##### GET/HEAD
+
+The response body is a `struct Report` representing the report.
```
cjpatton, line 93: Who needs this functionality, and for what purpose? How do we think about how long this report needs to be kept around?

This might be a bit tricky for Daphne because we may want to shard reports by timestamp.

api-resources.md
```diff
@@ -0,0 +1,337 @@
...
+} AggregateJob
+```
+
+See "POST", below, for explanation and justification of the `round` field.
+
+#### Required HTTP methods
+
+##### GET/HEAD
+
+The response body is a `struct AggregateJob` representing the current state of the reports in the job.
```
cjpatton, line 143: Who needs to implement this: the Leader, Helper, or both? Presumably just the Helper? How long do we need to keep this around?

api-resources.md
```diff
@@ -0,0 +1,337 @@
...
+
+If the helper received the `POST` and advanced its state to the second round, then the leader should do `GET /tasks/{task-id}/aggregate_jobs/{aggregation-job-id}` so it can compute the second round broadcast message and then do `POST /tasks/{task-id}/aggregate_jobs/{aggregation-job-id}` to have the helper move to the finished state. If the helper did not receive the `POST`, then the leader should re-send the first-round broadcast prepare message.
+
+But in DAP-02, the leader has no way to know what happened in the helper, leaving it unable to recover from this state. Adding a `round` field to the `AggregateJob` message enables the leader to reliably and idempotently find out what state the helper is in by doing `GET /tasks/{task-id}/aggregate_jobs/{aggregation-job-id}`, and then taking the appropriate next step.
```
cjpatton, line 193: Nice idea.

api-resources.md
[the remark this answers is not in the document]
cjpatton: Hmm, are you confusing "batchID" with "aggJobID" here? I don't see as batch ID in the agg job endpoint.

api-resources.md
```diff
@@ -0,0 +1,337 @@
...
+} CollectReq;
+```
+
+If successful, the response from the helper is a `struct Collection`. That same share may later be obtained by a GET request on the resource.
+
+This is analogous to DAP-02's `POST /collect`.
+
+###### A note on `Query.fixed_size`
+
+`struct Query` is essentially identical to `struct BatchSelector`, so the note about `BatchSelector.fixed_size` from `PUT aggregate_share` applies here, except that a collection's ID is chosen by the collector.
```
cjpatton, line 317: This would change if we took #313.

api-resources.md
```diff
@@ -0,0 +1,337 @@
...
+
+This is analogous to DAP-02's `DELETE` on a collect job URI.
+
+#### Alternatives: `collection-id`, `PUT` vs. `POST` and idempotence?
+
+This proposal introduces the notion of a `collection-id`, which is chosen by the collector when it sends `CollectReq` to helper. In DAP-02, a collection is already uniquely identified by the combination of the aggregation parameter and the query, meaning we can get by without `collection-id`. Let's sketch out what the API would look like so we can weigh pros and cons.
+
+`POST /tasks/{task-id}/collections`
+
+The body is a `CollectReq`, defined as above. The request method is `POST` instead of `PUT` because I don't think this request can be idempotent. Let's say a collector makes a time interval query over some interval _i_ and it is received by the leader at time _t1_. At _t1_, the leader has _n1_ reports that fall into the interval _i_. Then, suppose at least one more report that falls into _i_ arrives and the leader and helper prepare it. Then, at _t2 > t1_, the collector sends a query with the same interval _i_ again.
```
cjpatton, line 331: nit: Avoid using italics "_blah_" in I-Ds. this doesn't get rendered the way you want. (Look at the draft-ietf-ppm-dap.txt.)

cjpatton (commented):
We obviously have some details to work out, as you've noted, but overall I strongly support this change. The main concern I have is that there may be hidden costs in terms of implementation complexity. (I haven't done a thorough review yet, but I'll be looking out for things that might require big refactors in Daphne.) If we can mitigate that, I think the obvious benefit is a much, much cleaner spec that will be easier to implement fresh.

High-level things:

1. It would be useful to flesh out a bit more the requirements for resource limits. How long do I have to keep data artifacts around before I can throw them away?
2. (Not sure what the answer is here.) Does this close the door on DAP deployments with multiple helpers, or does the design anticipate this?
3. I think this definitely warrants discussion at IETF 115, if at all possible.

api-resources.md
tgeoghegan: From DAP-02 [5.4.1](https://www.ietf.org/archive/id/draft-ietf-ppm-dap-02.html#section-5.4.1):

> Furthermore, the aggregators must store data related to a task as long as the current time has not passed this task's task_expiration. Aggregator MAY delete the task and all data pertaining to this task after task_expiration. Implementors SHOULD provide for some leeway so the collector can collect the batch after some delay.

So if a task expires, then from that point, a helper MAY respond to `PUT /tasks/{task-id}/aggregation_jobs/agg-job-1` with an error. It could also choose to continue providing the results of the aggregation job indefinitely. It's up to the implementation to decide when it discards data belonging to a task like aggregate results.

I think there will have to be some text explaining that leaders should be prepared for helpers to have discarded the results of an aggregation job, but I don't think this is an idempotence problem, because the state change (helper discarding aggregate job results) wasn't a result of repeated `PUT` requests by the leader.

api-resources.md
```diff
@@ -0,0 +1,337 @@
...
+
+Generally, if a resource supports the `GET` method, it also supports `HEAD`. The response to `HEAD` is exactly the same as `GET`, but the response contains no body.
+
+If an HTTP method is listed in a resource's required HTTP methods, implementations MUST provide responses as indicated. Otherwise, implementations MAY implement methods on resources however they wish, including returning an error like HTTP 405 Method Not Allowed.
+
+We use HTTP `PUT` requests for idempotent creation of resources. Server responses to `PUT` requests follow [RFC 9110 section 9.3.4](https://httpwg.org/specs/rfc9110.html#rfc.section.9.3.4).
+
+Finally, a quick reminder of some HTTP semantics: `GET` and `PUT` are idempotent (i.e., the same request may be made multiple times without changing state in the server more than once, and the response will be the same each time). `POST` is not idempotent. `GET` requests may not have bodies.
+
+We care a lot about idempotence because we need to account for cases where a client makes a request but never sees the response, and thus needs a way to recover from this state of uncertainty.
```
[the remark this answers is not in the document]
tgeoghegan: I think you (quite understandably) read this as meaning a DAP client, but here I am using "client" in the sense of an HTTP client, such that the leader is a client when it makes requests to the helper, and the collector is a client when it makes requests to the leader. I'll clarify this.

- [ ] clarify meaning of "client" to mean HTTP client, not DAP client

api-resources.md
tgeoghegan: This change will be complicated enough, so I think I will choose not to solve #248, too, so I lean toward leaving the `hpke_config` exactly as it is.

- [ ] put `hpke_config` back the way it was

api-resources.md
tgeoghegan: You're right, I was just putting it in because semantically it's obvious, but there's no protocol use case, so we shouldn't specify it (implementations are free to implement this if they find it useful).

- [ ] remove GETters we don't explicitly need in the protocol

tgeoghegan:
> * It would be useful to flesh out a bit more the requirements for resource limits. How long do I have to keep data artifacts around before I can throw them away?

I agree that DAP needs to deal with this, however I don't think any new ambiguities are introduced by this change. Yes, there are circumstances where a helper must respond to `PUT /tasks/{task-id}/aggregation_jobs/{aggregation-job-id}` with an error indicating it doesn't have the report shares anymore, but that exact same situation exists when it's `POST /aggregate` instead.

> * (Not sure what the answer is here.) Does this close the door on DAP deployments with multiple helpers, or does the design anticipate this?

I don't see any way this change forbids more than one helper, though I think DAP should commit to exactly two aggregators anyway.
 
> * I think this definitely warrants discussion at IETF 115, if at all possible.

WG involvement and consensus is obviously important. But I don't know if a slide deck and a few brief minutes of discussion will let us make much progress. I think a call to action to come hash out the design in the WG mailing list or here on GitHub will be most effective.

api-resources.md
tgeoghegan: Just the helper, as noted above, but we do need to--

- [ ] add some language discussing how long servers should have to respond to GET aggregate jobs, aggregate shares, collections, considering task expiration and DELETE requests against these

api-resources.md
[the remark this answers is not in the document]
tgeoghegan: This is aggregate share, not aggregate job, which does reference batch ID in [DAP-02](https://www.ietf.org/archive/id/draft-ietf-ppm-dap-02.html#section-4.5.2-3)

api-resources.md
tgeoghegan: - [ ] Do we need an API level concept of collection ID or aggregate share ID? If so, who chooses those IDs? Can the same solution work for both time interval and chunky DAP queries?

api-resources.md
tgeoghegan: Resolving since the list resources have been deleted anyway, but if/when we want something like this again, RFC 8288 link relations are the way to go.

api-resources.md
```diff
@@ -0,0 +1,414 @@
...
+
+The idea here is that all the values in `struct AggregateJobPutReq` never need to be transmitted again during the handling of an aggregate job, which is why the message is not `struct AggregateJob`.
+
+##### POST
+
+POST to an aggregate job is how the leader steps an aggregate job. The request body is a `struct AggregateJob` where `prepare_steps` contains the leader's current-round prepare messages, and `round` is the number of the current round. The response is a `struct AggregateJob` where `prepare_steps` contains the helper's next-round prepare messages. The response's `round` field will be one more than the request's `round`.
+
+We use a POST here because this request is _not_ idempotent. Suppose the leader sends `POST /tasks/task-id/aggregate_jobs/aggregation-job-id` where the body has `round = 1`, and that this is successfully handled by the helper. Besides generating the response, this will have the side effect of advancing the state in the helper to round 2. If the leader were to resend the `POST /tasks/task-id/aggregate_jobs/aggregation-job-id` with `round = 1`, then the request should be rejected by the helper, because it has advanced to the next round.
```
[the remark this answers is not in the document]
tgeoghegan: I like how this proposal eliminates an extra `GET` roundtrip. My intuition though is that the helper has to hold on not just to a hash of the previous request, but possibly also its response from the previous round. I've got to try to sketch out some failure scenarios in many-round VDAFs to see how this plays out.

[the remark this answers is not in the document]
tgeoghegan: I think I get it now -- your scheme assumes that the leader would record its state before trying to advance state in the helper, which means that in the case of a crash, the helper can never be ahead of the leader.

I think that same assumption eliminates the need for the `GET` in the `round` solution. Suppose the helper is on round _n_. After recovering from a crash, the leader sends `POST` to the aggregate job with either:

- `round = n`: helper has seen this message before the leader crashed and sends its current state (round _n+1_ prepare messages). Protocol continues from there. 
- `round = n+1`: helper advances its state to the next round. Protocol continues from there.
- any other value of `round`: the leader and helper are too far out of sync. The aggregate job must be abandoned. The leader could recover by creating a new aggregate job with the same parameters, essentially going back to the first round. 

So I think either notion (request hash or round number) can achieve the functionality we want. Question is: which is "better"?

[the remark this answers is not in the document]
branlwyd: > Where would the helper hashing the leader's PUT request help?

The only purpose of storing the (hash of) the previous request is to ensure that the response the Helper gives is coherent (i.e. matches the request).

While you are correct that the Leader should (if properly following DAP) always send the same request, there is nothing practically ensuring this if the Helper doesn't check somehow. A sufficiently-buggy Leader could send any request at all, and without the hash (or some other mechanism) to allow the request to be verified, the Helper would give the answer a non-buggy Leader would have received rather than the error the actually-buggy Leader should receive.

If folks don't think it's valuable to avoid this scenario, storage of the (hash of) the request can be dropped. IMO, I wouldn't recommend this (it's unlikely to come up in practice, but if it does, the results will likely be _very_ confusing to operators).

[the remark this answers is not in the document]
tgeoghegan: Yeah, I think you're on to something. I had hoped that cases where the leader is sending the wrong prepare messages after a restart would be caught by failures in the helper's `Vdaf.prep_next` call, but at best they'd be caught on the next round.

What do you think about considering this fault recovery case separately from the resource oriented API change? The PR that introduces that into the protocol text will already be quite large, and this scenario that we are discussing is nuanced enough that it would benefit from having a spotlight shone on it, which is to say, a PR dedicated to making just that change.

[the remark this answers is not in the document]
branlwyd: Sure -- I think the issues are separable. Probably the resource-oriented API should be specified first, then can be tweaked if necessary to achieve idempotency or otherwise solve the fault-recovery issue.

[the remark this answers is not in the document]
branlwyd: It occurs to me it might wiser to leave it up to the implementation: the fault-recovery logic `MUST` check the request's round number, and `SHOULD`(?) check the entire request [worded to allow checking equality or by hash or some other technically-approximate method]. WDYT?

api-resources.md
wangshan: I misunderstood idempotency, your explanation is correct: it's not about the response being the same but the state not changed by multiple calls.

api-resources.md
```diff
@@ -0,0 +1,414 @@
...
+
+The `task_id` query parameter is optional, as described in https://www.ietf.org/archive/id/draft-ietf-ppm-dap-02.html#section-4.3.1.
+
+##### Alternatives
+
+An HPKE config has an ID, so we could use a path `hpke_configs/hpke-config-id`. However, a client should never need to get any HPKE config except for the "current" one, so does it make sense to expand the API surface with a path param for HPKE config ID?
+
+`GET /hpke_config` could redirect to `/hpke_config/current-hpke-config-id`, which would make this mostly transparent to clients, assuming they handle 3xx redirects properly.
+
+Instead of providing the task ID as a query parameter, we could also include it in the path, like `/tasks/task-id/hpke_config`, on the premise that the task ID is part of the resource's permanent identity. However, it's awkward to solve the global HPKE config case with this formulation. We could include a special task ID "global" so that clients could query `/tasks/global/hpke_config`.
```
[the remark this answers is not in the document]
wangshan: @tgeoghegan returning global HPKE config when taskID unrecognized should work. But I prefer the `/hpke_config` endpoint, because it's a resource identified by hpke config id, task_id as a query parameter seems appropriate here.

tgeoghegan:
I've been putting together some sequence diagrams of the sub-protocols to illustrate how the API flows, where side effects are incurred by POST requests, and some error recovery. Disclose any triangle for biiiiiiiig sequence diagram PNGs and also links to editable source diagrams.
  
<details>
  <summary>Aggregate sub-protocol with leader crashes</summary>

  image
  [editable source](https://sequencediagram.org/index.html#initialData=CoSwLgNgpgBAggcwQJygghmWBnArgIxgAdkB7MUgY1IhgHdwALGadAEymRkuXW0agAoQUXTIwISiFEA7MDAAyUdpxFiJU2fIASUCEVWDk+UgA8YpAG6dFyjsgA0u-ZwBcAZTCZY74HABKwIIm5lY2SirIHpQCbLjQMOgyiUioGBKkyQBWpPgAOjJJbDBxvPgQAJ7EnNgg2PLg2DAAkgAiDgUgMpQQuBzFqESk4tgOMFBglAB0ghH2ALQAfM4GUQAKAKrAMAD0XtgA1tg7AN77B-MgbAC+O+ipaJggmQD6OfjHAGakpMFmFtYuCs3ABhUgAWyIuCwMAAjPUYGRcDJiiQoEQCuCoNhsPdsYkUSVcGVKgVVrUEWABOCjCEATZgVFPN4YP4APIbAByrThf1CgNskQAxKg2K45jYeHwBNhBIz5iKoGKAMymCVReHyJGEtGiVAwLE4vGyulhLjq8V2GyoerqUbcTJ4CBgbBkmp1LDFW1YWn-M2C+weLww3wBILqpaM1ybbZ7PhHU7nS43O4PdLPGRvXJfH58+lAvSrVyM7gCShHarICmemDeqAdGSoagIGQgABe+PuCBg7xg31IYxtQxRrpkDCptYoorhCO1qMGmOxuIQ2Llhc4SwtmsRpGR8-RYlghuXq4jy3X6zZvl252OZ3jyduXbST1e7xzpAKACFSGwKq4dz3ABeWEG23OcYHwMh2EoPh5F1Q8DSXY0839KNqEhaFYAAJkJCDdUXI0VyaIoiRJCo3SrD0YCpKBwQKAoAB5qAgYZXCFYYkhXRZ3CuWAoE+T4oEoMBGJ2FjhkWACS1gmQCjYOpYOQYpPhAKstV3HVBkPQiTxIwlqHiAYBNwbAoAKCha04SxJBwAgzIAR1wKA5AKGMb3jO8kyuJ801fTN3x2ftfX5BkLyDFl2S5HlsNQgV1UVMV1W4Xh+FXeVEtcVULVwgZNP3PUj2Q4i4vCK0omSm0vBGMZqBkJ0XUo6slUnbwQvzAM3GZGEou5XkzyjNYr1jW9EwfHzUxQR4MgC7Mgp+b9f3-QCURAsDZ3yyDoLYWCEQQ-VjxQ00BSjGSywrRpK2ar1g3rAozPxKlMBgcADUwGJ8Qg2E1xcZBN3K1xcpWgrEMOkrjrKyJXDBTCYSBiCoNIGC4LJQYkKIztCVKdBygomRyWo2iaWMP14oB7rYF6mLZnKyNwqG6840OLzxpTZ9pozLMPnmz8ZB-P8ALnIDsIbeHNsR5G9u0g7itXCGC1+1xVMKCB21gLooXkfhEIIlFiRxyoruo-aWo16FRyYiSog43gZG43iOHGQThNE8SaEk6SL24JISkUsQ2Hu4TMjyvdiGloqMdHUjsFIN6qlQT5TNgSyzOQGzKHM+r7KgJyXPkBngAKJmE3vQ5H0ml8Zq5j8ClQXP6iacdmCF76SdChWiwpmAADFmk5Zp3G0ABRVoftWf6ofcHPnO6WBSE+GA1nDzx0QACmVupYgASna-0LS73v+8HkfBCAA)
</details>


<details>
  <summary>Collect protocol with time interval query</summary>
  
  image
  [editable source](https://sequencediagram.org/index.html#initialData=CoSwLgNgpgBAwgewtAxmGAHATgsCVIwDu4AFjGCALawgB2YUWAbgIYQwCOArkwJ4AoARlZZKKECIYwAMlFYATJsNHjJraQAkoEDMpFiQEqekTIoaBFiFnUeLAFoAfHMVMAXAAUAqsBgB6MFYAZwBrYP8AbyCwhxAFAF9-AnM0EAQ6YIAdOgAlKB4oYPRgvQkAMxAimGCEGgpqWgYmNggBACMEAA8YBGYmWXklLHcULHlGGFYYFLt0uhgAKwR2nI0FGAVuLFZ2iD5MJmCQYphwHIAKcGCYb1yASQAaM7oUCG4lDfGMKzBgx5y9DeHygG1YAHNweNwaxKBklitggBKASuYbOWwWezuTTAYCeGAAZgADISYOMFCBxmk6OCKAgcoEQuEojFQnFEskkHMMhFyggEB1ur1+lh4NysVZ3AARba7faHLDHYo5WZY24PASYyyOFxDDyeADyAGU-EywhFosyOUk1Wlef5+YK0UwMRKdTi8QSAEzE70CVhpNiTbX2AR0XCwLAgcGkdAIcri1LYxXpSkodgKn7Ieh0u3oO5PHLgqBgSi0mC4-EwX3e8lFH6ZIqo-W67S6A2+AJsy1sm3+CFQqAwxgAfWCpFEzc6PT6A3behGsp2ewOi+VYDWkOhsNgE6nCNWdCufw1TxewM+9Z+YmydCB7yvg53cIWy3ayIEC9dercIyrBIkmSFJUliub0oyPasta8RJM+w67uOk7jHyApCrOoqDH+MpyquiobluQ4jnuyGwIWAgQAKGAwNwDAgBw8HETUpGHmcNwKBkUAtn+zjfiMADiACiZpQVasSwQO24IWO+4oY6aF8c4Lr-l6NZ+gIUB0Ao3Hok4fHuEJInMr2MGcoxiGyUU8mCopv7DJ61a+sSOR8VMUlMZZAhKIGlDBrAoZWDY7r2EprZeCaRkWtB4mcvm8yoc6rZusmUoAWpzl0KG8z1sE3AQGAAhAA)
</details>

<details>
  <summary>Collect protocol with fixed batch query</summary>

  image
  [editable source](https://sequencediagram.org/index.html#initialData=CoSwLgNgpgBAwgewtAxmGAHATgsCVIwDu4AFjAGYgAeUAJjAEYCGYK5AjgK5RYCeAKAEZmWMCBQgRAO3QAZKMzq9ho8ZJnoAElAgYVIsRKnNZ8JKjxYhjBNRgIAbrxgKlvAFwARKAGcQAObSrLC+CAC2sFzSBMhQaPQAOtIsbOTMvjAARChcWFhQslkAdDBeecyMEHzJ+lj+vuhEpBLkqewwIJk5eQVFxQKIcWgIWAC0AHxuylgeAAoAqsAwAPRgGQDWvisA3uu+G2MgdAC+K7GWIAjSvskASlDcfui++pJUfjC5+YVgY+2kAS2exOFzTTxwAohGDML4WeLia4wABWCEYlFGyWy3z6YCyTFYHVMDDoFSqNWkdQa6HAMAAFOBMgs7gBJAA0nRiEC4yjoWIKGFGYF8HJAXJ59BhAQCBQCrCu0mSqMYvgAlANweMJkNLKMPFpgMA5jAAMwABhNMAKdBABTQYoCMDwyTWm22e02R1O53h9uu2woCAQQLsDmcWHMwys3jJ1UwvGpyQuCJgzJZg19VkmmvmAHkAMrLV0Hd37Q7HM7Jv03FaB4PAsNgxQzbx+QLBMChCKwXCkFzRKudvkpQnpbo434lMqxvjx+pdMDJZqtAlpTrj3qT4rJZIAHlieoAxKNTAEoBN88dYFAKBQEbufRBRhMPDBfFwVY8eGYCk9GrdZAQGBFiLMtS09CtH0uf1kkDCMJzMAFiBAZAYD4EBdBJEBb14X5kgKd8IDAARNUmHUET1A0jRgAAmM0aIEZh7UcaFyJGaxpFwWAsECUh0AQChI11WY5yuG0UGYZBZ0FZAHThKNU1ZNlkjPMBxGkR0qONOiaKtPxBRuPwSObXhJh0PRPBA1YwN2MsvTOZhpVlEIAH1fFIUQjIbUEI3MuoYywSo4ypBdkkcmUoDlTs3w8goUTRZIGWFRT2U5FBuV5PTBTEACxXSiUGHC5zEWkeKVVVAQ-NMqYTNmLTTQtPSbTtdTHWdaRiy2WyIO9IrItc9zPIDIMQxBcNXFqgKgukhNQuCJz+uiwa4rTAQnwQDAYGicQIClCKotCWLYGVdcYDoa4oGM9wtSq2YAHEAFFQLdbqDnslY+oOtyjuG4NbuzSb6p0gRCjoK6ZjM3R-Me56S1e8teoWr7lr8WsRv+mrrv1Q1tLNM1kluvbisOzyBGUJjxBY6K2KsIQadGAGsbmAtYa6j03sgwcFV+8HqvpuqcdovHkhphU9MI4igA)
</details>

tgeoghegan:
During the PPM WG session at IETF115, @martinthomson suggested that DAP clients shouldn't be allowed to choose report IDs. So instead of `PUT /tasks/{task-id}/reports/{report-id}`, clients would do `POST /tasks/{task-id}/reports`. The server would choose a report ID and store the report under it. This has to be a POST (I think) because it has the side effect of generating a report ID.

I like the idea of servers being able to synthesize the report IDs totally internally, but we need to think through how we would enable clients to retry report uploads without the risk of duplicate submissions.

I think it might be a bit easier to do this for aggregation jobs, since the partial batch selector might suffice to coalesce requests together.

martinthomson:
In a design that uses POST, the server takes more responsibility for managing duplicate reports. the idempotency key draft might provide some help there, but there isn’t really a magical answer.

tgeoghegan:
In the latest commit, I've taken a stab at making the helper responsible for constructing opaque aggregation job URIs. I think this may also be possible for reports, if we can convince ourselves that there is some means of de-duplicating uploaded reports without a report ID. For instance, the leader could hash the input share, or we could use an [idempotency-key header](https://datatracker.ietf.org/doc/draft-ietf-httpapi-idempotency-key-header/).

tgeoghegan:
<details>
  <summary>Updated aggregate sub-protocol sequence diagram, illustrating an n-round VDAF and better highlighting side effects.</summary>

  image
  [editable source](https://sequencediagram.org/index.html#initialData=CoSwLgNgpgBAdgagIwFoBOB7ArnAJjAQwHMi0oiCxYBnLAIxgAdMwMBjDCGAd3AAsY0ArihoYbNAWp8o1AFBzGBNGBBsQSuGBgAZKMNGLlq9Zu0AJKBEaG5BBhgBuo3fpFoANJeuiAXAGUwSlh-YABBACVgOToMAA8YJxc9AzQAthlcLGhCOEISMgpVDDyAKww6AB04AjwYLMk6CABPJlFqEGptcGoYAEkAEQ9qkDg2CCwRfDJGDBVqDxgoMDYAOjkU9xQAPm8bNIAFAFVgGAB6IOoAa2ozgG9Lq5QQXABfM+JSckoQEoB9cp0eSxBJJMR7PwAYRKXTQWDY2lq+S+RV+ZQqMCOET61WhAFtGFgqDAkF0YJgcPhmFBGNU8bJqMRZLl8A17C1qvsOmSwDI8XI0CDEs5wVZ9gEgsSIgB5I4AOQGJJi8WFyTcogAxGRcL5Ni4JFIZPIIWgUFqoDqAMxxPVpczAYAHGAAJgADEgYJCyMFcNUdOwfiVfDA7oCUFg0CBXtUAEIYXDNYOk7QUurUpRkGD06iMoiyZWgkWuVK69ViMhdYwLcQw7Jgaic9qdKj4StUAVCsHF9wS4IwUKRaK2nYm3zHU4XKQ3e6PZ5vD4Fb7FOAAirAlVd0cm8QyNg3NpobktmBtqDDOB4ygZZklGB0K8CahWKAIuYs6qfJjKAh4xYV2Z4L0vC8tUXRzBaJJkqmVIzFmDJMsaYqiDstq+Pajouu6nrei2foBsuwahhU4aRtGcBxgmwZwCgvLktgaYzMosDZrm+adkWqH4oSxLJnRlJ3pgwhsFIYCcoxmYsQhLL1BG7LNI2h7NjAvJQPygobhxZa9lKsoKkqw67EhhzSqEIZhhGUaxvGiZ8XgAC8SDnrx0ECRgQkiUw4nMfBebroWLijhwBJErAzp1C56Z0j5zK1KyslNPJcBckpKn8vYGkBUZGpzLUeYBC8sBQAAZkVL5gMG27CXA1S4J0wloNMNJzNo0hMb0sXVBw2SNUVWBPspGAnqIjhqFAoH0E+ACOWBQFoMAHCZpxEXQJGWXAwECNBDkdhloo+GkgR9jK8qKs6cgmihWlhdM9EwTSTFwTmCFyHZdn+FAIoEFweLgbZuDtZmUBxC+IW4K9O3+WItpeEZ2mwMdelwBsZYjrDC2mctq1kRRNlbXA57UbRLl0IJuDCWS6YPZJvkFqqe3ikF3GwIgqBE7dnk0lFT2+dJbIJQpR7KXydjsZl+3ZZIcB5f4BVLCVZUVUZ4i1DVdXKPg1Gs5t7OU2QXOse1dRdRAPV9WNWiDU+aAjWwNATVA02zaJcDo0t5mkTw-B-XZmsQ3TMCjodOknfAyC012trmjqtriJI0j5hdUe+NaqEszR2v8br3nc2xu3dn4McVkE8yLBwcC0BA9YC82EGnn7EdaUH8O6YqSMGaOrtmcRFnY9ZwZ4+e6d-a57kU15j0G+HRZbkr157r0PQHketeSmeoFQMyvKUDA4BZg+zIuUjF3bKnyBDxF4-U7nkP52kXEhaHWvDyTblkyJYk0hPUmxTJjQcklTYeTC3UjfVCTcYAI1bmHduaNFpdxWj3KylFvYswJmfNm-EX6j20FnL+NNRb0z8EVUYX0QAAC9YCjG4iePgD1Ip4Hii0JeSks74CoUSeQ6Ub4mglrlKA+URBy1KgiRW+1lZ5FqtQeqvo4BpwwQxe6ElooNhqHUagg1LytDIL1fqrAhrW1GuNIEDsZpzU7pjRBcAyCOy6EBL2eMw4gP9oHVeMAABifQ5R9H8OYAAogMc6RlLolneo7MYsAMBFXml5QINIAAUxC4CdEyAASnrppEJriPFeJ8f4uQQA)
</details>

tgeoghegan:
The discussion here has run its course and the changes are being pursued in #395, #398, #399 and #400. I'm closing this PR because we have no intention to merge it and to declutter the repository's open PR list.

PR #370, opened 2022-10-19
Align media type of aggregate request/response

easonchan1213:
In IANA all media types seem to be of "application" type instead of "message" type, this PR is to align the mismatch in the protocol description and adapt to the media types specified in IANA.


cjpatton (approved):
Nice catch!

PR #372, opened 2022-10-31
Use structs for AAD encoding

### LINKED ISSUE #371: Clarify encoding of AAD
Currently, the AAD for input shares is defined as `task_id || metadata || public_share`. The variable `public_share` is narratively introduced in 4.4.1.3 as "the public share sent to each aggregator", referring to a part of the report share. Two implementations interpreted the construction of the AAD differently, as either [including](https://github.com/cloudflare/daphne/blob/61ab97f0c07f9f4bbdb13ec666cebd22febf12ec/daphne/src/vdaf/mod.rs#L184) or [not including](https://github.com/divviup/janus/blob/42873cddb666dce39d27a2342adc3d7232b14eda/core/src/hpke.rs#L37-L49) a 32-bit length suffix preceding the public share.

While it would likely be okay to have a single variable-length byte array in the AAD without a length prefix, as the resulting AAD would still be unambiguous, I think it would be prudent to change this to explicitly include a length prefix, to simplify future changes.

We should define new structs for both kinds of AAD strings, and specify that the AAD is the result of encoding the corresponding struct. For example:

```
struct {
  TaskID task_id;
  ReportMetadata metadata;
} ReportShareAAD;
```

divergentdave:
This revises the sections dealing with encryption and decryption of input shares and aggregate shares, rewriting the construction of each AAD input to be in terms of RFC 8446-style structs instead of byte string concatenation. This fixes #371.

### DIFF
--- a/draft-ietf-ppm-dap.md
+++ b/draft-ietf-ppm-dap.md
@@ -810,14 +810,23 @@ as specified by the VDAF. It then encrypts each input share as follows:
 ~~~
 enc, payload = SealBase(pk,
   "dap-02 input share" || 0x01 || server_role,
-  task_id || metadata || public_share, input_share)
+  input_share_aad, input_share)
 ~~~
 
 where `pk` is the aggregator's public key; `server_role` is the Role of the
-intended recipient (`0x02` for the leader and `0x03` for the helper), `task_id`
-is the task ID, `metadata` is the report metadata, and `input_share` is the
-Aggregator's input share. The `SealBase()` function is as specified in {{!HPKE,
-Section 6.1}} for the ciphersuite indicated by the HPKE configuration.
+intended recipient (`0x02` for the leader and `0x03` for the helper),
+`input_share` is the Aggregator's input share, and `input_share_aad` is an
+encoded message of type `InputShareAad`, constructed from the same values as the
+corresponding fields in the report. The `SealBase()` function is as specified in
+{{!HPKE, Section 6.1}} for the ciphersuite indicated by the HPKE configuration.
+
+~~~
+struct {
+  TaskID task_id;
+  ReportMetadata metadata;
+  opaque public_share<0..2^32-1>;
+} InputShareAad;
+~~~
 
 The order of the encrypted input shares appear MUST match the order of the
 task's `aggregator_endpoints`. That is, the first share should be the leader's,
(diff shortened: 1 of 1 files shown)


cjpatton (approved):
Solid implementation, and a clear win 👍

tgeoghegan (approved):

draft-ietf-ppm-dap.md
```diff
@@ -1636,26 +1646,36 @@ done as follows:
...
 enc, payload = SealBase(pk, "dap-02 aggregate share" || server_role || 0x00,
-  AggregateShareReq.task_id || AggregateShareReq.batch_selector, agg_share)
+  agg_share_aad, agg_share)
 ~~~
 
 where `pk` is the HPKE public key encoded by the collector's HPKE key,
-`server_role` is `0x02` for the leader and `0x03` for a helper.
+`server_role` is `0x02` for the leader and `0x03` for a helper, and
+`agg_share_aad` is constructed from the `task_id` and `batch_selector` fields
+of the `AggregateShareReq`, with the following format:
```
ekr, line 1655:
```suggestion
`server_role` is `0x02` for the leader and `0x03` for a helper, and
`agg_share_aad` is a value of type `AggregateShareAAD` with its values set from the corresponding fields of the `AggregateShareReq`.
```

draft-ietf-ppm-dap.md
```diff
@@ -1636,26 +1646,36 @@ done as follows:
...
-request, and `server_role` is the role of the server that sent the aggregate
-share (`0x02` for the leader and `0x03` for the helper). The value of
-`batch_selector` is computed by the Collector from its query and the response to
-its query:
+where `sk` is the HPKE secret key, `server_role` is the role of the server that
+sent the aggregate share (`0x02` for the leader and `0x03` for the helper), and
+`agg_share_aad` is an `AggregateShareAAD` message constructed from the task ID
+in the collect request and a batch selector. The value of the batch selector
+used in `agg_share_aad` is computed by the Collector from its query and the
+response to its query:
```
ekr, line 1678:
```suggestion
response to its query as follows:
```

ekr (changes requested):
Generally LGTM.

draft-ietf-ppm-dap.md
```diff
@@ -810,13 +810,22 @@ as specified by the VDAF. It then encrypts each input share as follows:
...
+`input_share` is the Aggregator's input share, and `input_share_aad` is an
+encoded message of type `InputShareAAD`, constructed from the same values as
+the corresponding fields in the report:
+
+~~~
+struct {
+  TaskID task_id;
+  ReportMetadata metadata;
+  opaque public_share<0..2^32-1>;
+} InputShareAAD;
```
wangshan, line 827: nit: if we follow he same camel case convention of types, this should be `InputShareAad`. Same applies to `AggregateShareAad`

wangshan (approved):

divergentdave:
I squashed this back into two commits and then rebased to fix some conflicts.

cjpatton:
@ekr would you mind taking a quick look?

ekr (approved):
LGTM

PR #373, opened 2022-11-02
Adopt must-implement semantics for extensions

### LINKED ISSUE #369: What is the extension processing model
Are unknown/unsupported extensions ignored?

cjpatton:
Closes #369.

### DIFF
--- a/draft-ietf-ppm-dap.md
+++ b/draft-ietf-ppm-dap.md
@@ -772,7 +772,6 @@ structured as follows:
 struct {
   ReportID report_id;
   Time time;
-  Extension extensions<0..2^16-1>;
 } ReportMetadata;
 
 struct {
@@ -801,24 +800,37 @@ task ID and report "metadata". The metadata consists of the following fields:
   timestamp cannot be used to link a report back to the Client that generated
   it.
 
-* A list of extensions to be included with the report. (See
-  {{upload-extensions}}.)
+To generate a report, the Client begins by sharding its measurement into input
+shares as specified by the VDAF. It then wraps each input share in the following
+structure:
 
-To generate a report, the Client first shards its measurement into input shares
-as specified by the VDAF. It then encrypts each input share as follows:
+~~~
+struct {
+  Extension extensions<0..2^16-1>;
+  opaque payload<0..2^32-1>;
+} PlaintextInputShare;
+~~~
+
+Field `extensions` is set to the list of extensions intended to be consumed by
+the given Aggregator. (See {{upload-extensions}}.) Field `payload` is set to the
+Aggregator's input share output by the VDAF sharding algorithm.
+
+Next, the Client encrypts each PlaintextInputShare `plaintext_input_share` as
+follows:
 
 ~~~
 enc, payload = SealBase(pk,
   "dap-02 input share" || 0x01 || server_role,
-  input_share_aad, input_share)
+  input_share_aad, plaintext_input_share)
 ~~~
 
 where `pk` is the aggregator's public key; `server_role` is the Role of the
 intended recipient (`0x02` for the leader and `0x03` for the helper),
-`input_share` is the Aggregator's input share, and `input_share_aad` is an
-encoded message of type `InputShareAad`, constructed from the same values as the
-corresponding fields in the report. The `SealBase()` function is as specified in
-{{!HPKE, Section 6.1}} for the ciphersuite indicated by the HPKE configuration.
+`plaintext_input_share` is the Aggregator's PlaintextInputShare, and
+`input_share_aad` is an encoded message of type `InputShareAad`, constructed
+from the same values as the corresponding fields in the report. The `SealBase()`
+function is as specified in {{!HPKE, Section 6.1}} for the ciphersuite indicated
+by the HPKE configuration.
 
 ~~~
 struct {
@@ -862,14 +874,20 @@ provide for some small leeway, usually no more than a few minutes, to account
 for clock skew. If the leader rejects a report for this reason, it SHOULD abort
      the upload protocol and alert the client with error "reportTooEarly".
 
+If the Report contains an unrecognized extension, then the Leader MAY abort the
+upload request with error "unrecognizedMessage". Note that this behavior is not
+mandatory because it requires the Leader to decrypt its input share.
+
 ### Upload Extensions {#upload-extensions}
 
-Each Report carries a list of extensions that clients may use to convey
-additional, authenticated information in the report. [OPEN ISSUE: The extensions
-aren't authenticated. It's probably a good idea to be a bit more clear about how
-we envision extensions being used. Right now this includes client attestation
-for defeating Sybil attacks. See issue#89.] Each extension is a tag-length
-     encoded value of the following form:
+Each PlaintextInputShare carries a list of extensions that Clients use to convey
+additional information to the Aggregator. Some extensions might be intended for
+all Aggregators; others may only be intended for a specific Aggregator. (For
+example, a DAP deployment might use some out-of-band mechanism for an Aggregator
+to verify that Reports come from authenticated Clients. It will likely be useful
+to bind the extension to the input share via HPKE encryption.)
+
+Each extension is a tag-length encoded value of the following form:
 
 ~~~
 struct {
@@ -883,8 +901,12 @@ enum {
 } ExtensionType;
 ~~~
 
-"extension_type" indicates the type of extension, and "extension_data" contains
-information specific to the extension.
+Field "extension_type" indicates the type of extension, and "extension_data"
+contains information specific to the extension.
+
+Extensions are mandatory-to-implement: If an Aggregator receives a Report
+containing an extension it does not recognize, then it MUST reject the Report.
+(See {{early-input-share-validation}} for details.)
 
 ### Upload Message Security
 
@@ -1013,6 +1035,7 @@ enum {
   vdaf_prep_error(5),
   batch_saturated(6),
   task_expired(7),
+  unrecognized_message(8),
   (255)
 } ReportShareError;
 ~~~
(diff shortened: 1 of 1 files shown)


draft-ietf-ppm-dap.md
```diff
@@ -852,14 +867,21 @@ provide for some small leeway, usually no more than a few minutes, to account
 for clock skew. If the leader rejects a report for this reason, it SHOULD abort
      the upload protocol and alert the client with error "reportTooEarly".
 
+If the Report contains an unrecognized exgension, then the Leader MAY abort the
+upload request with error "unrecognizedExtension". Note that this behavior is
+not mandatory because it requires the Leader to decrypt its input share.
```
cjpatton, line 872: This is a bit awkward. The requirement I'm trying to meet here is: The Leader shouldn't have to have its HPKE decryption keys available during the upload sub-protocol. One way to avoid this would be to move the extensions out of the plaintext.

draft-ietf-ppm-dap.md
```diff
@@ -441,6 +441,7 @@ in the "type" field (within the DAP URN namespace
 | unrecognizedMessage        | The message type for a response was incorrect or the payload was malformed. |
 | unrecognizedTask           | An endpoint received a message with an unknown task ID. |
 | unrecognizedAggregationJob | An endpoint received a message with an unknown aggregation job ID. |
+| unrecognizedExtension      | An endpoint received a message with an unrecognized extension. |
```
tgeoghegan, line 444: Do we need a new error type here, as well as the corresponding prescription elsewhere that an implementation MUST provide this error type in some scenario?

The question I think we should apply to all error types (including the existing ones) is whether any protocol participant actually behaves differently than they would if they just got some kind of generic "Bad Request" error.

Suppose a leader rejects a report because it contains an unrecognized extension. If the client sees a problem document with `unrecognizedExtension`, what is it supposed to do? Try again, trying different subsets of the extensions until it finds a combination the leader accepts? That doesn't make sense to me. If the client is configured to send some list of extensions, but the leader is configured to accept a different list of extensions, then I don't think the protocol can recover in-band. The client and leader operators would need to dig into why the configuration disagrees and take some action that is outside the scope of DAP.

That leads me to conclude that an error of type `unrecognizedMessage`, in response to which the client simply stops trying to upload that report, should suffice for the case of unrecognized extensions. The detail field of the problem document constructed by the server can contain detailed information on what's wrong with the message that aids operators in debugging, but I don't think the protocol needs to dictate a particular error code.

draft-ietf-ppm-dap.md
```diff
@@ -852,14 +867,21 @@ provide for some small leeway, usually no more than a few minutes, to account
 for clock skew. If the leader rejects a report for this reason, it SHOULD abort
      the upload protocol and alert the client with error "reportTooEarly".
 
+If the Report contains an unrecognized exgension, then the Leader MAY abort the
```
tgeoghegan, line 870:
```suggestion
If the Report contains an unrecognized extension, then the Leader MAY abort the
```

draft-ietf-ppm-dap.md
```diff
@@ -852,14 +867,21 @@ provide for some small leeway, usually no more than a few minutes, to account
...
 ### Upload Extensions {#upload-extensions}
 
-Each Report carries a list of extensions that clients may use to convey
-additional, authenticated information in the report. [OPEN ISSUE: The extensions
-aren't authenticated. It's probably a good idea to be a bit more clear about how
-we envision extensions being used. Right now this includes client attestation
-for defeating Sybil attacks. See issue#89.] Each extension is a tag-length
-     encoded value of the following form:
+Each Report carries lists of extensions that Clients use to convey additional,
+authenticated information in ts Report. Extensions that need to be handled by
```
tgeoghegan, line 877:
```suggestion
authenticated information in the Report. Extensions that need to be handled by
```

draft-ietf-ppm-dap.md
```diff
@@ -852,14 +867,21 @@ provide for some small leeway, usually no more than a few minutes, to account
...
+all Aggregators are included in the ReportMetadata; extensions intended to be
+handled by a specific Aggregator are included in the PlaintextInputShare. (For
```
tgeoghegan, line 879: I worry that putting extensions in two different places will lead to complexity in the implementations and ambiguities, too: what if the same extension type appears in both places but with different values? Either DAP or the specification of each extension has to spell out how to resolve such conflicts.

Unless I'm missing something, the reason to allow extensions in two places is to avoid transmitting the same extension in each aggregator's ciphertext. If I'm right and this is just about that performance optimization, could we avoid doing this until operational experience shows it to be a problem worth the extra complexity?

draft-ietf-ppm-dap.md
```diff
@@ -1222,6 +1250,9 @@ following generic checks.
 1. Check if the report's timestamp has passed its task's `task_expiration` time,
    if so the Aggregator MAY mark the input share as invalid with the error
    "task_expired".
+1. Check if the Report contains unrecognized extensions, either in
+   ReportMetadata.extensions or PlaintextInputShare.extensions. If so, then the
+   Aggregator MUST abort with "unrecognized_extension".
```
tgeoghegan, line 1255: Same observation here that I had about the problem document types: it's not clear to me how the leader will handle this differently than some more generic error code.

tgeoghegan (changes requested):

draft-ietf-ppm-dap.md
cjpatton: We already have this problem for ReportMetadata.extensions, but I agree that de-duplicating in two places adds a bit more complexity. Basically instead of:

1. Parse report
2. De-duplicate Report.metadata.extensions
3. Decrypt input share

We have to do

1. Parse report
2. Decrypt input share
3. De-duplicate Report.metadata.extensions and plaintext_input_share.extensions

> Unless I'm missing something, the reason to allow extensions in two places is to avoid transmitting the same extension in each aggregator's ciphertext. If I'm right and this is just about that performance optimization, could we avoid doing this until operational experience shows it to be a problem worth the extra complexity?

Given how important a consideration bandwidth has been, I think it's worth adding a bit of complexity to avoid increasing overhead here. This is particularly concerning for the taskprov extension, FWIW.

draft-ietf-ppm-dap.md
cjpatton: I think it's better to reserve unrecognizedMessage for true parsing errors, but I take your point. There's no reason to spell out this extra error case because there's no automated response to it.

draft-ietf-ppm-dap.md
cjpatton: Here I have a different view: Knowing what went wrong is useful telemetry because if we observe a spike in rejection rates, we'll want to investigate what's going on.

draft-ietf-ppm-dap.md
cjpatton: I make the distinction between unrecognized extension and parsing error because it's possible to parse an unrecognized extension without knowing its structure. This is by virtue of the length tag.

draft-ietf-ppm-dap.md
```diff
@@ -804,19 +805,33 @@ task ID and report "metadata". The metadata consists of the following fields:
...
+structure:
+
+~~~
+struct {
+  Extension extensions<0..2^16-1>;
+  opaque payload<0..2^32-1>;
+} PlaintextInputShare;
+~~~
+
+Field `extension` is set to the list of extensions intended to be consumed by
```
divergentdave, line 819: nit: field name mismatch
```suggestion
Field `extensions` is set to the list of extensions intended to be consumed by
```

draft-ietf-ppm-dap.md
tgeoghegan: I maintain that for now, we should use `unrecognizedMessage`, and perhaps rename that error to just `badRequest` or something more generic. In particular I don't think this PR should add new error types.

draft-ietf-ppm-dap.md
tgeoghegan: Yes, I buy that we need a way for implementations to diagnose problems. In this case, you're arguing that you want the leader to be able to tell why a report share was rejected by the helper? This is handled reasonably gracefully anywhere we use HTTP problem documents because the `details` field allows adding arbitrary information for use in logging or metrics, but here we only have a single error code to insert in the list of `PrepareStep` structures. Maybe we could change `ReportShareError` so that instead of being just an enum, it's a structure that contains a value of the current `ReportShareError` enum and a `details` field that can optionally provide an implementation specific integer error code? Then the protocol wouldn't have to say anything specific about unrecognized extensions but your leader could still get something to track metrics on, and then ask your peer operating the helper why you're suddenly getting so many errors with `details = 18674749`, or whatever error code their implementation assigns to this case.

draft-ietf-ppm-dap.md
cjpatton: Done!

draft-ietf-ppm-dap.md
cjpatton: Hmmm, that might be OK. I'm sort of used to enumerating different error types, based on my experience with TLS. (See ["AlertDescription"](https://datatracker.ietf.org/doc/html/rfc8446#section-6). Though I take your point that we should do our best minimize these.

One idea is to change "unrecognized_extension" to "unrecognized_message" and let this be a catch-all for messages that could not be handled ((to match our decision for upload flow). This is slightly odd, though, because there are cases where the Helper would not abort with "unrecognizedMessage" but would reject a report with "unrecognized_message".

tgeoghegan: The precedent of TLS 1.3 does seem like a compelling argument, but I wonder if there were specific use cases for each of those errors.

I'll propose this as a way forward for this change:

- In this PR, use the generic `unrecognized_message` error. The principle here is to do the simplest thing that solves the problem we have, and to keep this PR focused on its problem statement.
- File an issue to extend `ReportShareError` as I suggest to allow implementation-specific error information in the aggregate sub-protocol.
- File an issue about whether we should use more specific error codes, and do more research on whether it's a good idea. We could for instance reach out to the TLS WG to see how they feel in retrospect about the error codes in TLS 1.3.

draft-ietf-ppm-dap.md
tgeoghegan: I'm still concerned about this, but I'm not sure how to break the tie between us. I'm OK moving forward with this change as it is (especially since you helpfully filed #375), but perhaps we could take this question to the WG to get more opinions?

tgeoghegan (changes requested):

draft-ietf-ppm-dap.md
```diff
@@ -804,19 +804,33 @@ task ID and report "metadata". The metadata consists of the following fields:
...
+
+~~~
+struct {
+  Extension extensions<0..2^16-1>;
+  opaque payload<0..2^32-1>;
+} PlaintextInputShare;
+~~~
+
+Field `extensions` is set to the list of extensions intended to be consumed by
+the given Aggregator. (See {{upload-extensions}}.) Field `payload` is set the
```
divergentdave, line 819: nit: typo
```suggestion
the given Aggregator. (See {{upload-extensions}}.) Field `payload` is set to the
```

draft-ietf-ppm-dap.md
divergentdave: I think separate places for per-report and per-share extensions makes sense. We may want to add guidance in the extensions section or security considerations section for extension designers. Specifically, I think each extension should have to define whether it is allowable as a per-report extension, per-share extension, or both. If an extension implicitly relies on both aggregators seeing the same extension value, then failure to reject that extension inside `PlaintextInputShare` would be an easy implementation mistake to make.

cjpatton: One datapoint from IETF 115: There is not much enthusiasm for taskprov extension as-it-is, so I think it would be fair to disregard that particular use case as a consideration. That said I imagine there might be use cases for which shared extensions are useful.

Let's take this to the list, as @tgeoghegan suggests. Hopefully someone will give us a perspective that makes the decision tree more clear.

cjpatton:
Rebased and squashed.

draft-ietf-ppm-dap.md
cjpatton: I took the suggestion to use "unrecognized_message" here ... we can file issues for follow-up things once this PR merges.

tgeoghegan (approved):

cjpatton:
Rebased and squashed.

draft-ietf-ppm-dap.md
cjpatton: Update the PR to remove shared extensions.

cjpatton:
@tgeoghegan I've removed the shared extensions as discussed. Can you take another look and re-approve?

tgeoghegan (approved):
Noticed a typo, otherwise LGTM.

PR #374, opened 2022-11-02
editorial: Add referneces for `SealBase()` and `OpenBase()`

### LINKED ISSUE #368: SealBase is not defined
I only tried searching for it, but all I found were places that it was used.  This makes it difficult to determine which inputs are which.  Which is the AAD, which is the plaintext, and which is the nonce?  Please also *describe* the inputs in text.

Note that at least one of the code blocks that use this is too wide at 76 characters.

cjpatton:
Closes #368.

### DIFF
--- a/draft-ietf-ppm-dap.md
+++ b/draft-ietf-ppm-dap.md
@@ -816,7 +816,8 @@ enc, payload = SealBase(pk,
 where `pk` is the aggregator's public key; `server_role` is the Role of the
 intended recipient (`0x02` for the leader and `0x03` for the helper), `task_id`
 is the task ID, `metadata` is the report metadata, and `input_share` is the
-Aggregator's input share.
+Aggregator's input share. The `SealBase()` function is as specified in {{!HPKE,
+Section 6.1}} for the ciphersuite indicated by the HPKE configuration.
 
 The order of the encrypted input shares appear MUST match the order of the
 task's `aggregator_endpoints`. That is, the first share should be the leader's,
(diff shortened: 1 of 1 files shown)


draft-ietf-ppm-dap.md
```diff
@@ -816,7 +816,8 @@ enc, payload = SealBase(pk,
 where `pk` is the aggregator's public key; `server_role` is the Role of the
 intended recipient (`0x02` for the leader and `0x03` for the helper), `task_id`
 is the task ID, `metadata` is the report metadata, and `input_share` is the
-Aggregator's input share.
+Aggregator's input share. The `SealBase()` function is as specified in
+{{!HPKE}} for the ciphersuite indicated by the HPKE configuration.
```
tgeoghegan, line 820: The strings "SealBase" and "OpenBase" never actually occur in RFC 9180. Instead, [section 6.1](https://www.rfc-editor.org/rfc/rfc9180.html#section-6.1) describes "Seal<MODE>" and "Open<MODE>" and then enumerates the possible modes, among which is "Base". Just to make this a little easier on folks reading DAP, could we refer specifically to RFC 9180 section 6.1?

tgeoghegan (changes requested):

draft-ietf-ppm-dap.md
cjpatton: Sure, mr. pedantic-pants

cjpatton: (seriously, great idea.)

tgeoghegan (approved):

PR #380, opened 2022-11-16
Add extension points to aggregate and collect flows

cjpatton:
Based on #373 (merge that first).

Generalizes "upload extensions" to "protocol extensions" and lifts this section to a subsection of the protocol definition.

Adds a list of extensions to AggregateInitializeReq and CollectReq.

### DIFF
--- a/draft-ietf-ppm-dap.md
+++ b/draft-ietf-ppm-dap.md
@@ -689,6 +689,31 @@ of the aggregators is configured with following parameters:
 Finally, the collector is configured with the HPKE secret key corresponding to
 `collector_hpke_config`.
 
+## Extensions {#protocol-extensions}
+
+Some DAP messages include a list of protocol extensions that are used to modify
+the behavior the core protocol. Each extension is a tag-length encoded value of
+the following form:
+
+~~~
+struct {
+  ExtensionType extension_type;
+  opaque extension_data<0..2^16-1>;
+} Extension;
+
+enum {
+  TBD(0),
+  (65535)
+} ExtensionType;
+~~~
+
+Field "extension_type" indicates the type of extension, and "extension_data"
+contains information specific to the extension.
+
+Extensions are mandatory-to-implement: If a party receives a message containing
+an extension it does not recognize, then it MUST handle it as an error.
+Precisely how the error is handled depends on the context.
+
 ## Uploading Reports {#upload-flow}
 
 Clients periodically upload reports to the leader, which then distributes the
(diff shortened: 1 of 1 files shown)


draft-ietf-ppm-dap.md
```diff
@@ -1080,6 +1076,7 @@ struct {
 struct {
   TaskID task_id;
   AggregationJobID job_id;
+  Extension extensions<0..2^16-1>;
```
tgeoghegan, line 1079: I think you should explain how this relates to the `extensions` field in `CollectReq`. In particular, I think the only reason that a `CollectReq` would ever contain extensions is so that they can be relayed to the helper. If the collector wanted to send something special to the leader, it could send an extra header in the HTTP request that sends the `CollectReq`. A leader that wants to add something non-standard to an `AggregateInitializeReq` could do much the same thing. It's only the collector<->helper case that needs some kind of in-band extensions scheme, since they never communicate directly (yet).

draft-ietf-ppm-dap.md
```diff
@@ -1470,6 +1472,10 @@ The named parameters are:
 
 * `task_id`, the DAP task ID. If the Leader does not recognize the task ID, it
   MUST abort with error `unrecognizedTask`.
+* `extensions`, a list of protocol extensions ({{protocol-extensions}}). If the
+  Leader does not recognize one of the extensions, or if two extensions have the
+  same ExtensionType, it MUST abort the request with error
+  "unrecognizedMessage".
```
tgeoghegan, line 1478: Continuing from my remark above: I think we should at least recommend against the leader ever parsing extensions. HTTP already provides plenty of mechanisms for the collector to provide extra information or parameters to the leader out-of-band from DAP protocol messages. Extensions should only ever be needed so the collector can say something extra to the helper, and at that point we have to think about how the leader could forge extensions. Maybe there should be a recommendation that anything put into an extension by the collector needs to be authenticated so the helper can verify where it came from?

draft-ietf-ppm-dap.md
wangshan: Is this `extensions` meant to replace the `ReportShare.encrypted_input_share .payload.extensions` eventually? If not, what's the relation between these two extensions, can people introduce an extension only relevant to `AggregateInitializeReq` but not clients?

cjpatton:
I've decided this is probably not a good idea. In any case, it's not needed for the use case I have in mind (taskprov).

PR #382, opened 2022-11-17
Allow implementations to relax anti-replay requirement

### LINKED ISSUE #362: DAP-02 has stricter anti-replay requirements than DAP-01
Anti-replay is enforced in [Section 4.4.1.4](https://ietf-wg-ppm.github.io/draft-ietf-ppm-dap/draft-ietf-ppm-dap.html#section-4.4.1.4). Both DAP-01 and DAP-02 read:

> Check if the report has already been aggregated with this aggregation parameter. If this check fails, the input share MUST be marked as invalid with the error report_replayed. This is the case if the report was used in a previous aggregate request and is therefore a replay.

What's not fully specified, at least in this section, is what it means for a report to "have been aggregated (with this aggregation parameter)". For that we have been referring to the anti-replay section. [DAP-02, Section 4.5.7](https://www.ietf.org/archive/id/draft-ietf-ppm-dap-02.html#name-anti-replay) reads:

> To detect replay attacks, each aggregator keeps track of the set of report IDs pertaining to reports that were previously aggregated for a given task.

The report ID is defined as

```
uint8 ReportID[16];
```

On the other hand, [DAP-01, Section 4.4.6](https://www.ietf.org/archive/id/draft-ietf-ppm-dap-01.html#name-anti-replay) reads:

> To detect replay attacks, each aggregator keeps track of the set of nonces pertaining to reports that were previously aggregated for a given task.

The nonce is defined as

```
struct {
  Time time;
  uint8 rand[16];
} Nonce;
```

Thus in DAP-01 we are required to check for collisions of the timestamp+random value, whereas in DAP-02 we are required to check for collisions of just a random value (report ID). The latter is stricter, since it disallows repeated random values.

After discussing some implementation stuff with @MichaelScaria, I realized that this may have been an unintended regression. From a security standpoint, I think the DAP-02 requirement is better, since it forces the client implementation to generate a fresh report ID each time it generates a report. Otherwise it is easier to fingerprint reports from the same client. 

Does this violate assumptions in anyone's implementation? Where I can imagine this being problematic is if your implementation currently shards nonce sets into time windows. [Daphne 0.1.2](https://crates.io/crates/daphne) does this, but we have to amend this behavior to implement fixed-size queries anyway.

Reply to this thread if this causes an issue for you and you think we should reconsider this requirement.

cjpatton:
Closes #362.

Having to track all report IDs is a burden for sharding anti-replay storage across time. This change clarifies that, while tracking report IDs is sufficient, it is also sufficient to track any report metadata that contains the reprot ID.

### DIFF
--- a/draft-ietf-ppm-dap.md
+++ b/draft-ietf-ppm-dap.md
@@ -790,8 +790,9 @@ public share, but Poplar1 does not. See {{!VDAF}}.) The header consists of the
 task ID and report "metadata". The metadata consists of the following fields:
 
 * A report ID used by the Aggregators to ensure the report appears in at most
-  one batch. (See {{anti-replay}}.) The Client MUST generate this by generating
-  16 random bytes using a cryptographically secure random number generator.
+  one batch. (See {{input-share-validation}}.) The Client MUST generate this by
+  generating 16 random bytes using a cryptographically secure random number
+  generator.
 
 * A timestamp representing the time at which the report was generated.
   Specifically, the `time` field is set to the number of seconds elapsed since
@@ -860,7 +861,7 @@ error of type 'outdatedConfig'. Clients SHOULD invalidate any cached aggregator
 does not succeed, clients MUST abort and discontinue retrying.
 
 The Leader MUST ignore any report pertaining to a batch that has already been
-collected. (See {{early-input-share-validation}} for details.) Otherwise,
+collected. (See {{input-share-validation}} for details.) Otherwise,
 comparing the aggregate result to the previous aggregate result may result in a
 privacy violation. (Note that the Helpers enforce this as well.) The Leader MAY
 ignore any reports whose timestamp is past the task's `task_expiration`. When it
@@ -907,7 +908,7 @@ contains information specific to the extension.
 
 Extensions are mandatory-to-implement: If an Aggregator receives a Report
 containing an extension it does not recognize, then it MUST reject the Report.
-(See {{early-input-share-validation}} for details.)
+(See {{input-share-validation}} for details.)
 
 ### Upload Message Security
 
@@ -1054,7 +1055,7 @@ report shares as follows:
 1. Decrypt the input share for each report share as described in
    {{input-share-decryption}}.
 1. Check that the resulting input share is valid as described in
-   {{early-input-share-validation}}.
+   {{input-share-validation}}.
 1. Initialize VDAF preparation as described in {{input-share-prep}}.
 
 If any step yields an invalid report share, the leader removes the report share
@@ -1148,12 +1149,12 @@ as follows:
 1. Decrypt the input share for each report share as described in
    {{input-share-decryption}}.
 1. Check that the resulting input share is valid as described in
-   {{early-input-share-validation}}.
+   {{input-share-validation}}.
 1. Initialize VDAF preparation and initial outputs as described in
    {{input-share-prep}}.
 
 [[OPEN ISSUE: consider moving the helper report ID check into
-{{early-input-share-validation}}]]
+{{input-share-validation}}]]
 
 Once the helper has processed each valid report share in
 `AggregateInitializeReq.report_shares`, the helper then creates an
@@ -1230,7 +1231,7 @@ fails, the aggregator marks the report share as invalid with the error
 `plaintext_input_share`. The `OpenBase()` function is as specified in {{!HPKE,
 Section 6.1}} for the
 
-#### Early Input Share Validation {#early-input-share-validation}
+#### Input Share Validation {#input-share-validation}
 
 Validating an input share will either succeed or fail. In the case of failure,
 the input share is marked as invalid with a corresponding ReportShareError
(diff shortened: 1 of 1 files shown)


cjpatton:
cc/ @bhalleycf

bhalleycf:
LGTM

draft-ietf-ppm-dap.md
```diff
@@ -1845,22 +1845,22 @@ about the client's measurement, violating the privacy goal of DAP. To prevent
...
-To detect replay attacks, each aggregator keeps track of the set of report IDs
-pertaining to reports that were previously aggregated for a given task. If the
-leader receives a report from a client whose report ID is in this set, it
-either ignores it or aborts the upload sub-protocol as described in
-{{upload-flow}}. A Helper who receives an encrypted input share whose report ID
-is in this set rejects the report as described in
+To detect replay attacks in a given time window, each aggregator is required to
+keep track of the set of reports it has processed for a given task and that have
+timestamps that fall in the time window. Reports that are replayed MUST be
+rejected during the aggreagtion sub-protocol as described in
```
tgeoghegan, line 1851:
```suggestion
rejected during the aggregation sub-protocol as described in
```

draft-ietf-ppm-dap.md
```diff
@@ -1845,22 +1845,22 @@ about the client's measurement, violating the privacy goal of DAP. To prevent
...
 
-[OPEN ISSUE: This has the potential to require aggregators to store report ID
-sets indefinitely. See issue#180.]
+Because honest clients choose the report ID at random, it is sufficient to store
+the set of IDs of processed reports. However, implementations may find it
+helpful to track additional information, like the timestamp, so that the storage
+used for anti-replay can be sharded efficiently.
 
 A malicious aggregator may attempt to force a replay by replacing the report ID
 generated by the client with a report ID its peer has not yet seen. To prevent
```
tgeoghegan, line 1860: I acknowledge that I'm critiquing text that was already present and is not modified by your change but--
```suggestion
A malicious leader may attempt to force a replay by replacing the report ID
generated by the client with a report ID the helper has not yet seen. To prevent
```
Is there any reason not to be explicit about which aggregators we're talking about?

draft-ietf-ppm-dap.md
```diff
@@ -1845,22 +1845,22 @@ about the client's measurement, violating the privacy goal of DAP. To prevent
 such replay attacks, this specification requires the aggregators to detect and
 filter out replayed reports.
 
-To detect replay attacks, each aggregator keeps track of the set of report IDs
-pertaining to reports that were previously aggregated for a given task. If the
-leader receives a report from a client whose report ID is in this set, it
-either ignores it or aborts the upload sub-protocol as described in
-{{upload-flow}}. A Helper who receives an encrypted input share whose report ID
-is in this set rejects the report as described in
+To detect replay attacks in a given time window, each aggregator is required to
```
tgeoghegan, line 1848: I think we need to be more clear about what the "given time window" is. My reading of the arguments in #362 is that we're instructing implementations to choose a window of time relative to the instant at which a report being evaluated for anti-replay, and only check for unique report IDs within that window. Is the behavior the same regardless of time interval or fixed batch tasks? Is there a risk of the helper and leader applying inconsistent anti-replay policy, either because they construct their time windows differently or because the leader might be evaluating anti-replay at upload time, well before the helper gets a chance to do so at aggregation time? Finally, if we say nothing about how to construct the time windows, then doesn't that allow a conforming implementation to not do anti-replay checks at all?

branlwyd (approved):

draft-ietf-ppm-dap.md
cjpatton: I think this is useful if at some point we move to direct-upload architecture.

draft-ietf-ppm-dap.md
tgeoghegan: I don't think so: the point of this PR is to allow aggregators to evaluate anti-replay on a narrower scope than the entire task. Bob's [argument](https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/issues/362#issuecomment-1301108181), I think, is that aggregators only need to look at reports that wouldn't get rejected for being too old or from too far in the future, and I think perhaps this paragraph could be written in those terms.

cjpatton: The "given time window" refers to the window of time for which an aggregator chooses to check for replayed reports; any reports outside of that timestamp must be rejected.

chris-wood: Perfect. Then how about we just keep "given time window" as is, say that implementations can choose this to fit their needs, and add a line at the end of the paragraph saying "Any report that is outside this time window MUST be rejected" (or whatever). Would that work?

cjpatton: I've pushed @chris-wood's suggestion of simply dropping discussion about the time window at that point. To @tgeoghegan point about not needing to do anti-replay for reports too far in the past or in the future, this is already premitted in {{early-input-share-validation}}.

In fact, {{early-input-share-validation}} is really the main text that matters. I think the only normative text we want to add is that anti-replay MUST be enforced per nonce, but MAY be enforced per nonce/timestamp. What do y'all think about axing this {{anti-replay}} altogether and merging this normative text into {{early-input-share-validation}}?

chris-wood: If we rename that section to "Input Share Validation" and promote the replay rejection step to the first thing in the list, I'd be OK with that.

cjpatton: Going with "Report Share Validation" :)

tgeoghegan (approved):

draft-ietf-ppm-dap.md
cjpatton: Done. I ended up going with "Input Share Validation", as @chris-wood suggested, as this better matches the language in that section.

tgeoghegan (approved):

Issue #360, opened 2022-09-26
`taskid` should be optional in problem documents


tgeoghegan:
Section 3.2 discusses how to construct HTTP problem documents. In particular, [this paragraph](https://www.ietf.org/archive/id/draft-ietf-ppm-dap-02.html#section-3.2-4) states:

>The problem document MUST also include a "taskid" member which contains the associated DAP task ID (this value is always known, see [Section 4.2](https://www.ietf.org/archive/id/draft-ietf-ppm-dap-02.html#task-configuration)), encoded in Base 64 using the URL and filename safe alphabet with no padding defined in sections 5 and 3.2 of [[RFC4648](https://www.ietf.org/archive/id/draft-ietf-ppm-dap-02.html#RFC4648)].

It's not true that the task ID is always known. [4.3.1](https://www.ietf.org/archive/id/draft-ietf-ppm-dap-02.html#section-4.3.1-3) allows implementations to respond to requests to `/hpke_config` that don't have a `task_id` query parameter with an error, and in that scenario I don't think there is any sensible task ID value for the server to use. We should change this MUST to allow some problem documents to omit `taskid`.

tgeoghegan:
We might also consider making the key `task_id` for consistency with the query parameter on `/hpke_config`.

cjpatton:
I think this is text-ready.

tgeoghegan:
Thinking about this further, I think this is actually a dupe of #278. Currently, we require that problem documents contain the `taskid` and and `instance` value that "MUST be the endpoint to which the request was targeted". The examples of `instance` values in RFC 7807 look like this:

```
{
             "type": "https://example.com/probs/out-of-credit",
             "title": "You do not have enough credit.",
             "detail": "Your current balance is 30, but that costs 50.",
             "instance": "/account/12345/msgs/abc",
             "balance": 30,
             "accounts": ["/account/12345",
                          "/account/67890"]
}
```
Here, the value of `instance` is an HTTP path pointing to the resource that caused the error. Our requirement that `instance` be the "endpoint" seems unhelpful. If I hit some API relative to `https://aggregator.example/dap-api`, and it fails, and I get back a problem document with `"instance": "https://aggregator.example/dap-api"`, then that's not very helpful, because I already knew what URI I accessed when I made the request.

[...]

PR #389, opened 2022-12-08
Address Problem Details issues pointed out in #209

### LINKED ISSUE #209: Alignment with RFC 7807 Problem Details
Some of the requirements and recommendations in section 3.1 conflict with the spirit of RFC 7807, which it references.

There is presently a requirement that:

> The "instance" value MUST be the endpoint to which the request was targeted.

The "instance" key is defined as one of the standard members in RFC 7807, and it says:

> "instance" (string) - A URI reference that identifies the specific occurrence of the problem.  It may or may not yield further information if dereferenced.

As a consequence of the current text, PPM aggregators would always provide the same URI in the "instance" field. I would recommend that we remove the MUST requirement here. Alternately, we could define a new extension member (say, "endpoint") and put the aggregator's endpoint in that member, if we still wish to have it as part of the problem details. (For what it's worth, RFC 7807 allows the "type" and "instance" members to be relative URIs, so there is already a baked-in assumption that problem detail consumers retain the request URL associated with a JSON document)

Second, there's a SHOULD recommendation that the "detail" field be populated, and no mention of the "title" field. These are both standard members, defined as follows:

> "title" (string) - A short, human-readable summary of the problem type.  It SHOULD NOT change from occurrence to occurrence of the problem, except for purposes of localization (e.g., using proactive content negotiation; see [[RFC7231], Section 3.4](https://www.rfc-editor.org/rfc/rfc7231#section-3.4)).
> "detail" (string) - A human-readable explanation specific to this occurrence of the problem.

I think we should add a recommendation that, when one of the error types from Table 1 is used, the accompanying description from the table (or a localization thereof) should be provided in the "title" member. The "detail" member could then provide more particularized information, for example, a message mentioning the old and new configuration IDs involved in an `urn:ietf:params:ppm:error:outdatedConfig` error.

Third, there is currently the following requirement for a PPM-specific extension member:

> The problem document MUST also include a "taskid" member which contains the associated PPM task ID (this value is always known, see {{task-configuration}}).

Assuming that an aggregator services multiple PPM tasks at one HTTP endpoint, there are edge cases where we may want to use `urn:ietf:params:ppm:error:unrecognizedMessage`, but not know the task ID, for example, if a request body is too short to encode a task ID. With the present text, I think aggregators would have to respond with a non-PPM error type, or just a 400 Bad Request response with an empty body.

As "taskid" is the only extension member we define on our problem types, I think we should clarify the MUST requirement here, and only require it when the "type" is one of the error types in the PPM URN namespace.

For the edge cases where "taskid" is not known we could relax the MUST requirement, and only require a "taskid" member when the PPM Task ID is known. Alternately, we could suggest that the error in such error cases be of type `about:blank` (the default type, indicating no additional semantics beyond that of the HTTP status code) and with an appropriate status line, like "400 Bad Request".

cjpatton:
Closes #209.

### DIFF
--- a/draft-ietf-ppm-dap.md
+++ b/draft-ietf-ppm-dap.md
@@ -454,13 +454,14 @@ in the "type" field (within the DAP URN namespace
 
 This list is not exhaustive. The server MAY return errors set to a URI other
 than those defined above. Servers MUST NOT use the DAP URN namespace for errors
-not listed in the appropriate IANA registry (see {{urn-space}}). Clients SHOULD
-display the "detail" field of all errors. The "instance" value MUST be the
-endpoint to which the request was targeted. The problem document MUST also
-include a "taskid" member which contains the associated DAP task ID (this value
-is always known, see {{task-configuration}}), encoded in Base 64 using the URL
-and filename safe alphabet with no padding defined in sections 5 and 3.2 of
-{{!RFC4648}}.
+not listed in the appropriate IANA registry (see {{urn-space}}). The "detail"
+member of the Problem Details document includes additional diagnostic
+information.
+
+When the task ID is known (see {{task-configuration}}), the problem document SHOULD
+include an additional "taskid" member containing the ID encoded in Base 64 using
+the URL and filename safe alphabet with no padding defined in sections 5 and 3.2
+of {{!RFC4648}}.
 
 In the remainder of this document, the tokens in the table above are used to
 refer to error types, rather than the full URNs. For example, an "error of type


draft-ietf-ppm-dap.md
```diff
@@ -454,13 +454,17 @@ in the "type" field (within the DAP URN namespace
...
-include a "taskid" member which contains the associated DAP task ID (this value
-is always known, see {{task-configuration}}), encoded in Base 64 using the URL
-and filename safe alphabet with no padding defined in sections 5 and 3.2 of
-{{!RFC4648}}.
+not listed in the appropriate IANA registry (see {{urn-space}}).
+
+For errors defined in this document, the "title" member MUST be set to the
+Description corresponding to the error type in the table above. The "detail"
+member includes additional diagnostic information. The "instance" member is
+typically the endpoint to which the request was targeted.
```
cjpatton, line 462: @divergentdave is this guidance "instance" useful, but sufficiently ambiguous?

draft-ietf-ppm-dap.md
```diff
@@ -454,13 +454,17 @@ in the "type" field (within the DAP URN namespace
...
-display the "detail" field of all errors. The "instance" value MUST be the
-endpoint to which the request was targeted. The problem document MUST also
-include a "taskid" member which contains the associated DAP task ID (this value
-is always known, see {{task-configuration}}), encoded in Base 64 using the URL
-and filename safe alphabet with no padding defined in sections 5 and 3.2 of
-{{!RFC4648}}.
+not listed in the appropriate IANA registry (see {{urn-space}}).
+
+For errors defined in this document, the "title" member MUST be set to the
+Description corresponding to the error type in the table above. The "detail"
```
tgeoghegan, line 460: This requirement on `title` doesn't make sense: if the `title` field is a deterministic function of the problem type, then there's no value in including it in problem documents because everyone can just look it up in this table. [RFC 7807 3.1](https://www.rfc-editor.org/rfc/rfc7807#section-3.1) says the title field is "human readable", which I think can be read as "implementation defined" and "not to be relied upon by machines". So I don't think DAP should say anything at all about `title`.

tgeoghegan (changes requested):

draft-ietf-ppm-dap.md
divergentdave: I think this is still pointing implementers in the wrong direction compared to RFC 7807. The RFC says that the instance URI "will always identify a specific occurrence of a problem". Thus, the URI would have to include some identifier pointing back to error information from the particular request/trace.

I think it would make sense for DAP to be silent on the `instance` member. It's optional and it's sufficiently specified in RFC 7807 to the extent that DAP doesn't need to add anything further. Plus I have a hunch most DAP implementations won't go to the extent of storing error information such that consumers could dereference `instance` URIs for "further information" as described in RFC 7807 section 3.1.

draft-ietf-ppm-dap.md
cjpatton: Done. I was following @divergentdave recommendation in the issue, but I see his thumbs up here :)

draft-ietf-ppm-dap.md

draft-ietf-ppm-dap.md
```diff
@@ -454,13 +454,14 @@ in the "type" field (within the DAP URN namespace
...
-endpoint to which the request was targeted. The problem document MUST also
-include a "taskid" member which contains the associated DAP task ID (this value
-is always known, see {{task-configuration}}), encoded in Base 64 using the URL
-and filename safe alphabet with no padding defined in sections 5 and 3.2 of
-{{!RFC4648}}.
+not listed in the appropriate IANA registry (see {{urn-space}}). The "detail"
+member of the Problem Details document includes additional diagnostic
+information.
+
+When the task ID is known (see {{task-configuration}}), the problem document SHOULD
```
cjpatton, line 461: @tgeoghegan would this close #360?
```suggestion
When the task ID is known (see {{task-configuration}}), the problem document MAY
```

tgeoghegan: Yeah I think it does!

tgeoghegan (approved):

PR #390, opened 2022-12-08
Make "taskid" member of Problem Details optional

### LINKED ISSUE #360: `taskid` should be optional in problem documents

cjpatton:
Closes #360

### DIFF
--- a/draft-ietf-ppm-dap.md
+++ b/draft-ietf-ppm-dap.md
@@ -458,7 +458,7 @@ not listed in the appropriate IANA registry (see {{urn-space}}). The "detail"
 member of the Problem Details document includes additional diagnostic
 information.
 
-When the task ID is known (see {{task-configuration}}), the problem document SHOULD
+When the task ID is known (see {{task-configuration}}), the problem document MAY
 include an additional "taskid" member containing the ID encoded in Base 64 using
 the URL and filename safe alphabet with no padding defined in sections 5 and 3.2
 of {{!RFC4648}}.


tgeoghegan (approved):
I'm not opposed to this change but I think the `SHOULD` in conjunction with the acknowledgement that task ID is not always known was good enough.

cjpatton:
> I'm not opposed to this change but I think the `SHOULD` in conjunction with the acknowledgement that task ID is not always known was good enough.

Ah, OK, let's leave it there then. I'll close that issue.

PR #391, opened 2022-12-09
Add more input share validation items

### LINKED ISSUE #255: Consider specifying aggregator behavior on message-deserialization failures.
The DAP specification defines a number of error modes, which ultimately end up mapped to error codes defined in section 3.1.

Currently, the specified behavior does not describe how to treat message-deserialization failures, in any of client-aggregator, aggregator-aggregator, or collector-aggregator communications. We should consider specifying this.

Initial thoughts:
* Moving to `FAILED` (with a new error code?) or `INVALID` makes sense to me, for deserialization errors during the aggregation process.
* It'd be very nice if one "bad" client report didn't fail the entire aggregation job, e.g. we should probably specify behavior that limits the "blast radius" of a badly-serialized message.
* We could leave this unspecified; I suppose the risk is that different implementations might treat this error mode with different semantics (e.g. some will fail a single report, some will fail the entirety of the aggregation job).

tgeoghegan:
- Allows helpers to reject reports from too far in the future, matching the leader check detailed in "Upload Requests". We add `ReportShareError.report_too_early` for this case, and allow the leader to try the report again later.
 - Requires helpers to mark undecodable decrypted input shares as invalid with `ReportShareError.unrecognized_message` (#255).

Resolves #255

### DIFF
--- a/draft-ietf-ppm-dap.md
+++ b/draft-ietf-ppm-dap.md
@@ -1048,6 +1048,7 @@ enum {
   batch_saturated(6),
   task_expired(7),
   unrecognized_message(8),
+  report_too_early(9),
   (255)
 } ReportShareError;
 ~~~
@@ -1152,7 +1153,7 @@ To begin this process, the helper first checks if it recognizes the task ID, if
 not then it MUST abort with error `unrecognizedTask`. Then the helper checks
 that the report IDs in `AggregateInitializeReq.report_shares` are all distinct.
 If two ReportShare values have the same report ID, then the helper MUST abort
-with error "unrecognizedMessage". If this check succeeds, the helper then
+with error `unrecognizedMessage`. If this check succeeds, the helper then
 attempts to recover each input share in `AggregateInitializeReq.report_shares`
 as follows:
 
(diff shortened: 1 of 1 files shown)


divergentdave (approved):

draft-ietf-ppm-dap.md
```diff
@@ -1276,6 +1280,12 @@ following generic checks.
       but during the collect sub-protocol. (See {{batch-validation}}.) If both
       checks succeed, the input share is not marked as invalid.
 
+1. Aggregators SHOULD NOT accept reports whose timestamp is too far into the
+   future. Implementors MAY provide for some small leeway, usually no more than
+   a few minutes, to account for clock skew. If a report is rejected for this
+   reason, the Aggregator SHOULD mark the input share as invalid with the error
+   `report_too_early`.
```
chris-wood, line 1287:
```suggestion
1. Check if the report is too far into the future. Implementors can provide for
   some small leeway, usually no more than a few minutes, to account for clock
   skew. If a report is rejected for this reason, the Aggregator SHOULD mark
   the input share as invalid with the error `report_too_early`.
```
To both relax normative language around something that's hard to be normative about, and use consistent language with the other checks.

chris-wood (approved):

cjpatton (approved):
👍

PR #395, opened 2023-01-04
Introduce resource oriented API

### LINKED ISSUE #278: Resource oriented HTTP API (was: Conveying task and job IDs)

tgeoghegan:
Changes the DAP API to orient the upload, aggregate and collect sub-protocols around some HTTP resources, namely:

Upload:
 - HPKE configurations
 - Reports

Aggregate:
 - Aggregation jobs

Collect:
 - Collections
 - Aggregate shares

There's a lot of change here, most of which is changing the spelling of HTTP endpoints from e.g. `POST /upload` to
`PUT /tasks/{task-id}/reports/{report-id}`. But there are some more important improvements made here:

 - move some identifiers from message bodies into resource paths, removing the need to partially parse messages to get task ID
 - aggregation job messages now contain a `round` field to allow aggregators to recover from message loss

The other big goal here is to make better use of HTTP primitives and conventions. However, it's important not to get too precious about this: we expect that anything using this API is a DAP implementation rather than a generic HTTP client like a web browser, so it doesn't matter that much if we choose exactly the right HTTP method, because ultimately DAP gets to dictate the semantics of any message.

Resolves #278

### DIFF
--- a/draft-ietf-ppm-dap.md
+++ b/draft-ietf-ppm-dap.md
@@ -489,6 +489,7 @@ in the "type" field (within the DAP URN namespace
 | unauthorizedRequest        | Authentication of an HTTP request failed (see {{request-authentication}}). |
 | missingTaskID              | HPKE configuration was requested without specifying a task ID. |
 | queryMismatch              | Query type indicated by a message does not match the task's query type. |
+| roundMismatch              | The aggregators disagree on the current round of the VDAF preparation protocol. |
 
 This list is not exhaustive. The server MAY return errors set to a URI other
 than those defined above. Servers MUST NOT use the DAP URN namespace for errors
@@ -512,12 +513,42 @@ implies HTTP status code 400 Bad Request unless explicitly specified otherwise.
 
 # Protocol Definition
 
-DAP has three major interactions which need to be defined:
+DAP is structured as an HTTP application consisting of three major interactions
+or sub-protocols:
 
-* Uploading reports from the client to the aggregators, specified in
-  {{upload-flow}}
-* Computing the results of a given measurement, specified in {{aggregate-flow}}
-* Collecting aggregated results, specified in {{collect-flow}}
+* Uploading reports from the client to the aggregators, using the HPKE
+  configuration ({{hpke-config-resource}}) and report ({{report-resource}})
+  resources, as described in {{upload-flow}}.
+* Computing the aggregate over sets of reports, using the aggregation job
+  resource ({{aggregation-job-resource}}), as described in {{aggregate-flow}}.
+* Collecting aggregation results, using the aggregate share
+  ({{aggregate-share-resource}}) and collection ({{collection-resource}})
+  resources, as described in {{collect-flow}}.
+
+In this section, we discuss each of the resources and the messages used to act
+on these resources. A resource's path is resolved relative to a server's
+endpoint to construct a resource URI. A server's endpoint MAY contain
+arbitrarily many path components before the DAP resources.
+
+Resource paths are specified as templates like:
+
+~~~
+/resource_type/{resource-id}
+~~~
+
+Templates may also include optional query parameters, indicated by square
+brackets ("[" and "]"), as in:
+
+~~~
+/resource_type/{resource-id}[?query_parameter={resource-id}]
+~~~
+
+DAP resource identifiers are always opaque byte strings, so anywhere
+`{resource-id}` occurs in a template (e.g., `{task-id}` or `{report-id}`) MUST
+be expanded to the URL-safe, unpadded Base 64 representation of the resource's
+identifier, as specified in sections 5 and 3.2 of {{!RFC4648}}.
+
+## Basic Type Definitions
 
 The following are some basic type definitions used in other messages:
 
@@ -538,6 +569,12 @@ struct {
 
 /* An ID used to uniquely identify a report in the context of a DAP task. */
 uint8 ReportID[16];
+
+/* Metadata describing a report. */
+struct {
+  ReportID report_id;
+  Time time;
+} ReportMetadata;
 
 /* The various roles in the DAP protocol. */
 enum {
@@ -718,8 +755,8 @@ In addition, in order to facilitate the aggregation and collect protocols, each
 of the aggregators is configured with following parameters:
 
 * `collector_config`: The {{!HPKE=RFC9180}} configuration of the collector
-  (described in {{hpke-config}}); see {{compliance}} for information about the
-  HPKE configuration algorithms.
+  (described in {{hpke-config-resource}}); see {{compliance}} for information
+  about the HPKE configuration algorithms.
 * `vdaf_verify_key`: The VDAF verification key shared by the aggregators. This
   key is used in the aggregation sub-protocol ({{aggregate-flow}}). [OPEN ISSUE:
   The manner in which this key is distributed may be relevant to the VDAF's
@@ -730,37 +767,31 @@ Finally, the collector is configured with the HPKE secret key corresponding to
 
 ## Uploading Reports {#upload-flow}
 
-Clients periodically upload reports to the leader, which then distributes the
-individual shares to each helper.
+Clients periodically upload reports to the Leader, which then distributes the
+individual shares to each Helper. The upload sub-protocol involves the HPKE
+configurations ({{hpke-config-resource}}) and reports ({{report-resource}})
+resources.
 
-### HPKE Configuration Request {#hpke-config}
+### HPKE Configuration Resource {#hpke-config-resource}
 
 Before the client can upload its report to the leader, it must know the HPKE
 configuration of each aggregator. See {{compliance}} for information on HPKE
-algorithm choices.
+algorithm choices. Clients retrieve the HPKE configuration from each aggregator
+by accessing the HPKE configuration resource.
 
-Clients retrieve the HPKE configuration from each aggregator by sending an HTTP
-GET request to `[aggregator]/hpke_config`, where `[aggregator]` is the
-aggregator's endpoint URL, obtained from the task parameters. Clients MAY
-specify a query parameter `task_id` when sending an HTTP GET request to
-`[aggregator]/hpke_config?task_id=[task-id]`, where `[task-id]` is the task ID
-obtained from the task parameters, encoded in Base 64 with URL and filename safe
-alphabet with no padding, as specified in sections 5 and 3.2 of {{!RFC4648}}. If
-the aggregator does not recognize the task ID, then it MUST abort with error
-`unrecognizedTask`.
+#### Required methods
+
+##### GET `/hpke_config[?task_id={task-id}]`
 
 An aggregator is free to use different HPKE configurations for each task with
-which it is configured. If the task ID is missing from a client's request, the
-aggregator MAY abort with an error of type `missingTaskID`, in which case the
-client SHOULD retry the request with a well-formed task ID included.
+which it is configured. Clients MAY specify a query parameter `task_id` whose
+value is the task ID whose HPKE configuration they want.
 
-An aggregator responds to well-formed requests with HTTP status code 200 OK and
-an `HpkeConfigList` value. The `HpkeConfigList` structure contains one or more
-`HpkeConfig` structures in decreasing order of preference. This allows a server
-to support multiple HPKE configurations simultaneously.
+If the task ID is missing from a Client's request, the Aggregator MAY abort with
+an error of type `missingTaskID`, in which case the Client SHOULD retry the
+request with a `task_id` query parameter.
 
-[TODO: Allow aggregators to return HTTP status code 403 Forbidden in deployments
-that use authentication to avoid leaking information about which tasks exist.]
+An aggregator responds to well-formed requests with an `HpkeConfigList` value:
 
 ~~~
 HpkeConfig HpkeConfigList<1..2^16-1>;
@@ -781,25 +812,32 @@ uint16 HpkeKdfId;  /* Defined in [HPKE] */
 
 [OPEN ISSUE: Decide whether to expand the width of the id.]
 
-Aggregators SHOULD allocate distinct id values for each `HpkeConfig` in a
+The `HpkeConfigList` structure contains one or more `HpkeConfig` structures in
+decreasing order of preference. This allows a server to support multiple HPKE
+configurations simultaneously.
+
+[TODO: Allow aggregators to return HTTP status code 403 Forbidden in deployments
+that use authentication to avoid leaking information about which tasks exist.]
+
+Aggregators SHOULD allocate distinct `id` values for each `HpkeConfig` in an
 `HpkeConfigList`. The RECOMMENDED strategy for generating these values is via
-rejection sampling, i.e., to randomly select an id value repeatedly until it
+rejection sampling, i.e., to randomly select an `id` value repeatedly until it
 does not match any known `HpkeConfig`.
 
-The client MUST abort if any of the following happen for any HPKE config
+The Client MUST abort if any of the following happen for any HPKE config
 request:
 
 * the GET request failed or did not return a valid HPKE configuration; or
 * the HPKE configuration specifies a KEM, KDF, or AEAD algorithm the client does
   not recognize.
 
-Aggregators SHOULD use HTTP caching to permit client-side caching of this
+Aggregators SHOULD use HTTP caching to permit Client-side caching of this
 resource {{!RFC5861}}. Aggregators SHOULD favor long cache lifetimes to avoid
 frequent cache revalidation, e.g., on the order of days. Aggregators can control
 this cached lifetime with the Cache-Control header, as follows:
 
 ~~~
-  Cache-Control: max-age=86400
+Cache-Control: max-age=86400
 ~~~
 
 Clients SHOULD follow the usual HTTP caching {{!RFC9111}} semantics for key
@@ -809,48 +847,37 @@ Note: Long cache lifetimes may result in clients using stale HPKE
 configurations; aggregators SHOULD continue to accept reports with old keys for
 at least twice the cache lifetime in order to avoid rejecting reports.
 
-### Upload Request
+### Report Resource {#report-resource}
 
-Clients upload reports by using an HTTP POST to `[leader]/upload`, where
-`[leader]` is the first entry in the task's aggregator endpoints. The payload is
-structured as follows:
+Only the Leader supports this resource. The Leader transmits report shares to
+the Helper during the aggregate sub-protocol ({{aggregate-flow}}).
 
-~~~
-struct {
-  ReportID report_id;
-  Time time;
-} ReportMetadata;
+#### Required Methods
 
-struct {
-  TaskID task_id;
-  ReportMetadata metadata;
-  opaque public_share<0..2^32-1>;
-  HpkeCiphertext encrypted_input_shares<1..2^32-1>;
-} Report;
-~~~
+##### PUT `/tasks/{task-id}/reports/{report-id}`
 
-This message is called the Client's report. It consists of the task ID, report
-metadata, the "public share" output by the VDAF's input-distribution algorithm,
-and the encrypted input share of each of the Aggregators. (Note that the public
-share might be empty, depending on the VDAF. For example, Prio3 has an empty
-public share, but Poplar1 does not. See {{!VDAF}}.) The header consists of the
-task ID and report "metadata". The metadata consists of the following fields:
+`report-id` is a `ReportId` used by the Aggregators to ensure the report appears
+in at most one batch (see {{input-share-validation}}). The Client MUST generate
+this using a cryptographically secure random number generator.
 
-* A report ID used by the Aggregators to ensure the report appears in at most
-  one batch. (See {{input-share-validation}}.) The Client MUST generate this by
-  generating 16 random bytes using a cryptographically secure random number
-  generator.
-
-* A timestamp representing the time at which the report was generated.
-  Specifically, the `time` field is set to the number of seconds elapsed since
-  the start of the UNIX epoch. The client SHOULD round this value down to the
-  nearest multiple of `time_precision` in order to ensure that that the
-  timestamp cannot be used to link a report back to the Client that generated
-  it.
+[Note/TODO: it's important that the _Client_ chooses the report ID. We use the
+report ID as the nonce, and VDAF-04 will include some guidance explaining that
+if the report ID is chosen with knowledge of the VDAF verify key, privacy could
+be violated. For that reason, we can't allow either aggregator to choose report
+IDs.]
 
 To generate a report, the Client begins by sharding its measurement into input
-shares as specified by the VDAF. It then wraps each input share in the following
-structure:
+shares and computing the public share using the VDAF's
+`measurement_to_input_shares` algorithm ({{!VDAF, Section 5.1}}):
+
+~~~
+(public_share, input_shares) = VDAF.measurement_to_input_shares(measurement)
+~~~
+
+[TODO: in VDAF-04, `measurement_to_input_shares` will take a nonce argument, for
+which we will use the report ID #394]
+
+The Client then wraps each input share in the following structure:
 
 ~~~
 struct {
@@ -860,11 +887,11 @@ struct {
 ~~~
 
 Field `extensions` is set to the list of extensions intended to be consumed by
-the given Aggregator. (See {{upload-extensions}}.) Field `payload` is set to the
-Aggregator's input share output by the VDAF sharding algorithm.
+the Aggregator receiving the input share (see {{upload-extensions}}). Field
+`payload` is set to the Aggregator's input share output by the VDAF sharding
+algorithm.
 
-Next, the Client encrypts each PlaintextInputShare `plaintext_input_share` as
-follows:
+Next, the Client encrypts each `PlaintextInputShare` as follows:
 
 ~~~
 enc, payload = SealBase(pk,
@@ -874,10 +901,10 @@ enc, payload = SealBase(pk,
 
 where `pk` is the aggregator's public key; `server_role` is the Role of the
 intended recipient (`0x02` for the leader and `0x03` for the helper),
-`plaintext_input_share` is the Aggregator's PlaintextInputShare, and
+`plaintext_input_share` is the Aggregator's `PlaintextInputShare`, and
 `input_share_aad` is an encoded message of type `InputShareAad`, constructed
-from the same values as the corresponding fields in the report. The `SealBase()`
-function is as specified in {{!HPKE, Section 6.1}} for the ciphersuite indicated
+from the corresponding values in the report. The `SealBase()` function is as
+specified in {{!HPKE, Section 6.1}} for the ciphersuite indicated
 by the HPKE configuration.
 
 ~~~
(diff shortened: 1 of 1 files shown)


draft-ietf-ppm-dap.md
```diff
@@ -512,12 +513,42 @@ implies HTTP status code 400 Bad Request unless explicitly specified otherwise.
...
+brackets ("[" and "]"), as in:
+
+~~~
+/resource_type/{resource-id}[?query_parameter={resource-id}]
+~~~
+
+DAP resource identifiers are always opaque byte strings, so anywhere
+`{resource-id}` occurs in a template (e.g., `{task-id}` or `{report-id}`) MUST
+be expanded to the URL-safe, unpadded Base 64 representation of the resource's
+identifier, as specified in sections 5 and 3.2 of {{!RFC4648}}.
```
tgeoghegan, line 549: [RFC 6570](https://datatracker.ietf.org/doc/rfc6570/) discusses URI templating, and I considered using it here, but decided not to because (1) we don't need a template scheme as complicated as the one in 6570 and (2) 6570 has rules about how variables in templates should be percent encoded to be URL safe, whereas we have already settled on using URL-safe unpadded base64.

draft-ietf-ppm-dap.md
```diff
@@ -1391,124 +1320,227 @@ Otherwise, `out` is the pair `(prep_state, prep_msg)`.
...
+pertain to the same DAP task and compiling them into an `AggregationJob` (see
+{{aggregation-job-representation}}. The Leader can prepare many reports in
+parallel by concurrently creating multiple aggregation jobs. Creating
+aggregation jobs is how the Leader transmits report shares to the helper, so
+each `PrepareStep.prepare_step_state` MUST be set to `PrepareStepState.start`
+and the `PrepareStep` MUST contain a `ReportShare`. `AggregationJob.round` MUST
+be 0.
+
+If successful, the Helper's responds with HTTP status 303 See Other and a
+Location header containing a URI for the aggregation job. The structure of this
```
tgeoghegan, line 1528: I'm not certain if explicitly spelling out HTTP 303 See Other and a Location header is a good idea. I think we could just say something like "the Helper redirects to a URI for the aggregation job", and rely on existing HTTP specifications for things like the status code and the headers and so on. On the other hand, if we specify this explicitly as I have here, then there's less ambiguity and less code to write for DAP implementations (with this text, they can simply ignore any 3xx status that isn't 303 instead of having to reason about the nuanced difference between 301 and 303). So I lean towards being specific.

tgeoghegan:
See also https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/pull/367 for some more discussion and analysis of this change.

draft-ietf-ppm-dap.md
```diff
@@ -997,28 +1040,26 @@ of relevant: issue #89]]
...
 
 The exact structure of the aggregation job flow depends on the VDAF.
 Specifically:
 
-* Some VDAFs (e.g., Prio3) allow the leader to start aggregating reports
+* Some VDAFs (e.g., Prio3) allow the Leader to start aggregating reports
   proactively before all the reports in a batch are received. Others (e.g.,
-  Poplar1) require all the reports to be present and must be initiated by the
-  collector.
+  Poplar1) cannot be prepared until the collector provides the aggregation
```
cjpatton, line 1061:
```suggestion
  Poplar1) cannot be prepared until the Collector provides the aggregation
```

draft-ietf-ppm-dap.md
```diff
@@ -1048,161 +1089,76 @@ Aggregate request (continued, Job = M) ------------------>  | 11-20
...
 ### Aggregate Initialization {#agg-init}
 
-The leader begins an aggregation job by choosing a set of candidate reports that
-pertain to the same DAP task and a unique job ID. The job ID is a 32-byte value,
-structured as follows:
+The Leader begins an aggregation job by choosing a set of candidate reports that
+pertain to the same DAP task. The Leader can prepare many reports in parallel by
+concurrently creating multiple aggregation jobs. After choosing the set of
+candidates, the Leader begins aggregation by splitting each report into "report
+shares", one for each aggregator. The Leader and Helpers then run the aggregate
```
cjpatton, line 1111:
```suggestion
shares", one for each Aggregator. The Leader and Helpers then run the aggregate
```

draft-ietf-ppm-dap.md
```diff
@@ -1391,124 +1320,227 @@ Otherwise, `out` is the pair `(prep_state, prep_msg)`.
...
+
+* The sequence of report shares to aggregate. The `encrypted_input_share` field
+  of the report share is the `HpkeCiphertext` whose index in
+  `Report.encrypted_input_shares` is equal to the index of the aggregator in the
+  task's `aggregator_endpoints` to which the AggregateInitializeReq is being
+  sent.
+
+#### Required methods
+
+##### PUT `/tasks/{task-id}/aggregation_jobs` {#aggregation-job-put}
```
cjpatton, line 1516: * Editorial: It might be helpful to define the resource and message format before getting into the protocol flow.
* Where is the aggregation job ID? We still need a way to link requests for the same job across rounds.

cjpatton: Summary of offline discussion: The aggregation job ID is missing here because it has been folded into the redirect URI that the Leader is supposed to post to next to move aggregation forward. (In other words, the aggreagtion job ID is now an implementation detail.) The ideas is to make aggreagtion jobs "asynchronous" the way collection jobs are. However there is an important difference between the two: Here the Helper computes its response immediately, i.e., aggregation is inherently synchronous. Collection is inherently asynchronous because the Leader usually won't be finished computing the aggregate at the time the request is made.

We decided that we should re-align this part with DAP-03, i.e., move the aggregation job ID into the API here.

draft-ietf-ppm-dap.md
```diff
@@ -512,12 +513,42 @@ implies HTTP status code 400 Bad Request unless explicitly specified otherwise.
...
+or sub-protocols:
 
-* Uploading reports from the client to the aggregators, specified in
-  {{upload-flow}}
-* Computing the results of a given measurement, specified in {{aggregate-flow}}
-* Collecting aggregated results, specified in {{collect-flow}}
+* Uploading reports from the client to the aggregators, using the HPKE
+  configuration ({{hpke-config-resource}}) and report ({{report-resource}})
+  resources, as described in {{upload-flow}}.
+* Computing the aggregate over sets of reports, using the aggregation job
```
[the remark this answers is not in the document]
tgeoghegan: I think this would be a bit confusing, because in this bullet, we are discussing the aggregation sub-protocol, which works in terms of aggregation jobs, and those aren't necessarily equivalent to batches.

draft-ietf-ppm-dap.md
```diff
@@ -512,12 +513,42 @@ implies HTTP status code 400 Bad Request unless explicitly specified otherwise.
...
+  resource ({{aggregation-job-resource}}), as described in {{aggregate-flow}}.
+* Collecting aggregation results, using the aggregate share
+  ({{aggregate-share-resource}}) and collection ({{collection-resource}})
```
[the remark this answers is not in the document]
tgeoghegan: I feel like that's trying to sneak a verb ("job") into a resource, which should be a noun. However this is purely a naming quibble so I can live with either. I'm curious what others think; I'd be happy to go with the majority consensus.

draft-ietf-ppm-dap.md
```diff
@@ -781,25 +812,32 @@ uint16 HpkeKdfId;  /* Defined in [HPKE] */
 
 [OPEN ISSUE: Decide whether to expand the width of the id.]
 
-Aggregators SHOULD allocate distinct id values for each `HpkeConfig` in a
+The `HpkeConfigList` structure contains one or more `HpkeConfig` structures in
+decreasing order of preference. This allows a server to support multiple HPKE
+configurations simultaneously.
+
+[TODO: Allow aggregators to return HTTP status code 403 Forbidden in deployments
+that use authentication to avoid leaking information about which tasks exist.]
```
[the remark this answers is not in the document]
tgeoghegan: No, I guess not, since we don't have any text that forbids this.

draft-ietf-ppm-dap.md
```diff
@@ -809,48 +847,37 @@ Note: Long cache lifetimes may result in clients using stale HPKE
...
-task ID and report "metadata". The metadata consists of the following fields:
-
-* A report ID used by the Aggregators to ensure the report appears in at most
-  one batch. (See {{input-share-validation}}.) The Client MUST generate this by
-  generating 16 random bytes using a cryptographically secure random number
-  generator.
-
-* A timestamp representing the time at which the report was generated.
-  Specifically, the `time` field is set to the number of seconds elapsed since
-  the start of the UNIX epoch. The client SHOULD round this value down to the
```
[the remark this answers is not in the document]
tgeoghegan: I moved it down to line 928, in the discussion of `struct Report`:

~~~
* `time` is the time at which the report was generated. The Client SHOULD round
  this value down to the nearest multiple of the task's `time_precision` in
  order to ensure that that the timestamp cannot be used to link a report back
  to the Client that generated it.
~~~

draft-ietf-ppm-dap.md
tgeoghegan: We reached a similar conclusion about the helper's aggregate share resource: the leader should always be able to wait until all relevant aggregation jobs are driven to completion before requesting an aggregate share from the helper. So servicing such a request should just be a matter of collapsing a bunch of output shares into an aggregate share (or, depending on the implementation, collapsing a bunch of partial aggregates over, say, an hour's worth of output shares into the requested aggregate share), which should be doable synchronously for the helper. So I'm also going to move the aggregate share resource back to how it worked in DAP-03 (though we'll still change the request path).

tgeoghegan: > Editorial: It might be helpful to define the resource and message format before getting into the protocol flow.

I'll give that a try after the dust has settled on other changes.

- [ ] put the description of aggregate job endpoints and messages before the prose discussion of the protocol

draft-ietf-ppm-dap.md
```diff
@@ -1391,124 +1317,230 @@ Otherwise, `out` is the pair `(prep_state, prep_msg)`.
...
+} PrepareStep;
 
-~~~
 struct {
+  u16 round;
+  opaque agg_param<0..2^32-1>;
+  PartialBatchSelector part_batch_selector;
   PrepareStep prepare_steps<1..2^32-1>;
-} AggregateContinueResp;
+} AggregationJob;
```
tgeoghegan, line 1479: I've been working on implementing this in Janus (https://github.com/divviup/janus/pull/731/files) and I'm finding this definition of `AggregationJob` a little awkward. The idea here is to have a single representation of an aggregation job for both helper and leader across multiple rounds of preparation. But we end up with some extra complexity in implementations:

- Not all `PrepareStep.prepare_step_state` values are valid in all rounds of preparation. For instance, when the leader first sends an `AggregationJob` to the helper, the helper must verify that each `PrepareStep` is in the `start` state. In subsequent rounds, the `start` state is illegal.
- `PrepareStep`s in the `start` state contain a `ReportShare`, which contains `ReportMetadata`, which has `ReportID` in it. That duplicates the `ReportID` field in the outer `PrepareStep` structure, which is wasteful.
- the `agg_param` and `part_batch_selector` fields are only really useful in the first leader->helper messages. Subsequently, the helper has to check that they haven't changed since an earlier round.

All of these mean extra code for servers to write to validate messages. I think we were better off the way things were before, with distinct representations for the initial creation of an aggregation job and then the subsequent stepping. I think it's less REST-y, but that doesn't seem like a good enough tradeoff to me against the implementation complexity and increased message size.

draft-ietf-ppm-dap.md
```diff
@@ -730,37 +767,31 @@ Finally, the collector is configured with the HPKE secret key corresponding to
...
-obtained from the task parameters, encoded in Base 64 with URL and filename safe
-alphabet with no padding, as specified in sections 5 and 3.2 of {{!RFC4648}}. If
-the aggregator does not recognize the task ID, then it MUST abort with error
-`unrecognizedTask`.
+algorithm choices. Clients retrieve the HPKE configuration from each aggregator
+by accessing the HPKE configuration resource.
+
+#### Required methods
+
+##### GET `/hpke_config[?task_id={task-id}]`
```
cjpatton, line 784: For consistency with other resources? Also, the return value is a list of HPKE configs.
```suggestion
##### GET `/hpke_configs[?task_id={task-id}]`
```

draft-ietf-ppm-dap.md
```diff
@@ -1391,124 +1317,230 @@ Otherwise, `out` is the pair `(prep_state, prep_msg)`.
...
+
+##### POST `/tasks/{task-id}/aggregation_jobs/{aggregation-job-id}` {#aggregation-job-post}
+
+The Leader drives the continuation of input preparation using POST requests to
+aggregation job resources (see {{agg-continue-flow}}).
+
+###### DELETE `/tasks/{task-id}/aggregation_jobs/{aggregation-job-id}`
+
+Once an aggregation job is deleted, the Helper MAY abandon it and discard all
+state related to it (e.g., report shares, prepare messages, preparation state).
```
cjpatton, line 1543: Is the Helper required to keep the agg job state around until it is DELETEd? As a deployer, I don't want to have to count on my Leader to manage my state responsibly.

draft-ietf-ppm-dap.md
cjpatton: +1, I think we should go back to what we had before (more or less). My main concern: We should not transmit the agg param more than once per agg job, since for Poplar this might be relatively large. (Think 1-2MBs.)

draft-ietf-ppm-dap.md
```diff
@@ -1524,82 +1556,108 @@ the requirements in {{request-authentication}}.
...
 Once complete, the collector computes the final aggregate result as specified in
 {{collect-finalization}}.
 
-### Collection Initialization {#collect-init}
+### Collection Resource {#collection-resource}
 
-To initiate collection, the collector issues a POST request to
-`[leader]/collect`, where `[leader]` is the leader's endpoint URL. The body of
-the request is structured as follows:
+This resource allows the Collector to create new collections in the Leader and
```
cjpatton, line 1577: I don't recall if we agreed to be consistent about this, but just in case:
```suggestion
This resource allows the Collector to create new collection jobs in the Leader and
```

draft-ietf-ppm-dap.md
```diff
@@ -1524,82 +1556,108 @@ the requirements in {{request-authentication}}.
...
+check on their state. Only the Leader supports this resource.
 
-[OPEN ISSUE: Decide if and how the collector's request is authenticated. If not,
-then we need to ensure that collect job URIs are resistant to enumeration
-attacks.]
+#### Required Methods
+
+##### PUT `/tasks/{task-id}/collections` {#collections-put}
```
cjpatton, line 1582:
```suggestion
##### PUT `/tasks/{task-id}/collection_jobs` {#collections-put}
```

draft-ietf-ppm-dap.md
```diff
@@ -1524,82 +1556,108 @@ the requirements in {{request-authentication}}.
...
+valid using the procedure in {{batch-validation}}. If so, it immediately sends
+the Collector a response with HTTP status 303 See Other and a Location header
+containing a URI for the individual collection that can be polled by the
+Collector, as described in {{collection-resource}}. The structure of this URI is
+not specified, so implementations may use any unique identifier they wish for
+collections.
+
+The Leader then asynchronously begins working with the Helper to prepare the
+input shares satisfying the query (or continues this process, depending on the
+VDAF) as described in {{aggregate-flow}}, and then invokes the aggregate share
```
cjpatton, line 1620:
```suggestion
The Leader then asynchronously begins working with the Helper to aggregate the reports satisfying the query (or continues this process, depending on the
VDAF) as described in {{aggregate-flow}}, and then invokes the aggregate share
```

draft-ietf-ppm-dap.md
```diff
@@ -1524,82 +1556,108 @@ the requirements in {{request-authentication}}.
...
+##### POST `{collection URI}`
+
+The collector makes a POST request to the collection resource to check whether
+it is available yet. A collection is available if:
 
-The leader then begins working with the helper to prepare the shares satisfying
-the query (or continues this process, depending on the VDAF) as described in
-{{aggregate-flow}}.
+* The Leader has all the output shares it needed to construct its own aggregate
+  share.
```
cjpatton, line 1630:
```suggestion
* The Leader and Helper have aggregated a sufficient number of reports.
```

draft-ietf-ppm-dap.md
```diff
@@ -1524,82 +1556,108 @@ the requirements in {{request-authentication}}.
...
+
+The Leader then responds to the Collector with a `Collection`:
 
 ~~~
 struct {
   PartialBatchSelector part_batch_selector;
   uint64 report_count;
   HpkeCiphertext encrypted_agg_shares<1..2^32-1>;
-} CollectResp;
+} Collection;
```
cjpatton, line 1660: ?
```suggestion
} CollectJobResult;
```

draft-ietf-ppm-dap.md
```diff
@@ -1612,37 +1670,55 @@ This structure includes the following:
...
+  in the same order as the aggregator endpoints list of the task parameters.
+
+If obtaining Helper aggregate shares fails, then the Leader responds to
+subsequent POST requests to the collection with an HTTP error status and a
+problem document as described in {{errors}}.
+
+The Leader MAY respond with 204 No Content for requests to a collection if the
+results have been deleted due to age.
+
+###### A note on idempotence
```
cjpatton, line 1683:
```suggestion
###### A Note on Idempotence
```

draft-ietf-ppm-dap.md
```diff
@@ -1612,37 +1670,55 @@ This structure includes the following:
...
+
+The reason we use a POST instead of a GET to poll the state of a collection is
+because of the fixed-size query mode (see {{fixed-size-query}}). Collectors may
+make a query against the current batch, and it is the Leader's responsibility to
+keep track of what batch is current for some task. Polling a collection is the
+only point at which it is safe for the Leader to change its current batch, since
+it constitutes acknowledgement on the Collector's part that it received the
+response to some previous PUT request to the collections resource.
```
cjpatton, line 1691: I don't think I follow. For starters, you're talking about polling the collect URI here, correct? If yes, then the contents of the response, that is the aggregate result, is not going to change once computed.

draft-ietf-ppm-dap.md
```diff
@@ -1214,64 +1167,37 @@ as follows:
...
-Otherwise, the PrepareStep is either marked as continued with the output
-`prep_msg`, or is marked as finished if the VDAF preparation process is finished
-for the report share.
+Once the Helper has processed each valid report share in
+`AggregationJob.prepare_steps`, it responds with an `AggregationJob` (see
+{{aggregation-job-representation}}). The order of its `prepare_steps` MUST match
+the Leader's request. Each report that was marked as invalid is assigned the
+`PrepareStepState` `failed`. Otherwise, the `PrepareStepState` is either
+`continued` with the output `prep_msg`, or `finished` if the VDAF preparation
+process is finished for the report share.
```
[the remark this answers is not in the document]
tgeoghegan: As explained [here](https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/pull/395#discussion_r1065098317), we're going to change this.

draft-ietf-ppm-dap.md
```diff
@@ -512,12 +513,42 @@ implies HTTP status code 400 Bad Request unless explicitly specified otherwise.
...
+endpoint to construct a resource URI. A server's endpoint MAY contain
+arbitrarily many path components before the DAP resources.
+
+Resource paths are specified as templates like:
+
+~~~
+/resource_type/{resource-id}
+~~~
+
+Templates may also include optional query parameters, indicated by square
```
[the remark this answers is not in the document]
tgeoghegan: I don't know, it depends on what query params we add in the future. For now I'm trying to make the path templates no more complicated than what is needed to describe the API we currently have.

draft-ietf-ppm-dap.md
```diff
@@ -1214,64 +1167,37 @@ as follows:
...
-sequence of the AggregateInitializeReq. If any message appears out of order, is
-missing, has an unrecognized report ID, or if two messages have the same report
-ID, then the leader MUST abort with error "unrecognizedMessage".
+Upon receipt of a Helper's `AggregationJob` message, the Leader checks that the
+sequence of `PrepareStep` messages corresponds to the `PrepareStep` sequence of
+the request. If any `PrepareStep` appears out of order, is missing, has an
+unrecognized report ID, is in the `start` state, or if two prepare steps have
+the same report ID, then the Leader MUST abort with error `unrecognizedMessage`.
+If `AggregationJob.round` is not 1, then the Leader MUST abort with error
+`roundMismatch`.
```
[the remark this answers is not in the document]
tgeoghegan: I don't think that can work: in the case of the helper replying to the leader, there's no request path in which to indicate the round.

draft-ietf-ppm-dap.md
tgeoghegan: Yes, excellent idea.

draft-ietf-ppm-dap.md
tgeoghegan: Yes, but that's not what affects idempotence. What matters is that polling a collection has the _side effect_ of changing what the leader considers to be the current batch. There's a much more verbose explanation of this [here](https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/blob/134d5305849f52a59061a543e843387564b24a69/api-resources.md?plain=1#L341).

draft-ietf-ppm-dap.md
tgeoghegan: I prefer "collections" over "collection jobs". The collection _job_ is the means by which we get the thing we actually care about, which is the collection. But at the end of the day, it doesn't really make a difference which string we use here.

draft-ietf-ppm-dap.md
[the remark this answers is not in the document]
cjpatton: Discussed offline: This is just spelling.

draft-ietf-ppm-dap.md
```diff
@@ -1214,64 +1269,174 @@ as follows:
...
+either of these operations fails, then the Leader marks the report as invalid.
+Otherwise, if the VDAF has reached the finished state, then `out` will be the
+Leader's output share, in which case the Leader finishes and stores its output
+share for further processing as described in {{collect-flow}}. Otherwise, `out`
+is the pair `(new_state, prep_msg)`, where `new_state` is the Leader's updated
+state and `prep_msg` is the next VDAF message (which will be `leader_outbound`
+in the next round of continuation). For the latter case, the Leader sets
+`prep_state` to `new_state`.
+
+The Leader SHOULD store each report share's current preparation state and the
```
branlwyd, line 1350: Shouldn't this storage happen after receiving the response from the helper?

draft-ietf-ppm-dap.md
```diff
@@ -1214,64 +1269,174 @@ as follows:
...
+{{aggregation-job-delete}}).
+
+Otherwise, for each `PrepareStep` received from the Leader in
+`AggregationJob.prepare_steps`, the Helper checks
+`PrepareStep.prepare_step_state` to determine if the report share should
+continue being prepared. If the report share is in state `failed` or `finished`,
+then the Helper marks the report as failed with error `unrecognized_message`.
+
+If the state is `continued`, then the Helper continues with preparation for the
+report share. The Helper MUST check its current preparation state against the
```
branlwyd, line 1377:
```suggestion
report share. The Helper MUST check its current round against the
```

draft-ietf-ppm-dap.md
```diff
@@ -1214,64 +1269,174 @@ as follows:
...
+   `(prep_state, prep_msg)`, in which case the Helper replies to the leader with
+   `PrepareStep` in the `continued` state containing `prep_msg`.
 
-The message is a sequence of PrepareStep values, the order of which matches that
-of the ReportShare values in `AggregateInitializeReq.report_shares`. Each report
-that was marked as invalid is assigned the PrepareStepResult `failed`.
-Otherwise, the PrepareStep is either marked as continued with the output
-`prep_msg`, or is marked as finished if the VDAF preparation process is finished
-for the report share.
+After computing each report share's preparation state and prepare message, the
```
branlwyd, line 1400: Recovering from round skew only requires storing one old round of responses; I think it would be worthwhile to mention this.

draft-ietf-ppm-dap.md
```diff
@@ -1214,64 +1269,174 @@ as follows:
...
+Helper SHOULD store them, tagged with the round they were computed in, so that
```
branlwyd, line 1401: Here we say the helper should store the answer, but below we say the helper can either store or recompute the response (paragraph starting L1404).

I'd suggest dropping the "recompute" part -- storing the answer is simpler & less error-prone. And recomputing the response still requires storing information to allow the recomputation to occur, I think.

draft-ietf-ppm-dap.md
```diff
@@ -1214,64 +1269,174 @@ as follows:
...
+The Helper's response to the Leader is an `AggregationJob` (see
+{{aggregation-job-representation}}) compiled from the `PrepareStep`s. The order
+of the Helper's `prepare_steps` MUST match that of the Leader's.
+
+[[OPEN ISSUE: consider relaxing this ordering constraint. See issue#217.]]
+
+##### DELETE `/tasks/{task-id}/aggregation_jobs/{aggregation-job-id}` {#aggregation-job-delete}
+
+Indicates to the Helper that it MAY abandon the aggregation job and discard all
+state related to it. Whether or not it deletes any data, the Helper MUST respond
```
branlwyd, line 1423: I think this text should allow returning e.g. `500` on error -- as written, the Helper must unconditionally respond with a `204`.

draft-ietf-ppm-dap.md
```diff
@@ -1524,82 +1568,108 @@ the requirements in {{request-authentication}}.
...
 
-Upon receipt of a `CollectReq`, the leader begins by checking that the request
-meets the requirements of the batch parameters using the procedure in
-{{batch-validation}}. If so, it immediately sends the collector a response with
-HTTP status 303 See Other and a Location header containing a URI identifying the
-collect job that can be polled by the collector, called the "collect job URI".
+Upon receipt of a `CollectReq`, the Leader begins by checking that the query is
+valid using the procedure in {{batch-validation}}. If so, it immediately sends
+the Collector a response with HTTP status 303 See Other and a Location header
+containing a URI for the individual collect job that can be polled by the
```
branlwyd, line 1625: Since we are defining URIs for everything else, is there any reason not to introduce a `CollectJobID` and treat collect jobs as we do e.g. reports or aggregation jobs? (i.e. define a specific URI scheme, including the collect job ID, something like `/tasks/{task-id}/collect_jobs/{collect-job-id}`)

This would match the rest of the API much more closely, and I don't think there are any downsides. The collector would be generating the collect job IDs, but I don't think that is a problem.

draft-ietf-ppm-dap.md
```diff
@@ -1048,81 +1086,87 @@ Aggregate request (continued, Job = M) ------------------>  | 11-20
...
 
-The leader can run this process for many candidate reports in parallel as
-needed. After choosing the set of candidates, the leader begins aggregation by
-splitting each report into "report shares", one for each aggregator. The leader
-and helpers then run the aggregate initialization flow to accomplish two tasks:
+struct {
+  ReportID report_id;
+  PrepareStepState prepare_step_state;
+  select (PrepareStep.prepare_step_state) {
+    case start: ReportShare;
```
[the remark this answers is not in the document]
tgeoghegan: This is an error, left over from a previous iteration where a single `AggregationJob` message was used for all stages of the aggregate sub-protocol. There shouldn't be a `start` state here. Thank you for spotting that.

draft-ietf-ppm-dap.md
```diff
@@ -1214,64 +1269,174 @@ as follows:
...
+In the continuation phase, the Leader drives the VDAF preparation of each share
+in the candidate report set until the underlying VDAF moves into a terminal
+state, yielding an output share for all aggregators or an error. This phase may
+involve multiple rounds of interaction depending on the underlying VDAF. Each
+round trip is initiated by the Leader making a POST request to an aggregation
+job resource.
+
+###### Leader Continuation {#aggregation-leader-continuation}
+
+The Leader begins each round of continuation with an `AggregationJob` received
```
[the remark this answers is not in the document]
tgeoghegan: You're quite right that a leader could validate the helper's `AggregationJob` immediately. The reason I put this text here is so that I don't have to duplicate it at the bottom of both "Leader Initialization" and also "Leader Continuation" (in both cases, the Leader gets an `AggregationJob` from the helper and must validate it). But I think what we could do is add a section "Aggregation Job Validation" as a peer to "Input Share Preparation" and "Input Share Validation" and then refer to that from both "Leader Initialization" and "Leader Continuation".

- [ ] refactor aggregation job validation text into another section

draft-ietf-ppm-dap.md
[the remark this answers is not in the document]
tgeoghegan: I put it back to "collect job" in the most recent round of changes. If you both like it better, we can change it again to "collection job".

draft-ietf-ppm-dap.md
```diff
@@ -1048,81 +1086,87 @@ Aggregate request (continued, Job = M) ------------------>  | 11-20
...
+Executing the aggregation flow consists of the Leader interacting with the
+Helper's aggregate job resource. Only the Helper supports this resource.
+
+#### Representation {#aggregation-job-representation}
 
 ~~~
-opaque AggregationJobID[32];
-~~~
+enum {
+  continued(0),
```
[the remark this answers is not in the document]
tgeoghegan: As noted [here](https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/pull/395#discussion_r1067551046), there should not be a start state.

draft-ietf-ppm-dap.md
tgeoghegan: I think there's a few different points at which the leader could store different things. What we care about is that the leader checkpoint state so that it can recover from crashes, but the more I think about it, the more I think it's inappropriate for the spec to say anything about what an implementation stores or when. So maybe we shouldn't say anything about this at all.

draft-ietf-ppm-dap.md
tgeoghegan: - [ ] include implementation note explaining that helper only needs to store the most recent set of prepare messages for each aggregation job

draft-ietf-ppm-dap.md
tgeoghegan: - [ ] clean up language about HTTP status code in {{aggregation-job-delete}}

draft-ietf-ppm-dap.md
tgeoghegan: Yeah the argument from consistency seems reasonable. The counter-argument is that not specifying the collect job URI is the least change from DAP-03, and thus the least disruptive to implementations. I can go either way on this. I'm curious what @cjpatton thinks.

draft-ietf-ppm-dap.md
tgeoghegan: I think this is similar to the equivalent leader text discussed [here](https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/pull/395#discussion_r1067581728) so I'll pursue discussion there.

draft-ietf-ppm-dap.md
tgeoghegan: We have the same problem in the equivalent helper text (noted by Brandon [here](https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/pull/395#discussion_r1067587961). These two passages are awkward, because there's a few different ways for implementations to checkpoint state such that they can implement round skew recovery, and I don't think it's appropriate for the spec to dictate implementation details (because maybe there's some nuance about a particular implementation's storage system that makes the prescribed strategy difficult or impossible). Maybe these SHOULDs about storing state should be deleted and instead we could have an implementation note elsewhere (Operational Considerations?) discussing the intent of the `round` field. and suggesting one possible solution.

draft-ietf-ppm-dap.md
branlwyd: The only part of this thread that's different from the linked thread is that this thread suggests dropping the suggestion to "recompute" -- IIUC, recomputing is a strictly inferior strategy to storing the previously-computed result (since even in a recompute strategy, you still need to store the data required to recompute the result, i.e. recomputation also requires storage).

Agreed w/ your comment in the other thread that we should probably not be very specific about when storage occurs.

draft-ietf-ppm-dap.md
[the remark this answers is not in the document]
branlwyd: I'd be happy with either "collect job" or "collection job". (I have a slight preference for "collection job" as it matches "aggregation job" & is slightly more grammatically correct, but this is very quibbly.)

draft-ietf-ppm-dap.md
branlwyd: I hear you; my thoughts are that we should aim to design things as they "should be" if we were writing the design down for the first time, without necessarily worrying about the size of the diff against previous drafts.

draft-ietf-ppm-dap.md
tgeoghegan: I deleted the SHOULDs about storing anything and punted the discussion of round skew to a new "Recovering From Round Skew".

cjpatton (commented):
Thank you, @tgeoghegan, for all of your hard work on this. I remain supportive of this change and think it's going in a good direction. I have two big-picture concerns that I would like to see addressed before we merge.

**Editorial flow changes.** This PR makes significant changes to the flow of the document in order to re-arrange the text around the new concept of "resources". Below is the TOC of this PR on the left; TOC of main on the right:
image

While I understand the reasoning for this, it has a major downside. The text that's changing here is hard-fought, and quite delicate; and changing it so much makes it very difficult to review for correctness. In addition, I fear that the protocol text will no longer align with people's mental model for how the protocol works. We are due for a major overhaul of the document flow, but I think we need to do so incrementally.

I'll also just say that I'm not sure how useful it is to re-arrange the protocol flow in terms of "resources". It makes sense to describe the API in these terms, and even to make protocol changes that align better with this idea semantically; but at the end of the day we need to describe the sequence of interactions in some chronological fashion. The flow in this PR seems a bit at odds with this.

What I believe our goal should be is to try to preserve all of the API changes, wire changes,  renaming of structs, and behavior changes that this PR entails -- that is, keep *all* of the substantive changes that are made -- but minimize code movement to the greatest possible extent.

Here's what that this might look like. Starting on a fresh branch:

First, before tthe upload, aggregation, and collection sections, add a new section specifying the new API endpoints and summarize what they do.

Second, go through each of the sections line-by-line modifying them as necessary. For example, rewrite
  
  > Clients retrieve the HPKE configuration from each aggregator by
  > sending an HTTP GET request to [aggregator]/hpke_config, where
  > [aggregator] is the aggregator's endpoint URL, obtained from the task
  > parameters.
  
as

 > Clients retrieve the HPKE configuration via the Aggregator's 
> HPKE-config resource API ({{refer-to-section-where-this-is-defined}}).

It also be necessary to add text in certain places, e.g., the text around encoding and validating the round number in the aggregation flow. Also add any related, but technically orthogonal changes you want might want to to suggest.

Third, in the API definition sections, add a forward reference to where the body of the request/response is defined in the text, e.g., "The body of the HTTP request is specifed in {{refer-to-section-where-this-is-defined}}."

All of that said: Perhaps you feel you've done your best to minimize code movement and don't think we can minimize this change any more than you already have. I'm happy to engage on this.

**Recovery.** Enabling data recovery mechanisms is a primary goal of this PR. However, I think these bits of new protocol behavior should be lifted into a future PR so that they can be discussed on their own merits. Here I'm including things like "rewinding" the Helper to a previous state: The costs associated with this, and any security implications, need to be discussed. On the other hand, "plumbing" changes that are required to specify such mechanisms, like the round number, should be kept.

draft-ietf-ppm-dap.md
tgeoghegan: OK, that sounds like a consensus.

- [ ] specify a `CollectionJobId` and collection job URIs

tgeoghegan:
Your point about how disruptive this change is is well-taken. I'll work on chunking it into less drastic changes.

tgeoghegan:
@cjpatton per your suggestion, #398 is just the message struct changes and request paths, without the round skew recovery stuff or any other behavioral changes.

draft-ietf-ppm-dap.md
```diff
@@ -1214,64 +1273,154 @@ as follows:
...
+###### Leader Continuation {#aggregation-leader-continuation}
+
+The Leader begins each round of continuation with a sequence of `PrepareStep`s
+received from the Helper and its locally computed prepare message.
+
+If `PrepareStep` is of type `failed`, then the Leader acts based on the value of
+the `ReportShareError`:
+
+* If the error is `ReportShareError.report_too_early`, then the Leader MAY try
+  to re-send the report in a later `AggregationJob`.
```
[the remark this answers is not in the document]
tgeoghegan: This behavior is not new. The equivalent passage in DAP-03: https://datatracker.ietf.org/doc/html/draft-ietf-ppm-dap-03#section-4.4.2.1-2.1

tgeoghegan:
I've downgraded this PR to a draft since any changes that get merged will happen via #398, #399 and #400.

tgeoghegan:
Closing; the changes discussed here have materialized elsewhere.

PR #398, opened 2023-01-19
Barebones resource oriented API

tgeoghegan:
Following from discussion in #395, this PR aims to only include the changes to request paths and message structures needed to absolve implementations of needing to partially parse messages to figure out how to parse the remainder. Further behavioral changes included in #395 will be submitted as further PRs.

### DIFF
--- a/draft-ietf-ppm-dap.md
+++ b/draft-ietf-ppm-dap.md
@@ -519,6 +519,25 @@ DAP has three major interactions which need to be defined:
 * Computing the results of a given measurement, specified in {{aggregate-flow}}
 * Collecting aggregated results, specified in {{collect-flow}}
 
+In this section, we discuss each of the resources and the messages used to act
+on these resources. A resource's path is resolved relative to a server's
+endpoint to construct a resource URI.
+
+Resource paths are specified as templates like:
+
+~~~
+{role}/resource_type/{resource-id}
+~~~
+
+`{role}` is one of the API endpoints in the task's `aggregator_endpoints` (see
+{{task-configuration}}). The remainder of the path is resolved relative to the
+endpoint.
+
+DAP resource identifiers are opaque byte strings, so any occurrence of
+`{resource-id}` in a URL template (e.g., `{task-id}` or `{report-id}`) MUST
+be expanded to the URL-safe, unpadded Base 64 representation of the corresponding resource
+identifier, as specified in sections 5 and 3.2 of {{!RFC4648}}.
+
 The following are some basic type definitions used in other messages:
 
 ~~~
@@ -537,7 +556,7 @@ struct {
 } Interval;
 
 /* An ID used to uniquely identify a report in the context of a DAP task. */
-uint8 ReportID[16];
+opaque ReportID[16];
 
 /* The various roles in the DAP protocol. */
 enum {
@@ -740,13 +759,9 @@ configuration of each aggregator. See {{compliance}} for information on HPKE
 algorithm choices.
 
 Clients retrieve the HPKE configuration from each aggregator by sending an HTTP
-GET request to `[aggregator]/hpke_config`, where `[aggregator]` is the
-aggregator's endpoint URL, obtained from the task parameters. Clients MAY
-specify a query parameter `task_id` when sending an HTTP GET request to
-`[aggregator]/hpke_config?task_id=[task-id]`, where `[task-id]` is the task ID
-obtained from the task parameters, encoded in Base 64 with URL and filename safe
-alphabet with no padding, as specified in sections 5 and 3.2 of {{!RFC4648}}. If
-the aggregator does not recognize the task ID, then it MUST abort with error
+GET request to `{aggregator}/hpke_config`. Clients MAY specify a query parameter
+`task_id` whose value is the task ID whose HPKE configuration they want. If the
+aggregator does not recognize the task ID, then it MUST abort with error
 `unrecognizedTask`.
 
 An aggregator is free to use different HPKE configurations for each task with
@@ -812,9 +827,11 @@ at least twice the cache lifetime in order to avoid rejecting reports.
 
 ### Upload Request
 
-Clients upload reports by using an HTTP POST to `[leader]/upload`, where
-`[leader]` is the first entry in the task's aggregator endpoints. The payload is
-structured as follows:
+Clients upload reports by using an HTTP PUT to
+`{leader}/tasks/{task-id}/reports`, where `{leader}` is the first entry in the
+task's aggregator endpoints.
+
+The payload is structured as follows:
 
 ~~~
 struct {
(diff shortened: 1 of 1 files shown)


draft-ietf-ppm-dap.md
```diff
@@ -519,6 +519,26 @@ DAP has three major interactions which need to be defined:
 * Computing the results of a given measurement, specified in {{aggregate-flow}}
 * Collecting aggregated results, specified in {{collect-flow}}
 
+In this section, we discuss each of the resources and the messages used to act
+on these resources. A resource's path is resolved relative to a server's
+endpoint to construct a resource URI. A server's endpoint MAY contain
+arbitrarily many path components before the DAP resources.
```
cjpatton, line 525: This guidance doesn't seem necessary to me given the definition of `[role]` below.

```suggestion
endpoint to construct a resource URI.
```

draft-ietf-ppm-dap.md
```diff
@@ -519,6 +519,26 @@ DAP has three major interactions which need to be defined:
...
+
+Resource paths are specified as templates like:
+
+~~~
+[role]/resource_type/{resource-id}
```
cjpatton, line 530: Suggestion: For consistency with the other "variables" in paths, use curly braces instead of square braces. Unless, do you intend `{}` to denote a variable that is expanded into base64 encoding?
```suggestion
{role}/resource_type/{resource-id}
```

draft-ietf-ppm-dap.md
```diff
@@ -812,47 +834,49 @@ at least twice the cache lifetime in order to avoid rejecting reports.
...
+report ID as the nonce, and VDAF-04 will include some guidance explaining that
+if the report ID is chosen with knowledge of the VDAF verify key, privacy could
+be violated. For that reason, we can't allow either aggregator to choose report
+IDs.]
+
+The payload is structured as follows:
 
 ~~~
 struct {
-  ReportID report_id;
```
cjpatton, line 821: Hrmm, I'm a bit wary. of moving the report ID out of the `Report` struct. Can you explain your thinking here?

In any case, we need to make sure that the task ID, report ID, timestamp, and any other metadata that we might have in the future are all included in the report metadata.

draft-ietf-ppm-dap.md
```diff
@@ -1665,9 +1685,6 @@ struct {
 
 The message contains the following parameters:
 
-* The DAP task ID. If the Helper does not recognize the task ID, it MUST abort
-  with error `unrecognizedTask`.
```
cjpatton, line 1669: Reviewer note: This MUST has been moved below.

cjpatton (approved):
👍  Thanks for taking the time to refactor. This is a nice, clean start.

cjpatton (changes requested):
Oops, I was a bit too "approval happy" :) I just have one blocking comment: https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/pull/398#discussion_r1081893133

draft-ietf-ppm-dap.md
tgeoghegan: No, I was just preserving the existing usage of `[leader]`, `[helper]` and `[aggregator]`. Using `{}` is more consistent, though.

draft-ietf-ppm-dap.md
tgeoghegan: The report ID now occurs in the request path `[leader]/tasks/{task-id}/reports/{report-id}`, so it does not need to occur in the request body. `struct ReportMetadata` does still contain `ReportID` so that it can be used to construct `InputShareAad` and `ReportShare`.

I suppose it's not essential that the report ID be in the request path: thus far, it's only the task ID that's needed by servers to figure out how to parse the rest of the message. So we could put report ID back into the body, and then we can use `ReportMetadata` in `Report` again, which is a little nicer.

tgeoghegan: The downside of this change is that it's inconsistent with the URIs for aggregation jobs and collections, but I don't really think that matters.

branlwyd (approved):

draft-ietf-ppm-dap.md
cjpatton: Consistency is a good idea, I think. The curly brackets look nicer.

draft-ietf-ppm-dap.md
cjpatton: My vote would be to move report ID back into body so that the metadata appears in both report and report share. If you decide to to keep the report ID where it is, then my only request is to make sure to spell out how the metadata is constructed where we specify encryption of the input shares.

cjpatton: Oh looks like you already updated this! Thanks.

cjpatton (approved):

draft-ietf-ppm-dap.md
```diff
@@ -519,6 +519,25 @@ DAP has three major interactions which need to be defined:
 * Computing the results of a given measurement, specified in {{aggregate-flow}}
 * Collecting aggregated results, specified in {{collect-flow}}
 
+In this section, we discuss each of the resources and the messages used to act
+on these resources. A resource's path is resolved relative to a server's
+endpoint to construct a resource URI.
```
divergentdave, line 524: Suggested rephrasing, to introduce the resources up front:
```suggestion
In this section, we discuss each of these interactions, the resources they act on (report, aggregation job, and collection job), and the messages used to do so. A resource's path is resolved relative to a server's
endpoint to construct a resource URI.
```

draft-ietf-ppm-dap.md
```diff
@@ -519,6 +519,25 @@ DAP has three major interactions which need to be defined:
...
+~~~
+
+`{role}` is one of the API endpoints in the task's `aggregator_endpoints` (see
+{{task-configuration}}). The remainder of the path is resolved relative to the
+endpoint.
+
+DAP resource identifiers are always opaque byte strings, so anywhere
+`{resource-id}` occurs in a template (e.g., `{task-id}` or `{report-id}`) MUST
+be expanded to the URL-safe, unpadded Base 64 representation of the resource's
+identifier, as specified in sections 5 and 3.2 of {{!RFC4648}}.
```
divergentdave, line 539: rephrasing:
```suggestion
DAP resource identifiers are opaque byte strings, so any occurrence of
`{resource-id}` in a URL template (e.g., `{task-id}` or `{report-id}`) MUST
be expanded to the URL-safe, unpadded Base 64 representation of the corresponding resource
identifier, as specified in sections 5 and 3.2 of {{!RFC4648}}.
```

draft-ietf-ppm-dap.md
```diff
@@ -1539,68 +1536,77 @@ aggregate result. This entire process is composed of two interactions:
...
-After receiving the response to its CollectReq, the collector makes an HTTP GET
-request to the collect job URI to check on the status of the collect job and
-eventually obtain the result. If the collect job is not finished yet, the leader
-responds with HTTP status 202 Accepted. The response MAY include a Retry-After
-header field to suggest a pulling interval to the collector.
+After receiving the response to its `CollectionReq`, the collector makes an HTTP
+`POST` request to the collection job URI to check on the status of the collect
+job and eventually obtain the result. If the collection  job is not finished
+yet, the leader responds with HTTP status 202 Accepted. The response MAY include
+a Retry-After header field to suggest a pulling interval to the collector.
```
divergentdave, line 1596: pre-existing typo
```suggestion
a Retry-After header field to suggest a polling interval to the collector.
```

draft-ietf-ppm-dap.md
```diff
@@ -823,36 +840,42 @@ struct {
...
 
-* A timestamp representing the time at which the report was generated.
-  Specifically, the `time` field is set to the number of seconds elapsed since
-  the start of the UNIX epoch. The client SHOULD round this value down to the
-  nearest multiple of `time_precision` in order to ensure that that the
-  timestamp cannot be used to link a report back to the Client that generated
-  it.
+* `public_share` is the public share output by the VDAF sharding algorithm. Note
+  that the public share might be empty, depending on the VDAF. For example,
+  Prio3 has an empty public share, but Poplar1 does not. See {{!VDAF}}.
```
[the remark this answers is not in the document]
tgeoghegan: We can make that change in a separate PR where DAP will adapt to VDAF-04, but I want to keep this PR focused on its stated goal.

draft-ietf-ppm-dap.md
tgeoghegan: I don't want to have to keep this list of resources in sync with the remainder of the document.

PR #400, opened 2023-01-20
Aggregation: round skew recovery

tgeoghegan:
Introduces the `round` field to `AggregationJob` and discusses how aggregators should recover from round skew.

### DIFF
--- a/draft-ietf-ppm-dap.md
+++ b/draft-ietf-ppm-dap.md
@@ -489,6 +489,7 @@ in the "type" field (within the DAP URN namespace
 | unauthorizedRequest        | Authentication of an HTTP request failed (see {{request-authentication}}). |
 | missingTaskID              | HPKE configuration was requested without specifying a task ID. |
 | queryMismatch              | Query type indicated by a message does not match the task's query type. |
+| roundMismatch              | The aggregators disagree on the current round of the VDAF preparation protocol. |
 
 This list is not exhaustive. The server MAY return errors set to a URI other
 than those defined above. Servers MUST NOT use the DAP URN namespace for errors
@@ -1207,13 +1208,21 @@ This message consists of:
 * The sequence of report shares to aggregate. The `encrypted_input_share` field
   of the report share is the `HpkeCiphertext` whose index in
   `Report.encrypted_input_shares` is equal to the index of the aggregator in the
-  task's `aggregator_endpoints` to which the AggregateInitializeReq is being
+  task's `aggregator_endpoints` to which the `AggregationJobInitReq` is being
   sent.
 
-Let `{aggregator}` denote the helper's API endpoint. The leader sends a PUT
+Let `{aggregator}` denote the Helper's API endpoint. The Leader sends a PUT
 request to `{aggregator}/tasks/{task-id}/aggregation_jobs/{aggregation-job-id}`
 with its `AggregationJobInitReq` message as the payload. The media type is
-"application/dap-aggregation-job-init-req".
+"application/dap-aggregation-job-init-req". The Leader's aggregation job is now
+in round 0.
+
+The Helper's response will be an `AggregationJobResp` message (see
+{{aggregation-helper-init}}, which the Leader validates according to the
+criteria in {{aggregation-job-validation}}. If the message is valid, the Leader
+moves to the aggregation job continuation phase with the enclosed prepare steps,
+as described in {{agg-continue-flow}}. Otherwise, the Leader should abandon the
+aggregation job entirely.
 
 #### Helper Initialization {#aggregation-helper-init}
 
@@ -1225,10 +1234,10 @@ carrying a VDAF-specific message for each report share.
 
 To begin this process, the helper first checks if it recognizes the task ID, if
 not then it MUST abort with error `unrecognizedTask`. Then the helper checks
-that the report IDs in `AggregateInitializeReq.report_shares` are all distinct.
+that the report IDs in `AggregationJobInitReq.report_shares` are all distinct.
 If two ReportShare values have the same report ID, then the helper MUST abort
 with error `unrecognizedMessage`. If this check succeeds, the helper then
-attempts to recover each input share in `AggregateInitializeReq.report_shares`
+attempts to recover each input share in `AggregationJobInitReq.report_shares`
 as follows:
 
 1. Decrypt the input share for each report share as described in
@@ -1243,7 +1252,7 @@ as follows:
 
 Once the helper has processed each valid report share in
 `AggregationJobInitReq.report_shares`, the helper then creates an
-`AggregationJob` message to complete its initialization. This message is
+`AggregationJobResp` message to complete its initialization. This message is
 structured as follows:
 
 ~~~
@@ -1266,7 +1275,7 @@ struct {
 
 struct {
   PrepareStep prepare_steps<1..2^32-1>;
-} AggregationJob;
+} AggregationJobResp;
 ~~~
 
 The message is a sequence of PrepareStep values, the order of which matches that
@@ -1274,20 +1283,11 @@ of the ReportShare values in `AggregationJobInitReq.report_shares`. Each report
 that was marked as invalid is assigned the `PrepareStepState` `failed`.
 Otherwise, the `PrepareStep` is either marked as continued with the output
 `prep_msg`, or is marked as finished if the VDAF preparation process is finished
-for the report share.
+for the report share. The Helper's aggregation job is now in round 0.
 
 On success, the helper responds to the leader with HTTP status code 201 Created
-and a body consisting of the `AggregationJob`, with media type
-"application/dap-aggregation-job".
-
-Upon receipt of a helper's `AggregationJob` message, the leader checks
-that the sequence of `PrepareStep` messages corresponds to the `ReportShare`
-sequence of the `AggregationJobInitReq`. If any message appears out of order, is
-missing, has an unrecognized report ID, or if two messages have the same report
-ID, then the leader MUST abort with error `unrecognizedMessage`.
-
-[[OPEN ISSUE: the leader behavior here is sort of bizarre -- to whom does it
-abort?]]
+and a body consisting of the `AggregationJobResp`, with media type
+"application/dap-aggregation-job-resp".
 
 #### Input Share Decryption {#input-share-decryption}
 
@@ -1416,6 +1416,23 @@ If either step fails, the aggregator marks the report as invalid with error
 is the last round of the VDAF, then `out` is the aggregator's output share.
 Otherwise, `out` is the pair `(prep_state, prep_msg)`.
 
+#### Aggregation Job Validation {#aggregation-job-validation}
+
+During the aggregation job initialization ({{leader-init}}) or continuation
+({{agg-continue-flow}}) phases, the Leader will receive an `AggregationJobResp`
+message from the Helper, which should be validated before the Leader can move to
+the next phase of the aggregation protocol.
+
+An `AggregationJobResp` is valid only if it satisfies the following requirement:
+
+* The Helper's `prepare_steps` MUST include exactly the same report IDs in the
+  same order as either the `report_shares` in the Leader's
+  `AggregationJobInitReq` (if this is the first round of continuation) or the
+  `prepare_steps` in the Leader's `AggregationJobContinueReq` (if this is a
+  subsequent round).
+
+[[OPEN ISSUE: consider relaxing this ordering constraint. See issue#217.]]
+
 ### Aggregate Continuation {#agg-continue-flow}
 
 In the continuation phase, the leader drives the VDAF preparation of each share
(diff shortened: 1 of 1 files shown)


draft-ietf-ppm-dap.md
```diff
@@ -1490,33 +1521,73 @@ if the report share should continue being prepared:
...
+
+When implementing a round skew recovery strategy, it is important for Helpers to
+ensure that the Leader's `AggregationJob` message did not change when it was
+re-sent (i.e., the two messages must contain the same set of report IDs and
+prepare messages). This rules out attacks in which the Leader could violate the
+privacy of a report by re-running a VDAF preparation round with different
+prepare messages. One way a Helper could address this would be to store a digest
+of the Leader's request, indexed by aggregation job ID and round, and refuse to
+service a request for a given aggregation job round unless it matches the
+previously seen request (if any).
```
tgeoghegan, line 1587: [@cjpatton points out](https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/pull/395/files#r1070186916):

> Whether there is an attack here is speculation. I'm also not certain the defense is sufficient.

Should we drop this requirement until we have a stronger motivation for it? Or should we err on the side of caution?

branlwyd: IMO, we should keep some language suggesting that the helper check that the retransmitted message matches the original message. Less strongly, IMO, this should be a requirement (but perhaps this could be a suggestion instead, e.g. a `SHOULD` instead of a `MUST`).

My concern is less about an attack (I'm also not sure there is an attack here[1]), and more about consistency/coherence -- it would be _extremely_ confusing if some buggy Leader behavior caused them to retransmit a different message the second time around, and they got an answer as if they had sent the original message.

[1] Not caused by this PR, but it looks like neither the leader nor the helper checks that the request & response have the same reports in the same order during aggregation continuance. (The spec does dictate the helper's responses `MUST` be in the same order, but no one is required to check this.) Am I missing something or should this be fixed?

cjpatton: I'm fine with the requirement for checking consistency. One question I have: What triggers the consistency check? Is it noticing that for a given agg job, the current agg request has the same round number as the current one?

I also agree with Brandon: Consistency/coherence is a stronger case for this requirement. I think SHOULD is appropriate; if we want to lift to MUST, then I think we had better be clearer about the actual mechanism.

That said, I think we should leave a note about the potential attack vector. The problem is the Leader's ability to "rewind" and the Helper and retry on different inputs. For instance, if the VDAF verification key changes, then the Leader gets to see different verifier shares produced by the Helper. This is potentially an avenue of attack, and I'm not sure it's ruled out by the consistency check. We'll have to wait for security analysis to get more clarity here.

I think our best bet is to leave this as an OPEN ISSUE. Something like:

> [[OPEN ISSUE: Allowing the Leader to "rewind" the state of an aggregation job may be an avenue of attack on privacy. For instance, if the VDAF verification key changes, the preparation shares in the Helper's response would change even if the consistency check is made. Security analysis is required.]]

> @branlwyd [1] Not caused by this PR, but it looks like neither the leader nor the helper checks that the request & response have the same reports in the same order during aggregation continuance. (The spec does dictate the helper's responses MUST be in the same order, but no one is required to check this.) Am I missing something or should this be fixed?

Hmm, I believe this is supposed to be enforced by the Helper. It should be fixed if not.

tgeoghegan: Chris:
> What triggers the consistency check? Is it noticing that for a given agg job, the current agg request has the same round number as the current one?

Yes. We want to make sure that the leader isn't doing exactly what you describe, rewinding state and re-running an aggregation round with different broadcast prep messages.

I think I like the idea of making this a SHOULD so that implementations can defend against innocent bugs, and we'll leave an OPEN ISSUE about whether this needs to be upgraded to a MUST as an attack mitigation.

Bran:
> [1] Not caused by this PR, but it looks like neither the leader nor the helper checks that the request & response have the same reports in the same order during aggregation continuance. (The spec does dictate the helper's responses MUST be in the same order, but no one is required to check this.) Am I missing something or should this be fixed?

At [the bottom of "Helper Initialization"](https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/blob/71ccb43c4356228c70219bf87f6c1ee4abadf8c2/draft-ietf-ppm-dap.md?plain=1#L1288) and [the bottom of "Helper Continuation"](https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/blob/71ccb43c4356228c70219bf87f6c1ee4abadf8c2/draft-ietf-ppm-dap.md?plain=1#L1560), we have text that states the leader has to apply the criteria in [`{{aggregation-job-validation}}`](https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/blob/71ccb43c4356228c70219bf87f6c1ee4abadf8c2/draft-ietf-ppm-dap.md?plain=1#L1421). Admittedly it's confusing that this leader behavior appears in sections for the helper. The intent is that the text flows like this:

- describe leader request
- describe helper handling of request and how helper responds
- describe leader handling of helper's response

Which follows the order in which this happens. We could move those paragraphs into "Leader Initialization" and "Leader Continuation", respectively, but then the flow of text would be:

- describe leader request
- describe leader handling of helper's response
- describe helper handling of request and how helper responds

I'll give that a try and see how it reads.

The other thing is that currently, the leader is allowed to change the order of reports across aggregation rounds. I believe that's already the case in DAP-03 (I don't see anything in [DAP-03 leader continuation](https://datatracker.ietf.org/doc/html/draft-ietf-ppm-dap-03#section-4.4.2.1) governing the order of `AggregateContinueReq.prepare_steps`. If we want to add ordering requirements on the leader side, I think we should discuss that in a separate PR.

tgeoghegan: I put in a SHOULD and a TODO as discussed, and wrote up #401 to track the security analysis we discussed. Additionally I remixed the text on how the leader validates `AggregationJob` messages and I think it reads better now.

branlwyd (approved):

draft-ietf-ppm-dap.md
branlwyd: Thanks -- IMO it's easier to follow with the leader checks in the Leader Initialization/Continuation sections, I like this change.

draft-ietf-ppm-dap.md
```diff
@@ -1239,30 +1247,27 @@ struct {
...
-The message is a sequence of PrepareStep values, the order of which matches that
-of the ReportShare values in `AggregateInitializeReq.report_shares`. Each report
-that was marked as invalid is assigned the PrepareStepResult `failed`.
-Otherwise, the PrepareStep is either marked as continued with the output
-`prep_msg`, or is marked as finished if the VDAF preparation process is finished
-for the report share.
+The message consists of:
+
+* `round` is the VDAF preparation round that the job is on, which is 0 when the
+  Helper responds to the Leader's `AggregationJobInitReq`.
```
branlwyd, line 1258:
```suggestion
  Helper responds to the Leader's `AggregationInititalizationReq`.
```

draft-ietf-ppm-dap.md
```diff
@@ -1239,30 +1247,27 @@ struct {
 } PrepareStep;
 
 struct {
+  u16 round;
```
branlwyd, line 1250: nit: since `AggregateInitializeResp` is only used for the first round, `round` will always be `0`. We could drop the field entirely from `AggregateInitializeResp` and save a few bytes.

draft-ietf-ppm-dap.md
```diff
@@ -1189,7 +1190,14 @@ Let `[aggregator]` denote the helper's API endpoint. The leader sends a POST
 request to `[aggregator]/aggregate` with its AggregateInitializeReq message as
 the payload. The media type is "application/dap-aggregate-initialize-req".
 
-#### Helper Initialization
+The Helper's response will be an `AggregationJob` message (see
```
branlwyd, line 1193:
```suggestion
The Helper's response will be an `AggregateInitializeResp` message (see
```

draft-ietf-ppm-dap.md
```diff
@@ -1390,6 +1395,28 @@ If either step fails, the aggregator marks the report as invalid with error
 is the last round of the VDAF, then `out` is the aggregator's output share.
 Otherwise, `out` is the pair `(prep_state, prep_msg)`.
 
+#### Aggregation Job Validation {#aggregation-job-validation}
+
+During the aggregation job initialization ({{leader-init}}) or continuation
+({{agg-continue-flow}}) phases, the Leader will receive an `AggregationJob`
```
branlwyd, line 1401: The `AggregationJob` message no longer exists, I think it was replaced by `Aggregate{Initialize,Continue}{Req,Resp}` based on context. (applies to various references to `AggregationJob` throughout this PR)

draft-ietf-ppm-dap.md
```diff
@@ -1473,43 +1500,74 @@ prepared:
...
+allow an attack on privacy. For instance, if the VDAF verification key changes,
+the preparation shares in the Helper's response would change even if the
+consistency check is made. Security analysis is required. See #401.]]
+
+One way a Helper could address this would be to store a digest of the Leader's
+request, indexed by aggregation job ID and round, and refuse to service a
+request for a given aggregation job round unless it matches the previously seen
+request (if any).
+
+As there are several implementation strategies for round skew recovery,
```
branlwyd, line 1569: nit: IMO, this sentence could be dropped -- we don't specify a solution to many implementation problems, and we don't need to call out every instance. (We _could_ also drop the preceding paragraph, as it does specify one possible implementation strategy for the round-skew recovery check.)

branlwyd (changes requested):

draft-ietf-ppm-dap.md
tgeoghegan: No, `AggregationJob` is what we want. This is also because I forgot to rebase on `main` after merging #398.

draft-ietf-ppm-dap.md
tgeoghegan: This is confusing, but `AggregationJobInitReq` is correct. The diff looked funny when you reviewed because I forgot to rebase on `main` after merging #398.

draft-ietf-ppm-dap.md
tgeoghegan: `AggregationJob` is correct, I forgot to rebase this on `main` after merging #398.

draft-ietf-ppm-dap.md
tgeoghegan: You're right. And on subsequent rounds, we require the helper to echo back the round value that the leader sent in its request, so even there it carries no information and is just an opportunity for the helper to have a bug. So it seems like we'd want to stop using `struct AggregationJob` in three places and go back to `AggregateContinueReq` and `AggregateContinueResp`. I was trying to do a REST thing here -- the servers send a REpresentation of the STate of the aggregation job back and forth -- but I'm backing away from that particular ledge now.

tgeoghegan: I re-introduced `AggregationJobContinueReq` and `AggregationJobResp` messages. `AggregationJobContinueReq` contains the round field and is sent by the leader to helper.  `AggregationJobResp` omits the round field and is used as the response to both `AggregationJobInitReq` and `AggregationJobContinueReq`. I now wish I hadn't collapsed all the message types together back in #398, but so it goes.

draft-ietf-ppm-dap.md
```diff
@@ -1462,19 +1478,38 @@ finishes and stores its output share for further processing as described in
...
+`AggregationJobContinueReq`:
+
+~~~
+struct {
+  u16 round;
+  PrepareStep prepare_Steps<1..2^32-1>;
+} AggregationJobContinueReq;
+~~~
+
+`round` MUST be set to the Leader's current round and the `prepare_steps` field
```
branlwyd, line 1494: Consider describing the expectation on the sequence of `round` values, i.e. that the first `AggregationJobContinueReq` will have a round of 1 and subsequent requests will increment the round by 1 (I think).

To make the later text around "If the Leader's `round` is behind or more than one round ahead of the Helper's current round" work out, I think we also need to describe the Helper is considered to be at round 0 after aggregation initialization.

tgeoghegan: Good point. I think I have made this clear in the latest commit.

draft-ietf-ppm-dap.md
```diff
@@ -1492,33 +1531,74 @@ if it should continue preparing the report share:
...
+Aggregators to ensure that their peer is on an expected round of the VDAF
+preparation algorithm. In particular, the intent is to allow recovery from a
+scenario where the Helper successfully advances from round `n` to `n+1`, but its
+`AggregationJobResp` response to the Leader gets dropped due to something like a
+transient network failure. The Leader could then resend the request to have the
+Helper advance to round `n+1` and the Helper should be able to retransmit the
+`AggregationJobContinueReq` that was previously dropped. To make that kind of
+recovery possible, Aggregator implementations SHOULD checkpoint preparation
+state to durable storage such that Leaders can re-construct continuation
+requests and Helpers can re-construct continuation responses as needed.
```
cjpatton, line 1585: * What can go wrong if an Aggregator doesn't follow this SHOULD? Should this be a MUST?
* What state is "checkpointed"? It would probably be helpful to be specific here: Do we need to put off committing nonces, or output shares, until a certain point?

cjpatton (commented):
I think a bit of clarification around how round skew recovery works would be helpful.

cjpatton (approved):

branlwyd (approved):

tgeoghegan:
Nobody's asked for any changes on this for about week, so let's merge it!

PR #403, opened 2023-02-01
Add `interval` to struct `Collection`

### LINKED ISSUE #397: Aggregate results should include the interval of time that the constituent reports span
[In DAP-03](https://datatracker.ietf.org/doc/html/draft-ietf-ppm-dap-03#section-4.5.2-13), the results of an aggregation are presented to the collector as:

```
struct {
  uint64 report_count;
  HpkeCiphertext encrypted_agg_shares<1..2^32-1>;
} CollectResp;
```

where `encrypted_agg_shares` is the encrypted form of [a VDAF's `AggResult` associated type](https://datatracker.ietf.org/doc/html/draft-irtf-cfrg-vdaf-03#name-constants-and-types-defined). In the case of a time interval task, the collector will have explicitly requested collection over some interval of time, but in the case of a fixed size task, that aggregate result will have been compiled from a set of reports chosen by the leader. So the collector can only infer that the reports are from no later than when the collect request came in and no sooner than when the collector got the previous batch.

DAP should explicitly tell the collector what time interval is covered by a collected batch.

tgeoghegan:
The Leader now informs the Collector of what time interval is spanned by the reports aggregated into a collection, regardless of the query type. This is particularly useful for fixed size tasks as otherwise a Collector can only guess at the report times based on the timing of collection requests, but also useful in time interval tasks since the aggregated reports could span a *smaller* interval that what was requested (e.g., if the collector requested a one hour time interval, but reports only arrived during minutes 30-40 of that hour).

Resolves #397

### DIFF
--- a/draft-ietf-ppm-dap.md
+++ b/draft-ietf-ppm-dap.md
@@ -1617,6 +1617,7 @@ job with HTTP status code 200 OK and a body consisting of a `Collection`:
 struct {
   PartialBatchSelector part_batch_selector;
   uint64 report_count;
+  Interval interval;
   HpkeCiphertext encrypted_agg_shares<1..2^32-1>;
 } Collection;
 ~~~
@@ -1630,6 +1631,12 @@ This structure includes the following:
   [OPEN ISSUE: What should the Collector do if the query type doesn't match?]
 
 * The number of reports included in the batch.
+
+* The smallest interval of time that contains the timestamps of all reports
+  included in the batch, such that the interval's start and duration are
+  both multiples of the task's `time_precision` parameter. Note that in the case
+  of a `time_interval` type query (see {{query}}), this interval can be smaller
+  than the one in the corresponding `CollectionReq.query`.
 
 * The vector of encrypted aggregate shares. They MUST appear in the same order
   as the aggregator endpoints list of the task parameters.


draft-ietf-ppm-dap.md
```diff
@@ -1631,6 +1632,11 @@ This structure includes the following:
 
 * The number of reports included in the batch.
 
+* The smallest interval of time that contains the timestamps of all reports
+  included in the collection. Note that in the case of a `time_interval` type
```
divergentdave, line 1636: I think we should specify that the interval be aligned to `time_precision`, for clarity.
```suggestion
* The smallest interval of time that contains the timestamps of all reports
  included in the collection, such that the interval's start and duration are
  both multiples of the task's `time_precision` parameter. Note that in the case of a `time_interval` type
```

divergentdave (approved):

draft-ietf-ppm-dap.md
tgeoghegan: Won't this be the case already if the client rounded the timestamp before upload? But then I suppose there's no harm in the leader doing that again just in case the client didn't for some reason.

divergentdave: The chief ambiguity I want to squash here is that, if every timestamp is equal, we should be reporting only `Interval { ts, time_precision }`, not, say, `Interval { ts, 1 }`.

branlwyd (approved):

draft-ietf-ppm-dap.md
```diff
@@ -1631,6 +1632,12 @@ This structure includes the following:
 
 * The number of reports included in the batch.
 
+* The smallest interval of time that contains the timestamps of all reports
+  included in the collection, such that the interval's start and duration are
```
cjpatton, line 1636:
```suggestion
  included in the batch, such that the interval's start and duration are
```

cjpatton (approved):

Issue #384, opened 2022-11-29
Aggregation phase naming


simon-friedberger:
We currently have AggregateInitializeReq/-Resp and AggregateContinueReq/-Resp containing the word "Aggregate" even though no aggregation is performed in these steps. In some places we also speak of verification, the VDAF draft calls it simply preparation.
"Verification" would also be a natural choice. VDAF chose preparation because of the option to have a non-V DAF and because other tasks, like format conversion, are included as well.

We might want to adopt the term preparation as well. 

I've started a draft of a report flow here which might help with the discussion:

![image](https://user-images.githubusercontent.com/103195467/204466721-96bd7139-c1a7-42e1-b91f-d9e9b3d0b054.png)

[...]

PR #410, opened 2023-02-23
Update requirements for share validation

### LINKED ISSUE #408: VDAF security analysis: Requirements for report processing
The VDAF analysis [posted on the CFRG list](https://mailarchive.ietf.org/arch/msg/cfrg/-errBjFRvCqi7KuAxoZwH6iY4Vc/) suggests the following restrictions is needed:

> Poplar1: Applications MUST ensure that reports are aggregated at most once
> per-level. Repeated usage of the authenticator at a given level may lead to
> subtle attacks on privacy. In particular, the current
> "max_batch_query_count" mechanism of DAP [2, Section 4.5.6] is not
> sufficient.

By "level" we mean the level of the IDPF tree. The basic idea is that the client generates some randomness for each level of the tree, and it is not safe to use this randomness more than once.

Relatedly: 

> Prio3: Applications SHOULD ensure that each report is processed at most
> once. It's probably safe to allow re-processing, but (1) this is not
> captured in our model; (2) this gets hairy in terms of differential privacy
> (see issue https://github.com/cfrg/draft-irtf-cfrg-vdaf/issues/94); and (3)
> there doesn't seem to be a good reason to allow this. (It is explicitly
> disallowed in DAP.)

In other words, the safest bet for Prio3 is to make sure we don't use a proof more than once.

In general, VDAFs may need to restrict how a report is processed based on the aggregation parameters with which it was processed in the past.

simon-friedberger:
Fixes #408

### DIFF
--- a/draft-ietf-ppm-dap.md
+++ b/draft-ietf-ppm-dap.md
@@ -1323,36 +1323,10 @@ the input share is marked as invalid with a corresponding ReportShareError
 error.
 
 Before beginning the preparation step, Aggregators are required to perform the
-following generic checks.
+following checks:
 
 1. Check that the decrypted input share can be decoded. If not, the input share
    MUST be marked as invalid with the error `unrecognized_message`.
-
-1. Check if the report has already been aggregated with this aggregation
-   parameter. If this check fails, the input share MUST be marked as invalid
-   with the error `report_replayed`. This is the case if the report was used in
-   a previous aggregate request and is therefore a replay.
-    * Implementation note: To detect replay attacks, each Aggregator is required
-      to keep track of the set of reports it has processed for a given task.
-      Because honest Clients choose the report ID at random, it is sufficient to
-      store the set of IDs of processed reports. However, implementations may
-      find it helpful to track additional information, like the timestamp, so
-      that the storage used for anti-replay can be sharded efficiently.
-
-1. Check if the report has never been aggregated but is contained by a batch
-   that has been collected. If this check fails, the input share MUST be marked
-   as invalid with the error `batch_collected`. This prevents additional reports
-   from being aggregated after its batch has already been collected.
-
-1. Depending on the query type for the task, additional checks may be
-   applicable:
-    * For fixed_size tasks, the Aggregators need to ensure that each batch is
-      roughly the same size. If the number of reports aggregated for the current
-      batch exceeds the maximum batch size (per the task configuration), the
-      Aggregator MAY mark the input share as invalid with the error
-      "batch_saturated". Note that this behavior is not strictly enforced here
-      but during the collect sub-protocol. (See {{batch-validation}}.) If both
-      checks succeed, the input share is not marked as invalid.
 
 1. Check if the report is too far into the future. Implementors can provide for
    some small leeway, usually no more than a few minutes, to account for clock
@@ -1371,11 +1345,53 @@ following generic checks.
    the same. If so, then the Aggregator MUST mark the input share as invalid
    with error "unrecognized_message".
 
+1. Check if the report may still be aggregated with the current aggregation
+   parameter. This can be done by looking up all aggregation parameters
+   previously used for this report and calling
+
+   ~~~
+   VDAF.is_valid(current_agg_param, previous_agg_params)
+   ~~~
+
+   If this returns false, the input share MUST be marked as invalid with the
+   error `report_replayed`.
+
+    * Implementation note: To detect replay attacks, each Aggregator is required
+      to keep track of the set of reports it has processed for a given task.
+      Because honest Clients choose the report ID at random, it is sufficient to
+      store the set of IDs of processed reports. However, implementations may
+      find it helpful to track additional information, like the timestamp, so
+      that the storage used for anti-replay can be sharded efficiently.
+
+1. Check that a report is only aggregated in a batch if it was included in all
+   previous collections for the batch. Some VDAFs allow a batch to be collected
+   multiple times; the aggregators must check that the exact same set of reports
+   is being aggregated each time. Checking that no reports are omitted
+   from a batch is covered in {{batch-validation}}.
+   If this check fails, the input share MUST be marked as invalid with the error
+   `batch_collected`. This prevents collectors from learning anything about
+   small numbers of reports that are uploaded between two collections of a
+   batch.
+   The helper must consider a batch as collected when it has sent the
+   `AggregateShare` to the leader.
+   [TODO: If a section to clarify report and batch states is added this can be
+   removed. See Issue #384]
+
+1. Depending on the query type for the task, additional checks may be
+   applicable:
+    * For fixed_size tasks, the Aggregators need to ensure that each batch is
+      roughly the same size. If the number of reports aggregated for the current
+      batch exceeds the maximum batch size (per the task configuration), the
+      Aggregator MAY mark the input share as invalid with the error
+      "batch_saturated". Note that this behavior is not strictly enforced here
+      but during the collect sub-protocol. (See {{batch-validation}}.) If both
+      checks succeed, the input share is not marked as invalid.
+
 1. Finally, if an Aggregator cannot determine if an input share is valid, it
-   MUST mark the input share as invalid with error `report_dropped`. This
-   situation arises if, for example, the Aggregator has evicted from long-term
-   storage the state required to perform the check. (See
-   {{reducing-storage-requirements}} for details.)
+   MUST mark the input share as invalid with error `report_dropped`. For
+   example, if the Aggregator has evicted the state required to perform the
+   check from long-term storage. (See {{reducing-storage-requirements}} for
+   details.)
 
 If all of the above checks succeed, the input share is not marked as invalid.
 


simon-friedberger:
Intended for #408

draft-ietf-ppm-dap.md
```diff
@@ -1338,11 +1362,11 @@ following generic checks.
...
       that the storage used for anti-replay can be sharded efficiently.
+    * This also covers report re-use for the Prio3 VDAF because its aggregation
+      parameter is nil.
 
-1. Check if the report has never been aggregated but is contained by a batch
-   that has been collected. If this check fails, the input share MUST be marked
-   as invalid with the error `batch_collected`. This prevents additional reports
-   from being aggregated after its batch has already been collected.
+1. VDAFs may require specific additional checks. For example, Poplar1 requires
+   that reports are aggregated at most once at each level of the IDPF tree.
```
tgeoghegan, line 1369: I don't think a prescription like this suffices. This should be written in terms of some function exposed by VDAFs, so that DAP implementations can remain generic. So I believe we're blocked on a corresponding VDAF change here.

draft-ietf-ppm-dap.md
```diff
@@ -1338,11 +1362,11 @@ following generic checks.
       store the set of IDs of processed reports. However, implementations may
       find it helpful to track additional information, like the timestamp, so
       that the storage used for anti-replay can be sharded efficiently.
+    * This also covers report re-use for the Prio3 VDAF because its aggregation
+      parameter is nil.
 
-1. Check if the report has never been aggregated but is contained by a batch
-   that has been collected. If this check fails, the input share MUST be marked
-   as invalid with the error `batch_collected`. This prevents additional reports
-   from being aggregated after its batch has already been collected.
```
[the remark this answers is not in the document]
simon-friedberger: Since the previous n which have been aggregated cannot be used anymore this is already covered. The only way to aggregate again is if there are another $threshold reports for the same time interval. In that case it's fine.

Maybe the phrasing is indeed the problem. Afaiu "is in a batch which has been collected" and "has never been aggregated" are contradictory. I assumed it meant "batch definition" i.e. "time interval".

draft-ietf-ppm-dap.md
simon-friedberger: Agreed, this is supposed to be a bandaid. Until we have the VDAF change we have the only requirement that we know of covered here.

cjpatton: I actually think that @simon-friedberger is more or less on the right track here, however I agree we want to be generic. Here is where we have converged on the VDAF spec so far (https://github.com/cfrg/draft-irtf-cfrg-vdaf/pull/161, please feel free to offer feeedback there as well):

> A VDAF *MAY* specify the "validity scope" of each aggregation parameter. Protocols using a VDAF (like DAP) *MUST* ensure that for each validity scope, a report is aggregated at most once with an aggregation parameter of that validity scope. If no validity scope is specified, a report *MUST NOT* be aggregated more than once with a given aggregation parameter.

This provides what we need for the current VDAFs:

* For Poplar1 the validity scope is the IDPF tree level, which is included in the aggregation parameter; the language here seems sufficient to ensure we don't evaluate at the same level more than once.

* For Prio3 we have no validity scope, however the only valid aggregation parameter is the empty string. The language implies we don't aggregate the same report more than once [1].

So how does this translate into normative text for input share validation? Maybe something like this:

> Check if the input share was previously aggregated with the aggregation parameter's validity scope {{reference-to-definition-in-the-vdaf-draft}}. If so, the Aggregator *MUST* mark the input share as invalid with the error "XXX".

cc/ @schoppmp

[1] The attack we're considering here: We want to make sure the attacker can't query the same input/proof on multiple distinct query randomnesses. If we can ensure by some other means that both the nonce and verify_key can't change overtime, then this may not be strictly necessary. Since we haven't really settled the details here, making sure we aggregate each report at most once seems like a reasonable conservative choice.

draft-ietf-ppm-dap.md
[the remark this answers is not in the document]
cjpatton: > Since the previous n which have been aggregated cannot be used anymore this is already covered. The only way to aggregate again is if there are another $threshold reports for the same time interval. In that case it's fine.

Hmm, I think I'm missing something important here. Can you clarify? Is there language elsewhere that covers this?

So the requirement is: "once a batch has been defined, do not add any more reports to that batch".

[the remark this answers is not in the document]
simon-friedberger: Okay, so how about
```
   A report must never be used in more than one batch. If a report with the same report ID is contained in a batch    for which the `AggregateShare` has been sent the input share MUST be marked as invalid with the error `batch_collected`.
```
This makes the requirement that Chris stated explicit, and removes the somewhat contradictory statement about a batch which has been collected but contains a report which has not been aggregated.

draft-ietf-ppm-dap.md
simon-friedberger: I agree with Tim, since this is VDAF specific it should be implemented on the VDAF side.
How about:
> Check if the input share was not previously aggregated with the aggregation parameter's validity scope by calling `VDAF.is_valid(report_id, agg_param)`. Otherwise, the Aggregator MUST mark the input share as invalid with the error "invalid_aggregation_parameter".

cjpatton: Something like that, yeah. However, `VDAF.is_valid()` would also need to take as input the the agg params that have been used for report_id in the past.

simon-friedberger: @schoppmp Yes, that sounds good, so `validity_scope = VDAF.validity_scope(agg_param)` would extract parts of the agg_param (or otherwise construct) an opaque bitstring and DAP ensures that (report_id, validity_scope) has never been aggregated before, right?

cjpatton: Yup, that sounds right to me.

draft-ietf-ppm-dap.md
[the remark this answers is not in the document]
tgeoghegan: I don't think that phrasing is right, because to me it doesn't capture that you can't add a report to a batch once it has been collected. How about this:

```
1. Some VDAFs allow a batch to be collected multiple times with different aggregation parameters, but the aggregators must check that the same set of reports is being aggregated each time. A report may not be aggregated into a collection unless it was included in all previous collections for the batch. If this check fails, the input share MUST be marked as invalid with the error `batch_collected`. This prevents collectors from learning anything about small numbers of reports that are uploaded between two collections of a batch.
```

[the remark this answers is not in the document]
simon-friedberger: I was trying to clarify what "once it has been collected" means with "for which the `AggregateShare` has been sent". Since the helper cannot tell if collection has actually finished (and if there is pro-active aggregation maybe even if it has started) it can only check if it has sent the `AggregateShare` to the Leader.

draft-ietf-ppm-dap.md
tgeoghegan: IIUC we want to write this section in terms of the interface being added in https://github.com/cfrg/draft-irtf-cfrg-vdaf/pull/161

simon-friedberger:
Update with comments so far.

draft-ietf-ppm-dap.md
```diff
@@ -1371,11 +1345,53 @@ following generic checks.
...
+   previous collections for the batch. (Checking that no batches are omitted
+   from a batch is covered in {{batch-validation}}.)
+   If this check fails, the input share MUST be marked as invalid with the error
+   `batch_collected`. This prevents collectors from learning anything about
+   small numbers of reports that are uploaded between two collections of a
+   batch.
+   The helper must consider a batch as collected when it has sent the
+   `AggregateShare` to the leader.
+
+The following additional checks may apply depending on the VDAF and query type.
```
cjpatton, line 1366: I would say the agg param validity scope check should be made for all VDAFs. Where we landed in 04 is to require that each VDAF specify a method for checking it: https://stackoverflow.com/questions/868568/what-do-the-terms-cpu-bound-and-i-o-bound-mean

draft-ietf-ppm-dap.md
```diff
@@ -1371,11 +1345,53 @@ following generic checks.
...
+
+1. Check if the report has already been aggregated with an aggregation
+   parameter which is equivalent to the current one. To determine this, compute
+
+   ~~~
+   validity_scope = VDAF.validity_scope(agg_param)
+   ~~~
```
cjpatton, line 1373: API is now `Vdaf.is_valid(agg_param: AggParam, previous_agg_params: Vec[AggParam]) -> Bool`. What we need to do is something like: "Look up all of the aggregation parameters with which the report was previously aggregated; if `Vdaf.is_valid(agg_param, previous_agg_params)` is false, then the input share MUST be marked ...."

draft-ietf-ppm-dap.md
```diff
@@ -1371,11 +1345,53 @@ following generic checks.
...
-1. Finally, if an Aggregator cannot determine if an input share is valid, it
-   MUST mark the input share as invalid with error `report_dropped`. This
-   situation arises if, for example, the Aggregator has evicted from long-term
-   storage the state required to perform the check. (See
-   {{reducing-storage-requirements}} for details.)
+1. If an Aggregator cannot determine if an input share is valid, it
+   MUST mark the input share as invalid with error `report_dropped`. For
+   example, if the Aggregator has evicted the state required to perform the
+   check from long-term storage. (See {{reducing-storage-requirements}} for
+   details.)
```
cjpatton, line 1352: I think this check should go last, since it's kind of a "catch all": I.e., if the Aggregator did not have enough information to perform even one of the previous checks, then we have to invalidate the report.

draft-ietf-ppm-dap.md
```diff
@@ -1371,11 +1345,53 @@ following generic checks.
...
+
+1. Some VDAFs allow a batch to be collected multiple times; the aggregators must
+   check that the exact same set of reports is being aggregated each time.
+   A report may not be aggregated into a batch unless it was included in all
+   previous collections for the batch. (Checking that no batches are omitted
+   from a batch is covered in {{batch-validation}}.)
+   If this check fails, the input share MUST be marked as invalid with the error
+   `batch_collected`. This prevents collectors from learning anything about
+   small numbers of reports that are uploaded between two collections of a
+   batch.
```
cjpatton, line 1362: nit: Wrap paragraphs at 80 characters.

cjpatton (changes requested):

draft-ietf-ppm-dap.md
```diff
@@ -1371,11 +1345,53 @@ following generic checks.
...
+   MUST mark the input share as invalid with error `report_dropped`. For
+   example, if the Aggregator has evicted the state required to perform the
+   check from long-term storage. (See {{reducing-storage-requirements}} for
+   details.)
+
+1. Some VDAFs allow a batch to be collected multiple times; the aggregators must
+   check that the exact same set of reports is being aggregated each time.
+   A report may not be aggregated into a batch unless it was included in all
+   previous collections for the batch. (Checking that no batches are omitted
+   from a batch is covered in {{batch-validation}}.)
```
tgeoghegan, line 1358:
```suggestion
   previous collections for the batch (checking that no reports are omitted
   from a batch is covered in {{batch-validation}}).
```

tgeoghegan (changes requested):

draft-ietf-ppm-dap.md
simon-friedberger: These seem like two sentences to me so I'm not sure why we would remove the split. I'm removing the parentheses instead.

draft-ietf-ppm-dap.md
simon-friedberger: I am a fan of fuzzy 80 but happy to fix. These all show as 80 or less for me, though.

draft-ietf-ppm-dap.md
```diff
@@ -1354,28 +1385,11 @@ following generic checks.
...
-1. Check if the ExtensionType of any two extensions in PlaintextInputShare is
-   the same. If so, then the Aggregator MUST mark the input share as invalid
-   with error "unrecognized_message".
-
-1. Finally, if an Aggregator cannot determine if an input share is valid, it
-   MUST mark the input share as invalid with error `report_dropped`. This
-   situation arises if, for example, the Aggregator has evicted from long-term
-   storage the state required to perform the check. (See
-   {{reducing-storage-requirements}} for details.)
+1. If an Aggregator cannot determine if an input share is valid, it
```
cjpatton, line 1388: nit: This ought to minimize the size of the patch.
```suggestion
1. Finally, if an Aggregator cannot determine if an input share is valid, it
```

draft-ietf-ppm-dap.md
```diff
@@ -1323,26 +1323,57 @@ the input share is marked as invalid with a corresponding ReportShareError
...
+
+1. Check if the PlaintextInputShare contains unrecognized extensions. If so,
+   then the Aggregator MUST mark the input share as invalid with error
+   "unrecognized_message".
+
+1. Check if the ExtensionType of any two extensions in PlaintextInputShare is
+   the same. If so, then the Aggregator MUST mark the input share as invalid
+   with error "unrecognized_message".
+
+1. Check if the report may still be aggregated with the current aggregation
```
cjpatton, line 1348: nit (grammar): Most of the items in this list follow the template: "Check if <SOMETHING BAD HAPPNES>." Can we change this to something like "Check if the aggregation parameter violates the validity scope of the report."

draft-ietf-ppm-dap.md
```diff
@@ -1323,26 +1323,57 @@ the input share is marked as invalid with a corresponding ReportShareError
...
       that the storage used for anti-replay can be sharded efficiently.
 
-1. Check if the report has never been aggregated but is contained by a batch
-   that has been collected. If this check fails, the input share MUST be marked
-   as invalid with the error `batch_collected`. This prevents additional reports
-   from being aggregated after its batch has already been collected.
+1. Some VDAFs allow a batch to be collected multiple times; the aggregators must
+   check that the exact same set of reports is being aggregated each time.
+   A report may not be aggregated into a batch unless it was included in all
+   previous collections for the batch. Checking that no batches are omitted
```
cjpatton, line 1369:
```suggestion
   previous collections for the batch. Checking that no reports are omitted
```

draft-ietf-ppm-dap.md
```diff
@@ -1323,26 +1323,57 @@ the input share is marked as invalid with a corresponding ReportShareError
...
+   from a batch is covered in {{batch-validation}}.
+   If this check fails, the input share MUST be marked as invalid with the error
+   `batch_collected`. This prevents collectors from learning anything about
+   small numbers of reports that are uploaded between two collections of a
+   batch.
+   The helper must consider a batch as collected when it has sent the
+   `AggregateShare` to the leader.
```
cjpatton, line 1376: nit: IMO this doesn't need to be clarified here.
```suggestion
```

cjpatton (commented):
Looking good, just a few minor comments left.

draft-ietf-ppm-dap.md
tgeoghegan: I agree with Chris, but also agree with Simon that DAP should be more clear about the lifecycle or state machine of various objects in the protocol, because otherwise it's kinda hard for implementors to figure out stuff like when a helper aggregate share is considered to be collected. I'd like to have a section in DAP where we discuss all that, so that then other parts of the protocol like this could refer to it, and then it'd be clear just what we mean when we say that a report is aggregated or a batch is collected. We talked about that a bit in https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/issues/384. In any case, I certainly don't want to solve that problem in this PR. Maybe we could take Simon's text and include a TODO referencing #384?

tgeoghegan (approved):
LGTM, modulo Chris' observations.

draft-ietf-ppm-dap.md
cjpatton: Yeah make sense :)

simon-friedberger:
I have two more concerns:
1. We have `report_too_early` do we want to add `report_too_late`? This would cover the "validation state is no longer available" which we currently use as an example of `report_dropped`. It would be clearer to define a lifetime for the state and emit `report_too_late` when it has expired.
2. I think we should maybe change the names. `batch_redefined` would be more accurate than `batch_collected`. Or maybe `batch_overlap`. Same for `report_replayed`, this might be better as `agg_param_invalid` now.

cjpatton:
>     1. We have `report_too_early` do we want to add `report_too_late`? This would cover the "validation state is no longer available" which we currently use as an example of `report_dropped`. It would be clearer to define a lifetime for the state and emit `report_too_late` when it has expired.
>     2. I think we should maybe change the names. `batch_redefined` would be more accurate than `batch_collected`. Or maybe `batch_overlap`. Same for `report_replayed`, this might be better as `agg_param_invalid` now.

I like these suggestions, but they should be discussed in a different PR. Note that  @tgeoghegan has in the past wanted to err on the side of not adding too many error variants.

`batch_overlap` seems better to me :)

cjpatton (approved):

cjpatton:
@simon-friedberger please squash these changes into a single commit, then I'll merge it.

simon-friedberger:
> @simon-friedberger please squash these changes into a single commit, then I'll merge it.

Didn't I already do that?

Issue #161, opened 2021-10-28
Securely provisioning `vdaf_verify_key`


tgeoghegan:
The [VDAF spec](https://datatracker.ietf.org/doc/html/draft-patton-cfrg-vdaf-00) describes how to generate `verify_params` for aggregators in `prio3` and `poplar1`. The verification parameters MUST be kept secret from other clients, which means that they may need to be rotated either if some aggregator's verification parameter is exposed or if a sufficient number of inputs are processed under some verification parameter (this would take a huge number of inputs to be a problem).

cjpatton:
I think this is going to need to be specified. If it's done incorrectly, it could become an attack vector.

Both Prio v1 [[CGB17]] and Heavy Hitters [[BBCGGI21]] require per-report randomness for validating the recovered output shares. For robustness, it is important that the aggregators agree on fresh randomness that is unpredictable to the client. Otherwise it might be possible for a malicious client to garble the aggregate result.

For privacy the picture is murkier. The key question is whether a malicious aggregator can, by controlling the random input the helper consumes, cause the helper to leak information about the measurement. It's possible that the answer depends on the protocol:
- For Prio v1 [[CGB17]], it seems to be just fine for the leader to just pick the randomness and send it to the helper. (See [[CGB17], Section 4.2, Step 3a].)
- For Heavy Hitters [[BBCGGI21]], the process by which the shared randomness is established is unspecified (See [[BBCGGI21], Appendix C.4.].) It may be OK for the leader to just pick it, but this is not guaranteed.
- For schemes based on techniques in [[BBCGGI19]], including `prio3` in the VDAF draft, as well as the code running in ENPA, the situation is similar: The distributed ZKP system described in [[BBCGGI19], Section 6.2] makes use of an "ideal coin-flipping functionality" (see Definition 6.3), which in practice would have to be realized by an actual protocol. Incidentally, I believe ENPA effectively realizes this ideal coin-flipping functionality by having a third party choose the randomness. (Can you confirm, @tgeoghegan?)

[...]

PR #411, opened 2023-02-27
Add security requirements for verification key.

### LINKED ISSUE #407: VDAF security analysis: Requirements for the verification key
The analysis posted [on the CFRG mailing list](https://mailarchive.ietf.org/arch/msg/cfrg/-errBjFRvCqi7KuAxoZwH6iY4Vc/) describes the security consideration for DAP:

> Some care is required in choosing the VDAF verification key. The
> Aggregators may use any strategy they wish to exchange this key, but the
> application MUST ensure that the key that is picked independently of the
> reports. For DAP in particular, this means it is safe for the Leader to
> pick the key and distribute it, but the Aggregators need to "commit" to the
> verification key used for a given task and use the same key for the
> duration of the task. (Rotation requires rolling the task configuration.)

Assuming we agree this is an issue that needs to be addressed, what should we do about it?

simon-friedberger:
Closes #407

### DIFF
--- a/draft-ietf-ppm-dap.md
+++ b/draft-ietf-ppm-dap.md
@@ -741,9 +741,8 @@ of the aggregators is configured with following parameters:
   (described in {{hpke-config}}); see {{compliance}} for information about the
   HPKE configuration algorithms.
 * `vdaf_verify_key`: The VDAF verification key shared by the aggregators. This
-  key is used in the aggregation sub-protocol ({{aggregate-flow}}). [OPEN ISSUE:
-  The manner in which this key is distributed may be relevant to the VDAF's
-  security. See issue#161.]
+  key is used in the aggregation sub-protocol ({{aggregate-flow}}). The security
+  requirements are described in {{verification-key}}.
 
 Finally, the collector is configured with the HPKE secret key corresponding to
 `collector_hpke_config`.
@@ -2476,6 +2475,17 @@ if a majority of runs agree, and a single aggregator appears in every run that
 disagrees). See
 [#22](https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/issues/22) for
 discussion.
+
+## Verification key requirements {#verification-key}
+
+The verification key for a task SHOULD be chosen before any reports are generated.
+It SHOULD be fixed for the lifetime of the task and not be rotated. One way
+to ensure this is to include the verification key in a derivation of the task ID.
+
+This consideration comes from current security analysis for existing VDAFs. For example,
+to ensure that the security proofs for Prio3 hold, the verification key MUST be
+chosen independently of the generated reports. This can be achieved as
+recommended above.
 
 ## Infrastructure diversity
 


draft-ietf-ppm-dap.md
```diff
@@ -2477,6 +2476,17 @@ disagrees). See
 [#22](https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/issues/22) for
 discussion.
 
+## Verification key requirements {#verification-key}
+
+The verification key for a task SHOULD be chosen before any reports are generated.
+It SHOULD be fixed for the lifetime of the task and cannot be rotated. One way
+to ensure this is to include the verification key in a derivation of the task ID.
+
+To ensure that the security proofs for Prio3 hold, the verification key MUST be
```
cjpatton, line 2485: I think it's worth citing Prio3 as an example, but in the interest of being generic, I think we should add a little more context here.

```suggestion
This consideration comes from current security analysis for existing VDAFs. For example,
to ensure that the security proofs for Prio3 hold, the verification key MUST be
```

draft-ietf-ppm-dap.md
```diff
@@ -2477,6 +2476,17 @@ disagrees). See
...
+chosen independently of the generated reports. This can be achieved as
+recommended above. It would also be possible for the aggregators to use a coin
+flipping protocol if key rotation is necessary.
```
cjpatton, line 2488: I don't think "coin flipping protocol" means much to most readers. Do we have a concrete suggestion here? If not, I think we should either cut this suggestion or add an OPEN ISSUE for tracking discussion on this. My preference would be the former.

cjpatton (changes requested):

draft-ietf-ppm-dap.md
tgeoghegan: > if key rotation is necessary.

We say above that verification keys cannot be rotated, so it seems like this sentence can be deleted. And I agree with Chris: the specification is to be read by implementers who aren't necessarily cryptography experts and need concrete advice on how to do this. If I remember right, it's sufficient for the Leader to randomly generate a verification key with a CSPRNG and then transmit it to the Helper over a secure channel (like TLS). If we want to provide a concrete recommendation, does that suffice?

tgeoghegan (changes requested):

draft-ietf-ppm-dap.md
```diff
@@ -2477,6 +2476,17 @@ disagrees). See
 [#22](https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/issues/22) for
 discussion.
 
+## Verification key requirements {#verification-key}
+
+The verification key for a task SHOULD be chosen before any reports are generated.
+It SHOULD be fixed for the lifetime of the task and cannot be rotated. One way
```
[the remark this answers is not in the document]
simon-friedberger: Hm, so this was intended as a default but if both aggregators agree to use a coin flipping protocol instead that would also work. I'm not sure what the typical protocol text for such an OOB agreement would be.

We can make this a MUST but then key rotation really cannot be done anymore.

draft-ietf-ppm-dap.md
cjpatton: That would be sufficient, and I think it would be worth RECOMMENDING this. However this is kind of separate from what @simon-friedberger is suggesting here: If we specified some sort of protocol for exchanging the key whereby each Aggregator committed to and provided some entropy, then we might be able weaken the "don't rotate while running a task" requirement.

draft-ietf-ppm-dap.md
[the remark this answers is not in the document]
cjpatton: Let's go with SHOULD for now. A nudge is sufficient for the time being.

draft-ietf-ppm-dap.md
tgeoghegan: Yeah, I was under the impression that VDAF-04 had already forbidden verification key rotation, but that's not the case. Then I agree, we should leave the door open for coin-flipping schemes.

draft-ietf-ppm-dap.md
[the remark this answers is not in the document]
tgeoghegan: Yes, I was jumping the gun on forbidding verify key rotation. This is fine as is.

draft-ietf-ppm-dap.md
```diff
@@ -2477,6 +2476,18 @@ disagrees). See
...
```
cjpatton, line 2482:
```suggestion
It SHOULD be fixed for the lifetime of the task and not be rotated. One way
```

draft-ietf-ppm-dap.md
simon-friedberger: Would a reference be sufficient clarification? Maybe this?
> @inproceedings{Blum1981CoinFB,
  title={Coin Flipping by Telephone.},
  author={Manuel Blum},
  booktitle={Annual International Cryptology Conference},
  year={1981}
}

Or we could add a description of such a protocol, @cjpatton proposed one in #161.

cjpatton: The Blum paper is the first crypto paper I ever read! Very cool, but very outdated. Moreover, I am not ready to endorse the solution we sketched in #161. I think what we would need here is a concrete proposal + security analysis.

I really think we should do nothing here until we've had a chance to actually work it out.

simon-friedberger: Removed.

cjpatton (approved):

tgeoghegan (approved):

PR #412, opened 2023-02-27
Use the report ID as the nonce during sharding

### LINKED ISSUE #394: Clients need to provide nonce when sharding measurements
In the upcoming VDAF-04, [`Vdaf.measurement_to_input_shares` now takes a `nonce` argument](https://github.com/cfrg/draft-irtf-cfrg-vdaf/blob/main/draft-irtf-cfrg-vdaf.md#sharding-sec-vdaf-shard). The intention is that clients will use the report ID they generate as the nonce, corresponding to [what we already do when we call `Vdaf.prep_init`](https://datatracker.ietf.org/doc/html/draft-ietf-ppm-dap-03#section-4.4.1.5). Once VDAF-04 is out, we will need to align DAP's discussion of report construction with it.

tgeoghegan:
Catches us up with recent VDAF changes ([1], [2], [3]). We also require that VDAFs used in DAP have `NONCE_SIZE = 16`.

[1]: https://github.com/cfrg/draft-irtf-cfrg-vdaf/issues/127
[2]: https://github.com/cfrg/draft-irtf-cfrg-vdaf/pull/149
[3]: https://github.com/cfrg/draft-irtf-cfrg-vdaf/pull/171

Resolves #394

### DIFF
--- a/draft-ietf-ppm-dap.md
+++ b/draft-ietf-ppm-dap.md
@@ -582,6 +582,11 @@ struct {
 struct {} Empty;
 ~~~
 
+DAP uses the 16 byte `ReportID` as the nonce parameter for the VDAF
+`measurement_to_input_shares` and `prep_init` methods (see {{!VDAF, Section 5}}.
+Thus for a VDAF to be compatible with DAP, it MUST specify a `NONCE_SIZE` of 16
+bytes.
+
 ## Queries {#query}
 
 Aggregated results are computed based on sets of report, called batches. The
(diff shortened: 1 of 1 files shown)


cjpatton (approved):

draft-ietf-ppm-dap.md
```diff
@@ -582,6 +582,11 @@ struct {
 struct {} Empty;
 ~~~
 
+DAP uses the 16 byte `ReportID` as the nonce parameter for the VDAF
+`measurement_to_input_shares` and `prep_init` methods (see {{!VDAF, Section 5}}.
```
divergentdave, line 586:
```suggestion
`measurement_to_input_shares` and `prep_init` methods (see {{!VDAF, Section 5}}).
```

divergentdave (approved):

PR #417, opened 2023-03-06
edit: conventions: Update terminology to match VDAF

cjpatton:
Based on #416.

Update the definitions to align better with terminology in the VDAF draft. Also, add "aggregation paramter" and "public 
share".


draft-ietf-ppm-dap.md
```diff
@@ -191,81 +191,83 @@ seen in the clear by any server.
...
+: A plaintext input emitted by a Client (e.g., a count, summand, or string),
+  before any encryption or secret sharing is applied. Depending on the VDAF in
+  use, multiple valus may be grouped into a single measurement. As defined in
+  {{!VDAF}}.
 
 Minimum batch duration:
-: The minimum batch duration permitted for a DAP task, i.e., the minimum time
-  difference between the oldest and newest report in a batch.
+: For a given DAP task, the minimum time difference between the oldest and
+  newest report in a batch.
```
divergentdave, line 256: I think this is incorrect, because the actual times of reports doesn't affect the validity of a batch, only the definition of the batch interval. (In the most common case, batches will consist of a set of reports with the same client timestamp, due to client-side rounding. Such a batch would be valid, because the collector's query spanned the right amount of time, even though the time difference between the oldest and newest report is zero.) Moreover, this was renamed to "time precision" elsewhere in the document, so we should update and realphabetize this glossary entry to match.

draft-ietf-ppm-dap.md
```diff
@@ -191,81 +191,83 @@ seen in the clear by any server.
...
+Public share:
+: The output of the VDAF sharding algorithm broadcast to each of the
+  Aggregators. As defined {{!VDAF}}.
 
-Proof:
-: A value generated by the client and used by the aggregators to verify the
-  client's input.
+Leader:
+: A distinguished Aggregator that coordinates aggregation and collection
+  amongst the Aggregators.
```
divergentdave, line 267: nit: this is out of alphabetical order

draft-ietf-ppm-dap.md
```diff
@@ -191,81 +191,83 @@ seen in the clear by any server.
...
 
 Report:
-: Uploaded to the leader from the client. A report contains the secret-shared
-  and encrypted input and proof.
-
-Server:
-: An aggregator.
+: A cryptographically protected measuremnet uploaded to the Leader by a Client.
```
divergentdave, line 270: nit: spelling
```suggestion
: A cryptographically protected measurement uploaded to the Leader by a Client.
```

draft-ietf-ppm-dap.md
```diff
@@ -191,81 +191,83 @@ seen in the clear by any server.
...
+  execution of the VDAF preparation phase. Many output shares are combined into
+  an aggregate share during the VDAF aggregation phase. As defined in
+  {{!VDAF}}.
 
 Measurement:
-: A single value (e.g., a count) being reported by a client. Multiple
-  measurements may be grouped into a single protocol input.
+: A plaintext input emitted by a Client (e.g., a count, summand, or string),
+  before any encryption or secret sharing is applied. Depending on the VDAF in
+  use, multiple valus may be grouped into a single measurement. As defined in
```
divergentdave, line 251: nit: spelling
```suggestion
  use, multiple values may be grouped into a single measurement. As defined in
```

draft-ietf-ppm-dap.md
cjpatton: Removed minimum batch duration, since it's out of date. I don't think we need to define time precision here, since it's pretty technical.

draft-ietf-ppm-dap.md
```diff
@@ -191,81 +191,79 @@ seen in the clear by any server.
...
 Batch interval:
-: A parameter of the collect or aggregate-share request that specifies the time
-  range of the reports in the batch.
+: A parameter of a query issued by the Collector that specifies the time range
+  of the reports in the batch.
 
 Client:
 : The endpoint from which a user sends data to be aggregated, e.g., a web
-  browser.
+  browser. As defined in {{!VDAF}}.
```
tgeoghegan, line 228: This doesn't seem right. There's more to a DAP client than VDAF's definition of a client, so perhaps rephrase to explain that a DAP client is also a VDAF client.

Additionally, while you're here, referring to the client as an "endpoint" is confusing.

draft-ietf-ppm-dap.md
```diff
@@ -191,81 +191,79 @@ seen in the clear by any server.
...
+: Parameter used to prepare a set of measurements for aggregation (e.g., the
+  candidate prefixes for Poplar1 from {{Section 8 of !VDAF}}). As defined in
+  {{!VDAF}}.
 
 Aggregator:
-: An endpoint that runs the input-validation protocol and accumulates input
-  shares.
+: An endpoint that recieves input shares from Clients and validates and
+  aggregates them with the help of the other Aggregators. As defined in
+  {{!VDAF}}.
```
tgeoghegan, line 214: There's more to a DAP aggregator than VDAF's definition of an aggregator, so perhaps rephrase to explain that a DAP aggregator is also a VDAF aggregator.

draft-ietf-ppm-dap.md
```diff
@@ -191,81 +191,79 @@ seen in the clear by any server.
...
 
 Client:
 : The endpoint from which a user sends data to be aggregated, e.g., a web
-  browser.
+  browser. As defined in {{!VDAF}}.
 
 Collector:
-: The endpoint that receives the output of the aggregation function.
+: The endpoint that selects the aggregation parameter and receives the
+  aggregate result. As defined in {{!VDAF}}.
```
tgeoghegan, line 232: There's more to a DAP collector than VDAF's definition of a collector, so perhaps rephrase to explain that a DAP collector is also a VDAF collector.

draft-ietf-ppm-dap.md
```diff
@@ -191,81 +191,79 @@ seen in the clear by any server.
...
-: An aggregator's share of the output of the VDAF
-  {{!VDAF=I-D.draft-irtf-cfrg-vdaf-03}} preparation step. Many output shares
-  are combined into an aggregate share via the VDAF aggregation algorithm.
-
-Proof:
-: A value generated by the client and used by the aggregators to verify the
-  client's input.
+Public share:
+: The output of the VDAF sharding algorithm broadcast to each of the
+  Aggregators. As defined {{!VDAF}}.
```
tgeoghegan, line 263:
```suggestion
  Aggregators. As defined in {{!VDAF}}.
```

tgeoghegan (changes requested):

draft-ietf-ppm-dap.md
```diff
@@ -191,81 +191,79 @@ seen in the clear by any server.
 
 {::boilerplate bcp14-tagged}
 
-The following terms are used:
-
 Aggregate result:
-: The output of the aggregation function over a given set of reports.
+: The output of the aggregation function over a given set of measurements and
+  aggregation parameter. As defined in {{!VDAF=I-D.draft-irtf-cfrg-vdaf-03}}.
```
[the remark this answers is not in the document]
cjpatton: I was thinking we'd wait to bump this until VDAF-05 is cut.

draft-ietf-ppm-dap.md
cjpatton: Are you alright with "A party that uploads a report?" I believe the "endpoint" language stems from the distinction we made early on between "user" and "client".

draft-ietf-ppm-dap.md
cjpatton: Done, I just removed the reference.

divergentdave (approved):

draft-ietf-ppm-dap.md
tgeoghegan: My problem with "endpoint" is that I read it to mean "a URL relative to which an API can be accessed", which isn't right in the case of the client. "A party that uploads a report" sounds fine.

tgeoghegan (approved):

PR #422, opened 2023-03-06
edit: aggregation nits and distinguish prep_msg from prep_share

cjpatton:
Continue editorial pass up to the collection flow. Along, clarify an
ambiguity in the mapping of VDAF preparation to DAP aggregation
initialization/continuation.

In particular, `VDAF.prep_next()` is meant to take in the "preperation
message" from the previous round and output either (1) the Aggregator's
output share or (2) the Aggregator's next preparation sate and next
"preparation message-share".

We are currently using the term "inbound message" for the former and
"outbound message" for the latter. This seems to be causing us to
conflate "preparation message" with "preparation message-share", which
may result in misunderstanding the protocol flow.

### DIFF
--- a/draft-ietf-ppm-dap.md
+++ b/draft-ietf-ppm-dap.md
@@ -1033,20 +1033,21 @@ of relevant: issue #89]]
 
 ## Verifying and Aggregating Reports {#aggregate-flow}
 
-Once a set of clients have uploaded their reports to the leader, the leader can
-send them to the helpers to be verified and aggregated. In order to enable the
-system to handle very large batches of reports, this process can be performed
-incrementally. Verification of a set of reports is referred to as an aggregation
-job. Each aggregation job is associated with exactly one DAP task, and a DAP
-task can have many aggregation jobs. Each job is associated with an ID that is
-unique within the context of a DAP task in order to distinguish different jobs
-from one another. Each aggregator uses this ID as an index into per-job storage,
-e.g., to keep track of report shares that belong to a given aggregation job.
+Once a set of Clients have uploaded their reports to the Leader, the Leader can
+begin the process of verifying and aggregating them with the Helpers. To enable
+the system to handle very large batches of reports, this process can be
+parallelized across smaller sets of reports. Verification of a set of reports is
+referred to as an "aggregation job". Each aggregation job is associated with
+exactly one DAP task, and a DAP task can have many aggregation jobs. Each job is
+associated with an ID that is unique within the context of a DAP task in order
+to distinguish different jobs from one another. Each aggregator uses this ID as
+an index into per-job storage, e.g., to keep track of report shares that belong
+to a given aggregation job.
 
-To run an aggregation job, the leader sends a message to each helper containing
-the report shares in the job. The helper then processes them (verifying the
-proofs and incorporating their values into the ongoing aggregate) and replies to
-the leader.
+To run an aggregation job, the Leader sends a request to each Helper containing
+the report shares in the job. Each Helper then processes them (verifying the
+proofs and incorporating their values into the ongoing aggregate) and responds
+to the Leader.
 
 The exact structure of the aggregation job flow depends on the VDAF.
 Specifically:
@@ -1064,7 +1065,7 @@ the next batch are coming in. This is because each report is validated
 independently.
 
 This process is illustrated below in {{aggregation-flow-illustration}}. In this
-example, the batch size is 20, but the leader opts to process the reports in
+example, the batch size is 20, but the Leader opts to process the reports in
 sub-batches of 10. Each sub-batch takes two round-trips to process.
 
 ~~~
@@ -1088,7 +1089,7 @@ The aggregation flow can be thought of as having three phases for transforming
 each valid input report share into an output share:
 
 - Initialization: Begin the aggregation flow by sharing report shares with each
-  helper. Each aggregator, including the leader, initializes the underlying VDAF
+  Helper. Each Aggregator, including the Leader, initializes the underlying VDAF
   instance using these report shares and the VDAF configured for the
   corresponding measurement task.
 - Continuation: Continue the aggregation flow by exchanging messages produced by
@@ -1099,7 +1100,7 @@ each valid input report share into an output share:
 
 ### Aggregate Initialization {#agg-init}
 
-The leader begins an aggregation job by choosing a set of candidate reports that
+The Leader begins an aggregation job by choosing a set of candidate reports that
 pertain to the same DAP task and a unique job ID. The job ID is a 16-byte value,
 structured as follows:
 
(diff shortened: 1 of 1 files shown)


draft-ietf-ppm-dap.md
```diff
@@ -1185,13 +1186,11 @@ necessary.]]
 
 This message consists of:
 
-* The opaque, VDAF-specific aggregation parameter provided during the collection
-  flow (`agg_param`),
-
-  [[OPEN ISSUE: Check that this handling of `agg_param` is appropriate when the
-  definition of Poplar is done.]]
```
cjpatton, line 1192: Reviewer note: I think we have already resolve this.

draft-ietf-ppm-dap.md
```diff
@@ -1248,11 +1253,8 @@ as follows:
 1. Initialize VDAF preparation and initial outputs as described in
    {{input-share-prep}}.
 
-[[OPEN ISSUE: consider moving the helper report ID check into
-{{input-share-validation}}]]
```
cjpatton, line 1252: Reviewer note: I believe this pertains to checking for distinctness of report IDs in the request: If so, this check does not go in {{input-share-validation}}, since that section is really about about per-report checks.

draft-ietf-ppm-dap.md
```diff
@@ -1364,22 +1367,28 @@ following checks:
...
+      has completed a collection job for a CollectReq message from the
+      Collector; the Helper considers a batch to be collected once it has
+      responded to an AggregateShareReq message from the Leader. A batches is
+      determined by query ({{query}}) conveyed in these messages. Queries must
+      satisfy the criteria covered in {{batch-validation}}. These criteria are
+      meant to restrict queries in a way make it easy to determine wither a
+      report pertains to a batch that was collected.
+
+      [TODO: If a section to clarify report and batch states is added this can be
+      removed. See Issue #384]
```
cjpatton, line 1387: Reviewer note: I have tried to rework with this chunk so that the actual thing to check is mentioned in the first sentence. @simon-friedberger please double check that I haven't inadvertently changed the meaning here.

cjpatton:
Reviewer note: This change is meant to be strictly editorial, so I'm hoping it'll be straightforward to review. That said, it's quite big, so I'd be happy to split it up if desired.

draft-ietf-ppm-dap.md
```diff
@@ -1033,20 +1033,21 @@ of relevant: issue #89]]
...
-from one another. Each aggregator uses this ID as an index into per-job storage,
-e.g., to keep track of report shares that belong to a given aggregation job.
-
-To run an aggregation job, the leader sends a message to each helper containing
-the report shares in the job. The helper then processes them (verifying the
-proofs and incorporating their values into the ongoing aggregate) and replies to
-the leader.
+Once a set of Clients have uploaded their reports to the Leader, the Leader can
+begin the process of verifying and aggregating them with the Helpers. In order
+to enable the system to handle very large batches of reports, this process can
```
tgeoghegan, line 1038:
```suggestion
begin the process of verifying and aggregating them with the Helpers. To enable the system to handle very large batches of reports, this process can
```

draft-ietf-ppm-dap.md
```diff
@@ -1227,19 +1231,20 @@ aggregation job entirely.
...
 #### Helper Initialization {#aggregation-helper-init}
 
-Each helper begins their portion of the aggregate initialization phase with the
-set of candidate report shares obtained in an `AggregationJobInitReq` message
-from the leader. It attempts to recover and validate the corresponding input
-shares similar to the leader, and eventually returns a response to the leader
-carrying a VDAF-specific message for each report share.
+Each Helper begins their portion of the aggregate initialization when they
+receive an AggregationJobInitReq message from the Leader. For each ReportShare
+conveyed by this message, the Helper attempts to initiailze VDAF preparation
```
tgeoghegan, line 1236:
```suggestion
conveyed by this message, the Helper attempts to initialize VDAF preparation
```

draft-ietf-ppm-dap.md
```diff
@@ -1421,24 +1430,24 @@ prep_state = VDAF.prep_init(vdaf_verify_key,
 out = VDAF.prep_next(prep_state, None)
 ~~~
 
-`vdaf_verify_key` is the VDAF verification key shared by the aggregators;
+`vdaf_verify_key` is the VDAF verification key shared by the Aggregators;
 `agg_id` is the aggregator ID (`0x00` for the Leader and `0x01` for the helper);
-`agg_param` is the opaque aggregation parameter distributed to the aggregators
-by the collector; `public_share` is the public share generated by the client and
-distributed to each aggregator; and `plaintext_input_share` is the aggregator's
+`agg_param` is the opaque aggregation parameter distributed to the Aaggregators
```
tgeoghegan, line 1435:
```suggestion
`agg_param` is the opaque aggregation parameter distributed to the Aggregators
```

draft-ietf-ppm-dap.md
```diff
@@ -1478,30 +1487,35 @@ processed further and MUST be removed from the candidate set.
...
-{{collect-flow}}. Otherwise, `out` is the pair `(new_state, prep_msg)`, where
-`new_state` is its updated state and `prep_msg` is its next VDAF message (which
-will be `leader_outbound` in the next round of continuation). For the latter
-case, the helper sets `prep_state` to `new_state`.
+where `[leader_prep_share, helper_prep_share]` is a vector of two elements. If
+either of these operations fails, then the leader marks the report as invalid
+with error "prep_share_failed". Otherwise it interprets `out` as follows: If
+this is the last round of the VDAF, then `out` is the Leader's output share, in
+which case it stores the output share for further processing as described in
+{{collect-flow}}. Otherwise, `out` is the pair `(next_prep_state,
```
tgeoghegan, line 1507: I don't like the "interprets `out` as follows" phrasing, and I deliberately changed it the last time I was in this paragraph. The reason is that the text you have suggests that the leader is supposed to check what round it currently is ("[I]f this is the last round of the VDAF") and then interpret `out` accordingly. That's backwards: the way that the Leader knows it's on the last round is if `VDAF.prep_next` returns an output share. It's not up to DAP to decide if we're on the last round of the VDAF or if `out` is an output share. VDAF makes that decision.

draft-ietf-ppm-dap.md
```diff
@@ -1525,39 +1539,40 @@ Otherwise, the Leader should abandon the aggregation job entirely.
...
 #### Helper Continuation {#aggregation-helper-continuation}
 
-If the helper does not recognize the task ID, then it MUST abort with error
+If the Helper does not recognize the task ID, then it MUST abort with error
 `unrecognizedTask`.
 
 Otherwise, the Helper continues with preparation for a report share by combining
-the Leader's input message in `AggregationJobReq` and its current preparation
-state (`prep_state`). This step yields one of three outputs:
+the previous message round's prepeare message (carried by the AggregationJobReq)
```
tgeoghegan, line 1546:
```suggestion
the previous message round's prepare message (carried by the AggregationJobReq)
```

tgeoghegan (changes requested):

draft-ietf-ppm-dap.md
cjpatton: Hmm. can you take another look? I haven't changes this phrasing AFAICT. In particular the following pattern is in both paragraphs:
> If either of these operations fails, then ... . Otherwise, it interprets `out` as follows: If this is the last round of the VDAF, ... . Otherwise, `out` is the pair ...

Alternatively, do you have a suggestion of how you think this should be worded?

tgeoghegan: Sorry, I think I got my wires crossed while reviewing this. You did indeed not change the wording here. I may or may not have a point about the interpretation thing, but we don't have to address that in this change.

tgeoghegan (approved):

PR #430, opened 2023-03-10
Add changelog for DAP-04

tgeoghegan:
We need #429 for this change to make sense.

### DIFF
--- a/draft-ietf-ppm-dap.md
+++ b/draft-ietf-ppm-dap.md
@@ -12,7 +12,7 @@ venue:
   github: "ietf-wg-ppm/draft-ietf-ppm-dap"
   latest: "https://ietf-wg-ppm.github.io/draft-ietf-ppm-dap/draft-ietf-ppm-dap.html"
 
-v: 3
+v: 4
 
 author:
  -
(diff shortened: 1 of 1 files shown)


cjpatton (changes requested):
For better alignment with previous diffs, could you make each sentence begin with a command, sort of like commit messages? E.g., instead of "Resource oriented API"  something like "Change API endpoints to ..."

draft-ietf-ppm-dap.md
```diff
@@ -114,6 +114,25 @@ seen in the clear by any server.
 
 (\*) Indicates a change that breaks wire compatibility with the previous draft.
 
+04:
+
+- Introduce resource oriented HTTP API. (#278, #398, #400) (\*)
```
cjpatton, line 119: I would add a sentence or explaining what "resource oriented" means and the motivation for this change.

cjpatton (approved):

draft-ietf-ppm-dap.md
tgeoghegan: There's ample discussion of that in the referenced issue and PRs. I don't think it's productive to re-hash it here.

PR #393, opened 2022-12-15
Specify one Helper & compute two VDAF rounds per DAP aggregation step.

branlwyd:
This is based on the "ping-pong" communication pattern described in the
VDAF specification, and reduces the number of network round-trips
required to complete an aggregation by about half for all VDAFs, from
`ROUNDS + 1` to `ceil((ROUNDS + 1) / 2)`. In particular, for 1-round
VDAFs like those in the Prio family, this reduces aggregation to a
single network round-trip.
    
Implementing this change requires specifying to exactly two aggregators,
i.e. exactly one Helper. The text of the aggregation section is updated
to explicitly specify one Helper.

PR #447, opened 2023-04-27
Only use helper endpoints

martinthomson:
Aggregator = leader + helpers; don't overuse it.


tgeoghegan:
I believe all this language gets cleaned up in https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/pull/393

cjpatton (approved):
👍

PR #454, opened 2023-04-28
Clarify how resource URIs are constructed

### LINKED ISSUE #441: URI template is confusing
The `{role}/resource_type/{resource-id}` template doesn't seem to hold.

> A resource's path is resolved relative to a server's endpoint to construct a resource URI. Resource paths are specified as templates like:
>
> ```
> {role}/resource_type/{resource-id}
> ```
>
> `{role}` is one of the API endpoints in the task's aggregator_endpoints (see [Section 4.2](https://datatracker.ietf.org/doc/html/draft-ietf-ppm-dap-04#task-configuration)). The remainder of the path is resolved relative to the endpoint.
>
> DAP resource identifiers are opaque byte strings, so any occurrence of `{resource-id}` in a URL template (e.g., `{task-id}` or `{report-id}`) MUST be expanded to the URL-safe, unpadded Base 64 representation [...]

The literal string "resource_type" does not appear in the document.  Instead, we have things like `{aggregator}/hpke_config` (with no apparent `{resource-id}`) and `{leader}/tasks/{task-id}/reports`.

It seems like you might benefit from the base64url note as a general rule, but the idea that you might define a common template form is probably not right.  Maybe you can limit this text to an explanation that resources are identified using templates relative to a root URL that each aggregator defines.

cjpatton:
Closes #441.

### DIFF
--- a/draft-ietf-ppm-dap.md
+++ b/draft-ietf-ppm-dap.md
@@ -547,22 +547,6 @@ DAP has three major interactions which need to be defined:
 Each of these interactions is defined in terms of "resources". In this section
 we define these resources and the messages used to act on them.
 
-A resource's path is resolved relative to a server's endpoint to construct a
-resource URI. Resource paths are specified as templates like:
-
-~~~
-{role}/resource_type/{resource-id}
-~~~
-
-`{role}` is one of the API endpoints in the task's `aggregator_endpoints` (see
-{{task-configuration}}). The remainder of the path is resolved relative to the
-endpoint.
-
-DAP resource identifiers are opaque byte strings, so any occurrence of
-`{resource-id}` in a URL template (e.g., `{task-id}` or `{report-id}`) MUST be
-expanded to the URL-safe, unpadded Base 64 representation of the corresponding
-resource identifier, as specified in {{Sections 5 and 3.2 of !RFC4648}}.
-
 The following are some basic type definitions used in other messages:
 
 ~~~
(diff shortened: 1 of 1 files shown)


cjpatton:
Please have a look @martinthomson!

draft-ietf-ppm-dap.md
```diff
@@ -777,6 +761,29 @@ of the Aggregators is configured with following parameters:
...
+* Variables `{task-id}`, `{aggregation-job-id}`, and `{collection-job-id}` are
+  replaced with the task ID ({{task-configuration}}), aggregation job ID
+  ({{agg-init}}), and collection job ID ({{collect-init}}) respectively. The
+  value MUST be encoded in its URL-safe, unpadded Base 64 representation as
+  specified in {{Sections 5 and 3.2 of !RFC4648}}.
+
+For example, resource URI `{leader}/tasks/{task-id}/reports` might be expanded
+into:
+~~~
+https://example.com/tasks/8BY0RzZMzxvA46_8ymhzycOB9krN-QIGYvg_RsByGec=/reports
```
martinthomson, line 784:
```suggestion
https://example.com/tasks/8BY0RzZMzxvA46_8ymhzycOB9krN-QIGYvg_RsByGec/reports
```

draft-ietf-ppm-dap.md
```diff
@@ -788,11 +795,12 @@ Before the Client can upload its report to the Leader, it must know the HPKE
...
 
-Clients retrieve the HPKE configuration from each aggregator by sending an HTTP
-GET request to `{aggregator}/hpke_config`. Clients MAY specify a query parameter
-`task_id` whose value is the task ID whose HPKE configuration they want. If the
-Aggregator does not recognize the task ID, then it MUST abort with error
-`unrecognizedTask`.
+Clients retrieve the HPKE configuration from the Leader by sending an HTTP GET
+request to `{leader}/hpke_config`. Similarly, Clients retrieve the HPKE
+configuration from the Helper by sending an HTTP GET request to
+`{helper}/hpke_config`. Clients MAY specify a query parameter `task_id` whose
```
martinthomson, line 801: It is a bit odd to have this structure.  I would have said global = `{aggregator}/hpke_config` and task-specific = `{aggregator}/tasks/{task-id}/hpke_config`.  Though I suspect that achieving consistency would be harder for the latter form, if you ever had to worry about that.

martinthomson (approved):
Much improved, thanks.

draft-ietf-ppm-dap.md
cjpatton: Nice idea! I'll create an issue to track.

cjpatton:
Squashed.

PR #458, opened 2023-04-29
Don't RECOMMEND how to generate IDs

### LINKED ISSUE #452: Why should job IDs be pseudorandom?
>   1.  Generate a fresh AggregationJobID.  This ID MUST be unique within
       the context of the corresponding DAP task.  It is RECOMMENDED
       that this be set to a random string output by a cryptographically
       secure pseudorandom number generator.

This seems unhelpful. Either these should be cryptographically random
in which case this should be a MUST, or it doesn't matter, in which case
why is this text here? How am I supposed to know.

cjpatton:
Closes #452.

The current recommendation is to generate it using a "cryptographically secure random number generator", which is usually reserved for purposes that need a uniform random value. All we need is uniqueness; the purpose of this text is merely to suggest a convenient way of meeting the requirement.

To avoid confusion, just drop the recommendation altogether.

### DIFF
--- a/draft-ietf-ppm-dap.md
+++ b/draft-ietf-ppm-dap.md
@@ -727,9 +727,8 @@ the configuration for each task. A task is uniquely identified by its task ID:
 opaque TaskID[32];
 ~~~
 
-A `TaskID` is a globally unique sequence of bytes. It is RECOMMENDED that this
-be set to a random string output by a cryptographically secure pseudorandom
-number generator. Each task has the following parameters associated with it:
+The task ID value MUST be a globally unique sequence of bytes. Each task has
+the following parameters associated with it:
 
 * `aggregator_endpoints`: A list of URLs relative to which each Aggregator's API
   endpoints can be found. Each endpoint's list MUST be in the same order. The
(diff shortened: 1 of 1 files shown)


draft-ietf-ppm-dap.md
```diff
@@ -1249,9 +1248,8 @@ The Leader and Helper initialization behavior is detailed below.
 The Leader begins the aggregate initialization phase with the set of candidate
 report shares as follows:
 
-1. Generate a fresh AggregationJobID. This ID MUST be unique within the context
-   of the corresponding DAP task. It is RECOMMENDED that this be set to a random
-   string output by a cryptographically secure pseudorandom number generator.
+1. Generate a fresh AggregationJobID. The ID value MUST be unique within the
+   context of the corresponding DAP task.
```
tgeoghegan, line 1252:
```suggestion
   scope of the corresponding DAP task.
```

draft-ietf-ppm-dap.md
```diff
@@ -1782,9 +1780,7 @@ First, the Collector chooses a collection job ID:
 opaque CollectionJobID[16];
 ~~~
 
-This ID MUST be unique within the context of the corresponding DAP task. It is
-RECOMMENDED that this be set to a random string output by a cryptographically
-secure pseudorandom number generator.
+This ID value MUST be unique within the context of the corresponding DAP task.
```
tgeoghegan, line 1783:
```suggestion
This ID value MUST be unique within the scope of the corresponding DAP task.
```

tgeoghegan (approved):
"Scope" feels more natural than "context" when discussing uniqueness of a name to me, but feel free to overrule me, especially if others disagree.

cjpatton:
Rebased and merged.

PR #461, opened 2023-05-02
Encapsulate details of VDAF preparation into "prep steps"

### LINKED ISSUE #451: What's the difference between a prep message and a prep share
~~~
        | AggregationJobInitReq                               |
      I | agg_param, helper_report_share, leader_prep_share_1 |
     N1 |---------------------------------------------------->| I
        |                                  AggregationJobResp | N1
        |                   prep_msg_1, [helper_prep_share_2] | P1
        |<----------------------------------------------------| [N2]
~~~

In this diagram the helper sends both `prep_msg_1` and `helper_prep_share_2`, but the latter only in a multiround exchange? Why is this a distinction that needs to be made at the DAP level? It seems much more intuitive to just have the leader and helper each send opaque messages. This is especially confusing because Section 4.2.2.1 does not seem to make use
of the helper's prep message at all, just its prep share.

As a separate note, having the messages numbered so that the helper's prep shares are even but its prep messages are odd is very confusing.

cjpatton:
Closes #451.

The VDAF spec distinguishes between two types of intermediate values computed and exchanged amongst the Aggregators: "prep shares" and "prep messages". This distinction is largely irrelevant in the current aggregation flow (i.e., that of draft04), since the Leader always sends the prep message to the Helper and the Helper always sends its prep share to the Leader.

In the proposed "ping-pong" flow (PR #393), the distinction is important because the Leader and Helper take turns sending these values. As a result, whether the host is processing a prep share or prep message depends on which step of preparation is being computed.

In PR #393 (by @branlwyd), these details are encapsulated by the `PrepareStep` message, which defines a variant for each combination of values encountered during the aggregation flow. In order to simplify the overview, this change hoists this encapsulation up to the overview section, thereby removing the low details that are irrelevant to the overall structure.

While at it, this change defines two procedures, `DAPPrepInit()` and `DAPPrepNext()`, that specify explicitly the prep step processing rules. Once we merge this PR, we will rewrite #393 to use these functions. Something like this:
1. During aggregation initialization, each Aggregator decrypts its input share, performs the early validation steps, then runs `DAPPrepInit()`. Then each Aggregator runs `DAPPrepNext(inbound)` where `inbound == None` in case of the Leader and, in case of the Helper, `inbound` is the the prep step sent by the Leader in the first request.
2. During aggregation continuation, each Aggregator calls `DAPPrepNext(inbound)`, where `inbound` is the prep step sent from the peer.


draft-ietf-ppm-dap.md
```diff
@@ -1096,84 +1096,169 @@ refinement process. We instead think of these as properties of the output
...
+      opaque prep_msg<0..2^32-1>;
+      opaque prep_share<0..2^32-1>;
+    case finish:
+      opaque prep_msg<0..2^32-1>;
+    case finished:
+      Empty;
+    case reject:
+      ReportShareError report_share_error;
+  };
+} PrepareStep;
```
cjpatton, line 1224: @branlwyd: Please weigh in on weather there is a clean way to integrate this structure into #393. I think we definitely not avoiding repeating report IDs, either to avoid sending extra bits or having an extra validation step (pick your poison!).

draft-ietf-ppm-dap.md
```diff
@@ -1186,27 +1271,157 @@ the last round is reached and each Aggregator computes an output share. All
...
+  |                |                    |
+  |                |                    v
+  +--> REJECTED <--+     FINISHED(out_share)
+~~~
+{: #vdaf-prep-state-machine title="State machine for VDAF preparation."}
+
+State transitions are made when the state is acted upon by the host's local
+inputs and/or a prep step sent by the peer. The initial state is `START`.
+
+~~~ transition
```
cjpatton, line 1287: @branlwyd: Please double check that both of these transition rules align with your understanding.

draft-ietf-ppm-dap.md
```diff
@@ -1186,27 +1271,157 @@ the last round is reached and each Aggregator computes an output share. All
...
+    else:
+      return PrepareStep.continue(prep_msg, host_prep_share)
+  except:
+    state REJECT
+    return None
+~~~
+
+State `FINISHED` represents the state in which an output share (denoted
+`out_share`) was successfully recovered by the host.
+
```
cjpatton, line 1385: Reviewer note: I'm not sure if we want to keep the figure below. It was helpful for working out the state machine details, but it has a lot of redundant information (cf. {{agg-flow}}).

draft-ietf-ppm-dap.md
```diff
@@ -1186,27 +1271,157 @@ the last round is reached and each Aggregator computes an output share. All
...
+                 host_input_share):
+  try:
+    host_prep_state = Vdaf.prep_init(
+      vdaf_verify_key,
+      0 if is_leader else 1,
+      agg_param,
+      report_id,
+      public_share,
+      host_input_share,
+    )
```
branlwyd, line 1313: This method would then need to call `prep_next(host_prep_state, None)` as is currently done during initialization: https://www.ietf.org/archive/id/draft-ietf-ppm-dap-04.html#name-input-share-preparation

branlwyd: I think the intent might be to move this `prep_next` call to https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/pull/461/files#diff-8dbbcb0d121b441734a70cc867271480731a494b50eefdd758c249fdfb5ab84eR1347-R1350 -- if so, this works out for Helper initialization, but would leave the Leader needing to make an additional `prep_next` call after `VdafPrepInit` but before sending its aggregation-initialization message to the Helper.

That's not ideal since users of the rest of the document now need to care about both the "wrapper" interface and the normal VDAF interface; that is, the encapsulation of the VDAF interface is leaking.

Unfortunately, if we do specify `VdafPrepInit` as including the initial `prep_next` call, this leaves the Helper initialization as requiring a `VdafPrepInit` followed by a function similar to `VdafPrepNext` except that the initial `prep_next` call is omitted. (that is, it would call `prep_shares_to_prep` followed by `prep_next`) I think having to specify another arbitrary wrapping function would probably tip this to being a complexity loss -- maybe there's some clever way to cleanly include this case into `VdafPrepNext`?

draft-ietf-ppm-dap.md
```diff
@@ -1186,27 +1271,157 @@ the last round is reached and each Aggregator computes an output share. All
...
+
+def VdafPrepNext(is_leader, agg_param, inbound):
+  if type(inbound) == PrepareStep.initialize:
+    prep_msg, peer_prep_share = None, inbound.prep_share
+  elif type(inbound) = PrepareStep.continue:
+    prep_msg, peer_prep_share = inbound.prep_msg, inbound.prep_share
+  elif type(inbound) = PrepareStep.finish:
+    prep_msg, peer_prep_share = inbound.prep_msg, None
+  else:
+    prep_msg, peer_prep_share = None, None
```
branlwyd, line 1344: Should this be an error? I'm not sure we want to pass a `finished` or `reject` PrepareStep to this method.

draft-ietf-ppm-dap.md
```diff
@@ -1186,27 +1271,157 @@ the last round is reached and each Aggregator computes an output share. All
...
+      if round == Vdaf.ROUNDS:
+        state FINISHED(out)
+        return PrepareStep.finish(prep_msg)
+      (host_prep_state, host_prep_share) = out
+
+    state CONTINUED(round, host_prep_state)
+    if is_leader and round == 1:
+      return PrepareStep.initialize(host_prep_share)
+    else:
+      return PrepareStep.continue(prep_msg, host_prep_share)
```
branlwyd, line 1377: As noted in https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/issues/451#issuecomment-1531741943, writing things this way effectively requires the Leader to store both the `prep_msg` and the `prep_share` to durable storage, barring implementors effectively splitting this function up to run part of it in one DAP round and part of it in another (which IMO may be a stretch to expect of implementors). #393 is written in a way that stores only `prep_msg`.

Maybe we can include an implementation note that the final `prep_next` call can actually be made after storing `prep_msg` to storage, for implementations that "checkpoint" between rounds?

draft-ietf-ppm-dap.md
branlwyd: I don't object to keeping this information around, but I think this ladder diagram should be merged with the one above (https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/pull/461/files#diff-8dbbcb0d121b441734a70cc867271480731a494b50eefdd758c249fdfb5ab84eR1107-R1138).

branlwyd (commented):
I left a few comments. One meta-comment is that, if the original problem was confusion between prep messages & prep shares, we haven't actually resolved the issue -- DAP implementors still have to care about the distinction between the two opaque message types (and prep state), place the correct opaque messages in the correct fields of DAP messages, etc. Truly fixing DAP to just work on opaque VDAF messages would take a change to VDAF to e.g. introduce a 2-party interface & special messages for that interface, which is an undesirable level of complexity to add to VDAF.

If we can come up with a nice encapsulation that works as a simplification, that is separately valuable (but see my comments -- I think the encapsulation currently has a flaw, and I don't see a way to fix it without either making the encapsulation leaky or much more complicated).

draft-ietf-ppm-dap.md
branlwyd: Thinking about this a little more, another approach would to update `VdafPrepInit` to compute `prep_next` conditionally, based on whether it is called by the Leader or the Helper. This function would return different kinds of results based on the role of the caller, but this may still be a relatively-clean way to fix up this issue.

draft-ietf-ppm-dap.md
cjpatton: Removed.

draft-ietf-ppm-dap.md
cjpatton: The value would be the ability to describe the state machine without having to loop in the decryption and early validation steps.

draft-ietf-ppm-dap.md
cjpatton: The intent is that the Leader can call `VdafPrepInit()`, then immediately call `VdafPrepNext(inbound=None)` to produce the first prep step. The Helper would call `VdafPrepInit()`, then immediately call `VdafPrepNext(inbound=peer_prep_step)`. Then in subsequent rounds, each just calls `VdafPrepNext()`.

So aggregation initialization becomes:
1. Decrypt the report share
1. Do early validation (report replay, etc.)
1. Call `VdafPrepInit()`
2. Call `VdafPrepNext(inbound=None)` if Leader else `VdafPrepNext(inbound=peer_prep_step)`.
3. Send prep step produced from step (2.).

Aggregation continuation becomes:
1. Call `VdafPrepNext(inbound=peer_prep_step)`.
2. Send the prep step produced.

draft-ietf-ppm-dap.md
cjpatton: The first time the Leader calls `VdafPrepNext()` it is immediately after calling `VdafPrepInit()`. It has not received a prep_msg or prep share from its peer, so there's nothing to do. The outbound message on this call is the leader's first prep share.

draft-ietf-ppm-dap.md
cjpatton: > As noted in [#451 (comment)](https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/issues/451#issuecomment-1531741943), writing things this way effectively requires the Leader to store both the `prep_msg` and the `prep_share` to durable storage, barring implementors effectively splitting this function up to run part of it in one DAP round and part of it in another (which IMO may be a stretch to expect of implementors). #393 is written in a way that stores only `prep_msg`.

I don't quite understand what you mean by durable storage. The intent of these methods is to describe the semantics of 393, i.e., I don't expect the protocol flow to be different.

draft-ietf-ppm-dap.md
branlwyd: Based on in-person discussion, this decomposition works:
* Leader initialization is DapPrepInit -> DapPrepNext -> send/recv -> DapPrepNext.
* Helper initialization is DapPrepInit -> DapPrepNext.
* Leader continuation is send/recv -> DapPrepNext.
* Helper continuation is DapPrepNext.

draft-ietf-ppm-dap.md
branlwyd: I think what is returned by `DapPrepNext` is effectively what a leader stores for retries -- `DapPrepNext` is the last step for leader initialization & continuation, and the result of that call is what is sent to the helper at the beginning of the next round of continuation.

But an implementation could effectively "pause" computation of `DapPrepNext` before the final `VDAF.prep_next` call, and only need to keep around the `prep_msg`, recomputing the `host_prep_share` if/when a retry occurs. This would help implementations that checkpoint each round to durable storage to e.g. survive unexpected process failures.

This is an implementation choice, so I don't think we need to make a change, but an implementation note would be helpful as I think otherwise it would be challenging for implementors to realize this optimization is possible.

cjpatton: (As discussed offline) The plan here will be to first try to make this point clear in the normative protocol text. Let's keep this section focused on the state machine.

cjpatton:
Rebased and squashed.

branlwyd (approved):
I traced through a 1- and 2-round VDAF and I think this works, both in terms of VDAF-message dataflow and round/state tracking.

draft-ietf-ppm-dap.md
```diff
@@ -1099,84 +1099,170 @@ refinement process. We instead think of these as properties of the output
...
+
+~~~ ladder
+  report, agg_param
+   |
+   v
++--------+                                         +--------+
+| Leader |                                         | Helper |
++--------+                                         +--------+
+   | AggregationJobInitReq                               |
+   | agg_param, report_share, prep_step(initialize)      |
```
ekr, line 1118: Is this saying:
"Send an AggregationJobInitReq consisting of[agg_param, report_share, and the output of prep_step()]"

If so, maybe
```suggestion
   | AggregationJobInitReq:                           |
   |    agg_param, report_share, prep_step(initialize)      |
```
e

draft-ietf-ppm-dap.md
```diff
@@ -1099,84 +1099,170 @@ refinement process. We instead think of these as properties of the output
...
+   |                                                     |
+   v                                                     v
+  leader_out_share                         helper_out_share
+~~~
+{: #agg-flow title="Overview of the DAP aggregation sub-protocol."}
+
+Each of these "preparation steps", or "prep steps" for short, has an associated
+type: one of "initialize", "continue", "finish", or "finished". The type of
+each step, and how many steps there are, depends on the VDAF. The message
+structure and processing rules are specified in {{preparation-messages}}.
```
ekr, line 1147: I don't understand what this is trying to say. These are arguments to some `prep_step()` function? How does the caller know what argument to provide? Why can't the encapsulated VDAF just do the right thing?

I'm deliberately not referring to the VDAF spec here, because the idea here is that you should be able to read this bit and not know that.

draft-ietf-ppm-dap.md
```diff
@@ -1099,84 +1099,170 @@ refinement process. We instead think of these as properties of the output
...
+      opaque prep_msg<0..2^32-1>;
+      opaque prep_share<0..2^32-1>;
+    case finish:
+      opaque prep_msg<0..2^32-1>;
+    case finished:
+      Empty;
+    case reject:
+      ReportShareError report_share_error;
+  };
+} PrepareStep;
```
ekr, line 1227: I still don't understand why this information has to be surfaced at this level.

Given that this message is self-contained, why can't this message be what's passed to the VDAF and let it figure out what to do. Please don't tell me that that's how the VDAF API works, because I am suggesting a different API.

branlwyd: My understanding for the reason for this complexity is:
* DAP is moving towards specifying to exactly two aggregators (this work is part of that effort).
* VDAF is currently specified to an arbitrary number of aggregators.

... so somewhere we need to specify how to adapt from the arbitrary-aggregator VDAF interface to the 2-aggregator DAP protocol, in a way that unlocks the performance benefit allowed by 2-aggregator DAP. This PR contains that complexity, and attempts to encapsulate it into the DAPPrepInit/DAPPrepNext functions (replacing the similar VDAF-interface VDAF.prep_init & VDAF.prep_next functions).

If we want to avoid this complexity entirely in DAP, the most likely change would be to the VDAF specification, either to specify VDAF to 2 aggregators or to specify a second VDAF interface which is equivalent to the 2-aggregator DAPPrepInit/DAPPrepNext interface described in this PR. I'm less confident to speak to changes to VDAF, but VDAF editors have expressed a desire to maintain that VDAF can work with an arbitrary number of aggregators. Perhaps we should consider specifying the functionality of this PR in VDAF? (I do not hold a strong opinion about where the better place for this complexity is, I'm just highlighting that that seems to be the tradeoff.)

cjpatton:
Rebased!

draft-ietf-ppm-dap.md
cjpatton: That's right. Done.

draft-ietf-ppm-dap.md
cjpatton: You call `DAPPrepInit()` and `DAPPrepNext()` to do the right thing, which are defined in this document. These are actually wrapper functions around the VDAF API (i.e., `VDAF.prep_init()`, `VDAF.prep_next()`, and `VDAF.prep_shares_to_prep()`) that are specific to the 2-Aggregator, Leader-Helper structure of DAP. (VDAFs still support >2 Aggregators.)

I think we have to specify these wrapper methods somewhere. I'd be fine moving them to VDAF if you think that would be better.

draft-ietf-ppm-dap.md
cjpatton: It's easy enough to move the wrapper code to VDAF, I'll create a PR there today.

cjpatton:
I have moved the wrapper code to a PR against VDAF: https://github.com/cfrg/draft-irtf-cfrg-vdaf/pull/240

draft-ietf-ppm-dap.md
cjpatton: https://github.com/cfrg/draft-irtf-cfrg-vdaf/pull/240

branlwyd (approved):

cjpatton (commented):
https://github.com/cfrg/draft-irtf-cfrg-vdaf/pull/240 looks to be close to ready. I'd like to propose that we merge this PR as-is then start reworking #393 to use the new API there. (We'll need to update the references to VDAF once we cut a new version, but I think we can wait to do that until we finish #393.) What do you think @ekr, @branlwyd?

branlwyd:
> [cfrg/draft-irtf-cfrg-vdaf#240](https://github.com/cfrg/draft-irtf-cfrg-vdaf/pull/240) looks to be close to ready. I'd like to propose that we merge this PR as-is then start reworking #393 to use the new API there. (We'll need to update the references to VDAF once we cut a new version, but I think we can wait to do that until we finish #393.) What do you think @ekr, @branlwyd?

Merging this PR sounds good to me.

cjpatton:
I'm going to merge this today.

PR #464, opened 2023-05-15
Use `invalidMessage` instead of `unrecognized`

### LINKED ISSUE #453: Why do duplicate report ids have error unrecognizedMessage

tgeoghegan:
The error type `unrecognizedMessage` and the `ReportShareError` variant `unrecognized_message` are changed to `invalidMessage` and `invalid_message`, respectively, as this better captures the range of conditions in which this error is used: some messages are recognizable in that a protocol participant can parse them, but invalid in that the values they contain are illegal.

Resolves #453


divergentdave (approved):

tgeoghegan:
This change has been up here for a week, the discussion about it was going on for two weeks prior to that in #453 and it has approval from two DAP editors, and nobody has objected, so I'm going to merge.

PR #465, opened 2023-05-22
Remove queryMismatch, in favor of invalidMessage.

branlwyd:
The aggregator behavior on receiving this error is the same either way, and there are many other message-based errors that do not get their own code. This change effectively leaves it up to implementations to decide the level of detail they include in error reports.


tgeoghegan (approved):
I support this change, and agree that it'll be even better as "invalidMessage". I've just merged #464, so could you please rebase?

branlwyd:
> I support this change, and agree that it'll be even better as "invalidMessage". I've just merged #464, so could you please rebase?

Rebased.

cjpatton (approved):

divergentdave (approved):

PR #468, opened 2023-06-13
Idempotent aggregate shares endpoint

### LINKED ISSUE #226: Aggregation idempotency
From @tgeoghegan:

Suppose we have a task with max_batch_lifetime = 1. Consider:

1. Leader makes an AggregateShareReq for some batch interval.
2. Helper services the request, and marks that batch interval as having been collected once.
3. Helper attempts to transmit an AggregateShareResp to leader, but the message is truncated.

Now, if the leader retries its AggregateShareReq, the helper will refuse the request because the batch interval's lifetime has been consumed. So, to allow the leader to retry this request, the helper has to be willing to resend previously computed aggregate shares.

Issue from #223.

tgeoghegan:
Require that the helper's `/tasks/{task-id}/aggregate_shares` endpoint be idempotent.

Resolves #226

### DIFF
--- a/draft-ietf-ppm-dap.md
+++ b/draft-ietf-ppm-dap.md
@@ -2039,6 +2039,11 @@ struct {
 computed above and `encrypted_aggregate_share.ciphertext` is the ciphertext
 `encrypted_agg_share` computed above.
 
+The helper's handling of this request MUST be idempotent. That is, if multiple
+identical, valid `AggregateShareReq`s are received, they should all yield the
+same response while only consuming one unit of the task's
+`max_batch_query_count` (see {{batch-validation}}).
+
 After receiving the Helper's response, the Leader uses the HpkeCiphertext to
 finalize a collection job (see {{collect-finalization}}).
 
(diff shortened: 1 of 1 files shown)


draft-ietf-ppm-dap.md
```diff
@@ -1947,6 +1947,11 @@ struct {
 computed above and `encrypted_aggregate_share.ciphertext` is the ciphertext
 `encrypted_agg_share` computed above.
 
+The helper's handling of this request MUST be idempotent. That is, if multiple
```
tgeoghegan, line 1950: I think a MUST (rather than a SHOULD) is appropriate here. If this is optional, then there's no way for the leader to recover from losing the first response to `POST /tasks/{task-id}/aggregate_shares`, because the helper can refuse to service subsequent aggregate share requests due to exhaustion of the max batch query count.

branlwyd: I think this is roughly equivalent to saying that `max_batch_query_count` consumption is based on _distinct_ `AggregateShareReq`s, which is what we want, so 👍🏻.

branlwyd (approved):

divergentdave (approved):

draft-ietf-ppm-dap.md
```diff
@@ -2039,6 +2039,11 @@ struct {
...
```
cjpatton, line 2042:
```suggestion
The Helper's handling of this request MUST be idempotent. That is, if multiple
```

cjpatton (approved):

PR #469, opened 2023-06-13
Be explicit about task param immutability

### LINKED ISSUE #237: Decide which Task parameters might be mutable.
Tasks are defined with a number of parameters; eventually, it might be helpful to determine which might be mutated (and any restrictions on how they might be mutated).

One initial thought is that any of the parameters that end up distributed to clients are likely effectively immutable, if only due to the challenge of updating all of the clients.

[This fell out of a discussion on https://github.com/divviup/janus/pull/142.]

tgeoghegan:
I've satisfied myself by reading the document that nothing implies that task params can be mutated or rotated, so all we need is a paragraph explicitly stating this.

Resolves #237

### DIFF
--- a/draft-ietf-ppm-dap.md
+++ b/draft-ietf-ppm-dap.md
@@ -747,6 +747,10 @@ of the Aggregators is configured with following parameters:
 Finally, the Collector is configured with the HPKE secret key corresponding to
 `collector_hpke_config`.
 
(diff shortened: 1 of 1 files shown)


branlwyd (approved):

divergentdave (approved):

cjpatton (approved):
Good idea.

PR #471, opened 2023-06-13
Explicitly forbid mutating agg jobs, collections

### LINKED ISSUE #406: What error should servers provide if someone tries to mutate an immutable resource?
There are a few places in DAP where a client (in the HTTP sense, not the DAP sense) creates an immutable resource using a PUT request involving a unique identifier:

- DAP clients create reports with a `report_id` in the body
- DAP leaders create aggregation jobs with an `aggregation-job-id` in the path
- DAP collectors create collection jobs with a `collection-job-id` in the path

These objects cannot be further mutated using a PUT request (e.g., you can't change the aggregation parameter on a collection job after creating it). The question is: what should the error look like?

tgeoghegan:
Resolves #406

### DIFF
--- a/draft-ietf-ppm-dap.md
+++ b/draft-ietf-ppm-dap.md
@@ -1461,6 +1461,16 @@ The Helper responds to the Leader with HTTP status code 201 Created and a body
 consisting of the `AggregationJobResp`, with media type
 "application/dap-aggregation-job-resp".
 
+Changing an aggregation job's parameters is illegal, so further requests to
+`PUT /tasks/{tasks}/aggregation_jobs/{aggregation-job-id}` for the same
+`aggregation-job-id` but with a different `AggregationJobInitReq` in the body
+MUST fail with an HTTP client error status code.
+
+Additionally, it is not possible to rewind or reset the state of an aggregation
+job. Once an aggregation job has been continued at least once (see
+{{agg-continue-flow}}), further requests to initialize that aggregation job MUST
+fail with an HTTP client error status code.
+
 Finally, if `state == Continued(prep_state)`, then the Helper stores `state` to
 prepare for the next continuation step ({{aggregation-helper-continuation}}).
 Otherwise, if `state == Finished(out_share)`, then the Helper stores `out_share`
@@ -1870,6 +1880,11 @@ requirements of the batch parameters using the procedure in
 The Leader then begins working with the Helper to aggregate the reports
 satisfying the query (or continues this process, depending on the VDAF) as
 described in {{aggregate-flow}}.
+
+Changing a collection job's parameters is illegal, so further requests to
+`PUT /tasks/{tasks}/collection_jobs/{collection-job-id}` for the same
+`collection-job-id` but with a different `CollectionReq` in the body MUST fail
+with an HTTP client error status code.
 
 After receiving the response to its `CollectionReq`, the Collector makes an HTTP
 `POST` request to the collection job URI to check on the status of the collect


draft-ietf-ppm-dap.md
```diff
@@ -1355,6 +1355,12 @@ On success, the Helper responds to the Leader with HTTP status code 201 Created
 and a body consisting of the `AggregationJobResp`, with media type
 "application/dap-aggregation-job-resp".
 
+Aggregation job initialization MUST be idempotent. That is, further requests to
```
branlwyd, line 1358: A few things to consider including:
* Mention that "further requests with a different body" would be grouped together by aggregation job ID (I suppose).
* We don't want to allow re-initialization -- so if a continuation step has occurred, we should no longer allow initialization.

(& identical suggestions for the collection job case, with aggregation job ID -> collection job ID)

branlwyd (approved):

draft-ietf-ppm-dap.md
tgeoghegan: >Mention that "further requests with a different body" would be grouped together by aggregation job ID (I suppose).

I think I know what you mean. Is this paragraph more clear?

```
Aggregation job initialization MUST be idempotent. That is, further requests to
initialize some aggregation job whose body is identical to the first request
MUST succeed. However, changing an aggregation job's parameters is illegal, so
further requests to initialize an existing aggregation job with different parameters
MUST fail with an HTTP client error status code.
```

>We don't want to allow re-initialization -- so if a continuation step has occurred, we should no longer allow initialization.

So we're talking about a sequence of requests like this:

1. `PUT /tasks/{task-id}/aggregation_jobs/{aggregation-job-id}`
2. `POST /tasks/{task-id}/aggregation_jobs/{aggregation-job-id}`
3. `PUT /tasks/{task-id}/aggregation_jobs/{aggregation-job-id}`

When you say we should no longer allow initialization, do you mean that the request at (3) should fail with something like HTTP 400 Bad Request, or just that it shouldn't rewind or reset the state of the aggregation job? I think that the request at (3) should still get a 201 Created response, because in many cases, the leader can then follow up with a `POST` to continue the aggregation job and gracefully resume processing the job.

branlwyd: > Mention that "further requests with a different body" would be grouped together by aggregation job ID (I suppose).

I'm suggesting that the text "further requests to initialize an existing aggregation job with different parameters" is ambiguous, because it does not define how to determine if a new request matches to an existing aggregation job. I think the right way to determine this is by aggregation job ID -- if the request includes an aggregation job ID matching to an existing aggregation job, it is a repeated request (possibly with different parameters); if the request includes a unique aggregation job ID, it is a request for a new aggregation job. My suggestion is that we state that explicitly.


> When you say we should no longer allow initialization, do you mean that the request at (3) should fail with something like HTTP 400 Bad Request, or just that it shouldn't rewind or reset the state of the aggregation job?

Yes -- IMO request (3) should fail [and it shouldn't rewind/reset the state of the aggregation job]. I would suggest we don't allow re-initialization of existing aggregation jobs because we don't need to: if a further "continuation" request has been made, the Leader has definitely received the result of the prior "initialization" step. And if we allow aggregation jobs to be re-initialized, we will have to deal with "weird" issues like "what if the Leader drives an agg job all the way to completion, then re-initializes it, then drives it to completion once again -- should the output shares recovered by the second completion contribute to the eventual aggregate 'again'?" (probably not, but we'd need text to that effect, and implementations would have to add some complexity to deal with such issues)

Or, framing things in terms of idempotency, request (2) changes the state of the aggregation job, so afterwards re-sending request (1) has no expectation of idempotency, as it is a request to rewind the aggregation job -- we would at that point expect request (2) to be idempotent rather than request (1).

tgeoghegan: > I'm suggesting that the text "further requests to initialize an existing aggregation job with different parameters" is ambiguous, because it does not define how to determine if a new request matches to an existing aggregation job.

OK, I'll rewrite to be more explicit.

> Yes -- IMO request (3) should fail [and it shouldn't rewind/reset the state of the aggregation job].

I'm coming around to your point of view: reviewing the changes that will land in #393, it appears we don't allow for the `AggregationJobResp` a leader gets in response to initialization to ever contain `PrepareResp`s in the finished state, so as it stands, we can't allow `PUT /tasks/{task-id}/aggregation_jobs/{aggregation-job-id}` to succeed regardless of the aggregation job's state. I was thinking that maybe an implementation could have multiple aggregation job workers racing to initialize and continue agg jobs, but the implementation would have to have a pretty serious state synchronization bug to have one worker step the job past round 1 and have the other still think the job needs to be initialized, so I don't think it's worth having the specification bend over backwards to accommodate this.

In short: requests to initialize an aggregation job that has been continued at least once are invalid.

draft-ietf-ppm-dap.md
```diff
@@ -1461,6 +1461,19 @@ The Helper responds to the Leader with HTTP status code 201 Created and a body
 consisting of the `AggregationJobResp`, with media type
 "application/dap-aggregation-job-resp".
 
+Aggregation job initialization MUST be idempotent. That is, multiple requests to
+`PUT /tasks/{tasks}/aggregation_jobs/{aggregation-job-id}` for the same
+`aggregation-job-id` and the same request body MUST succeed. However, changing
+an aggregation job's parameters is illegal, so further requests to
+`PUT /tasks/{tasks}/aggregation_jobs/{aggregation-job-id}` for the same
+`aggregation-job-id` but with a different request body MUST fail with an HTTP
+client error status code.
```
cjpatton, line 1470: This seems to imply that if the Helper gets an agg request and marks the reports as processed, then in the second agg request it has to "forget" this previous state change and allow the reports to be processed. Is this intended?

Today, Daphne won't even let you retry the second request:
```rust
if !self
    .put_helper_state_if_not_exists(task_id, &agg_job_id, &state)
    .await?
{
    // TODO spec: Consider an explicit abort for this case.
    return Err(DapAbort::BadRequest(
        "unexpected message for aggregation job (already exists)".into(),
    ));
}
```

Another question: what would be the recommended mechanism for implementing this MUST? For multi-round VDAFs it would make sense to at least require the Helper to wait as long as possible before recording the reports that were processed (i.e., until the last request). But what if the aggregation job just takes one request/response cycle, e.g., if the VDAF is 1-round? 

Finally, how do we check that the message hasn't changed? We previously discussed keeping track of the hash of the request body. If we're going to require this then I think we need to spell it out.

draft-ietf-ppm-dap.md
```diff
@@ -1871,6 +1884,14 @@ The Leader then begins working with the Helper to aggregate the reports
 satisfying the query (or continues this process, depending on the VDAF) as
 described in {{aggregate-flow}}.
 
+Collection job creation MUST be idempotent. That is, multiple requests to
```
wangshan, line 1887: does it matter if the task has expired and leader deleted it's collection_jobs?

draft-ietf-ppm-dap.md
```diff
@@ -1871,6 +1884,14 @@ The Leader then begins working with the Helper to aggregate the reports
...
+`PUT /tasks/{tasks}/collection_jobs/{collection-job-id}` for the same
+`collection-job-id` and the same request body MUST succeed. However, changing
+a collection job's parameters is illegal, so further requests to
```
wangshan, line 1890: Does this conflict with max_batch_query_count? if a collection-job-id has been queried more than max_batch_query_count, the Leader is suppose to fail the CollectionReq with "batchQueriedTooManyTimes".

tgeoghegan:
After thinking about this further, and having re-read what RFC 9110 already says about idempotence, safety and the different HTTP verbs, I think this change isn't necessary. We already explicitly reference RFC 9110 and it's not DAP's job to re-iterate HTTP concepts.

draft-ietf-ppm-dap.md
```diff
@@ -1461,6 +1461,16 @@ The Helper responds to the Leader with HTTP status code 201 Created and a body
 consisting of the `AggregationJobResp`, with media type
 "application/dap-aggregation-job-resp".
 
+Changing an aggregation job's parameters is illegal, so further requests to
+`PUT /tasks/{tasks}/aggregation_jobs/{aggregation-job-id}` for the same
+`aggregation-job-id` but with a different `AggregationJobInitReq` in the body
+MUST fail with an HTTP client error status code.
+
+Additionally, it is not possible to rewind or reset the state of an aggregation
```
cjpatton, line 1469:
```suggestion
Additionally, it is illegal to rewind or reset the state of an aggregation
```

tgeoghegan:
I've refocused the change to be about the DAP level semantics and to avoid restating HTTP semantics, which is confusing to readers. RFC 9110 among others does a better job of explaining HTTP and its methods than DAP can.

draft-ietf-ppm-dap.md
```diff
@@ -1461,6 +1461,16 @@ The Helper responds to the Leader with HTTP status code 201 Created and a body
 consisting of the `AggregationJobResp`, with media type
 "application/dap-aggregation-job-resp".
 
+Changing an aggregation job's parameters is illegal, so further requests to
+`PUT /tasks/{tasks}/aggregation_jobs/{aggregation-job-id}` for the same
+`aggregation-job-id` but with a different `AggregationJobInitReq` in the body
+MUST fail with an HTTP client error status code.
```
cjpatton, line 1467: What requirement are you trying to meet here? To me the subtext reads like "we should allow retries as long as retrying doesn't change the parameters". If so, how about making it more explicit?

Something like this:

> Changing an aggregation job's parameters is illegal. If another request to `PUT /tasks/{tasks}/aggregation_jobs/{aggregation-job-id}` for the same `aggregation-job-id` is made, the Helper MAY ignore the request if the message payload is the same as the previous request. Otherwise, if the message payload has changed or the Helper did not check, the Helper MUST abort the request.

cjpatton, line 1467: > ... MUST fail with an HTTP client error status code.

I think we should be more prescriptive than this. It seems like a situation where the Leader would want to take a different action based on the feedback from the Helper.

draft-ietf-ppm-dap.md
```diff
@@ -1461,6 +1461,16 @@ The Helper responds to the Leader with HTTP status code 201 Created and a body
...
+
+Additionally, it is not possible to rewind or reset the state of an aggregation
+job. Once an aggregation job has been continued at least once (see
+{{agg-continue-flow}}), further requests to initialize that aggregation job MUST
+fail with an HTTP client error status code.
```
cjpatton, line 1472: Can you check to make sure this doesn't contradict {{aggregation-step-skew-recovery}}?

draft-ietf-ppm-dap.md
tgeoghegan: The requirement is that once an aggregation job has been initialized, you can't change its parameters, i.e. its aggregation parameter or the partial batch selector. We're not discussing retries anymore.

tgeoghegan: I don't agree. What this paragraph says is that changing an aggregation job's parameters is illegal, undefined, outside of what this protocol discusses. So it does not make sense to design in-band mechanisms to recover from this scenario. If this ever happens, then one of the protocol participants has a bug and the way to recover is to make software changes.

draft-ietf-ppm-dap.md
tgeoghegan: It does not. That section concerns itself with the rounds of the continuation phase of the aggregation sub-protocol. This paragraph is saying you can't go back from the continuation phase to the initialization phase. See [discussion with @branlwyd](https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/pull/471#discussion_r1229862877) for context.

draft-ietf-ppm-dap.md
cjpatton: Ah Ok, then it that case, I wouldn't mention the payload explicitly: Just abort if the PUT is repeated for a particular aggregation job.

tgeoghegan: That doesn't work, because it would make it impossible to retry an aggregation job initialization request, which would make DAP basically impossible to operate.

branlwyd (approved):

draft-ietf-ppm-dap.md
cjpatton: Bump

draft-ietf-ppm-dap.md
cjpatton: Bump

tgeoghegan: This does not conflict with max batch query count. Each of the queries against a given batch is a distinct collection with a distinct collection ID.

draft-ietf-ppm-dap.md
cjpatton: We discussed this offline. The intention is to require idempotence here without spelling out how it's supposed to be achieved. There are a few options (quoting @tgeoghegan):
> 1. You could store the entire aggregation job init req that led to the creation of the aggregation job, and then compare the next request.
> 2. You could store a hash of the aggregation job init req and check it against hashes of next request.
> 3. You could reevaluate the aggregation job init req and check if its results match what you have stored from the last time you initialized the request.

I'm a little uncomfortable about leaving this up to implementations, especially since (3.) is a bit fiddly, but it's probably less fiddly to try to spell it out.

draft-ietf-ppm-dap.md
tgeoghegan: I don't understand how task expiration complicates this. We already have this text:

> Upon receipt of a `CollectionReq`, the Leader begins by checking that it
> recognizes the task ID in the request path. If not, it MUST abort with error
> `unrecognizedTask`.

So if task expires, then the leader should refuse to allow new collection jobs to be created in it, regardless of whether the request has been repeated.

cjpatton (approved):

divergentdave (approved):

PR #474, opened 2023-06-15
Use `Empty` for empty variants

cjpatton:
Define `struct {} Empty` and use it wherever we have an enum variant with no bytes following it.

This was removed in c5868ea9854cef7379cef6712aa2d702888f01c9. Our thinking was that unlisted variants implicitly have no bytes following them. This seems like a logical interpretation based on the following example (from RFC8446, Appendix B.3):

```
      struct {
          HandshakeType msg_type;    /* handshake type */
          uint24 length;             /* bytes in message */
          select (Handshake.msg_type) {
              case client_hello:          ClientHello;
              case server_hello:          ServerHello;
              case end_of_early_data:     EndOfEarlyData;
              case encrypted_extensions:  EncryptedExtensions;
              case certificate_request:   CertificateRequest;
              case certificate:           Certificate;
              case certificate_verify:    CertificateVerify;
              case finished:              Finished;
              case new_session_ticket:    NewSessionTicket;
              case key_update:            KeyUpdate;
          };
      } Handshake;
```

`HandshakeType` defines variants that are not listed here; presumably it is because there are no more bytes to read following the `select` statement in these variants.

However, as explained at the top of Appendix B, those variants suffixed by `RESERVED` are defined for backwards compatibiliby only and are not meant to be sent on the wire in TLS 1.3. Thus if the peer sends one of these variants, the host is supposed to treat it as a fatal error.

Similarly for the `message_hash` variant: This is used to construct an input to a hash function, but is never used in a message written to the wire.

### DIFF
--- a/draft-ietf-ppm-dap.md
+++ b/draft-ietf-ppm-dap.md
@@ -578,6 +578,9 @@ struct {
   opaque enc<1..2^16-1>;     /* encapsulated HPKE key */
   opaque payload<1..2^32-1>; /* ciphertext */
 } HpkeCiphertext;
+
+/* Represent a zero-length byte string. */
+struct {} Empty;
 ~~~
 
 DAP uses the 16-byte `ReportID` as the nonce parameter for the VDAF
@@ -621,6 +624,7 @@ struct {
   FixedSizeQueryType query_type;
   select (query_type) {
     by_batch_id: BatchID batch_id;
+    current_batch: Empty;
   }
 } FixedSizeQuery;
 
@@ -1247,6 +1251,7 @@ follows:
 struct {
   QueryType query_type;
   select (PartialBatchSelector.query_type) {
+    case time_interval: Empty;
     case fixed_size: BatchID batch_id;
   };
 } PartialBatchSelector;
@@ -1376,6 +1381,7 @@ struct {
   PrepareRespState prepare_resp_state;
   select (PrepareResp.prepare_resp_state) {
     case continue: opaque payload<0..2^32-1>;
+    case finished: Empty;
     case reject:   PrepareError prepare_error;
   };
 } PrepareResp;


tgeoghegan (approved):
Strict adherence to RFC 8446 aside, I think this is simply more clear than omitting the variant.

branlwyd:
Hmm -- while what you say is true about the `Handshake` message, see e.g. the `PreSharedKeyExtension` in [RFC 8446 B.3.1](https://datatracker.ietf.org/doc/html/rfc8446#appendix-B.3.1) -- this message omits all variants it does not care about, including many that are not `RESERVED`. So I think omitting "empty" cases is supported by the examples of RFC 8446. (whether or not this is clearer is up to y'all, I suppose)

cjpatton:
> Hmm -- while what you say is true about the `Handshake` message, see e.g. the `PreSharedKeyExtension` in [RFC 8446 B.3.1](https://datatracker.ietf.org/doc/html/rfc8446#appendix-B.3.1) -- this message omits all variants it does not care about, including many that are not `RESERVED`. So I think omitting "empty" cases is supported by the examples of RFC 8446. (whether or not this is clearer is up to y'all, I suppose)

I'm happy to look at more examples, but eventually we ought to take this to the TLS mailing list :) (Incidentally there is a somewhat lively discussion happening there right now about the presentation language.)

In this particular case, I think the correct interpretation is that any missing variant is treated as an error. (Similarly to Appendix B.3.) Here's the definition:
```
      struct {
          select (Handshake.msg_type) {
              case client_hello: OfferedPsks;
              case server_hello: uint16 selected_identity;
          };
      } PreSharedKeyExtension;
```

We can see here that the payload of this message depends on which handshake message the extension is included in. The message format is only defined for `ClientHello` and `ServerHello`; it is undefined for any other handshake message.

But is it really an error? Well, I happen to know that this is the extension of a payload that MUST NOT appear in any handshake message other than `ClientHello` and `ServerHello`. From Section 4.2: 

>   The table below ["pre_shared_key" is "CH, SH" ] indicates the messages where a given extension may
>   appear, using the following notation: CH (ClientHello),
>   SH (ServerHello), EE (EncryptedExtensions), CT (Certificate),
>   CR (CertificateRequest), NST (NewSessionTicket), and
>   HRR (HelloRetryRequest).  If an implementation receives an extension
>   which it recognizes and which is not specified for the message in
>   which it appears, it MUST abort the handshake with an
>   "illegal_parameter" alert.

Perhaps more generally we can interpret a missing variant as "unspecified behavior" that the rest of the spec must fill in somehow?

divergentdave (approved):

branlwyd (approved):

PR #475, opened 2023-06-15
Rename ReportShareError to PrepareError.

branlwyd:
The old name is confusing to me because almost none of the errors represented by this type are directly related to the report share. I chose the name PrepareError because this error is used during the report preparation (aka aggregation) process, matching other types similarly used during the report preparation process.


branlwyd:
(a nitpicky rename I noticed during #393, happy to take or leave this PR based on how folks feel about the rename.)

cjpatton (approved):
Good idea.

tgeoghegan (approved):

PR #483, opened 2023-06-26
Only valid reports count for min/max batch size

### LINKED ISSUE #456: min_batch_size
> The query configuration specifies the minimum batch size, `min_batch_size`, and
maximum batch size, `max_batch_size`. The Aggregator checks that `len(X) >=
min_batch_size` and `len(X) <= max_batch_size`, where `X` is the set of reports
in the batch.

Shouldn't this be *valid* reports in the batch?

tgeoghegan:
Resolves #456

### DIFF
--- a/draft-ietf-ppm-dap.md
+++ b/draft-ietf-ppm-dap.md
@@ -2236,7 +2236,7 @@ shares recovered during the aggregation sub-protocol.
 ##### Size Check
 
 The query configuration specifies the minimum batch size, `min_batch_size`. The
-Aggregator checks that `len(X) >= min_batch_size`, where `X` is the set of
+Aggregator checks that `len(X) >= min_batch_size`, where `X` is the set of valid
 reports in the batch.
 
 #### Fixed-size Queries {#fixed-size-batch-validation}
--- a/draft-ietf-ppm-dap.md
+++ b/draft-ietf-ppm-dap.md
@@ -2236,8 +2236,8 @@ shares recovered during the aggregation sub-protocol.
 ##### Size Check
 
 The query configuration specifies the minimum batch size, `min_batch_size`. The
-Aggregator checks that `len(X) >= min_batch_size`, where `X` is the set of valid
-reports in the batch.
+Aggregator checks that `len(X) >= min_batch_size`, where `X` is the set of
+reports successfully aggregated into the batch.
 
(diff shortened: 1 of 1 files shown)


draft-ietf-ppm-dap.md
```diff
@@ -2236,7 +2236,7 @@ shares recovered during the aggregation sub-protocol.
 ##### Size Check
 
 The query configuration specifies the minimum batch size, `min_batch_size`. The
-Aggregator checks that `len(X) >= min_batch_size`, where `X` is the set of
+Aggregator checks that `len(X) >= min_batch_size`, where `X` is the set of valid
```
tgeoghegan, line 2239: I considered throwing in a reference to either `{{aggregate-flow}}` or `{{validating-inputs}}` to clarify just what we mean by "valid", but I think it's clear enough: the document overall discusses report validity all over the place.

divergentdave: We should do the same under `{#fixed-size-batch-validation}` as well below.

divergentdave (approved):

branlwyd (commented):
I think saying "valid" isn't quite specific enough -- even if a given report is "valid", that doesn't necessarily mean it has completed the aggregation process. (One might argue that we must complete the aggregation process to determine if a report is valid, but several of the initial checks pre-aggregation determine "validity" of the report share.)

To avoid any wiggle room here, I think these checks should specifically check the number of reports which contributed to the values in the `CollectionReq`/`AggregateShareReq`, i.e. the number of reports in the batch which successfully completed aggregation (I believe).

tgeoghegan:
> I think saying "valid" isn't quite specific enough -- even if a given report is "valid", that doesn't necessarily mean it has completed the aggregation process. (One might argue that we must complete the aggregation process to determine if a report is valid, but several of the initial checks pre-aggregation determine "validity" of the report share.)
> 
> To avoid any wiggle room here, I think these checks should specifically check the number of reports which contributed to the values in the `CollectionReq`/`AggregateShareReq`, i.e. the number of reports in the batch which successfully completed aggregation (I believe).

Fair enough. I changed the language from "valid reports" to "successfully aggregated reports".

branlwyd (approved):

cjpatton (approved):

PR #486, opened 2023-07-06
Prepare to cut draft 05

cjpatton:


draft-ietf-ppm-dap.md
```diff
@@ -1951,9 +1974,9 @@ state related to it.
 
 #### A Note on Idempotence
 
-The reason we use a POST instead of a GET to poll the state of a collection job
-is because of the fixed-size query mode (see {{fixed-size-query}}). Collectors
-may make a query against the current batch, and it is the Leader's
+The reason we a POST is used to poll the state of a collection job instead of a
```
branlwyd, line 1977:
```suggestion
The reason a POST is used to poll the state of a collection job instead of a
```

draft-ietf-ppm-dap.md
```diff
@@ -377,15 +398,15 @@ Leader:
...
+: The Aggregator assisting the Leader with the computation. The protocol is
+  designed so that the Helper is relatively lightweight, with most of the
+  operational burdern born by the Leader.
 
 {:br}
 
 The basic unit of DAP is the "task" which represents a single measurement
-process (though potentially taken over multiple time windows). The definition of
-a task includes the following parameters:
+process (though potentially aggregating multiple batches of measurments). The
```
branlwyd, line 408:
```suggestion
process (though potentially aggregating multiple batches of measurements). The
```

draft-ietf-ppm-dap.md
```diff
@@ -211,8 +232,8 @@ seen in the clear by any server.
 {::boilerplate bcp14-tagged}
 
 Aggregate result:
-: The output of the aggregation function over a given set of measurements and
-  aggregation parameter. As defined in {{!VDAF}}.
+: The output of the aggregation function computed over a batch of measurements
+  and aggregation parameter. As defined in {{!VDAF}}.
```
branlwyd, line 236:
```suggestion
  and an aggregation parameter. As defined in {{!VDAF}}.
```

draft-ietf-ppm-dap.md
```diff
@@ -114,6 +114,27 @@ seen in the clear by any server.
...
+
+- Bump draft-irtf-cfrg-vdaf-05 to 06 {{!VDAF}}. (\*)
+
+- Specialize the protocol for two-party VDAFs (i.e., one Leader and One
+  Helper). Accordingly, update the aggregation sub-protocol to use the new
+  "ping-pong" interface for two-party VDAFs introduced in
+  draft-irtf-cfrg-vdaf-06. (\*)
+
+- Require the following actions to be idempotent: aggregation job creation,
+  collection job creation, and requesting the Helper's aggregate share.
```
branlwyd, line 127: I don't see that collection job creation is required to be idempotent. (agg job creation & agg shares are explicitly requried to be so)

branlwyd: Taking another look, I think agg job creation is also not required to be idempotent; the only related text I can find is:

> Changing an aggregation job's parameters is illegal, so further requests to PUT /tasks/{tasks}/aggregation_jobs/{aggregation-job-id} for the same aggregation-job-id but with a different AggregationJobInitReq in the body MUST fail with an HTTP client error status code.

... and this text allows, but does not require, idempotency.

cjpatton: Why doesn't this require idempotency?

branlwyd: It says that additional `PUT` requests to the same aggregation job ID but otherwise different parameters `MUST` fail (which is sensible & good). For idempotency to be required, we would need a different condition: that repeated identical requests (same aggregation job ID, same additional parameters) `MUST` succeed.

cjpatton: Fixed, PTAL and make sure it's correct. (Also feel free to suggest a clearer description.)

draft-ietf-ppm-dap.md
```diff
@@ -114,6 +114,27 @@ seen in the clear by any server.
...
+05:
+
+- Bump draft-irtf-cfrg-vdaf-05 to 06 {{!VDAF}}. (\*)
+
+- Specialize the protocol for two-party VDAFs (i.e., one Leader and One
+  Helper). Accordingly, update the aggregation sub-protocol to use the new
+  "ping-pong" interface for two-party VDAFs introduced in
+  draft-irtf-cfrg-vdaf-06. (\*)
+
+- Allow the following actions can be safely retried: aggregation job creation,
```
branlwyd, line 126:
```suggestion
- Allow the following actions to be safely retried: aggregation job creation,
```

(one last editorial nit)

branlwyd (approved):

draft-ietf-ppm-dap.md
```diff
@@ -114,6 +114,27 @@ seen in the clear by any server.
...
+  "ping-pong" interface for two-party VDAFs introduced in
+  draft-irtf-cfrg-vdaf-06. (\*)
+
+- Allow the following actions to be safely retried: aggregation job creation,
+  collection job creation, and requesting the Helper's aggregate share.
+
+- Merge error types that are related.
+
+- Drop recommendation to generate IDs using a cryptographically secure
+  pseudorandom number generator wherever pusedorandomness is not required.
```
tgeoghegan, line 132:
```suggestion
  pseudorandom number generator wherever pseudorandomness is not required.
```

tgeoghegan (approved):

PR #494, opened 2023-08-18
Square up with new VDAF ping-pong functions

tgeoghegan:
Recent VDAF changes replaced the ping-pong family of functions from VDAF-06. We now use the new ones which will appear in VDAF-07.

One consequence of this is that various aggregation protocol sub-messages now have fields of type `Message` (which is defined in TLS syntax in VDAF and thus can be included in DAP's message definitions) instead of being opaque byte buffers.

Additionally, we use the form `Vdaf.function_name()` instead of `VDAF.function_name()` to harmonize with the VDAF document.

### DIFF
--- a/draft-ietf-ppm-dap.md
+++ b/draft-ietf-ppm-dap.md
@@ -931,7 +931,7 @@ shares and the public share using the VDAF's sharding algorithm ({{Section 5.1
 of !VDAF}}), using the report ID as the nonce:
 
 ~~~
-(public_share, input_shares) = VDAF.measurement_to_input_shares(
+(public_share, input_shares) = Vdaf.measurement_to_input_shares(
     measurement, /* plaintext measurement */
     report_id,   /* nonce */
     rand,        /* randomness for sharding algorithm */
@@ -1193,14 +1193,12 @@ it from the set of candidate reports.
 Next, for each report the Leader executes the following procedure:
 
 ~~~
-state = VDAF.ping_pong_start(vdaf_verify_key,
-                             True,
-                             agg_param,
-                             report_id,
-                             public_share,
-                             plaintext_input_share.payload)
-if state != Rejected():
-  (state, outbound) = VDAF.ping_pong_req(agg_param, state, None)
+(state, outbound) = Vdaf.ping_pong_leader_init(
+    vdaf_verify_key,
+    agg_param,
+    report_id,
+    public_share,
+    plaintext_input_share.payload)
 ~~~
 
 where:
@@ -1214,12 +1212,12 @@ where:
 
 The methods are defined in {{Section 5.8 of !VDAF}}. This process determines
 the initial per-report `state`, as well as the initial `outbound` message to
-send to the Helper. (These are coalesced into a single HTTP request to the
-Helper as described below.) If `state == Rejected()`, then the report is
-rejected and removed from the set of candidate reports.
+send to the Helper. If `state` is of type `Rejected`, then the report is
+rejected and removed from the set of candidate reports, and no message is sent
+to the Helper.
 
-Next, for each candidate report the Leader constructs a `PrepareInit` message
-structured as follows:
+If `state` is of type `Continued`, then  the Leader constructs a `PrepareInit`
+message structured as follows:
 
 ~~~
 struct {
@@ -1245,8 +1243,11 @@ Each of these messages is constructed as follows:
 
   * `payload` is set to the `outbound` message computed by the previous step.
 
-Next, the Leader creates an `AggregationJobInitReq` message structured as
-follows:
+It is not possible for `state` to be of type `Finished` during Leader
+initialization.
+
+Once all the report shares have been initialized, the Leader creates an
+`AggregationJobInitReq` message structured as follows:
 
 ~~~
 struct {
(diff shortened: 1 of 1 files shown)


tgeoghegan:
The corresponding VDAF changes: https://github.com/cfrg/draft-irtf-cfrg-vdaf/pull/281

Ideally we'll wait for a VDAF including that PR to be published before we take this change to DAP.

draft-ietf-ppm-dap.md
```diff
@@ -1193,14 +1193,11 @@ it from the set of candidate reports.
...
-                             report_id,
-                             public_share,
-                             plaintext_input_share.payload)
-if state != Rejected():
-  (state, outbound) = VDAF.ping_pong_req(agg_param, state, None)
+(state, outbound) = Vdaf.ping_pong_leader_init(vdaf_verify_key,
+                                               agg_param,
+                                               report_id,
+                                               public_share,
+                                               plaintext_input_share.payload)
```
cjpatton, line 1200: I'm pretty sure the `make` winds if "art" (lines terminated by "~~~") is longer than 70 columns.

draft-ietf-ppm-dap.md
```diff
@@ -1230,7 +1227,7 @@ struct {
 
 struct {
   ReportShare report_share;
-  opaque payload<0..2^32-1>;
+  Message ping_pong_message;
```
cjpatton, line 1230: Hmm, we need to pick a more distinctive name than `Message`. Another option is to continue to handle this as an opaque byte string, including the length prefix. In fact, I think I'd prefer no to reference messages defined in another document.

draft-ietf-ppm-dap.md
```diff
@@ -1243,10 +1240,14 @@ Each of these messages is constructed as follows:
...
     Helper's) encrypted input share.
 
-  * `payload` is set to the `outbound` message computed by the previous step.
+  * `ping_pong_message` is set to the `outbound` message computed by the
+     previous step.
 
-Next, the Leader creates an `AggregationJobInitReq` message structured as
-follows:
+It is not possible for `state` to be of type `Finished` during Leader
+initialization.
```
cjpatton, line 1247: IMO this doesn't need to be said.

draft-ietf-ppm-dap.md
```diff
@@ -1314,17 +1315,27 @@ Otherwise, the Leader MUST abort the aggregation job.
 
 Otherwise, the Leader proceeds as follows with each report:
 
-1. If the inbound prep step has type "continue", then the Leader computes
+1. If the inbound prep response has type "continue", then the Leader computes
 
    ~~~
-   (state, outbound) = VDAF.ping_pong_req(agg_param, state, inbound)
+   (state, outbound) = Vdaf.ping_pong_continued(agg_param,
+                                                true, # is_leader
```
cjpatton, line 1322:
```suggestion
                                                True, # is_leader
```

cjpatton (commented):
I propose we continue to handle the outputs of `Vdaf` as opaque bytes strings, including the length prefix. It's somewhat wasteful in terms of number of bytes on the wire, but it's useful to keep a clean separation between messages defined here and messages defined in VDAF.

draft-ietf-ppm-dap.md
tgeoghegan: If we do that, then we should change the function definitions in VDAF so that instead of accepting and returning objects of type `Message`, they take opaque byte buffers. So instead of:

```
def ping_pong_continued(
            Vdaf,
            is_leader: bool,
            agg_param: bytes,
            state: State,
            inbound: Message,
        ) -> (State, Optional[Message])
```

We would have

```
def ping_pong_continued(
            Vdaf,
            is_leader: bool,
            agg_param: bytes,
            state: State,
            inbound: bytes,
        ) -> (State, Optional[bytes])
```

Which is less expressive, but keeps anything to do with `Message` inside of VDAF.

cjpatton: Sounds like a reasonable solution to me.

draft-ietf-ppm-dap.md
tgeoghegan: We end up with this:

```
   Next, for each report the Leader executes the following procedure:

(state, outbound) = Vdaf.ping_pong_leader_init(vdaf_verify_key,
                                               agg_param,
                                               report_id,
                                               public_share,
                                               plaintext_input_share.payload)

   where:
```

which didn't make the renderer bug out, but it is weird how the centered text goes past the left margin. I re-indented and now we get:

```
   Next, for each report the Leader executes the following procedure:

   (state, outbound) = Vdaf.ping_pong_leader_init(
       vdaf_verify_key,
       agg_param,
       report_id,
       public_share,
       plaintext_input_share.payload)

   where:
```

which I think is better.

draft-ietf-ppm-dap.md
tgeoghegan: I dunno, we're telling DAP authors here to go look at the `State` enum in VDAF and how to deal with its `Continued` and `Rejected` variants, so they'll quite naturally wonder what they should do about `Finished`, won't they?

cjpatton: Yeah fair enough.

draft-ietf-ppm-dap.md
tgeoghegan: OK, I made this change, with a corresponding change on my VDAF PR (commit [2066205](https://github.com/cfrg/draft-irtf-cfrg-vdaf/pull/281/commits/2066205c572de1eb625ece322b9437d677948509) https://github.com/cfrg/draft-irtf-cfrg-vdaf/pull/281) to change the ping-pong interfaces.

draft-ietf-ppm-dap.md
```diff
@@ -1314,17 +1315,27 @@ Otherwise, the Leader MUST abort the aggregation job.
 
 Otherwise, the Leader proceeds as follows with each report:
 
-1. If the inbound prep step has type "continue", then the Leader computes
+1. If the inbound prep response has type "continue", then the Leader computes
 
    ~~~
-   (state, outbound) = VDAF.ping_pong_req(agg_param, state, inbound)
+   (state, outbound) = Vdaf.ping_pong_continued(True, # is_leader
```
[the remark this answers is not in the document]
tgeoghegan: Sure, that seems more expressive than a boolean. I'll make that change in VDAF and sync up here.

draft-ietf-ppm-dap.md
```diff
@@ -1654,11 +1661,14 @@ job.
 
 Otherwise, the Leader proceeds as follows with each report:
 
-1. If the inbound prep step type is "continue" and the state is
+1. If the inbound prep response type is "continue" and the state is
    `Continued(prep_state)`, then the Leader computes
 
    ~~~
-   (state, outbound) = VDAF.ping_pong_req(agg_param, state, inbound)
+   (state, outbound) = Vdaf.ping_pong_continued(true, # is_leader,
```
divergentdave, line 1668: This should use `ping_pong_leader_continued`.

divergentdave (approved):

Issue #436, opened 2023-04-13
Do we still need `max_batch_query_count`?


cjpatton:
This seems redundant now that w'ere now enforcing this restriction at the VDAF level:

> Check if the report may still be aggregated with the current aggregation parameter. This can be done by looking up all aggregation parameters previously used for this report and calling
>
> `VDAF.is_valid(current_agg_param, previous_agg_params)`
>
> If this returns false, the input share MUST be marked as invalid with the error report_replayed.

divergentdave:
We'll need to fix #259 as well before eliminating `max_batch_query_count`. (this just came up in another discussion) Otherwise, `max_batch_query_count = 1` is load-bearing in terms of preventing replaying batches with a different set of reports.

cjpatton:
At IETF 118 I volunteered to implement multi-collection and resolve all related open issues.

PR #495, opened 2023-08-24
Clarify max_batch_query_count is counted by distinct aggregation params.

branlwyd:
I believe this is the correct check, as re-collecting a batch against an already-collected aggregation parameter will not reveal additional information -- it is the number of distinct aggregation parameters that determines the number of different collections that have been made against a given batch.

See https://github.com/divviup/janus/pull/1799 for context.

### DIFF
--- a/draft-ietf-ppm-dap.md
+++ b/draft-ietf-ppm-dap.md
@@ -2165,11 +2165,11 @@ determined by the query type. If the size check fails, then Helpers MUST abort
 with an error of type "invalidBatchSize". Leaders SHOULD wait for more reports
 to be validated and try the collection job again later.
 
-Next, the Aggregator checks that the batch has not been aggregated too many
-times. This is determined by the maximum number of times a batch can be queried,
-`max_batch_query_count`. Unless the query has been issued less than
-`max_batch_query_count` times, the Aggregator MUST abort with error of type
-"batchQueriedTooManyTimes".
+Next, the Aggregator checks that the batch has not been queried too many times.
+This is determined by the maximum number of times a batch can be queried,
+`max_batch_query_count`. If the batch has been queried with more than
+`max_batch_query_count` distinct aggregation parameters, the Aggregator MUST
+abort with error of type "batchQueriedTooManyTimes".
 
 Finally, the Aggregator checks that the batch does not contain a report that was
 included in any previous batch. If this batch overlap check fails, then the


tgeoghegan (approved):

cjpatton (approved):
This is definitely an improvement, but the actual requirement is o enforce aggregation parameter validation, as required by the specific VDAF. e.g.: https://www.ietf.org/archive/id/draft-irtf-cfrg-vdaf-06.html#section-8.2.3

For Poplar1, and even for Prio3, if the randomness generated by the Client is used more than once by the Aggregators during preparation, then there is a risk of exposing information about the measurement. See Remark 6 in https://eprint.iacr.org/2023/130.pdf.

branlwyd:
Hmm -- it seems like the larger change that is needed is for DAP to drop `max_batch_query_count` entirely, replacing it with a call to a new VDAF-specific function that captures the requirements for aggregation parameter validity.

In any case, I think we want to capture the idea that repeating the same collection request (same batch, same aggregation parameter) is valid. I think we'd need to tweak the `is_valid` definitions in VDAF to achieve this. Prio3's validity check currently checks that there are no previous aggregation parameters; the corrected check would take the union of `agg_param` into `previous_agg_params` (as a set), and check that the size of the resulting set is at most 1. I think Poplar1's check would be to partition the aggregation parameters (current & previous) by level, then check that each "level" has at most 1 distinct aggregation parameter.

cjpatton:
> In any case, I think we want to capture the idea that repeating the same collection request (same batch, same aggregation parameter) is valid. I think we'd need to tweak the `is_valid` definitions in VDAF to achieve this.

Why do we want to capture this? To make collection requests idempotent?

If so, I'm not sure if it makes sense to try to handle this at the VDAF level. `Vdaf.is_valid()` is meant to specify requirements for securely evaluating the VDAF. If a collection request is repeated, presumably we don't re-run the VDAF, but just return the aggregate result that was computed on the first run. So it sounds like all we want to say is that if the collect request with a given ID has the same content as the original, then we return the same result.


> Prio3's validity check currently checks that there are no previous aggregation parameters; the corrected check would take the union of `agg_param` into `previous_agg_params` (as a set), and check that the size of the resulting set is at most 1.

But the size of the union would only be 1 if `previous_agg_params` is empty, right?

 I think Poplar1's check would be to partition the aggregation parameters (current & previous) by level, then check that each "level" has at most 1 distinct aggregation parameter.

Similarly here: this would only be the case if `previous_agg_param` does not include the level of `agg_param`.

branlwyd:
> Why do we want to capture this? To make collection requests idempotent?

Indeed. I suppose another way to carve this out would be to explicitly state that (some?) request validation does not happen for repeated collection requests, and the expectation is that repeating a collection request (by batch ID & aggregation parameter) will simply receive the same results as last time.

> But the size of the union would only be 1 if `previous_agg_params` is empty, right?

Or if `previous_agg_params` has one element, and `agg_param` is equal to the element in `previous_agg_params`. Similarly for Poplar1: we could accept `agg_param` at a level that is already represented in `previous_agg_params` iff `agg_param` is equal to the previous agg param at the matching level.

divergentdave:
I think it would be safer to phrase this in terms of re-sending previously computed aggregate shares. Not only do we want the query and aggregation parameter to be the same, the set of reports that contributed to the aggregation must be the same as well. (this would be easier to get wrong for TimeInterval, though I suppose it could also be done wrong with FixedSize as well)

cjpatton:
> Indeed. I suppose another way to carve this out would be to explicitly state that (some?) request validation does not happen for repeated collection requests, and the expectation is that repeating a collection request (by batch ID & aggregation parameter) will simply receive the same results as last time.

Yeah, that sounds to me to be the right shape.


> > But the size of the union would only be 1 if `previous_agg_params` is empty, right?
> 
> Or if `previous_agg_params` has one element, and `agg_param` is equal to the element in `previous_agg_params`. 

Ah! But the intent though is to prevent an attacker from querying a single Prio3 proof more than once, so for Prio3 we really do want there to be one and only one acceptable agg param (the empty string). I guess the aggregation parameter is an awkward way to express this :/



Similarly for Poplar1: we could accept `agg_param` at a level that is already represented in `previous_agg_params` iff `agg_param` is equal to the previous agg param at the matching level.

tgeoghegan:
There's further work to be done here that might lead to removing `max_batch_query_count` (#436) but this change is good to go on its own.

Issue #502, opened 2023-10-06
Align VDAF interface with vdaf-07


wangshan:
The latest VDAF has simplified its API with verbs: https://github.com/cfrg/draft-irtf-cfrg-vdaf/pull/267.
A few places in DAP is still using the old ones.

cjpatton:
👍 please send a PR when you have the chance.

cjpatton:
Closed by #503.

PR #503, opened 2023-10-11
Align VDAF interface with vdaf-07

wangshan:
Fix issue https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/issues/502


cjpatton (approved):
Thanks!

PR #506, opened 2023-10-18
editorial: `outbound` doesnt't look right in .txt

cjpatton:


draft-ietf-ppm-dap.md
```diff
@@ -1,5 +1,4 @@
----
-title: "Distributed Aggregation Protocol for Privacy Preserving Measurement"
+--- title: "Distributed Aggregation Protocol for Privacy Preserving Measurement"
```
divergentdave, line 1: I think this change should be rolled back because it delimits the YAML document for front matter

divergentdave (approved):

draft-ietf-ppm-dap.md
cjpatton: whoops, fat fingers!

PR #507, opened 2023-10-19
Clarify Sybil attack mitigations

### LINKED ISSUE #499: Report selection
The Leader sees client IP addresses and can easily inject and drop reports. This puts the Leader in a powerful position for Sybil attacks. On the surface, it seems like an interesting suggestion to let the Helper select the reports for a certain aggregation instead.

Here are some counterarguments:
1.. The helper needs to explicitly mix reports for this to have any effect.
2. The leader can still single out reports by dropping and injecting sufficiently many to make the mixing ineffective.
3. An OHTTP server to obfuscate client identities might be more beneficial anyway.
4. Sybil attack protection would limit what the leader can do to single out reports.

chris-wood:
Closes #499

### DIFF
--- a/draft-ietf-ppm-dap.md
+++ b/draft-ietf-ppm-dap.md
@@ -1458,8 +1458,8 @@ where:
 * `plaintext_input_share` is the Helper's `PlaintextInputShare`
 
 This procedure determines the initial per-report `state`, as well as the
-initial `outbound` to send in response to the Leader. If `state` is of type
-`Rejected`, then the Helper responds with
+initial `outbound` message to send in response to the Leader. If `state` is of
+type `Rejected`, then the Helper responds with
 
 ~~~
 struct {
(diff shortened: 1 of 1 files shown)


cjpatton (approved):

draft-ietf-ppm-dap.md
```diff
@@ -2625,17 +2626,24 @@ reveal additional information about the honest measurements, leading to a
...
-Clients and to enforce policy on things like upload rates. Note that because the
-Helper never handles messages directly from the Clients, reports would have to
-use an extension ({{upload-extensions}}) to convey authentication information to
-the Helper.
+programmed with an identity), Client authentication can help Aggregators (or an
+authenticating proxy deployed between clients and the Aggregators; see
+{{anon-proxy}}) ensure that all reports come from authentic Clients. Note that
+because the Helper never handles messages directly from the Clients, reports
+would have to use an extension ({{upload-extensions}}) to convey authentication
+information to the Helper.
```
tgeoghegan, line 2646: We should link to https://github.com/cpriebe/draft-priebe-ppm-dap-reportauth if it ever makes it to the Datatracker.

tgeoghegan (approved):

PR #509, opened 2023-10-19
Add a PrepareError registry

### LINKED ISSUE #487: Registry for PrepareError
[draft-priebe-ppm-dap-reportauth](https://github.com/cpriebe/draft-priebe-ppm-dap-reportauth) (no I-D submitted yet) wants to define a new `PrepareError` type, but we do not have a registry for this, only the enum definition. In conjunction with setting up the necessary policy for `PrepareError` extensibility, we may want to expand the encoded size from one byte to two bytes. We are already using 4% of codepoints for DAP itself, and additional headroom would make managing this namespace easier.

chris-wood:
Closes #487

### DIFF
--- a/draft-ietf-ppm-dap.md
+++ b/draft-ietf-ppm-dap.md
@@ -3443,19 +3443,42 @@ Change controller:
 
 : IESG
 
-## Query Types Registry {#query-type-reg}
+## DAP Type Registries
+
+This document also requests creation of a new "Distributed Aggregation Protocol
+Parameters" page. This page will contain several new registries, described in the
+following sections.
+
+### Query Types Registry {#query-type-reg}
 
 This document requests creation of a new registry for Query Types. This registry
 should contain the following columns:
 
 [TODO: define how we want to structure this registry when the time comes]
 
-## Upload Extension Registry
+### Upload Extension Registry
 
 This document requests creation of a new registry for extensions to the Upload
 protocol. This registry should contain the following columns:
 
 [TODO: define how we want to structure this registry when the time comes]
+
+### Prepare Error Registry {#prepare-error-reg}
+
+This document requests creation of a new registry for PrepareError values. This
+registry should contain the following columns:
+
+Name:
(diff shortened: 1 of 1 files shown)


draft-ietf-ppm-dap.md
```diff
@@ -1404,7 +1404,7 @@ enum {
   task_expired(7),
   invalid_message(8),
   report_too_early(9),
-  (255)
+  (65535)
```
tgeoghegan, line 1407: Can you explain the choice to widen this error type? Are we ever going to define this many distinct error codes? Admittedly errors should be the minority of messages so this extra byte won't cost extra bandwidth in the common case of the protocol.

cjpatton, line 1407: Why bump 1-byte to 2-byte?

chris-wood:
@tgeoghegan @cjpatton the size increase is to account for random selection of new error types for experimentation. But we can do that separately, if at all, so I'll revert it from this PR.

draft-ietf-ppm-dap.md
chris-wood, line 1407:
```suggestion
  (255)
```

draft-ietf-ppm-dap.md
```diff
@@ -3414,20 +3414,43 @@ Change controller:
...
+### Prepare Error Registry {#prepare-error-reg}
+
+This document requests creation of a new registry for PrepareError values. This
+registry should contain the following columns:
+
+Name:
+: The name of the PrepareError value
+
+Value:
+: The 2-byte value of the PrepareError value
```
tgeoghegan, line 3446:
```suggestion
: The 1-byte value of the PrepareError value
```

tgeoghegan (approved):

cjpatton (approved):

cjpatton:
@chris-wood can you squash?

cjpatton:
Rebased and squashed.

Issue #505, opened 2023-10-17
Security considerations: privacy impact of per-task HpkeConfigs


divergentdave:
We may want to make note of the fact that the helper sees IP addresses of clients only during /hpke_config requests. Therefore the helper will see the combination of IP addresses and task IDs if using per-task HpkeConfigs, or IP addresses alone if not. This is a smaller difference (note that the leader sees them together either way) but we may want to note it as a consideration for choosing whether to use per-task HpkeConfigs. Note that OHTTP proxies would mitigate this.

h/t @tgeoghegan

cjpatton:
Relatedly: one benefit of 1-byte IDs for HpkeConfigs is that it reduces the "fingerprinting" surface by restricting the number of possible distinct IDs.

I don't recall if we did this mindfully; as I recall we just copy-pasted this from ECH. @chris-wood do you remember how we eneded up shortening the HpkeConfig ID in ECH to one byte? Was this about fingerprinting or something else?

chris-wood:
That was precisely the reason.

tgeoghegan:
This issue initially was about adding a security consideration, but I think we should do a couple other things in these paragraphs that seem related. There's a TODO lurking in this text:

> [TODO: Allow Aggregators to return HTTP status code 403 Forbidden in deployments that use authentication to avoid leaking information about which tasks exist.]

Additionally, I don't know if this prescription is necessary:

> Clients retrieve the HPKE configuration from each Aggregator by sending an HTTP GET request to {aggregator}/hpke_config. Clients MAY specify a query parameter task_id whose value is the task ID whose HPKE configuration they want. *If the Aggregator does not recognize the task ID, then it MUST abort with error unrecognizedTask.*

We don't need a DAP-layer error. An HTTP client error should suffice (as usual, implementations can put a problem document in the body of the HTTP 4XX if they want). Note that I still think we need the `missingTaskID` langauge that instructs clients to "upgrade" their HPKE config request if the aggregator doesn't use global HPKE configs.

tgeoghegan:
I and others have come around to a new proposal here: we should do away with per-task HPKE keys altogether. The primary motivation for this is to reduce the complexity of the protocol: if only global HPKE keys are supported, then we can remove the task ID query parameter from the `/hpke_config` route and remove the description of the fallback logic.

I have two more arguments in favour of doing this.

## Oblivious HTTP

[...]

PR #510, opened 2023-10-19
Make HPKE config fetching API consistent with other API endpoints

### LINKED ISSUE #459: Align task-specific HPKE config with other task-specific resources
It is a bit odd to have this structure.  I would have said global = `{aggregator}/hpke_config` and task-specific = `{aggregator}/tasks/{task-id}/hpke_config`.  Though I suspect that achieving consistency would be harder for the latter form, if you ever had to worry about that.

_Originally posted by @martinthomson in https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/pull/454#discussion_r1181203600_

chris-wood:
Closes #459

### DIFF
--- a/draft-ietf-ppm-dap.md
+++ b/draft-ietf-ppm-dap.md
@@ -834,10 +834,12 @@ configuration of each Aggregator. See {{compliance}} for information on HPKE
 algorithm choices.
 
 Clients retrieve the HPKE configuration from each Aggregator by sending an HTTP
-GET request to `{aggregator}/hpke_config`. Clients MAY specify a query parameter
-`task_id` whose value is the task ID whose HPKE configuration they want. If the
-Aggregator does not recognize the task ID, then it MUST abort with error
-`unrecognizedTask`.
+GET request to `{aggregator}/tasks/{task-id}/hpke_config`. If the Aggregator
+does not recognize the task ID, then it MUST abort with error `unrecognizedTask`
+and return the HTTP status code 404 Not Found. The Aggregator can choose to
+deny clients access to the HPKE configuration depending on the deployment model,
+e.g., if only authenticated clients can obtain HPKE configurations. In these
+cases, the Aggregator returns the HTTP status code 403 Forbidden.
 
 An Aggregator is free to use different HPKE configurations for each task with
 which it is configured. If the task ID is missing from  the Client's request,
@@ -849,9 +851,6 @@ an `HpkeConfigList` value, with media type "application/dap-hpke-config-list".
 The `HpkeConfigList` structure contains one or more `HpkeConfig` structures in
 decreasing order of preference. This allows an Aggregator to support multiple
 HPKE configurations simultaneously.
-
-[TODO: Allow Aggregators to return HTTP status code 403 Forbidden in deployments
-that use authentication to avoid leaking information about which tasks exist.]
 
 ~~~
 HpkeConfig HpkeConfigList<1..2^16-1>;


tgeoghegan (changes requested):
This breaks global HPKE configs, which are not bound to any specific task, and are necessary for taskprov to work. Previously we considered exposing both `GET {aggregator}/tasks/{task-id}/hpke_config` and `GET {aggregator}/hpke_config`, but I think we stuck with the query param layout because it was good enough.

chris-wood:
How does this break global configs? Aggregators can just put the same config behind each per-task resource, no?

tgeoghegan:
That would work, but it breaks taskprov, specifically: in that case, the aggregators may not yet know what the task ID is. The task configuration blob that lets them derive a task ID isn't provided until an upload request, but the client can't do an upload until it gets HPKE configs to encrypt report shares to.

chris-wood:
That's a taskprov problem to solve and shouldn't force the hand of the core protocol. That draft can reserve a special taskid for the global case, if it chooses to do it that way. Or it can introduce the global API endpoint that this particular change removes. Both would be valid.

tgeoghegan:
I agree, but I think we should defer this change past the draft we'll cut today(?) for IETF 118 so we can coordinate this with the taskprov authors.

cjpatton (approved):
Looks good to me.

Another way taskprov could address this is by adding the taskprov HTTP header to the GET request so that the Aggregator can derive the task ID. (cc/ @wangshan)

branlwyd:
This is related to #505, the gist of which is "task-specific configs leak the IPs of clients interested in the task to the Helper." With this change, client IPs would always be leaked to the Helper, even when the HPKE configs being returned are actually global -- deployments would have no way to avoid this downside.

We may want to decide if #505 is something deployments will care to address before taking this PR.

cjpatton:
Good call @branlwyd, though I think preserving the semantics of global mode should be possible. (@chris-wood  suggested a couple of strategies.)

chris-wood:
> This is related to #505, the gist of which is "task-specific configs leak the IPs of clients interested in the task to the Helper." With this change, client IPs would always be leaked to the Helper, even when the HPKE configs being returned are actually global -- deployments would have no way to avoid this downside.
> 
> We may want to decide if #505 is something deployments will care to address before taking this PR.

This doesn't really make sense to me. Whether or not the client IP is revealed to the Aggregator is orthogonal to how the API is structured. If that's something deployments care about, they should take steps to protect their IP address via proxies.

branlwyd:
> This doesn't really make sense to me. Whether or not the client IP is revealed to the Aggregator is orthogonal to how the API is structured. If that's something deployments care about, they should take steps to protect their IP address via proxies.

No matter how the HPKE config request is structured, the client IP is revealed to the Helper. If the request to the Helper for the HPKE config also includes the task ID, the Helper can infer that the requesting IP is interested in that specific task (which #505 considers a privacy leak).

I agree this might be better-mitigated by proxies, as it is challenging for me to envision a scenario where the client IP can be freely shared with the Leader but must not be shared with the Helper.

chris-wood:
@branlwyd you seem to be agreeing with me. This PR has no effect on whether or not client IPs are revealed. That's a totally orthogonal problem.

branlwyd:
@chris-wood to clarify:

* #505 points out that task-based HPKE config requests allow the Helper to build a map from client IP to tasks of interest for that IP, by the task ID included in the HPKE config request.
* With this PR, the issue pointed out by #505 would also be true for deployments using global HPKE configs, since with this PR all requests for an HPKE config include a task ID.

(that is, the risk isn't that client IPs are leaked, it is that the helper can determine which tasks each client IP is interested in.)

Separately, as you pointed out, an alternate strategy to mitigate this risk would be to deploy a proxy in front of the DAP aggregators which hides client IPs.

My suggested course of action would be:
* If a proxy is an acceptable mitigation, close #505 and take this PR.
* If a proxy is not an acceptable mitigation, don't take this PR.

chris-wood:
@branlwyd again, this PR doesn't change the situation here, since the spec _currently_ allows clients to fetch per-task configs. This PR is not a functionality change. It's an aesthetic one.

chris-wood:
> @chris-wood With this PR, _every_ request for an HPKE config now includes a task ID. Prior to this PR, only task-specific HPKE config requests include a task ID. So this PR expands the issue raised by #505 to affect deployments using global HPKE configs, too.

Again, this is an aesthetic change, not a functional one. IP address exposure is a distraction here.

chris-wood:
> Forget the IP address for a minute. Before this change, a helper would expose global HPKE configs at `/hpke_config`. A request to that path reveals nothing about what task(s) clients are interested in. After this change, it's at `/tasks/{task-id}/hpke_config`, and even if the helper serves up the exact same list of HPKE configs for any `task-id` value, it now can learn how many clients are interested in one task versus another. That's the information leak Brandon is pointing to.

Yeah, I agree with this, but what I'm trying to communicate is that, to me, this doesn't seem like a meaningful change because the _leader_ will already learn what task a client is interested in via the upload request. (I know that this particular PR changes what the _helper_ sees, but to me this doesn't feel like a notable distinction.) In any case, the solution to this general problem and #505 in particular seems to be the same -- toss in a proxy if you care about this leakage to any aggregator.

PR #511, opened 2023-10-19
Make the checksum optional

### LINKED ISSUE #446: Cheaper checksum
The current checksum algorithm is a little expensive.  It would be easier to just concatenate all of the report IDs and hash the resulting stream.  SHA-256 implementations usually have an IUF interface.

chris-wood:
Closes #446

### DIFF
--- a/draft-ietf-ppm-dap.md
+++ b/draft-ietf-ppm-dap.md
@@ -1988,13 +1988,7 @@ the set of current batches in the Leader, and thus using a GET is inappropriate.
 ### Obtaining Aggregate Shares {#collect-aggregate}
 
 The Leader must obtain the Helper's encrypted aggregate share before it can
-complete a collection job. To do this, the Leader first computes a checksum
-over the reports included in the batch. The checksum is computed by taking the
-SHA256 {{!SHS=DOI.10.6028/NIST.FIPS.180-4}} hash of each report ID from the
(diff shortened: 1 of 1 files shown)


cjpatton (commented):
In principle I have no objection to the checksum being optional, but I think we should figure out a way to signal to the Helper at _aggregation time_ that the the checksum will be requested later at collection time. I explain below.

The checksum is designed so that the order of reports doesn't matter: if both Aggregators have the same _set_ of reports in the batch, then the checksums will match and collection will succeed. This means that we can either compute it as we go at aggregation time or wait until collection time and compute it all at once. The former is strictly better because we already have the report IDs in memory during aggregation time, whereas if we wait until collection time, we'd have to make a bunch of database queries to look up the report IDs in the batch.

Computing it aggregation time is always better I think. If the Helper does not know whether the checksum will be requested, I suspect it'll always want to compute it ahead of time just in case.

We can avoid this by signaling ahead of time that the checksum will be requested. One way to do this would be to add the signal to the task config.

chris-wood:
Putting that signal in the task config seems like it'd work!

cjpatton (changes requested):
Moved to "request changes" while we wait on discussion. (This PR needs work as-is, but it could work with minor changes.)

chris-wood:
Closing per discussion at IETF 118.

PR #513, opened 2023-10-19
Rewrite verification key example

### LINKED ISSUE #414: Consider binding the task ID to the VDAF verification key
As of #411, the draft spells out requirements around exchanging the VDAF verification key. It includes the following suggestion:

>  One way to ensure this is to include the verification key in a derivation of the task ID.

The idea being that by deriving the task ID verification key, we assure clients that the key was picked prior to the task being configured and the key doesn't change over the course of the task.

We might consider spelling out a scheme for this, in order to remove this security critical choice from deployments.

See https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/issues/407 for more discussion

chris-wood:
Closes #414

### DIFF
--- a/draft-ietf-ppm-dap.md
+++ b/draft-ietf-ppm-dap.md
@@ -2665,14 +2665,29 @@ minimum batch size of 1 report).
 ### Verification key requirements {#verification-key}
 
 The verification key for a task SHOULD be chosen before any reports are
-generated. It SHOULD be fixed for the lifetime of the task and not be rotated.
-One way to ensure this is to include the verification key in a derivation of the
-task ID.
+generated so that it is independent of any reports. Moreover, it SHOULD be
+fixed for the lifetime of the task and not be rotated. One way to ensure
+that the key is independent is to derive the verification key (verify_key)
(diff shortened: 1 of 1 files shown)


cjpatton (approved):
🦆 - very nice to have a clear example of this. For the record, this is what we do in https://github.com/wangshan/draft-wang-ppm-dap-taskprov. If and when that gets adopted by the WG, we should add a reference to it here.

tgeoghegan (changes requested):
I think gesturing at the out of band agreement to a verify key seed is unnecessarily complicated for illustrating this. taskprov requires multiple independent actors to independently derive the same task ID, so you need a previously agreed upon verify key seed. But in the task provisioning protocol we worked out for Divvi Up, we have a single entity responsible for choosing the task ID and then distributing it to participating aggregators, so we just randomly generate the verify key and then take a SHA-256 of that.

chris-wood:
@tgeoghegan that's fine to include as an example. The point here is to not be prescriptive.

tgeoghegan (approved):

PR #516, opened 2023-10-24
Re-allow creation of aggregation jobs after collection job created.

branlwyd:
This wording was introduced by #512, please see that PR for context. Removing this sentence is desirable for a few reasons:

* For any VDAFs which require an aggregation parameter before aggregation can begin, or which require multiple aggregations with differing aggregation parameters, a collection job must be received before aggregation jobs can be created.
* Even for VDAFs which do not use an aggregation parameter, implementations may wish to use a more flexible job-scheduling strategy. For example, Janus will still create new aggregation jobs while awaiting completion of existing aggregation jobs, even if a collection job has been received for the relevant batch.
* I believe the technical requirement here is captured by the preceding sentence, i.e. aggregation jobs must be completed before an aggregate share request is made -- so it is safe to remove this sentence.

### DIFF
--- a/draft-ietf-ppm-dap.md
+++ b/draft-ietf-ppm-dap.md
@@ -1934,10 +1934,8 @@ collection job or tries again later, depending on which requirement in
 {{batch-validation}} was not met. If the Leader has a pending aggregation job
 that overlaps with the batch for the collection job, the Leader MUST first
 complete the aggregation job before proceeding and requesting an aggregate share
-from the Helper. After the collection job has been created, the Leader MUST NOT
-schedule new aggregation jobs that overlap with the batch for the collection
-job. This avoids a race condition between aggregation and collection jobs that
-can yield trivial batch mismatch errors.
+from the Helper. This avoids a race condition between aggregation and collection
+jobs that can yield trivial batch mismatch errors.
 
 Once both aggregate shares are successfully obtained, the Leader responds to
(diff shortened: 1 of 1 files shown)


branlwyd:
Nitpick around: "If the Leader has a pending aggregation job that overlaps with the batch for the collection job, ..."

I think the technical requirement here is that the aggregation job would need to overlap the batch *and aggregation parameter* of an existing collection job. Thoughts on updating text to capture that the aggregation parameter matters too?

tgeoghegan (approved):
Yes, this should reckon with `(batch, aggregation parameter)` tuples. It might be helpful to introduce into the glossary some noun that refers to that tuple and then use it in various places, but I can't think of a suitable noun right now. Regardless, we should take this change to address the regression introduces in the previous PR.

branlwyd:
> Yes, this should reckon with `(batch, aggregation parameter)` tuples. It might be helpful to introduce into the glossary some noun that refers to that tuple and then use it in various places, but I can't think of a suitable noun right now. Regardless, we should take this change to address the regression introduces in the previous PR.

100% agreed on coming up with better terminology (though it doesn't need to happen here, of course). I have seen "batch" variably refer to:

* A minimal unit of collection.
* A contiguous-in-time sequence of minimal units of collection, in the time-interval query type.
* Either of the above, along with an aggregation parameter.

This can get pretty confusing!

cjpatton (approved):

PR #518, opened 2023-10-24
Provide guidance for abandoning aggregation jobs

### LINKED ISSUE #241: Allow leader to explicitly cancel aggregate jobs
It's possible for a leader to abandon an aggregate job before it's finished, and the helper never gets to find out. For instance, suppose leader sends `AggregateInitReq`, then helper responds with a malformed `AggregateResp`, maybe just because there is a bug in the client's message encoding routine. Upon receipt of the malformed message, the leader will cease preparation of the reports in that aggregate job. From the helper's point of view, there will forever be an unfinished aggregate job, awaiting the next `AggregateReq` from leader.

This introduces an ambiguity when helper handles `AggregateShareReq`. It's possible that the helper might have some abandoned aggregate jobs whose reports would have contributed to the `AggregateShareReq`. We might want the helper to abort here, because this could indicate that the leader has jumped the gun on starting the collect protocol, but we can't, because there's no way for the helper to know whether any of those jobs were intentionally abandoned by the leader.

This could be fixed by having the leader explicitly send an aggregate job cancellation message to the helper if it sees an `AggregateResp` it doesn't like.

cjpatton:
Closes #241.

### DIFF
--- a/draft-ietf-ppm-dap.md
+++ b/draft-ietf-ppm-dap.md
@@ -1803,6 +1803,10 @@ prepare for the next continuation step ({{aggregation-helper-continuation}}).
 Otherwise, if `state == Finished(out_share)`, then the Helper stores `out_share`
 for use in the collection sub-protocol ({{collect-flow}}).
 
+If for whatever reason the Leader must abandon the aggregation job, it SHOULD
+send an HTTP DELETE request to the aggregation job URI so that the Helper knows
+it can clean up its state.
+
 #### Recovering from Aggregation Step Skew {#aggregation-step-skew-recovery}
 
 `AggregationJobContinueReq` messages contain a `step` field, allowing


branlwyd (approved):

draft-ietf-ppm-dap.md
```diff
@@ -1638,6 +1638,10 @@ Whether this phase is reached depends on the VDAF: for 1-round VDAFs, like
 Prio3, processing has already completed. Continuation is required for VDAFs
 that require more than one round.
 
+If for whatever reason the Leader must abandon the aggregation job, it SHOULD
+send an HTTP DELETE request to the aggregation job URI so that the Helper knows
+it can clean up its state.
```
tgeoghegan, line 1643: Since this recommendation applies to the leader, it should be in the "Leader Continuation" section, below.

cjpatton: Done!

tgeoghegan: Now I'm being difficult: I think it makes more sense to put this paragraph at the very end of "Leader Continuation" rather than the very beginning.

cjpatton: I don't fully agree given that it's a branch point that goes more naturally before constructing the next request. However I don't think it makes much of a practical difference, so I'm happy to follow you here.

cjpatton: Moved!

tgeoghegan (approved):

PR #524, opened 2023-11-07
Recommend that task configuration is transparent

### LINKED ISSUE #500: Making privacy-relevant task parameters visible to clients
The `min_batch_size` task parameter is relevant to the question of how well a given DAP tasks protects the privacy of its measurements. While agreement on the value of `min_batch_size` is out of scope of the protocol, both Aggregators must know its value. In some deployment scenarios, we may have the goal that that reverse engineering a Client provides enough information to assess how the privacy of measurements is protected. It would be easy to distribute `min_batch_size` to the Client as well, but it does not make use of this value, and any results would require trust in the means by which this `min_batch_size` was distributed to the client, on top of the Aggregator non-collusion assumption.

We could provide more robust transparency for this parameter in a couple ways: either by asking both aggregators, or by having the Client commit to its view of the parameter and honest Aggregators checking this commitment.

On-line verification verification of parameters would require both Aggregators to support some new endpoint, and return a select set of task parameters in response to requests for a particular task. Someone auditing a Client implementation could identify the task ID and aggregator endpoints, query both aggregators, confirm that they match, and then assume that aggregate shares will obey the `min_batch_size` restriction. It would not be necessary for clients to check this endpoint before uploading.

Having clients commit to `min_batch_size` is already accomplished by the [`taskprov`](https://datatracker.ietf.org/doc/draft-wang-ppm-dap-taskprov/) draft, because `min_batch_size` is hashed into the derivation of the task ID. If we assume that at least one aggregator is honest and that it only accepts task parameters via the `taskprov` header mechanism, then it must agree with the client on the value of `min_batch_size`, and we know it will follow that when releasing aggregate shares. (if there is one dishonest Aggregator claiming to use `taskprov` and one honest Aggregator that accepts any set of task parameters from the dishonest Aggregator via a non-`taskprov` means, then this falls apart)

There are other ways one could achieve the same commitment, without taking on the entirety of `taskprov`. For example, we could distribute `min_batch_size` to the client, and have it include it in some new extension in both report shares. We know this will be distributed to both Aggregators, since extensions are included in the HPKE AAD. Then, both Aggregators would check that the client's view of task parameters matches theirs, and proceed if it does. Alternately, we could consider adding `min_batch_size` to the report share AAD. This would save a few bytes per report, at the expense of requiring that Clients always be provided this parameter.

If we consider cases where the Aggregators add noise to aggregate shares for differential privacy, then we would want to include differential privacy parameters along with `min_batch_size` in the above mechanisms as well. As before, once we have confirmed that the Client and both Aggregators have a common view of the DP parameters, then we assume that at least one honest Aggregator will add noise following those parameters to all aggregate shares. (or, if noise is added inside MPC instead of in the clear by each Aggregator, we would have a stronger conclusion)

cjpatton:
Closes #500.

### DIFF
--- a/draft-ietf-ppm-dap.md
+++ b/draft-ietf-ppm-dap.md
@@ -2796,6 +2796,24 @@ participating Aggregators stored unencrypted input shares on the same cloud
 object storage service, then that cloud vendor would be able to reassemble all
 the input shares and defeat privacy.
 
+## Task Configuration Agreement and Consistency
+
+In order to execute a DAP task, it is necessary for honest parties to ensure
+they agree on the configuration of the task. The process by which tasks are
+configured is not in scope of this document. This means that, in general, it is
+possible for a party to participate in a task without knowing all of the
+relevant parameters. For example, a Client can upload a report
+({{upload-flow}}) without knowing the minimum batch size that is enforced by
+the Aggregators during collection ({{collect-flow}}).
+
+Depending on the deployment model, agreement can require that task parameters
+are visible to all parties such that each party can choose whether to participate
+based on the value of any parameter. This includes the parameters enumerated
+in {{task-configuration}} and any additional parameters implied by upload
+extensions {{upload-extensions}} used by the task. Since meaningful privacy
+requires that multiple clients contribute to a task, they should also share a
+consistent view of the task configuration.
+
 # IANA Considerations
 
 ## Protocol Message Media Types


draft-ietf-ppm-dap.md
```diff
@@ -809,6 +809,11 @@ A task's parameters are immutable for the lifetime of that task. The only way to
 change parameters or to rotate secret values like collector HPKE configuration
 or the VDAF verification key is to configure a new task.
 
+This document does not specify a mechanism for configuring each party with the
+task parameters. Such a mechanism SHOULD be transparent so that each party has
+an opportunity to opt out based on the value of any parameter. (See
```
cjpatton, line 814: Reviewer note: I considered refining this to "the value of any _privacy-relevant_ parameter", but I thought it would be a bit bike-sheddy to work out which parameters are privacy-relevant and which aren't.

npdoty:
I think it would be helpful to set that expectation of transparency, even if the implementation will depend on an extension or an out of band configuration. I suggested a small change to the language to make the choice clear (rather than "opt out").

cjpatton:
> I think it would be helpful to set that expectation of transparency, even if the implementation will depend on an extension or an out of band configuration. I suggested a small change to the language to make the choice clear (rather than "opt out").

Hmm, I don't see your suggestion. Did you create a comment and forget to publish it in the review?

tgeoghegan:
Task parameter negotiation and distribution is outside of DAP's scope. I don't know that it's wise for this document to contain SHOULDs about something it explicitly declares out of scope, especially since the definition of transparency is not especially strong and the recommendation for how to achieve it is not concrete. How could it be, when we deliberately don't define a representation of task parameters?

I think we should at most keep the security considerations item. I'm happy to be overruled by the other editors and the WG though.

npdoty:
The security considerations piece seems like the most significant to me. That section already includes other security-related considerations regarding task parameters, even with a note that while it's not defined in the spec, there are security implications.

cjpatton:
I've removed the paragraph from the task config section and took @npdoty's suggested wording change. Please take another look!

npdoty (approved):

cjpatton:
Rebased and squashed.

draft-ietf-ppm-dap.md
```diff
@@ -2797,6 +2797,22 @@ participating Aggregators stored unencrypted input shares on the same cloud
 object storage service, then that cloud vendor would be able to reassemble all
 the input shares and defeat privacy.
 
+## Transparency of the Task Configuration
```
chris-wood, line 2800:
```suggestion
## Task Configuration Agreement and Consistency
```

chris-wood: Transparency is likely to confuse people who may understand it to have the CT or KT-like meaning, which isn't what we're talking about here. Rather, I think what we're talking about is simply agreement.

draft-ietf-ppm-dap.md
```diff
@@ -2797,6 +2797,22 @@ participating Aggregators stored unencrypted input shares on the same cloud
...
+possible for a party to participate in a task without knowing all of the
+relevant parameters. For example, a Client can upload a report
+({{upload-flow}}) without knowing the minimum batch size that is enforced by
+the Aggregators during collection ({{collect-flow}}).
+
+The task configuration mechanism SHOULD ensure that all task parameters are
+visible to all parties such that each party can choose whether to participate
+in the task based on the value of any parameter. This includes the parameters
+enumerated in {{task-configuration}} and any additional parameters implied by
+upload extensions {{upload-extensions}} used by the task.
```
chris-wood, line 2814:
```suggestion
Depending on the deployment model, agreement can require that task parameters
are visible to all parties such that each party can choose whether to participate
based on the value of any parameter. This includes the parameters enumerated
in {{task-configuration}} and any additional parameters implied by upload
extensions {{upload-extensions}} used by the task. It also includes the task itself:
meaningful privacy requires that multiple clients contribute to a task and therefore
share a consistent view of the task configuration.
```

chris-wood (changes requested):
I think we ought to talk not just about agreement, but also about consistency of tasks amongst clients. DAP with a single client uploading reports is not great!

draft-ietf-ppm-dap.md
```diff
@@ -2797,6 +2797,24 @@ participating Aggregators stored unencrypted input shares on the same cloud
...
+({{upload-flow}}) without knowing the minimum batch size that is enforced by
+the Aggregators during collection ({{collect-flow}}).
+
+Depending on the deployment model, agreement can require that task parameters
+are visible to all parties such that each party can choose whether to participate
+based on the value of any parameter. This includes the parameters enumerated
+in {{task-configuration}} and any additional parameters implied by upload
+extensions {{upload-extensions}} used by the task. It also includes the task itself:
+meaningful privacy requires that multiple clients contribute to a task and therefore
+share a consistent view of the task configuration.
```
tgeoghegan, line 2816: This is confusing: what is a task but the parameters that constitute it? I think the text you had before was better, though I like that this version removes the all-caps SHOULD.

chris-wood: It's the set of all parameters. Can you offer suggestions?

cjpatton:
Squashed and rebased

draft-ietf-ppm-dap.md
tgeoghegan:
```suggestion
based on the value of any parameter. This includes the parameters enumerated
in {{task-configuration}} and any additional parameters implied by upload
extensions {{upload-extensions}} used by the task. Since meaningful privacy
requires that multiple clients contribute to a task, they should also share a
consistent view of the task configuration.
```

cjpatton:
Took @tgeoghegan's suggestion and squashed.

tgeoghegan (approved):

cjpatton:
@chris-wood can you take a last look?

chris-wood (approved):
🚢 🚢 🚢

PR #525, opened 2023-11-16
Clarify mapping of input shares to leader/helper, encoding requirements.

branlwyd:
Specifically, call out that the sharding algorithm will return two input shares, and the first one is the Leader's.

Also describe that DAP only works with VDAFs which have defined encodings for the various VDAF types which DAP embeds into its own types as opaque byte strings. I purposefully leave the set of messages which require a defined encoding ambiguous in hopes that this avoids the possibility of forgetting to update the list as the specification is changed.


branlwyd:
A few notes:
* These changes fell out of some implementation questions asked by @jbr.
* This point only matters if we decide we want to explicitly list the set of VDAF types which must have an encoding specified. Most of the VDAF types which require a specified encoding are clear -- they're the VDAF types which are embedded in some DAP type as an opaque byte string. The one VDAF type for which the encodability requirement isn't clear is the VDAF state: this type is never sent on the wire, and indeed an in-memory-only DAP implementation could get away without an encoding for the VDAF state at all. However, a DAP implementation which durably stores state between requests will need _some_ encoding for the VDAF state. IMO, if we decide to list VDAF types explicitly, we should call out the VDAF state as being only recommended for encoding.
* AFAICT, the DAP specification doesn't mention that VDAFs should be instantiated for 2 aggregators. IMO, this is OK to leave ambiguous.

draft-ietf-ppm-dap.md
```diff
@@ -639,9 +639,12 @@ struct {
...
 ~~~
 
-DAP uses the 16-byte `ReportID` as the nonce parameter for the VDAF
-`shard` and `prep_init` methods (see {{!VDAF, Section 5}}). Thus for a VDAF to
-be compatible with DAP, it MUST specify a `NONCE_SIZE` of 16 bytes.
+DAP uses the 16-byte `ReportID` as the nonce parameter for the VDAF `shard` and
+`prep_init` methods (see {{!VDAF, Section 5}}). Additionally, DAP includes
+messages defined in the VDAF specification encoded as opaque byte strings within
+various DAP messages. Thus, for a VDAF to be compatible with DAP, it MUST
+specify a `NONCE_SIZE` of 16 bytes, and must specify encodings all relevant
```
jbr, line 646: "…must specify encodings all relevant…" probably is missing a word, but I'm not sure what word was intended

draft-ietf-ppm-dap.md
```diff
@@ -639,9 +639,12 @@ struct {
...
 
-DAP uses the 16-byte `ReportID` as the nonce parameter for the VDAF
-`shard` and `prep_init` methods (see {{!VDAF, Section 5}}). Thus for a VDAF to
-be compatible with DAP, it MUST specify a `NONCE_SIZE` of 16 bytes.
+DAP uses the 16-byte `ReportID` as the nonce parameter for the VDAF `shard` and
+`prep_init` methods (see {{!VDAF, Section 5}}). Additionally, DAP includes
+messages defined in the VDAF specification encoded as opaque byte strings within
+various DAP messages. Thus, for a VDAF to be compatible with DAP, it MUST
+specify a `NONCE_SIZE` of 16 bytes, and must specify encodings for all relevant
+messages into byte strings.
```
tgeoghegan, line 647: Can we enumerate what these messages are? It's input shares, prepare shares and aggregate shares, right?

branlwyd: We can, with the downside that we're now on the hook to keep this list up-to-date. IMO, we might be better served by leaving the list ambiguous.

Note that the last two types (`PrepShare` and `PrepMessage`) require an encoding because of their inclusion in ping-pong messages, the serialization of which is specified in VDAF (which itself assumes that these types have a defined encoding). I think it's less confusing to list the specific VDAF types, since that is what someone trying to implement DAP would be able to reference when looking at any given VDAF.

branlwyd: Any thoughts on mentioning encodability of `PrepState`, either in this list or separately, given the considerations I mention above?

tgeoghegan: > I think it's less confusing to list the specific VDAF types, since that is what someone trying to implement DAP would be able to reference when looking at any given VDAF.

Yep, I think this is the right trade-off.

> Any thoughts on mentioning encodability of `PrepState`, either in this list or separately, given the considerations I mention above?

I don't think we need to discuss it in this protocol. I agree that any practical DAP implementation will need to be able to somehow encode and store `PrepState`, but since that representation doesn't have to be agreed upon by more than one implementation, this document doesn't need to be prescriptive about it.

tgeoghegan (approved):

draft-ietf-ppm-dap.md
divergentdave: Note that we need an encoding for the aggregation parameter as well, in collection requests and responses.

branlwyd: Gah, thanks, I mistakenly conflated AggParam with AggShare when I was generating the list.

draft-ietf-ppm-dap.md
```diff
@@ -639,9 +639,19 @@ struct {
...
 ~~~
 
-DAP uses the 16-byte `ReportID` as the nonce parameter for the VDAF
-`shard` and `prep_init` methods (see {{!VDAF, Section 5}}). Thus for a VDAF to
-be compatible with DAP, it MUST specify a `NONCE_SIZE` of 16 bytes.
+DAP uses the 16-byte `ReportID` as the nonce parameter for the VDAF `shard` and
+`prep_init` methods (see {{!VDAF, Section 5}}). Additionally, DAP includes
+messages defined in the VDAF specification encoded as opaque byte strings within
+various DAP messages. Thus, for a VDAF to be compatible with DAP, it MUST
+specify a `NONCE_SIZE` of 16 bytes, and must specify encodings for the following
```
cjpatton, line 646:
```suggestion
specify a `NONCE_SIZE` of 16 bytes, and MUST specify encodings for the following
```

cjpatton (approved):

PR #530, opened 2023-11-21
Make max batch size optional

wangshan:
As discussed in ietf118.

Not all applications require batch size to be tightly controlled within [min, max] batch size. For some applications, satisfy min batch size is sufficient. Making max batch size optional also allows `fixed_size` query to mimic `time_interval` query type.

Note the name `fixed_size` isn't strictly true anymore without max batch size, but I think it still conveys the right information: The aggregator should check one or two fixed batch size before outputting, therefore I opt to keep the current name.


draft-ietf-ppm-dap.md
```diff
@@ -754,16 +754,25 @@ it is interested in; in this case, it can also issue a query of type
 The Leader SHOULD select a batch which has not yet began collection.
 
 In addition to the minimum batch size common to all query types, the
-configuration includes a parameter `max_batch_size` that determines maximum
+configuration may includes a parameter `max_batch_size` that determines maximum
```
cjpatton, line 757:
```suggestion
configuration MAY include a parameter `max_batch_size` that determines maximum
```

draft-ietf-ppm-dap.md
```diff
@@ -2285,9 +2297,10 @@ ID:
 ##### Size Check
 
 The query configuration specifies the minimum batch size, `min_batch_size`, and
-maximum batch size, `max_batch_size`. The Aggregator checks that `len(X) >=
-min_batch_size` and `len(X) <= max_batch_size`, where `X` is the set of reports
-successfully aggregated into the batch.
+optionally maximum batch size, `max_batch_size`. The Aggregator checks that
```
cjpatton, line 2300:
```suggestion
optionally the maximum batch size, `max_batch_size`. The Aggregator checks that
```

cjpatton (approved):
Looks good to me.

Question, while at it: How do you think we should represent this optional information in taskprov?

chris-wood (approved):

Issue #526, opened 2023-11-16
Potential simplification of current-batch collection request semantics


branlwyd:
In the fixed-size query type, `current-batch` collection requests allow the Leader to choose an outstanding batch to associate with the request.

The current semantics are that the Leader can associate the same batch to multiple `current-batch` collection requests (DAP-07 4.1.2):

```
The Collector may not know which batch ID it is interested in; in this case, it can also issue a query of type current_batch, which allows the Leader to select a recent batch to aggregate. The Leader SHOULD select a batch which has not yet began collection.
```

The reason these semantics were chosen was to avoid data loss in the case that a Collector issued a collection request, then crashed before recording the collection job ID. Janus, for example, can associate the same batch to an arbitrary number of `current-batch` collection requests, until at least one of those collection requests is polled for the first time.

===

However, since the above semantics were chosen, DAP changed such that the Collector now determines the collection job ID itself (as part of the changes for the resource-oriented API). This means that we could similarly avoid data loss if we expect the Collector to durably store the collection job ID _before_ making the collection request for that ID, and lean on either idempotency or appropriate error codes to allow recovery in the face of process failure. This would allow aggregator implementations to associate each batch to exactly one `current-batch` request.

The upside of making this change is that Collectors would no longer need to deduplicate `current-batch` requests that happen to be mapped to the same batch. Along with the simplified aggregator semantics, I think this is pretty likely to be an overall complexity win.

branlwyd:
Thinking about what this would look in terms of textual change to the spec: I think we would just change "The Leader SHOULD select a batch which has not yet began collection." from a `SHOULD` to a `MUST`.

We could optionally choose to provide an implementation advice that the Collector note the collection job ID to durable storage before sending a collection request to avoid data loss.

cjpatton:
SGTM, please send a PR :)

Issue #450, opened 2023-04-28
Use POST instead of PUT


ekr:
RFC 9110 says:

> The fundamental difference between the POST and PUT methods is highlighted by the different intent for the enclosed representation. The target resource in a POST request is intended to handle the enclosed representation according to the resource's own semantics, whereas the enclosed representation in a PUT request is defined as replacing the state of the target resource. Hence, the intent of PUT is idempotent and visible to intermediaries, even though the exact effect is only known by the origin server.

As we are not replacing the resource at: `{leader}/tasks/{task-id}/reports` we should use `POST`

cjpatton:
I don't have a well-formed view point here. I think from a protocol flow perspective the main thing we need is to be able to distinguish the initial message for an aggregation job (or collection job) with a continue (or poll) message. I think we already get this from media types but I'll check if/when we put a PR for this.

cc/ @tgeoghegan

ekr:
Right. I think this generally should not be encoded in methods.

@martinthomson @mnot?

mnot:
Context?

tgeoghegan:
The semantic I focused on when I selected PUT for these requests is idempotence. I did consider doing the REST thing of transferring the entire representation of the resources in requests and responses, but concluded that this is quite wasteful in the aggregation sub-protocol: there's no reason for the helper to echo report shares and aggregation parameters back to the leader.

That said, my main takeaway from having spent time on this during DAP-04 is that it doesn't _really_ matter what methods we use because we don't expect DAP to be executed by a generic HTTP client (e.g. a web browser) but rather by dedicated DAP implementations. That means that it's not the end of the world if play a little fast and loose with method semantics.

Long story short: I don't object to using different methods here.

martinthomson:
If what you are doing is asking for a change to a resource, rather than its replacement PATCH works over PUT.  Idempotence is not guaranteed in that case, but it could be, depending on how you structure the request.  The same applies to POST, which could be idempotent and therefore safe to retry with that knowledge.

divergentdave:
One alternate resolution: we could keep using the `PUT` method, but change the URL to include the report ID (rather than it being part of the request body), i.e. `PUT /{leader}/tasks/{task-id}/reports/{report-id}`. (This was [previously considered](https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/pull/398#discussion_r1081893133) during the recent HTTP API redesign, but we ultimately reconsidered)

[...]

PR #532, opened 2023-11-27
Require current-batch queries to return distinct batches.

branlwyd:
Previously, the Leader could assign the same batch to an arbitrary number of current-batch queries, which required the Collector to dedupe. Now, each batch is assigned to exactly one current-batch collection request.

Closes https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/issues/526.

### DIFF
--- a/draft-ietf-ppm-dap.md
+++ b/draft-ietf-ppm-dap.md
@@ -781,7 +781,8 @@ To get the aggregate of a batch, the Collector issues a query specifying the
 batch ID of interest (see {{query}}). The Collector may not know which batch ID
 it is interested in; in this case, it can also issue a query of type
 `current_batch`, which allows the Leader to select a recent batch to aggregate.
-The Leader SHOULD select a batch which has not yet began collection.
+The Leader MUST select a batch that has not yet been associated with a
+collection job.
 
 In addition to the minimum batch size common to all query types, the
 configuration may include a parameter `max_batch_size` that determines maximum
@@ -1977,10 +1978,10 @@ Changing a collection job's parameters is illegal, so further requests to
 with an HTTP client error status code.
 
 After receiving the response to its `CollectionReq`, the Collector makes an HTTP
-`POST` request to the collection job URI to check on the status of the collect
-job and eventually obtain the result. If the collection job is not finished
-yet, the Leader responds with HTTP status 202 Accepted. The response MAY include
-a Retry-After header field to suggest a polling interval to the Collector.
+POST request to the collection job URI to check on the status of the collect job
+and eventually obtain the result. If the collection job is not finished yet, the
+Leader responds with HTTP status 202 Accepted. The response MAY include a
+Retry-After header field to suggest a polling interval to the Collector.
 
 Asynchronously from any request from the Collector, the Leader attempts to run
 the collection job. It first checks whether it can construct a batch for the
@@ -2042,20 +2043,6 @@ collection job if the results have been deleted.
 The Collector can send an HTTP DELETE request to the collection job, which
 indicates to the Leader that it can abandon the collection job and discard all
 state related to it.
-
-#### A Note on Idempotence
-
-The reason a POST is used to poll the state of a collection job instead of a
-GET is because of the fixed-size query mode (see {{fixed-size-query}}).
-Collectors may make a query against the current batch, and it is the Leader's
-responsibility to keep track of what batch is current for some task. Polling a
-collection job is the only point at which it is safe for the Leader to change
-its set of current batches, since it constitutes acknowledgement on the
-Collector's part that it received the response to some previous PUT request to
-the collection jobs resource.
-
-This means that polling a collection job can have the side effect of changing
-the set of current batches in the Leader, and thus using a GET is inappropriate.
 
 ### Obtaining Aggregate Shares {#collect-aggregate}
 


draft-ietf-ppm-dap.md
```diff
@@ -751,7 +751,13 @@ To get the aggregate of a batch, the Collector issues a query specifying the
 batch ID of interest (see {{query}}). The Collector may not know which batch ID
 it is interested in; in this case, it can also issue a query of type
 `current_batch`, which allows the Leader to select a recent batch to aggregate.
-The Leader SHOULD select a batch which has not yet began collection.
+The Leader MUST select a batch which has not yet began collection.
+
+It is RECOMMENDED that Collectors which use `current_batch` queries generate &
```
cjpatton, line 756:
```suggestion
It is RECOMMENDED that Collectors which use `current_batch` queries generate and
```

draft-ietf-ppm-dap.md
```diff
@@ -751,7 +751,13 @@ To get the aggregate of a batch, the Collector issues a query specifying the
...
+store collection job IDs as durably as they will store the eventual collection
```
cjpatton, line 757: "as durably"?

draft-ietf-ppm-dap.md
```diff
@@ -751,7 +751,13 @@ To get the aggregate of a batch, the Collector issues a query specifying the
...
+results before issuing a collection request. (The reason for this recommendation
+is that a Collector which issues a `current_batch` request and then loses the
```
cjpatton, line 759:
```suggestion
is that a Collector that issues a `current_batch` request and then loses the
```

draft-ietf-ppm-dap.md
```diff
@@ -751,7 +751,13 @@ To get the aggregate of a batch, the Collector issues a query specifying the
 batch ID of interest (see {{query}}). The Collector may not know which batch ID
 it is interested in; in this case, it can also issue a query of type
 `current_batch`, which allows the Leader to select a recent batch to aggregate.
-The Leader SHOULD select a batch which has not yet began collection.
+The Leader MUST select a batch which has not yet began collection.
+
```
cjpatton, line 755: Why RECOMMEND and not MUST? What are you trying to get implementations to do here?

cjpatton (commented):
I'd be OK with changing the POST to a GET, but we might err on the side of not messing with HTTP stuff until we've had a chance to do a thorough review (https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/issues/450#issuecomment-1815711315)

draft-ietf-ppm-dap.md
branlwyd: See the RECOMMEND vs MUST thread for more detail: what I'm trying to get at is that a collection job ID, even for an in-process collection job, is equally "important" as the eventual results from that collection job. Specifically, "important" in the sense that losing either would lose the results for the batch selected by the query, since in a `current-batch` query each batch is (now) assigned to exactly one collection job.

I use the term "durably" here in the sense of "durable storage", e.g. to suggest that the collection job ID be stored to the same datastore as the eventual collection result (as opposed to being stored in memory only, I suppose). Maybe there's a better way to phrase this?

branlwyd (commented):
> I'd be OK with changing the POST to a GET, but we might err on the side of not messing with HTTP stuff until we've had a chance to do a thorough review ([#450 (comment)](https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/issues/450#issuecomment-1815711315))

Ack, I've switched back to POST for now. I'll note the GET/POST question on that issue.

draft-ietf-ppm-dap.md
cjpatton: My chief complaint is that it's grammatically awkward. (When I get the tokens "as durably as", the next token I predict is "possible".)

draft-ietf-ppm-dap.md
cjpatton: I see. My gut feeling is that this paragraph is not really necessary: since a collection job is a resource associated with an ID, it seems obvious that you would have to remember that ID in order to complete the job later on.

Am I missing anything? Perhaps you encountered a bug in Janus that made you feel this text was needed?

cjpatton (approved):
Thanks for taking out the breaking change. I have no objection to merging this as-is.

draft-ietf-ppm-dap.md
```diff
@@ -751,7 +751,7 @@ To get the aggregate of a batch, the Collector issues a query specifying the
 batch ID of interest (see {{query}}). The Collector may not know which batch ID
 it is interested in; in this case, it can also issue a query of type
 `current_batch`, which allows the Leader to select a recent batch to aggregate.
-The Leader SHOULD select a batch which has not yet began collection.
+The Leader MUST select a batch that has not yet began collection.
```
tgeoghegan, line 754: Should this be "begun" instead of "began"?

More substantialy: are we being sufficiently clear about what it means for a batch to have begun collection? I think that once the leader has successfully handled `PUT /tasks/{task-id}/collection_jobs/{collection-job-id}` (regardless of whether the collector ever saw the response), the batch has "beg[u]n collection", but I'm not sure if we say that. Maybe it'd be better to say "a batch for which a collection job exists", because I think that's easier to understand in terms of the relation between a batch and a collection job resource?

tgeoghegan (commented):
In response to the earlier discussion on what happens if the collector loses collection job IDs: another solution besides requiring the collector to remember collection job IDs would be a leader endpoint `GET /tasks/{task-id}/collection_jobs` that would return a (paginated?) list of collection jobs for the task. I'm not sure if we want to go so far as to make this a MUST for implementations, but maybe it could go in `{{operational-capabilities}}`? Actually, discussion of collectors durably storing collection job IDs could go in "Operational Considerations/Collector capabilities".

draft-ietf-ppm-dap.md
branlwyd: I think you're right that it's "begun" rather than "began".

I updated the text to clarify that the selected batch "has not yet been associated with a collection job", which I think clarifies any ambiguity. I think that an aggregator implementation can choose when to associate a given `current-batch` collection request to a specific batch -- e.g. it might immediately do so as part of the `PUT` request handling, or it might choose to do so later -- but this text handles both of those choices.

On a side note: I think it's worth calling out that this leaves open whether we need to care about aggregation parameters here. That is, should the semantics be "the aggregator must select a batch which has not yet begun collection _for any aggregation parameter_" (which are the semantics we have as-written) or "not yet begun collection for the aggregation parameter in the collection request". IMO: I'm OK with going with any aggregation parameter, as this works just fine for all existing uses of DAP & the other choice is harder to implement and harder to understand.

branlwyd (commented):
> In response to the earlier discussion on what happens if the collector loses collection job IDs: another solution besides requiring the collector to remember collection job IDs would be a leader endpoint `GET /tasks/{task-id}/collection_jobs` that would return a (paginated?) list of collection jobs for the task. I'm not sure if we want to go so far as to make this a MUST for implementations, but maybe it could go in `{{operational-capabilities}}`?

I think this API is "a list of _uncollected_ collection jobs which are ready for collection", right? Otherwise we end up with unbounded growth in the result set (up to GC).

IMO: this idea is nice in that it potentially removes statefulness in Collector implementations, at the cost of adding a fair bit of complexity (& request traffic) to Aggregator implementations. One downside is that it's a lot of additional API surface to specify.

If we wanted to permit stateless Collector implementations, this would be one way to do so. OTOH, it's plausible that Collector implementations will need to store state anyway for implementation-specific reasons. (For example, even with the proposed list-of-collection-jobs API, if there are several independent Collector processes running they will need some form of coordination to avoid repeatedly collecting the same batch; without coordination, there is a race condition s.t. two different Collector processes list collection jobs & decide to collect the same job.)

But ultimately, I think this idea is orthogonal to this PR. Let's either file an issue or just write a separate PR for this, if we think it's a good idea?

draft-ietf-ppm-dap.md
tgeoghegan: I think it's got to be "for any aggregation parameter", because otherwise that allows a leader to return the same batch forever.

PR #533, opened 2023-11-29
Revise security considerations section.

branlwyd:
Mostly, I condense the information in the threat model section into the top-level security considerations section. I also moved a few sections around, eliminated a couple of no-longer-relevant sections, and gave the entire security considerations section a basic editorial pass.

### DIFF
--- a/draft-ietf-ppm-dap.md
+++ b/draft-ietf-ppm-dap.md
@@ -2329,13 +2329,13 @@ The DAP protocol has inherent constraints derived from the tradeoff between
 privacy guarantees and computational complexity. These tradeoffs influence how
 applications may choose to utilize services implementing the specification.
 
-## Protocol participant capabilities {#entity-capabilities}
+## Protocol Participant Capabilities {#entity-capabilities}
 
 The design in this document has different assumptions and requirements for
 different protocol participants, including Clients, Aggregators, and Collectors.
 This section describes these capabilities in more detail.
 
-### Client capabilities
+### Client Capabilities
 
 Clients have limited capabilities and requirements. Their only inputs to the
 protocol are (1) the parameters configured out of band and (2) a measurement.
@@ -2344,7 +2344,7 @@ they required to implement any sort of report upload retry mechanism. By design,
 the protocol in this document is robust against individual Client upload
 failures since the protocol output is an aggregate over all inputs.
 
-### Aggregator capabilities
+### Aggregator Capabilities
 
 Leaders and Helpers have different operational requirements. The design in this
 document assumes an operationally competent Leader, i.e., one that has no
@@ -2378,7 +2378,7 @@ In addition, for each DAP task, the Leader is required to:
 - Implement and store state for the form of inter- and intra-batch replay
   mitigation in {{input-share-validation}}.
 
-### Collector capabilities
+### Collector Capabilities
 
 Collectors statefully interact with Aggregators to produce an aggregate output.
 Their input to the protocol is the task parameters, configured out of band,
@@ -2389,24 +2389,40 @@ to the end as needed to produce the final aggregate output.
 Collectors must also maintain state for the lifetime of each task, which
 includes key material associated with the HPKE key configuration.
 
-## Data resolution limitations
+## VDAFs and Compute Requirements
 
-Privacy comes at the cost of computational complexity. While affine-aggregatable
-encodings (AFEs) can compute many useful statistics, they require more bandwidth
-and CPU cycles to account for finite-field arithmetic during input-validation.
-The increased work from verifying inputs decreases the throughput of the system
-or the inputs processed per unit time. Throughput is related to the verification
-circuit's complexity and the available compute-time to each Aggregator.
+The choice of VDAF can impact the computation required for a DAP task. For
+instance, the Poplar1 VDAF {{!VDAF}} when configured to compute a set of heavy
+hitters requires each measurement to be of the same bit-length which all parties
+need to agree on prior to VDAF execution. The computation required for such
+tasks increases superlinearly as a function of the chosen bit-length, as
+multiple rounds of collection are needed for each bit of the measurement value,
+and each bit of the measurement value requires computation which scales with the
+chosen bit-length.
 
-Applications that utilize proofs with a large number of multiplication gates or
-a high frequency of inputs may need to limit inputs into the system to meet
-bandwidth or compute constraints. Some methods of overcoming these limitations
-include choosing a better representation for the data or introducing sampling
-into the data collection methodology.
+When dealing with variable length measurements (e.g domain names), it is
+necessary to pad them to convert into fixed-length measurements. When computing
+the heavy hitters from a batch of such measurements, VDAF execution can be
+terminated early once once the padding region is reached for a given
+measurement: at this point, the padded measurement can be included in the last
+set of candidate prefixes. For smaller length measurements, this significantly
+reduces the cost of communication between Aggregators and the steps required for
+the computation. However, malicious Clients can still generate maximum length
+measurements forcing the system to always operate at worst-case performance.
+
+Therefore, care must be taken that a DAP deployment can handle VDAF execution of
+all possible measurement values for any tasks that the deployment is configured
+for. Otherwise, an attacker may deny service by uploading many expensive
+reports.
+
+Applications which require computationally-expensive VDAFs can mitigate the
+computation cost of aggregation in a few ways, such as producing aggregates over
+a sample of the data or choosing a representation of the data permitting a
+simpler aggregation scheme.
 
 [[TODO: Discuss explicit key performance indicators, here or elsewhere.]]
 
-## Aggregation utility and soft batch deadlines
+## Aggregation Utility and Soft Batch Deadlines
 
 A soft real-time system should produce a response within a deadline to be
 useful. This constraint may be relevant when the value of an aggregate decreases
@@ -2424,12 +2440,12 @@ privacy threshold defined by a minimum number of reports. One possible solution
 is to increase the reporting period so more samples can be collected, balanced
 against the urgency of responding to a soft deadline.
 
-## Protocol-specific optimizations
+## Protocol-specific Optimizations
 
 Not all DAP tasks have the same operational requirements, so the protocol is
 designed to allow implementations to reduce operational costs in certain cases.
 
-### Reducing storage requirements
+### Reducing Storage Requirements
 
 In general, the Aggregators are required to keep state for tasks and all valid
 reports for as long as collect requests can be made for them. In particular,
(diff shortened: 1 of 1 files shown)


draft-ietf-ppm-dap.md
```diff
@@ -2436,195 +2436,55 @@ suite:
...
-   * Clients can ensure that aggregate counts are non-sensitive by generating
-     reports independently of user behavior (see {{network-attacker}}.
-   * Clients, especially in deployments that cannot schedule report uploads at a
-     fixed time (e.g., an application that does not run persistently) can also
-     apply local differential privacy to measurements before constructing
-     reports.
+DAP aims to achieve the privacy and robustness security goals defined in
+{{Section 9 of !VDAF}}, even in the presence of an active attacker. It is
+assumed that the attacker may control the network and have the ability to
+control a subset of of Clients, one of the Aggregators, and the Collector for a
```
branlwyd, line 2442: Please note the change here -- I'm not sure it's useful to define an adversary who can arbitrarily control all of the parties. For example, under the old scenario, the attacker knows the Leader's TLS private key as well as both aggregators' HPKE private keys (since it knows the "secret state of any party"), and is in control of the network, and as such can MitM the Client<->Leader connections then decrypt each measurement as it arrives.

We could call this out as another item outside of our threat model, but I think it makes more sense to restrict attacker control to at most one aggregator -- DAP & VDAF provide no guarantees if the aggregators collude.

draft-ietf-ppm-dap.md
```diff
@@ -2680,35 +2542,53 @@ service has no obvious way to authenticate its report uploads.
...
+of the participants in the DAP protocol.
+
+## Differential privacy {#dp}
+
+DAP deployments can choose to ensure their aggregate results achieve
+differential privacy [Vad16]. A simple approach would require the Aggregators to
+add two-sided noise (e.g. sampled from a two-sided geometric distribution) to
+aggregate shares. Since each Aggregator is adding noise independently, privacy
+can be guaranteed even if all but one of the Aggregators is malicious.
+Differential privacy is a strong privacy definition, and protects users in
```
branlwyd, line 2561: Is this sentence accurate/not overstating things? If I know a single measurement-with-noise value, and the distribution used for noise, I can determine a probability distribution for the measurement-without-noise value. My understanding is that noise values are typically "small", so this distribution would determine a small likely range for the true measurement value.

draft-ietf-ppm-dap.md
```diff
@@ -2731,90 +2611,65 @@ chosen independently of the generated reports.
...
-measurement in a batch except for one, that one record is still formally
-protected.
+In order to execute a DAP task, it is necessary for all parties to ensure they
+agree on the configuration of the task. However, it is possible for a party to
+participate in the execution of DAP without knowing all of the task's
+parameters. For example, a Client can upload a report ({{upload-flow}}) without
+knowing the minimum batch size that is enforced by the Aggregators during
+collection ({{collect-flow}}).
 
-## Robustness in the presence of malicious servers
```
branlwyd, line 2777: I removed this section as the relevant issue is closed, and also because we now specify to two aggregators so aggregator subset selection no longer makes sense. (I think something like this idea could work if we had 3+ independent DAP aggregator deployments, sent each report to all deployments, and collected from all of them -- but IMO that's outside of the scope of what needs to be mentioned in the DAP spec.)

draft-ietf-ppm-dap.md
```diff
@@ -2436,195 +2436,55 @@ suite:
...
-has, and potential mitigations for attacks enabled by those capabilities.
-
-This model assumes that all participants have previously agreed upon and
-exchanged all shared parameters over some unspecified secure channel.
-
-### Client/user
-
-#### Assets
-
-1. Unsharded measurements. Clients are the only actor that can ever see the
```
branlwyd, line 2476: Note to self: move assets to an appendix (unless I figure out a clever way to keep it in the main text).

draft-ietf-ppm-dap.md
```diff
@@ -2436,195 +2436,55 @@ suite:
...
+   measurement value such that the existence of a report does not imply the
+   occurrence of a sensitive event.
+1. Aggregators can deny service by refusing to respond to collection requests or
+   aggregate share requests.
+1. Collection requests may leak information beyond the collection results. For
+   example, the Poplar1 VDAF {{!VDAF}} can be used to compute the set of heavy
+   hitters among a set of arbitrary bit strings uploaded by Clients. This
+   requires multiple evaluations of the VDAF, the results of which reveal
+   information to the Aggregators and Collector beyond what follows from the
+   heavy hitters themselves. Or the result of the Prio3Sum VDAF could leak
```
cjpatton, line 2481: Not specific to any VDAF. I think Prio3Sum is meant only as an example, but calling it out specifically may make a casual reader think it's more vulnerable than other VDAFs.
```suggestion
   heavy hitters themselves. Or the result could leak
```

draft-ietf-ppm-dap.md
```diff
@@ -2436,195 +2436,55 @@ suite:
...
+   information about outlier values. This leakage can be mitigated using
+   differential privacy ({{dp}}).
 
 [[OPEN ISSUE: Should we have authentication in either direction between the
 Helper and the Collector? #155]]
```
cjpatton, line 2486: This issue has been closed, so let's delete this paragraph.

draft-ietf-ppm-dap.md
```diff
@@ -2436,195 +2436,55 @@ suite:
...
-     regular intervals and independently of whether the event that the task is
-     tracking has not occurred, so that the absence of reports cannot be
-     distinguished from their presence.
-1. Tampering with network traffic. Attackers may drop messages or inject new
-   messages into communications between participants.
-   * DAP mitigates this by using standard HTTP semantics to allow requests to be
-     retried. However attacks that completely deny network access to
-     participants are outside of DAP's scope.
-
 [[OPEN ISSUE: The threat model for Prio --- as it's described in the original
```
cjpatton, line 2488: This open issue is redundant and can be removed. We also plan to call out this threat model in the VDAF draft: https://github.com/cfrg/draft-irtf-cfrg-vdaf/issues/304

draft-ietf-ppm-dap.md
```diff
@@ -2636,14 +2496,14 @@ they're practical.]]
...
-be measuring the heights of a human population and configure a VDAF to prove
-that measurements are values in the range of 80-250 cm. A malicious Client would not
-be able to claim a height of 400 cm, but they could submit multiple bogus
-reports inside the acceptable range, which would yield incorrect averages. More
-generally, DAP deployments are susceptible to Sybil attacks {{Dou02}}, especially
-when carried out by the Leader.
+Several attacks on privacy or robustness involve malicious Clients uploading
+reports that are valid under the chosen VDAF but incorrect. For example, a DAP
+deployment might be measuring the heights of a human population and configure a
+VDAF to prove that measurements are values in the range of 80-250 cm. A
```
cjpatton, line 2502:
```suggestion
deployment might be measuring the heights of a human population and configure a
variant of Prio3 to prove that measurements are values in the range of 80-250 cm. A
```

draft-ietf-ppm-dap.md
```diff
@@ -2636,14 +2496,14 @@ they're practical.]]
...
-Several attacks on privacy involve malicious clients uploading reports that are
-valid under the chosen VDAF but incorrect. For example, a DAP deployment might
-be measuring the heights of a human population and configure a VDAF to prove
-that measurements are values in the range of 80-250 cm. A malicious Client would not
-be able to claim a height of 400 cm, but they could submit multiple bogus
-reports inside the acceptable range, which would yield incorrect averages. More
-generally, DAP deployments are susceptible to Sybil attacks {{Dou02}}, especially
-when carried out by the Leader.
+Several attacks on privacy or robustness involve malicious Clients uploading
+reports that are valid under the chosen VDAF but incorrect. For example, a DAP
```
cjpatton, line 2500:
```suggestion
Several attacks on robustness or privacy involve malicious Clients uploading
reports that are valid under the chosen VDAF but incorrect.

For example, a DAP
```

draft-ietf-ppm-dap.md
```diff
@@ -2653,24 +2513,26 @@ reveal additional information about the honest measurements, leading to a
...
-Depending on the deployment and threat model, there are different ways to address
-Sybil attacks, including, though not limited to:
+Depending on the deployment and the specific threat being mitigated, there are
+different ways to address Sybil attacks, such as:
 
-1. Authenticating clients, as described in {{client-auth}};
-1. Removing client-specific metadata on individual reports, such as through the
-   use of anonymizing proxies in the upload flow, as described in {{anon-proxy}}; or
+1. Implementing Client authentication, as described in {{client-auth}}, likely
+   paired with rate-limiting uploads from individual Clients.
```
cjpatton, line 2520: We should reference the privacy pass draft: https://datatracker.ietf.org/doc/draft-ietf-privacypass-architecture/

It's in WGLC and shouldn't be a blocker for us.

draft-ietf-ppm-dap.md
```diff
@@ -2680,35 +2542,53 @@ service has no obvious way to authenticate its report uploads.
...
+information can be removed by having Clients submit reports to an anonymizing
+proxy server which would then use Oblivious HTTP
+{{!I-D.draft-ietf-ohai-ohttp-08}} to forward reports to the DAP Leader. In this
+scenario, Client authentication would be performed by the proxy rather than any
+of the participants in the DAP protocol.
+
+## Differential privacy {#dp}
+
+DAP deployments can choose to ensure their aggregate results achieve
+differential privacy [Vad16]. A simple approach would require the Aggregators to
```
cjpatton, line 2557:
```suggestion
differential privacy {{Vad16}}. A simple approach would require the Aggregators to
```

draft-ietf-ppm-dap.md
```diff
@@ -2680,35 +2542,53 @@ service has no obvious way to authenticate its report uploads.
...
+add two-sided noise (e.g. sampled from a two-sided geometric distribution) to
+aggregate shares. Since each Aggregator is adding noise independently, privacy
+can be guaranteed even if all but one of the Aggregators is malicious.
+Differential privacy is a strong privacy definition, and protects users in
+extreme circumstances: even if an adversary has prior knowledge of every
+measurement in a batch except for one, that one measurement is still formally
+protected.
+
+[[TODO: link to the Shan et al. I-D on differential privacy in DAP once it is
+published.]]
```
cjpatton, line 2567: I would probably not add this dependency because it creates a cyclic dependency. (One of these documents will become an RFC before the other.)

draft-ietf-ppm-dap.md
```diff
@@ -2653,24 +2513,26 @@ reveal additional information about the honest measurements, leading to a
...
 
-1. Authenticating clients, as described in {{client-auth}};
-1. Removing client-specific metadata on individual reports, such as through the
-   use of anonymizing proxies in the upload flow, as described in {{anon-proxy}}; or
+1. Implementing Client authentication, as described in {{client-auth}}, likely
+   paired with rate-limiting uploads from individual Clients.
+1. Removing Client-specific metadata on individual reports, such as through the
+   use of anonymizing proxies in the upload flow, as described in
+   {{anon-proxy}}.
 1. Applying differential privacy protections, as described in {{dp}}.
```
cjpatton, line 2524: We don't want to over-promise here. Whether/how DP is used to beat Sybil attacks is not straightforward.
```suggestion
1. Differential privacy ({{dp}}) can help mitigate Sybil attacks to some extent.
```

draft-ietf-ppm-dap.md
```diff
@@ -2680,35 +2542,53 @@ service has no obvious way to authenticate its report uploads.
...
-Clients submit reports to an anonymizing proxy server which would then use
-Oblivious HTTP {{!I-D.draft-ietf-ohai-ohttp-08}} to forward reports to the
-DAP Leader, without requiring any server participating in DAP to be aware of
-whatever Client authentication or attestation scheme is in use.
+agent, or Client authentication information (in deployments which use it, see
+{{client-auth}}). This metadata can be used by Aggregators to identify
+participating Clients or permit some attacks on robustness. This auxiliary
+information can be removed by having Clients submit reports to an anonymizing
+proxy server which would then use Oblivious HTTP
+{{!I-D.draft-ietf-ohai-ohttp-08}} to forward reports to the DAP Leader. In this
```
cjpatton, line 2550:
```suggestion
{{!I-D.draft-ietf-ohai-ohttp-10}} to forward reports to the DAP Leader. In this
```

draft-ietf-ppm-dap.md
```diff
@@ -2731,90 +2611,65 @@ chosen independently of the generated reports.
...
+[[TODO: Revisit this paragraph once
+https://github.com/cfrg/draft-irtf-cfrg-vdaf/issues/273 is resolved.]]
 
-Therefore, care must be taken that a DAP deployment can comfortably handle
-computation of measurements for arbitrarily large sizes, otherwise, it may
-result in a DoS possibility for the entire system.
+Therefore, care must be taken that a DAP deployment can handle VDAF execution of
+all possible measurement values for any tasks that the deployment is configured
+for. Otherwise, an attacker may deny service by uploading many expensive
+reports.
```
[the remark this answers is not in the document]
branlwyd: Hmm -- while availability/denial-of-service is often considered a security concern, I agree this section doesn't fit in well since all the other sections are more related to the robustness/privacy properties. Moved.

There was already a subsection ("Data Resolution Limitations") in the Operational Considerations section which touches on basically the same ideas; I pulled a little bit of information in from that subsection & eliminated it.

draft-ietf-ppm-dap.md
```diff
@@ -2436,195 +2436,55 @@ suite:
...
+{{Section 9 of !VDAF}}, even in the presence of an active attacker. It is
+assumed that the attacker may control the network and have the ability to
+control a subset of of Clients, one of the Aggregators, and the Collector for a
+given task.
 
-[[TODO: link to the Shan et al. I-D on differential privacy in DAP once it is
-published.]]
+In the presence of this adversary, there are some threats DAP does not defend
+against and which are considered outside of DAP's threat model. These are
+enumerated below, along with potential mitigations:
```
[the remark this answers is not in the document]
branlwyd: Sure -- I include "the client can lie, or submit multiple reports, to skew the results" in attacks on robustness. I think this is technically correct; VDAF defines robustness as "An attacker that controls the network and a subset of Clients cannot cause the Collector to compute anything other than the aggregate of the measurements of honest Clients." DAP will cause the Collector to compute the aggregate of the measurements of honest _and dishonest_ Clients.

draft-ietf-ppm-dap.md
branlwyd: IIUC, I think what I say is true for "Client DP" (noise added by client to original measurement), but not "Aggregator DP" (noise added by aggregator to aggregate share), looking to https://www.ietf.org/archive/id/draft-wang-ppm-differential-privacy-00.html.

draft-ietf-ppm-dap.md
branlwyd: Looking back to this list of assets, all of the assets are enumerated in the list of attacks. I think it's more useful to specify the assets with the context of how they can be used/compromised in an attack.

cjpatton (approved):

tgeoghegan (approved):

PR #540, opened 2024-01-02
Address some items from httpdir early review

tgeoghegan:
Mark Nottingham (@mnot) was kind enough to share early review feedback on DAP on behalf of httpdir ([1]). This PR addresses several of the items there. See individual commit messages for discussion.

[1]: https://datatracker.ietf.org/doc/review-ietf-ppm-dap-09-httpdir-early-nottingham-2023-12-29/

### DIFF
--- a/draft-ietf-ppm-dap.md
+++ b/draft-ietf-ppm-dap.md
@@ -301,8 +301,8 @@ Aggregation parameter:
   {{!VDAF}}.
 
 Aggregator:
-: An endpoint that receives input shares from Clients and validates and
-  aggregates them with the help of the other Aggregators.
+: A server that receives input shares from Clients and validates and aggregates
+  them with the help of the other Aggregators.
 
 Batch:
 : A set of reports (i.e., measurements) that are aggregated into an aggregate
@@ -319,8 +319,8 @@ Client:
 : A party that uploads a report.
 
 Collector:
-: The endpoint that selects the aggregation parameter and receives the
-  aggregate result.
+: The party that selects the aggregation parameter and receives the aggregate
+  result.
 
 Helper:
 : The Aggregator that executes the aggregation and collection sub-protocols as
@@ -434,12 +434,12 @@ Collector:
   by the Clients. Any given measurement task will have a single Collector.
 
 Client(s):
-: The endpoints which directly take the measurement(s) and report them to the
+: The parties which directly take the measurement(s) and report them to the
   DAP protocol. In order to provide reasonable levels of privacy, there must be
   a large number of Clients.
 
 Aggregator:
-: An endpoint which receives report shares. Each Aggregator works with its
+: A server which receives report shares. Each Aggregator works with its
   co-Aggregator to compute the aggregate result. Any given measurement task
   will have two Aggregators: a Leader and a Helper.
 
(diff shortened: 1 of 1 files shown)


tgeoghegan:
Draft for now -- I think I want to stack some related commits on this.

draft-ietf-ppm-dap.md
```diff
@@ -825,10 +825,10 @@ opaque TaskID[32];
 The task ID value MUST be a globally unique sequence of bytes. Each task has
 the following parameters associated with it:
 
-* `leader_aggregator_endpoint`: A URL relative to which the Leader's API
-  endpoints can be found.
-* `helper_aggregator_endpoint`: A URL relative to which the Helper's API
-  endpoints can be found.
+* `leader_aggregator_server`: A URL relative to which the Leader's API resources
```
branlwyd, line 832: I am not sure `server` is appropriate here (or for `helper_aggregator_server` below) -- I think either `{leader,helper}_aggregator_endpoint` or perhaps `{leader,helper}_aggregator_url` would be clearer. I'd expect a server to indicate a hostname, while this is a full URL including a protocol & path.

draft-ietf-ppm-dap.md
```diff
@@ -316,14 +316,18 @@ Batch interval:
   of the reports in the batch.
 
 Client:
-: A party that uploads a report.
+: DAP protocol role identifying a party that uploads a report. Note the
```
branlwyd, line 319:
```suggestion
: A DAP protocol role identifying a party that uploads a report. Note the
```

(editorial, matching structure of other items in this list)

branlwyd (approved):

draft-ietf-ppm-dap.md
branlwyd: (I'd go with "endpoint", but I think that an endpoint might also indicate a particular HTTP verb; if that's correct, I'd go with "URL.")

draft-ietf-ppm-dap.md
```diff
@@ -110,11 +110,11 @@ revealing any individual user's data.
 
 This document describes the Distributed Aggregation Protocol (DAP) for privacy
 preserving measurement. The protocol is executed by a large set of clients and a
-small set of servers. The servers' goal is to compute some aggregate statistic
-over the clients' inputs without learning the inputs themselves. This is made
-possible by distributing the computation among the servers in such a way that,
-as long as at least one of them executes the protocol honestly, no input is ever
-seen in the clear by any server.
+small set of aggregator servers. The aggregators' goal is to compute some
```
branlwyd, line 113: (This comment follows from off-PR discussion with/suggestion by @jbr.)

"small set of aggregator servers" is correct but perhaps imprecise -- nowadays, there are exactly two aggregator servers, the Leader & the Helper. Maybe "two aggregator servers, called the Leader & the Helper"?

tgeoghegan: Makes sense! I left out the discussion of leader and helper because those terms aren't used again in the introduction and they get introduced properly in the glossary.

draft-ietf-ppm-dap.md
tgeoghegan: Sure. "The" feels a bit more natural than "A" to me.

draft-ietf-ppm-dap.md
tgeoghegan: `_url` seems better since that's what the text on the right of the colon says. I want to avoid "endpoint" as the httpdir review flagged that word.

draft-ietf-ppm-dap.md
```diff
@@ -525,21 +529,21 @@ Client could not report 10^6s or -20s.
 
 # Message Transport {#message-transport}
 
-Communications between DAP participants are carried over HTTPS {{!RFC9110}}.
-HTTPS provides server authentication and confidentiality. Use of HTTPS is
-REQUIRED.
+Communications between DAP participants are carried over HTTP {{!RFC9110}}. Use
+of HTTPS {{!RFC8446}} is REQUIRED to provide server authentication and
```
cjpatton, line 533: Is this the correct reference? RFC8446 defines TLS, not HTTPS.

cjpatton (approved):

draft-ietf-ppm-dap.md
tgeoghegan: Hmm, good point. RFC 9110 obsoletes RFC 2818 "HTTP Over TLS", so maybe the reference to 9110 suffices?

PR #543, opened 2024-01-09
Use HTTP POST for report uploads

tgeoghegan:
Per httpdir early review of DAP ([1]), it's inappropriate to use PUT for a resource that cannot later be GET. Change the HTTP method for report uploads to POST, and add a note making it clear that uploads are idempotent and safe to retry.

[1]: https://datatracker.ietf.org/doc/review-ietf-ppm-dap-09-httpdir-early-nottingham-2023-12-29/

Previous description of this PR:
~For that reason, and also consistency with the aggregation job and collection job request paths, hoist the report ID out of the upload request body and into the request path. We don't want to repeat the report ID twice in one request, so `struct Report` no longer includes a `struct Metadata` and instead just puts the `time` inline in `Report`. `struct ReportMetadata` is still used in `InputShareAad` and elsewhere in DAP, so its definition is not changed, just moved farther down the document, and we explain how to construct it from the report request path and body.~

~With this change, there is now a unique URI for each report uploaded to a DAP leader. However we do not add any requirement that an Aggregator support GET requests on that path, because not all DAP implementations will necessarily store enough per-report information to be able to respond to such a request with a complete `struct Report`. Implementations are free to do that, but can also respond with HTTP 204, 404, 410 or some other error if they wish. As these are well-established HTTP semantics, DAP doesn't need to spend any more ink on explaining them.~

### DIFF
--- a/draft-ietf-ppm-dap.md
+++ b/draft-ietf-ppm-dap.md
@@ -846,6 +846,9 @@ the following parameters associated with it:
 * A unique identifier for the VDAF in use for the task, e.g., one of the VDAFs
   defined in {{Section 10 of !VDAF}}.
 
+Note that the `leader_aggregator_url` and `helper_aggregator_url` values may
+include arbitrary path components.
+
 In addition, in order to facilitate the aggregation and collect protocols, each
 of the Aggregators is configured with following parameters:
 
@@ -880,9 +883,13 @@ to the following rules:
   value MUST be encoded in its URL-safe, unpadded Base 64 representation as
   specified in {{Sections 5 and 3.2 of !RFC4648}}.
 
-For example, resource URI `{leader}/tasks/{task-id}/reports` might be expanded
-into
-`https://example.com/tasks/8BY0RzZMzxvA46_8ymhzycOB9krN-QIGYvg_RsByGec/reports`
+For example, given a helper URL "https://example.com/api/dap", task ID "f0 16 34
+47 36 4c cf 1b c0 e3 af fc ca 68 73 c9 c3 81 f6 4a cd f9 02 06 62 f8 3f 46 c0 72
+19 e7" and an aggregation job ID "95 ce da 51 e1 a9 75 23 68 b0 d9 61 f9 46 61
+28" (32 and 16 byte octet strings, represented in hexadecimal), resource URI
+`{helper}/tasks/{task-id}/aggregation_jobs/{aggregation-job-id}` would be
+expanded into
+`https://example.com/api/dap/tasks/8BY0RzZMzxvA46_8ymhzycOB9krN-QIGYvg_RsByGec/aggregation_jobs/lc7aUeGpdSNosNlh-UZhKA`.
 
 ## Uploading Reports {#upload-flow}
 
@@ -966,7 +973,7 @@ at least twice the cache lifetime in order to avoid rejecting reports.
 
 ### Upload Request
 
-Clients upload reports by using an HTTP PUT to
+Clients upload reports by using an HTTP POST to
 `{leader}/tasks/{task-id}/reports`. The payload is a `Report`, with media type
 "application/dap-report", structured as follows:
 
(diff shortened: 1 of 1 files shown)


branlwyd (commented):
I think this change is fine. However, as discussed off-PR (pointed out by @simon-friedberger, I think), switching from `PUT` to `POST` might be a smaller change. Are there downsides to using `POST`?

tgeoghegan:
I lean toward using PUT and changing the request path because:

1. It makes the report upload request path consistent with the aggregation job and collection job request paths, which include their respective unique job IDs.
1. Using PUT elegantly makes it clear that this request is idempotent and safe to retry (of course if we go to a POST we can just explain in the DAP text that it's an idempotent POST).
1. While this forces a wire-breaking change, I think we're going to do draft 10 of DAP anyway, and that will mean a wire-breaking change to update the HPKE domain separation strings. Making this change to request paths and message formats isn't zero risk, but I think it's a marginal lift for implementations if they're breaking wire compatibility anyway.

branlwyd (approved):

wangshan:
@tgeoghegan I read through the mailing list but don't get why this change is necessary? it seems specifying idempotent POST will do the same? Will we ever have a good reason to support GET a report?

draft-ietf-ppm-dap.md
```diff
@@ -1055,19 +1054,28 @@ where `pk` is the Aggregator's public key; `0x01` represents the Role of the
...
-the corresponding fields in the report. The `SealBase()` function is as
+message of type InputShareAad defined below. The `SealBase()` function is as
 specified in {{!HPKE, Section 6.1}} for the ciphersuite indicated by the HPKE
 configuration.
 
 ~~~
+struct {
+  ReportID report_id;
+  Time time;
+} ReportMetadata;
```
wangshan, line 1065: Does ReportMetadata still have meaning? would it be simpler if we simply put reportID and time in InputShareAad?

branlwyd:
For what it's worth, I think a `POST` is ultimately a smaller change that arrives at the desired behavior.

I don't see much use for a `GET` of uploaded reports, and implementations will plausibly delete report content as soon as possible to reduce storage costs (which would make `GET` not so useful).

cjpatton:
@tgeoghegan please rebase when you have a moment

draft-ietf-ppm-dap.md
tgeoghegan: `struct ReportMetadata` ends up being useful in the aggregation interaction (e.g. it goes into a `ReportShare`) so I think it's useful.

cjpatton (approved):
No objection to this change!

tgeoghegan:
This change has been up for quite a while and it seems to me that the most popular option, by dint of its simplicity, is to change the request upload to using POST without changing the request path or the request body. Accordingly I'm repurposing this PR to make that change.

cjpatton (approved):
Works for me. I think this is slightly better because it's clearer that reports aren't necessarily GETable once they are PUT (or POSTed, now).

draft-ietf-ppm-dap.md
```diff
@@ -880,9 +883,12 @@ to the following rules:
...
   specified in {{Sections 5 and 3.2 of !RFC4648}}.
 
-For example, resource URI `{leader}/tasks/{task-id}/reports` might be expanded
-into
-`https://example.com/tasks/8BY0RzZMzxvA46_8ymhzycOB9krN-QIGYvg_RsByGec/reports`
+For example, given a helper URL "https://example.com/api/dap", task ID "f0 16 34
+47 36 4c cf 1b c0 e3 af fc ca 68 73 c9 c3 81 f6 4a cd f9 02 06 62 f8 3f 46 c0 72
+19 e7" and an aggregation job ID "95 ce da 51 e1 a9 75 23 68 b0 d9 61 f9 46 61
+28" (32 and 16 byte octet strings, represented in hexadecimal), resource URI
+`{helper}/tasks/{task-id}/aggregation_jobs/{report-id}` would be expanded into
```
branlwyd, line 890:
```suggestion
`{helper}/tasks/{task-id}/aggregation_jobs/{aggregation-job-id}` would be expanded into
```

(editorial nit)

branlwyd (approved):

PR #544, opened 2024-01-10
Use GET rather than POST when polling collection jobs.

branlwyd:
This was first attempted as part of
571988ca27bc65dad3cd97544c340623330333b8, since the original reason for using POST was removed; but the switch to GET was reverted to avoid including (too many?) breaking wire changes.

Now that we are applying feedback from the HTTP directorate, we'll have a number of similar breaking changes in the next DAP draft. I think this is a good time to include this change.


tgeoghegan (approved):
Yes, I think this is the right time for this change. Thanks for remembering it.

cjpatton (approved):

tgeoghegan:
@branlwyd I think we have enough consensus to take this change now. Anything else you want to do before we merge?

Issue #541, opened 2024-01-08
Versioning PPM media types


suman-ganta:
PPM prescribed media types do not have version component in them. As DAP protocol progressing, the structures are changing in a non-compatible way. In order to consistently decode those structures, we need an indicator on what structure to use to decode them with.

I propose adding a MIME parameter, `version`, to express the version of the payload.

Ex: To represent Report structure of protocol version DAP-09, use the media type - `application/dap-report;version=09`

Leader/Helper need to rely on this MIME parameter to consistently parse the payload received, and should also be used to reject unsupported versions.

suman-ganta:
Agreed on path based ways to distinguish versions as an alternative.

But, given that we are defining a dedicated mime type (`application/dap-report`), and not using a generic mime type (`application/octet-stream`), the type should also indicate how to decode it. The encoding is purely based on a specific version and the type information is not part of the stream. So for the reasons of making it self describable as much as possible, I think it is important to express version there.

* It is not a problem for textual structured payloads such as json (`application/json`) since the structure has delimiters.
* It is not an issue with protobuf (`application/protobuf`) since the message encapsulates the field boundaries.

cjpatton:
@suman-ganta please prepare a PR when you have a moment.

mnot:
My .02 - it's fine to add a version parameter (or something else) for draft development purposes. When it ships, however, there shouldn't be any need for a version; the media type _is_ the version identifier.

mnot:
If it's a new major version, it needs a new media type.

suman-ganta:
> If it's a new major version, it needs a new media type.

That makes sense. But the current media types do not have any version embedded in them, So we need one of the following:

Option-1:
Add version to the existing mime types. Which means, change current report mime type from `application/dap-report` to `application/dap-report_01`.
Similar naming will be followed for major/incompatible revisions.

Option-2:
maintain a mandatory MIME parameter (version=) that points to a semver of the entity.

cjpatton:
It seems like a safe  bet that this document will be "DAP v1" and a future document with be "DAP v2". What about something like `application/dap1-report`?

mnot:
Remember that they're opaque strings; the first version could be `application/dap-report` and the second could be `application/dap2-report` or even `application/dap-results`.

wangshan:
I'd prefer we keep the first version as `application/dap-report` and the implementation can optionally add `;version=0x`.
[...]

PR #547, opened 2024-02-17
DAP Version info in mime types

suman-ganta:
This adds support for expressing dap versions in the mime types. This addresses the issue #541


cjpatton (commented):
The idea here is to version media types following "semver2", presumably as described in https://semver.org/. This seems like a reasonable way to address the issue, but we'll need two bits of information:

1. How do we construct media types with versions? Describe how the string is formatted.
2. What is the version of the current draft?

PR #550, opened 2024-02-28
Recommend batches above minimum size for Poplar1

### LINKED ISSUE #548: Recommend against min_batch_size = max_batch_size when aggregation parameters are non-trivial
It is possible that a report may pass VDAF preparation with one aggregation parameter, and later fail when prepared with another aggregation parameter. (for example, if a later correction word is corrupted in an otherwise-honestly generated Poplar1 report) We require that the same set of reports in any batch be used as input for aggregation with each aggregation parameter. The batch validation size check enforces that the number of successfully aggregated reports is at least `min_batch_size`. Therefore, if a deployment were to set `min_batch_size` equal to `max_batch_size`, a batch that has already been successfully collected with one aggregation parameter, and contains at least one such report, would be un-collectable with some other aggregation parameters, resulting in a loss of data, and an efficient DoS vector.

I think we should recommend that `max_batch_size` be strictly greater than `min_batch_size` for VDAFs that make use of the aggregation parameter. The other solution I can think of would be to somehow allow combining previously-collected batches (while still not allowing subdivision of batches) for subsequent collection with different aggregation parameters, but that would be a lot of complexity.

divergentdave:
This PR addresses the possibility that reports could be successfully prepared with a first aggregation parameter, but fail preparation with a subsequent aggregation parameter. A suggestion is added to increase actual batch size above the minimum batch size. The description of the fixed size query type is loosened -- the goal is no longer to create batches of the minimum batch size, but to create batches of a deployment-specific target batch size.

This closes #548.

### DIFF
--- a/draft-ietf-ppm-dap.md
+++ b/draft-ietf-ppm-dap.md
@@ -750,8 +750,8 @@ types have the following configuration parameters in common:
   see {{upload-flow}}. Additional semantics may apply, depending on the query
   type. (See {{batch-validation}} for details.)
 
-The parameters pertaining to specific query types are described in the relevant
-subsection below.
+The parameters pertaining to specific query types are described in
+{{time-interval-query}} and {{fixed-size-query}}.
 
 ### Time-interval Queries {#time-interval-query}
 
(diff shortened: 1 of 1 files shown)


draft-ietf-ppm-dap.md
```diff
@@ -804,19 +804,42 @@ the privacy guarantee is not affected by the sampling rate of the population,
...
-`min_batch_size == max_batch_size`. We should know better once we've had some
-implementation/deployment experience.]
+Implementation note: The goal for the Aggregators is to aggregate the same
+number of reports in each batch. The target batch size is deployment-specific,
+and may be equal to or greater than the minimum batch size. Deciding how soon
+batches should be output is also deployment-specific. Exactly sizing batches may
+be challenging for Leader deployments in which multiple, independent nodes
+running the aggregate interaction (see {{aggregate-flow}}) need to be
+coordinated. The difference between the minimum batch size and maximum batch
+size is in part intended to allow room for error, and allow a range of target
```
cjpatton, line 814:
```suggestion
size is in part intended to allow room for rejecting some number of invalid reports, and allow a range of target
```

draft-ietf-ppm-dap.md
```diff
@@ -804,19 +804,42 @@ the privacy guarantee is not affected by the sampling rate of the population,
...
+### Batch Size Considerations {#batch-size}
+
+If each batch will be collected only once (i.e. when using Prio3), then batch
+sizes may be equal to `min_batch_size` without issue. In the case of fixed size
+queries, `max_batch_size` may be equal to `min_batch_size`.
+
+If each batch may be collected more than once (i.e. when using Poplar1), then
+batches should be larger than `min_batch_size`, to allow for the possibility
+that some reports may successfully pass through VDAF preparation with the first
+aggregation parameter, but fail with a subsequent aggregation parameter. Once a
```
cjpatton, line 826: Instead of "pass/fail preparation", I prefer the verbiage "accept/reject".
```suggestion
that some reports may be accepted with the first
aggregation parameter, but be rejected with a subsequent aggregation parameter. Once a
```

draft-ietf-ppm-dap.md
```diff
@@ -804,19 +804,42 @@ the privacy guarantee is not affected by the sampling rate of the population,
...
+batch has been collected for the first time, subsequent collections of the same
+batch must process the same set of reports, and collections can only succeed if
+the number of successfully aggregated reports is at least `min_batch_size` (see
+{{time-interval-batch-validation}} and {{fixed-size-batch-validation}}). Thus,
+if enough reports fail validation that fewer than `min_batch_size` output shares
```
cjpatton, line 831:
```suggestion
if enough reports fail validation such that fewer than `min_batch_size` output shares
```

draft-ietf-ppm-dap.md
```diff
@@ -804,19 +804,42 @@ the privacy guarantee is not affected by the sampling rate of the population,
...
+coordinated. The difference between the minimum batch size and maximum batch
+size is in part intended to allow room for error, and allow a range of target
+batch sizes.
+
+### Batch Size Considerations {#batch-size}
+
+If each batch will be collected only once (i.e. when using Prio3), then batch
+sizes may be equal to `min_batch_size` without issue. In the case of fixed size
+queries, `max_batch_size` may be equal to `min_batch_size`.
+
```
cjpatton, line 822: This paragraph is a little long ... see if there is a way to split it up into more digestable chunks.

cjpatton (commented):
Looks good. Remember to use "SHOULD" instead of "should" when you want to clearly specify recommended behavior. Likewise for "may" and "must".

draft-ietf-ppm-dap.md
tgeoghegan: Rejecting invalid reports isn't the only reason an aggregator could deviate from the target batch size, so I think David's more general text is better.

tgeoghegan (approved):

draft-ietf-ppm-dap.md
cjpatton: Hmm, what else might happen?

divergentdave: In this sentence, "room for error" was previously referring to variance in batch size that may occur depending on how the aggregation jobs are ran by a distributed system. (depending on implementation decisions, etc.) I added the "in part" caveat in this PR because we now also need to allow for rejecting invalid reports. I can rephrase "error" to be more clear here.

cjpatton: Nah that's fine, thanks.

cjpatton (approved):

PR #554, opened 2024-04-23
Remove multi-collection of batches.

### LINKED ISSUE #259: Checking `batch-collected` at time of aggregate initalization is insufficient
I think that there is a hole in our anti-replay requirements, for tasks with `max_batch_lifetime > 1`. The issue is that currently, we only require that aggregators check if reports fall in intervals that have already been collected during the leader and helper aggregate initialization, see section `input-share-batch-validation`. We need a similar check either when handling aggregation continuations or the collect flow.

As an example of what we should be disallowing, consider a task with `min_batch_size = 10, max_batch_lifetime = 2`. Let's say that all aggregations are done with the same aggregation parameter, whatever it is.

- The leader sends out an AggregateInitializeReq for its first job, with ten reports.
  - The input share validation checks succeed, as there have been no collect reqeusts yet.
- The leader sends out an AggregateInitializeReq for its second job, with one report. This succeeds as well.
- The leader sends an AggregateContinueReq for its first job, with `finished` transitions.
- The leader sends an AggregateShareReq for an interval that covers all reports in both jobs.
- The leader sends an AggregateContinueReq for its second job, with `finished` transitions.
  - Note that we do not require the helper to do any checks against collect requests when handling aggregation continuations. (see `agg-continue-flow`)
- The leader sends a second AggregateShareReq, identical to the previous request.
  - Note that the checks in `batch-parameter-validation` pass.
    - The batch includes eleven reports, so `min_batch_size` is satisfied.
    - We assumed `max_batch_lifetime = 2`, and each report has been collected zero or one times thus far.
  - Aside from the `max_batch_lifetime` condition, the `collect-aggregate` section does not impose any other checks or conditions requiring the helper to consider previous AggregateShareReq requests when deciding if a request is valid, nor does it require that identical requests always get the same response.

After receiving the two aggregate shares produced above, the collector could compute the aggregate function over both the first ten reports and all eleven reports. For many functions, this would allow recovery of the eleventh input, breaking our privacy goal.

There is a paragraph that says:

> After issuing an aggregate-share request for a given batch interval, it is an error for the leader to issue any more aggregate or aggregate-init requests for additional reports in the batch interval. These reports will be rejected by helpers as described {{agg-init}}.

However, the referenced section only covers the "aggregate-init" part, and not aggregate continuation.

### LINKED ISSUE #316: Consider tying nonces to shares within a batch.
DAP uses a checksum over the report nonces to make sure that the same set of reports are included in repeated batches, but it does not verify the consistency of the ciphertexts  and input shares themselves between collect requests. In situations where max_batch_lifetime >=2, it may be possible for a a malicious leader to change the input share for a given nonce between two collect requests over the same batch interval. Replays are currently prevented by putting the nonce in the AAD of HPKE for client reports; however this does not protect against nonce replays with changed input shares by malicious clients. This could be of particular concern for Poplar, where multiple collect requests are used to explore a binary tree that is expected not to change between aggregations. 

Could we consider additional verification strategies for honest aggregators/helpers to ensure that every repeated batch interval contains the same input shares as well as the same nonces?

### LINKED ISSUE #405: Decoupling the aggregation parameter from aggregation jobs
Currently, the leader transmits aggregation parameters to the helper in an `AggregationJobInitReq`. Servicing a collection request can require arbitrarily many aggregation jobs (the leader chooses how many jobs to schedule), but all of them will have the same aggregation parameter value. This is fine for Prio3, because there is no aggregation parameter and thus it takes up no space in the wire message. But if a VDAF happens to have a large aggregation parameter (as is likely to be the case for Poplar1), then it's wasteful to retransmit the exact same aggregation parameter many times.

### LINKED ISSUE #409: Decouple transmission of report shares from leader to helper from aggregation job creation
Currently, the leader transmits encrypted report shares to the helper inline in an `AggregationJobInitReq` message. This is inefficient in the face of VDAFs that allow querying the same batch of reports multiple times with different aggregation parameters. Given the current protocol, those report shares must be retransmitted from leader to helper each time. This could be a significant waste of bandwidth, depending on the size of report shares and the number of collections that will be run against a batch.

### LINKED ISSUE #436: Do we still need `max_batch_query_count`?

branlwyd:
This is a prospective change to remove support for multi-collection of batches from DAP, as this was only required for the Heavy Hitters application which will no longer be supported by DAP.

Specifically:
 * Remove the `max_batch_query_count` task parameter, replacing uses of this parameter with semantics which are equivalent to a `max_batch_query_count` of 1.
 * Remove the fixed-size query type's `by_batch_id` query (as well as the `FixedSizeQueryType` message, since there is now only one possibility).
 * Remove the `PartialBatchSelector` from the `Collection` message, since this was used only to convey the `BatchId` to the `Collector` for later use of the `by_batch_id` query type, which is now gone. (With this change, Collectors no longer need to be aware of the concept of batch IDs.)

Closes #259.
Closes #316.
Closes #405.
Closes #409.
Closes #436.

### DIFF
--- a/draft-ietf-ppm-dap.md
+++ b/draft-ietf-ppm-dap.md
@@ -120,6 +120,11 @@ input is ever seen in the clear by any aggregator.
 
 (\*) Indicates a change that breaks wire compatibility with the previous draft.
 
+11:
+
+- Remove support for multi-collection of batches, as well as the fixed-size
+  query type's `by_batch_id` query. (\*)
+
 10:
 
 - Editorial changes from httpdir early review.
@@ -311,8 +316,7 @@ Aggregation function:
 : The function computed over the Clients' measurements. As defined in {{!VDAF}}.
 
 Aggregation parameter:
-: Parameter used to prepare a set of measurements for aggregation (e.g., the
-  candidate prefixes for Poplar1 from {{Section 8 of !VDAF}}). As defined in
+: Parameter used to prepare a set of measurements for aggregation. As defined in
   {{!VDAF}}.
 
 Aggregator:
@@ -396,16 +400,8 @@ else about the measurements. We call `F` the "aggregation function."
 
 This protocol is extensible and allows for the addition of new cryptographic
 schemes that implement the VDAF interface specified in
-{{!VDAF=I-D.draft-irtf-cfrg-vdaf-08}}. Candidates include:
-
-* Prio3 ({{Section 7 of !VDAF}}), which allows for aggregate statistics such as
-  sum, mean, histograms, etc.
-
-* Poplar1 ({{Section 8 of !VDAF}}), which allows for finding the most popular
-  strings uploaded by a set of Clients (e.g., the URL of their home page) as
-  well as counting the number of Clients that hold a given string. This VDAF is
-  the basis of the Poplar protocol of {{BBCGGI21}}, which is designed to solve
-  the heavy hitters problem in a privacy preserving manner.
+{{!VDAF=I-D.draft-irtf-cfrg-vdaf-08}}. This protocol only supports VDAFs which
+require a single collection to provide useful results.
 
 VDAFs rely on secret sharing to protect the privacy of the measurements. Rather
 than sending its input in the clear, each Client shards its measurement into a
@@ -600,7 +596,7 @@ standard tokens for use in the "type" field (within the DAP URN namespace
 | reportTooEarly             | Report could not be processed because its timestamp is too far in the future. |
 | batchInvalid               | The batch boundary check for Collector's query failed. |
 | invalidBatchSize           | There are an invalid number of reports in the batch. |
-| batchQueriedTooManyTimes   | The maximum number of batch queries has been exceeded for one or more reports included in the batch. |
+| batchQueriedMultipleTimes  | A batch was queried with multiple distinct aggregation parameters. |
 | batchMismatch              | Aggregators disagree on the report shares that were aggregated in a batch. |
 | unauthorizedRequest        | Authentication of an HTTP request failed (see {{request-authentication}}). |
 | missingTaskID              | HPKE configuration was requested without specifying a task ID. |
@@ -717,43 +713,27 @@ enum {
 
 The time_interval query type is described in {{time-interval-query}}; the
 fixed_size query type is described in {{fixed-size-query}}. Future
-specifications may introduce new query types as needed (see
-{{query-type-reg}}). Implementations are free to implement only a subset of the
-available query types.
+specifications may introduce new query types as needed (see {{query-type-reg}}).
+Implementations are free to implement only a subset of the available query
+types.
 
 A query includes parameters used by the Aggregators to select a batch of
 reports specific to the given query type. A query is defined as follows:
 
 ~~~
-opaque BatchID[32];
-
-enum {
-  by_batch_id(0),
-  current_batch(1),
-  (255)
-} FixedSizeQueryType;
-
-struct {
-  FixedSizeQueryType query_type;
-  select (FixedSizeQuery.query_type) {
-    case by_batch_id: BatchID batch_id;
-    case current_batch: Empty;
-  }
-} FixedSizeQuery;
-
 struct {
   QueryType query_type;
   select (Query.query_type) {
     case time_interval: Interval batch_interval;
-    case fixed_size: FixedSizeQuery fixed_size_query;
+    case fixed_size: Empty;
   }
 } Query;
 ~~~
 
 The query is issued in-band as part of the collect interaction
-({{collect-flow}}). Its content is determined by the "query type", which in
-turn is encoded by the "query configuration" configured out-of-band. All query
-types have the following configuration parameters in common:
+({{collect-flow}}). Its content is determined by the "query type", which in turn
+is encoded by the "query configuration" configured out-of-band. All query types
+have the following configuration parameters in common:
 
 - `min_batch_size` - The smallest number of reports the batch is allowed to
   include. In a sense, this parameter controls the degree of privacy that will
@@ -782,11 +762,9 @@ continuous, monotonically increasing, and have the same duration. For example,
 the sequence of batch intervals `(1659544000, 1000)`, `(1659545000, 1000)`,
 `(1659546000, 1000)`, `(1659547000, 1000)` satisfies these conditions. (The
 first element of the pair denotes the start of the batch interval and the second
-denotes the duration.) Of course, there are cases in which Collector may need to
-issue queries out-of-order. For example, a previous batch might need to be
-queried again with a different aggregation parameter (e.g, for Poplar1). In
-addition, the Collector may need to vary the duration to adjust to changing
-report upload rates.
+denotes the duration.) However, this is not a requirement--the Collector may
+decide to issue queries out-of-order. In addition, the Collector may need to
+vary the duration to adjust to changing report upload rates.
 
 ### Fixed-size Queries {#fixed-size-query}
 
@@ -801,12 +779,11 @@ that each batch has roughly the same number of reports. These batches are
 identified by opaque "batch IDs", allocated in an arbitrary fashion by the
 Leader.
 
-To get the aggregate of a batch, the Collector issues a query specifying the
-batch ID of interest (see {{query}}). The Collector may not know which batch ID
-it is interested in; in this case, it can also issue a query of type
-`current_batch`, which allows the Leader to select a recent batch to aggregate.
-The Leader MUST select a batch that has not yet been associated with a
-collection job.
+The Collector will not know the set of batch IDs available for collection. To
+get the aggregate of a batch, the Collector issues a query, which does not
+include any information specifying a particular batch (see {{query}}). The
+Leader selects a recent batch to aggregate. The Leader MUST select a batch that
+has not yet been associated with a collection job.
 
 In addition to the minimum batch size common to all query types, the
 configuration may include a parameter `max_batch_size` that determines maximum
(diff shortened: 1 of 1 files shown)


draft-ietf-ppm-dap.md
```diff
@@ -717,43 +715,29 @@ enum {
...
-enum {
-  by_batch_id(0),
-  current_batch(1),
-  (255)
-} FixedSizeQueryType;
-
-struct {
-  FixedSizeQueryType query_type;
-  select (FixedSizeQuery.query_type) {
-    case by_batch_id: BatchID batch_id;
```
branlwyd, line 739: It is OK to remove `by_batch_id` because this fixed-size query type was used only for repeated collection of the same batch with differing aggregation parameters. (Recovering from transient failures in collecting a given batch is enabled via re-querying the same collection job ID.)

draft-ietf-ppm-dap.md
```diff
@@ -829,35 +810,6 @@ coordinated. The difference between the minimum batch size and maximum batch
 size is in part intended to allow room for error, and allow a range of target
 batch sizes.
 
-### Batch Size Considerations {#batch-size}
```
branlwyd, line 832: I removed this section because:
 * The second & third paragraphs only apply to multiply-collected VDAFs.
 * The first paragraph boils down to "you should probably set `min_batch_size` and `max_batch_size` to the same value, and if you don't the Leader can set batches freely to any size between these two values."

Should we remove the concept of `max_batch_size`?

draft-ietf-ppm-dap.md
```diff
@@ -1683,42 +1624,8 @@ following checks:
    the same. If so, the Aggregator MUST mark the input share as invalid with
    error `invalid_message`.
 
-1. Check if the report may still be aggregated with the current aggregation
-   parameter. This can be done by looking up all aggregation parameters
-   previously used for this report and calling
-
-   ~~~
-   Vdaf.is_valid(current_agg_param, previous_agg_params)
```
branlwyd, line 1691: Is there a need for `Vdaf.is_valid` anymore? It _might_ be used to validate that an aggregation parameter itself is valid (irrespective of other aggregation parameters), but I think this might be better done at some point in evaluating the VDAF rather than via a separate method, for simplicity's sake.

draft-ietf-ppm-dap.md
```diff
@@ -2595,13 +2511,6 @@ Attacks on other properties of the system:
    existence of a report does not imply the occurrence of a sensitive event.
 1. Aggregators can deny service by refusing to respond to collection requests or
    aggregate share requests.
-1. Collection requests may leak information beyond the collection results. For
```
branlwyd, line 2598: I'm not sure there is a concrete example to cite here anymore, though I suppose there's nothing stopping a singly-collected VDAF with a nontrivial aggregation parameter from leaking information via the aggregation parameter.

draft-ietf-ppm-dap.md
```diff
@@ -401,12 +405,6 @@ schemes that implement the VDAF interface specified in
 * Prio3 ({{Section 7 of !VDAF}}), which allows for aggregate statistics such as
   sum, mean, histograms, etc.
 
-* Poplar1 ({{Section 8 of !VDAF}}), which allows for finding the most popular
```
branlwyd, line 404: I dropped references to Poplar1 throughout; I'm not sure if there is anything reasonable to replace it with. (Both PINE & Mastic, IIUC, are not yet to the point where we might cite them.)

draft-ietf-ppm-dap.md
tgeoghegan: Yes, I agree. The check for whether a report has been previously aggregated at all should suffice, and VDAF methods like `prep_init` or `prep_next` should safely reject invalid aggregation parameters. It's still useful for `Vdaf.is_valid` to be defined, because DAP-HH would use it, but DAP doesn't need to call it anymore.

draft-ietf-ppm-dap.md
tgeoghegan: I think we should leave this in, even without the explicit reference to Poplar1. Even when restricted to collecting batches once, DAP admits many future, unknown VDAFs, and pointing out that VDAFs may leak information (and not just to the collector!) and may need to be composed with DP is a legitimate security consideration. However I don't like the phrasing "Collection requests may leak information beyond the collection results" because it seems tautological. Let's try:

```suggestion
1. Some VDAFs could leak information to either Aggregator or the Collector
   beyond what the protocol intended to learn. It may be possible to mitigate
   such leakages using differential privacy ({{dp}}). 
```

tgeoghegan (commented):
Thanks for taking this on, Bran. Here's a first pass of comments, but I want to go back over the document looking for other things that may need to change.

draft-ietf-ppm-dap.md
cjpatton: We could keep the "counting the number of Clients that hold a given string" use case for Poplar1. This is basically a limited form of attributed based metrics (i.e., Prio3Count with attributes).

draft-ietf-ppm-dap.md
cjpatton: SGTM.

draft-ietf-ppm-dap.md
cjpatton: Let's consider removing `max_batch_size` in a separate issue.

draft-ietf-ppm-dap.md
```diff
@@ -1208,18 +1158,12 @@ Preparation has two purposes:
 1. To "refine" the input shares into "output shares" that have the desired
    aggregatable form. For some VDAFs, like Prio3, the mapping from input to
    output shares is some fixed, linear operation, but in general the mapping is
-   controlled dynamically by the Collector and can be non-linear. In
-   Poplar1, for example, the refinement process involves a sequence of
-   "candidate prefixes" and the output consists of a sequence of zeros and ones,
-   each indicating whether the corresponding candidate is a prefix of the
-   measurement from which the input shares were generated.
+   controlled dynamically by the Collector and can be non-linear.
```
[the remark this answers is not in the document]
cjpatton: No I don't think linearity is important. The suggestion looks good to me.

draft-ietf-ppm-dap.md
cjpatton: I think of validity of the agg param as being delegated to decoding on the wire.

draft-ietf-ppm-dap.md
```diff
@@ -1683,30 +1625,11 @@ following checks:
...
-
-   If this returns false, the input share MUST be marked as invalid with the
-   error `report_replayed`.
-
-    * Implementation note: To detect replay attacks, each Aggregator is required
-      to keep track of the set of reports it has processed for a given task.
-      Because honest Clients choose the report ID at random, it is sufficient to
-      store the set of IDs of processed reports. However, implementations may
-      find it helpful to track additional information, like the timestamp, so
-      that the storage used for anti-replay can be sharded efficiently.
```
cjpatton, line 1702: This note is still relevant I think.

draft-ietf-ppm-dap.md
```diff
@@ -2367,10 +2288,6 @@ For fixed_size tasks, the batch boundaries are defined by opaque batch IDs. Thus
 the Aggregator needs to check that the query is associated with a known batch
 ID:
```
cjpatton, line 2289: nit: Consider merging the text after the only remaining bullet into this paragraph.

draft-ietf-ppm-dap.md
```diff
@@ -2318,11 +2240,10 @@ determined by the query type. If the size check fails, then Helpers MUST abort
...
 
-Next, the Aggregator checks that the batch has not been queried too many times.
-This is determined by the maximum number of times a batch can be queried,
-`max_batch_query_count`. If the batch has been queried with more than
-`max_batch_query_count` distinct aggregation parameters, the Aggregator MUST
-abort with error of type "batchQueriedTooManyTimes".
+Next, the Aggregator checks that the batch has not been queried with multiple
+distinct aggregation parameters. If the batch has been queried with more than
+one distinct aggregation parameter, the Aggregator MUST abort with error of type
+"batchQueriedMultipleTimes".
```
tgeoghegan, line 2246: I think this could be stated more clearly in terms of the batch already having been collected.
```suggestion
Next, the Aggregator checks that the batch has not already been aggregated. If it
has, then the Aggregator MUST abort with error of type "batchAlreadyAggregated".
```

Note also the error type changes `batchQueriedMultipleTimes => batchAlreadyAggregated`.

tgeoghegan (commented):
We should delete [this paragraph](https://datatracker.ietf.org/doc/html/draft-ietf-ppm-dap-10#section-4.5.1.1-16) at L1345:

> [[OPEN ISSUE: Consider sending report shares separately (in parallel) to the aggregate instructions. Right now, aggregation parameters and the corresponding
report shares are sent at the same time, but this may not be strictly
necessary.]]

This isn't a concern anymore since we don't support multiple collection, and thus a report will be used exactly once.

We should revisit [VDAFs and compute requirements](https://datatracker.ietf.org/doc/html/draft-ietf-ppm-dap-10#name-vdafs-and-compute-requireme) so it doesn't discuss Poplar1 (or delete that and carry it over to the eventual DAP-HH draft).

branlwyd (commented):
> We should revisit [VDAFs and compute requirements](https://datatracker.ietf.org/doc/html/draft-ietf-ppm-dap-10#name-vdafs-and-compute-requireme) so it doesn't discuss Poplar1 (or delete that and carry it over to the eventual DAP-HH draft).

Done. I used a Prio3 example since that's the only concrete VDAF we have in hand at the moment, but some of the points regarding superlinearity of processing don't apply to any of the Prio3 variants (AFAIK).

Other than the changes suggested by review, I realized we can drop the partial batch selector from the `Collection` message, since this was used to convey the `BatchId` to the Collector for use in `by_batch_id` queries (which have now been removed). Other than making this message a little smaller, this also means that `BatchId`s are now private to the aggregators, which is a nice conceptual simplification for the Collector I suppose.

cjpatton:
> Other than the changes suggested by review, I realized we can drop the partial batch selector from the `Collection` message, since this was used to convey the `BatchId` to the Collector for use in `by_batch_id` queries (which have now been removed). Other than making this message a little smaller, this also means that `BatchId`s are now private to the aggregators, which is a nice conceptual simplification for the Collector I suppose.

This simplification sounds good to me.

tgeoghegan:
I think this change is in good shape, but we need to get a consensus call in the WG before we take it.

cjpatton (approved):

draft-ietf-ppm-dap.md
```diff
@@ -2318,11 +2237,10 @@ determined by the query type. If the size check fails, then Helpers MUST abort
 with an error of type "invalidBatchSize". Leaders SHOULD wait for more reports
 to be validated and try the collection job again later.
 
-Next, the Aggregator checks that the batch has not been queried too many times.
```
kenluck2001, line 2321: Is it possible to effect this repetitive query of the same batch by adopting a rate-limiting scheme that is not tightly bundled with the the underlying DAP. Could we make it loosely coupled?

cjpatton: What would be the goal of rate limiting?

The purpose of querying a batch multiple times is to allow a batch to be aggregated with different aggregation parameters. This enables the tree traversal at the heart of Poplar1's solution to heavy hitters: https://datatracker.ietf.org/meeting/interim-2024-ppm-01/materials/slides-interim-2024-ppm-01-sessa-supporting-heavy-hitters-00

branlwyd: This PR removes the ability to query the same batch multiple times. The purpose of the check this comment was made on is to ensure that we do not collect the same batch multiple times, since doing so may violate the privacy of the reports included in the batch. Thus, we need a solution which permanently blocks the undesirable queries, rather than temporarily rate-limiting them.

As Chris notes, before we removed multi-collection in this PR, each query of the same batch used a distinct aggregation parameter. For the Heavy Hitters application, there was a nontrivial rule to check whether a given aggregation parameter could be used given the aggregation parameters already used. Thus, any given solution to the multiple-queries problem would need to be aware of the aggregation parameters in use.

branlwyd:
Rebased & squashed. I'm going to hold off on merging this for a few days to allow for any final comments.

Issue #560, opened 2024-05-09
Discuss lifecycle of various objects and how they relate to each other


tgeoghegan:
On the mailing list, @wbl observed that it's hard for readers to understand the relationship of reports to aggregation jobs to batches to aggregate shares. We should add a section between the introduction and the the sections describing the sub-interactions which discusses the lifecycle and cardinality of these objects.

tgeoghegan:
Mailing list thread in question: https://mailarchive.ietf.org/arch/msg/ppm/E1OAX-HKwknXGevr5EmVjSEdOtE/

tgeoghegan:
One reason this can be confusing is the notion of "eager" aggregation, which is doable for VDAFs like Prio3 that don't have aggregation parameters. Simon noted in #385 that this is confusing, and perhaps we could clear this up by making the notion of eager aggregation more explicit in DAP.

branlwyd:
I think it would be useful to discuss lifecycles, but the DAP spec has historically avoided being too prescriptive about lifetimes/deletion -- IMO we should remain descriptive rather than prescriptive, and if we are prescriptive, say as little as possible. (I suppose the goal is to proscribe the minmimal amount of behavior leading implementations to be interoperable.)

cjpatton:
2024/10/24 interim: Now that draft 13 has substantial breaking changes planned, we are going to want to cut it sooner rather than later. I'm therefore punting this to the next draft.

PR #563, opened 2024-06-13
Rename "query type" to "batch mode", "fixed-size" to "leader-selected", remove max_batch_size.

### LINKED ISSUE #555: Consider removing `max_batch_size`.
Once multi-collection of batches has been removed, we should consider whether the `max_batch_size` parameter of the fixed-size query type might be removed.

See https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/pull/554/files#r1577003240 for context; in short, `max_batch_size`'s strongest use-case is in the case of a multiply-collected VDAF. It is unclear if there is a strong justification for this parameter in a DAP which supports only singly-collected VDAFs.

branlwyd:
Batch mode is a more descriptive name than query type. And leader-selected is a more accurate name than fixed-size, since the batch sizes are not fixed -- any size above min_batch_size (and, if specified, below max_batch_size) is acceptable.

The max_batch_size parameter of the leader-selected query type is removed. This parameter does not need to be enforced by both aggregators, so it can be made implementation-specific.

Also, restore the partial batch selector to the Collection message; the included batch ID is required for the Collector to decrypt the aggregate shares, as it is included in the AAD. This bug was introduced in #554 & effectively breaks collection for the fixed-size query type in DAP-11.

Closes #555.

### DIFF
--- a/draft-ietf-ppm-dap.md
+++ b/draft-ietf-ppm-dap.md
@@ -120,6 +120,20 @@ input is ever seen in the clear by any aggregator.
 
 (\*) Indicates a change that breaks wire compatibility with the previous draft.
 
+12:
+
+- Remove the `max_batch_size` parameter of the fixed-size query type.
+
+- Rename the "fixed-size" query type to "leader-selected", to align the name
+  with the behavior of this query type.
+
+- Rename "query type" to "batch mode", to align the name of this configuration
+  value with its functionality.
+
+- Restore the `part_batch_selector` field of the `Collection` structure, which
+  was removed in draft 11, as it is required to decrypt collection results in
+  some cases. (\*)
+
 11:
 
 - Remove support for multi-collection of batches, as well as the fixed-size
@@ -400,7 +414,7 @@ measurement (or set of measurements, e.g., counts of some user behavior). Given
 the input set of measurements `x_1, ..., x_n` held by `n` Clients, and an
 aggregation parameter `p` shared by the Aggregators, the goal of DAP is to
 compute `y = F(p, x_1, ..., x_n)` for some function `F` while revealing nothing
-else about the measurements. We call `F` the "aggregation function."
+else about the measurements. We call `F` the "aggregation function".
 
 This protocol is extensible and allows for the addition of new cryptographic
 schemes that implement the VDAF interface specified in
@@ -696,47 +710,51 @@ VDAF types:
 * PrepShare
 * PrepMessage
 
-## Queries {#query}
+## Batch Modes, Batches, and Queries {#batch-mode}
 
 Aggregated results are computed based on sets of reports, called "batches". The
-Collector influences which reports are used in a batch via a "query." The
+Collector requests the aggregated results for a given batch via a "query". The
 Aggregators use this query to carry out the aggregation flow and produce
 aggregate shares encrypted to the Collector.
 
-This document defines the following query types:
+Each measurement task has a preconfigured "batch mode". The batch mode defines
+both how reports may be partitioned into batches, as well as how these batches
+are addressed and the semantics of the query used for collection.
+
+This document defines the following batch modes:
 
 ~~~
 enum {
   reserved(0), /* Reserved for testing purposes */
   time_interval(1),
-  fixed_size(2),
+  leader_selected(2),
   (255)
-} QueryType;
+} BatchMode;
 ~~~
 
-The time_interval query type is described in {{time-interval-query}}; the
-fixed_size query type is described in {{fixed-size-query}}. Future
-specifications may introduce new query types as needed (see {{query-type-reg}}).
-Implementations are free to implement only a subset of the available query
-types.
+The time-interval batch mode is described in {{time-interval-batch-mode}}; the
+leader-selected batch mode is described in {{leader-selected-batch-mode}}.
+Future specifications may introduce new batch modes as needed (see
+{{batch-mode-reg}}). Implementations are free to implement only a subset of the
+available batch modes.
 
 A query includes parameters used by the Aggregators to select a batch of
-reports specific to the given query type. A query is defined as follows:
+reports specific to the given batch mode. A query is defined as follows:
 
 ~~~
 struct {
-  QueryType query_type;
-  select (Query.query_type) {
+  BatchMode batch_mode;
+  select (Query.batch_mode) {
     case time_interval: Interval batch_interval;
-    case fixed_size: Empty;
+    case leader_selected: Empty;
   }
 } Query;
 ~~~
 
 The query is issued in-band as part of the collect interaction
-({{collect-flow}}). Its content is determined by the "query type", which in turn
-is encoded by the "query configuration" configured out-of-band. All query types
-have the following configuration parameters in common:
+({{collect-flow}}). Its content is determined by the batch mode, which in turn
+is determined by the measurement task, which is configured out-of-band. All
+batch modes have the following configuration parameters in common:
 
 - `min_batch_size` - The smallest number of reports the batch is allowed to
   include. In a sense, this parameter controls the degree of privacy that will
@@ -745,20 +763,20 @@ have the following configuration parameters in common:
   measurements and aggregation function.
 
 - `time_precision` - Clients use this value to truncate their report timestamps;
-  see {{upload-flow}}. Additional semantics may apply, depending on the query
-  type. (See {{batch-validation}} for details.)
+  see {{upload-flow}}. Additional semantics may apply, depending on the batch
+  mode. (See {{batch-validation}} for details.)
 
-The parameters pertaining to specific query types are described in
-{{time-interval-query}} and {{fixed-size-query}}.
+The parameters pertaining to specific batch modes are described in
+{{time-interval-batch-mode}} and {{leader-selected-batch-mode}}.
 
-### Time-interval Queries {#time-interval-query}
+### Time-interval Batch Mode {#time-interval-batch-mode}
 
-The first query type, `time_interval`, is designed to support applications in
-which reports are collected over a long period of time. The Collector specifies
-a "batch interval" that determines the time range for reports included in the
-batch. For each report in the batch, the time at which that report was generated
-(see {{upload-flow}}) MUST fall within the batch interval specified by the
-Collector.
+The time-interval batch mode is designed to support applications in which
+reports are collected into batches grouped by an interval of time. The Collector
+specifies a "batch interval" that determines the time range for reports included
+in the batch. For each report in the batch, the time at which that report was
+generated (see {{upload-flow}}) MUST fall within the batch interval specified by
+the Collector.
 
 Typically the Collector issues queries for which the batch intervals are
 continuous, monotonically increasing, and have the same duration. For example,
@@ -769,45 +787,27 @@ denotes the duration.) However, this is not a requirement--the Collector may
 decide to issue queries out-of-order. In addition, the Collector may need to
 vary the duration to adjust to changing report upload rates.
 
-### Fixed-size Queries {#fixed-size-query}
+### Leader-selected Batch Mode {#leader-selected-batch-mode}
 
-The `fixed_size` query type is used to support applications in which the
-Collector needs the ability to strictly control the batch size. This is
-particularly important for controlling the amount of noise added to reports by
-Clients (or added to aggregate shares by Aggregators) in order to achieve
-differential privacy.
-
-For this query type, the Aggregators group reports into arbitrary batches such
-that each batch has roughly the same number of reports. These batches are
-identified by opaque "batch IDs", allocated in an arbitrary fashion by the
-Leader.
+The leader-selected batch mode is used to support applications in which it is
+acceptable for reports to be batched in an arbitrary fashion by the Leader. Each
+batch of reports is identified by an opaque "batch ID". Both the reports
+included in each batch and the ID for each batch are allocated in an arbitrary
+fashion by the Leader.
 
 The Collector will not know the set of batch IDs available for collection. To
 get the aggregate of a batch, the Collector issues a query, which does not
-include any information specifying a particular batch (see {{query}}). The
+include any information specifying a particular batch (see {{batch-mode}}). The
 Leader selects a recent batch to aggregate. The Leader MUST select a batch that
 has not yet been associated with a collection job.
 
-In addition to the minimum batch size common to all query types, the
-configuration may include a parameter `max_batch_size` that determines maximum
-number of reports per batch.
-
-If the configuration does not include `max_batch_size`, then the Aggregators
-can output any batch size that is larger than or equal to `min_batch_size`.
-This is useful for applications that are not concerned with sample size, i.e.,
-the privacy guarantee is not affected by the sampling rate of the population,
-therefore a larger than expected batch size does not weaken the designed privacy
-guarantee.
-
-Implementation note: The goal for the Aggregators is to aggregate the same
-number of reports in each batch. The target batch size is deployment-specific,
-and may be equal to or greater than the minimum batch size. Deciding how soon
-batches should be output is also deployment-specific. Exactly sizing batches may
-be challenging for Leader deployments in which multiple, independent nodes
+The Aggregators can output batches of any size that is larger than or equal to
+`min_batch_size`. The target batch size, if any, is implementation-specific, and
+may be equal to or greater than the minimum batch size. Deciding how soon
+batches should be output is also implementation-specific. Exactly sizing batches
+may be challenging for Leader deployments in which multiple, independent nodes
 running the aggregate interaction (see {{aggregate-flow}}) need to be
-coordinated. The difference between the minimum batch size and maximum batch
-size is in part intended to allow room for error, and allow a range of target
-batch sizes.
+coordinated.
 
 ## Task Configuration {#task-configuration}
 
@@ -825,10 +825,10 @@ the following parameters associated with it:
    can be found.
 * `helper_aggregator_url`: A URL relative to which the Helper's API resources
   can be found.
-* The query configuration for this task (see {{query}}). This determines the
-  query type for batch selection and the properties that all batches for this
-  task must have. The party MUST NOT configure the task if it does not
-  recognize the query type.
+* The batch mode for this task (see {{batch-mode}}). This determines how reports
+  are grouped into batches and the properties that all batches for this task
+  must have. The party MUST NOT configure the task if it does not recognize the
+  batch mode.
 * `task_expiration`: The time up to which Clients are expected to upload to this
   task. The task is considered completed after this time. Aggregators MAY reject
   reports that have timestamps later than `task_expiration`.
@@ -1323,10 +1323,10 @@ Once all the report shares have been initialized, the Leader creates an
 opaque BatchID[32];
 
 struct {
-  QueryType query_type;
-  select (PartialBatchSelector.query_type) {
+  BatchMode batch_mode;
+  select (PartialBatchSelector.batch_mode) {
     case time_interval: Empty;
-    case fixed_size: BatchID batch_id;
+    case leader_selected: BatchID batch_id;
   };
 } PartialBatchSelector;
 
@@ -1344,23 +1344,17 @@ This message consists of:
 * `part_batch_selector`: The "partial batch selector" used by the Aggregators to
   determine how to aggregate each report:
 
-    * For `fixed_size` tasks, the Leader specifies a "batch ID" that determines
-      the batch to which each report for this aggregation job belongs.
+    * For leader-selected tasks, the Leader specifies a "batch ID" that
+      determines the batch to which each report for this aggregation job
+      belongs.
 
-      [OPEN ISSUE: For fixed_size tasks, the Leader is in complete control over
-      which batch a report is included in. For time_interval tasks, the Client
-      has some control, since the timestamp determines which batch window it
-      falls in. Is this desirable from a privacy perspective? If not, it might
-      be simpler to drop the timestamp altogether and have the agg init request
-      specify the batch window instead.]
-
-  The indicated query type MUST match the task's query type. Otherwise, the
+  The indicated batch mode MUST match the task's batch mode. Otherwise, the
   Helper MUST abort with error `invalidMessage`.
 
   This field is called the "partial" batch selector because, depending on the
-  query type, it may not determine a batch. In particular, if the query type is
+  batch mode, it may not determine a batch. In particular, if the batch mode is
   `time_interval`, the batch is not determined until the Collector's query is
-  issued (see {{query}}).
+  issued (see {{batch-mode}}).
 
 * `prepare_inits`: the sequence of `PrepareInit` messages constructed in the
   previous step.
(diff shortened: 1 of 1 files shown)


draft-ietf-ppm-dap.md
```diff
@@ -2214,7 +2202,7 @@ from its query and the response to its query as follows:
 * For time_interval tasks, the batch selector is the batch interval specified in
   the query.
 
-* For fixed_size tasks, the batch selector is the batch ID assigned sent in the
+* For leader_selected tasks, the batch selector is the batch ID sent in the
```
branlwyd, line 2205: The aggregate share AAD includes the batch ID, so the batch ID needs to be communicated to the Collector to allow decryption, alas. I restored the batch ID to the `Collection` message.

However, I think maybe this isn't the right fix (just the least controversial one to unbreak things for now). DAP aims to provide robustness only if both aggregators are honest (and robustness is trivally breakable by a single dishonest aggregator -- either aggregator can forge results). If both aggregators are honest, there is no need for the aggregate share AAD to include the batch ID, since there is no need for the Collector to e.g. detect the aggregator attempting to repeatedly transmit an aggregate share. So I think the "right" fix would be to drop batch ID from both the `Collection` and `AggregateShareAad` messages.

However, this argument applies just as well to every other field in `AggregateShareAad` -- so what is the purpose of the AAD here? (best-effort protection against bugs, perhaps?)

If this isn't trivially resolvable, this might be worthy of an issue to allow discussion on what the right fix is.

branlwyd:
Closes #555.

draft-ietf-ppm-dap.md
branlwyd: (I think the same argument doesn't apply to `InputShareAad`, since this AAD is used to allow the Helper to ensure the encrypted report share it received from the Leader matches the rest of the report share.)

draft-ietf-ppm-dap.md
```diff
@@ -120,6 +120,13 @@ input is ever seen in the clear by any aggregator.
 
 (\*) Indicates a change that breaks wire compatibility with the previous draft.
 
+12:
+
+- Rename the fixed-size query type to leader-selected, to align the name with
+  the behavior of this query type.
+
+- Remove the `max_batch_size` parameter of the leader-selected query type.
```
cjpatton, line 128: nit: I'd re-order these, since you need the second as context for the first.

draft-ietf-ppm-dap.md
```diff
@@ -770,17 +777,12 @@ denotes the duration.) However, this is not a requirement--the Collector may
...
-Collector needs the ability to strictly control the batch size. This is
-particularly important for controlling the amount of noise added to reports by
-Clients (or added to aggregate shares by Aggregators) in order to achieve
-differential privacy.
-
-For this query type, the Aggregators group reports into arbitrary batches such
-that each batch has roughly the same number of reports. These batches are
-identified by opaque "batch IDs", allocated in an arbitrary fashion by the
+The `leader_selected` query type is used to support applications in which it is
+acceptable for reports to be batched in an arbitrary fashion. Each batch of
```
cjpatton, line 783:
```suggestion
acceptable for reports to be batched by the Leader in an arbitrary fashion. Each batch of
```

draft-ietf-ppm-dap.md
```diff
@@ -789,26 +791,21 @@ include any information specifying a particular batch (see {{query}}). The
...
-can output any batch size that is larger than or equal to `min_batch_size`.
-This is useful for applications that are not concerned with sample size, i.e.,
-the privacy guarantee is not affected by the sampling rate of the population,
-therefore a larger than expected batch size does not weaken the designed privacy
-guarantee.
+The Aggregators can output batches of any size that is larger than or equal to
+`min_batch_size`. This is useful for applications that are not concerned with
+sample size, i.e., the privacy guarantee is not affected by the sampling rate of
+the population, therefore a larger than expected batch size does not weaken the
+designed privacy guarantee.
```
cjpatton, line 798: This text now lacks important. There is only a minimum batch size, hence no application of DAP is allowed to be "concerned with the sample size".

I think what we want to do now instead is say something generic about how privacy guarantees of DAP depend on the batch size, perhaps when composed with some DP policy.

However I wonder if we still think this is accurate. My intuition is that choosing a larger batch size strictly increases privacy, regardless of how DP is implemented. (Though utility may be negatively impacted.) Is this correct?

cc/ @wangshan since according to `git blame`, you wrote this text originally.

draft-ietf-ppm-dap.md
```diff
@@ -789,26 +791,21 @@ include any information specifying a particular batch (see {{query}}). The
...
 
-Implementation note: The goal for the Aggregators is to aggregate the same
-number of reports in each batch. The target batch size is deployment-specific,
+Implementation note: The target batch size, if any, is implementation-specific,
 and may be equal to or greater than the minimum batch size. Deciding how soon
-batches should be output is also deployment-specific. Exactly sizing batches may
-be challenging for Leader deployments in which multiple, independent nodes
+batches should be output is also implementation-specific. Exactly sizing batches
+may be challenging for Leader deployments in which multiple, independent nodes
```
cjpatton, line 803: I don't object, but why change "deployment" to "implementation" here?

draft-ietf-ppm-dap.md
```diff
@@ -789,26 +791,21 @@ include any information specifying a particular batch (see {{query}}). The
...
 running the aggregate interaction (see {{aggregate-flow}}) need to be
-coordinated. The difference between the minimum batch size and maximum batch
-size is in part intended to allow room for error, and allow a range of target
-batch sizes.
+coordinated. Applications which include differential privacy may wish to control
+the size (or range of sizes) of generated batches in order to control the amount
+of noise added to reports by Clients (or added to aggregate shares by
+Aggregators).
```
cjpatton, line 808: here too, @wangshan (see above comment)

draft-ietf-ppm-dap.md
```diff
@@ -1356,15 +1353,16 @@ This message consists of:
...
-      has some control, since the timestamp determines which batch window it
-      falls in. Is this desirable from a privacy perspective? If not, it might
-      be simpler to drop the timestamp altogether and have the agg init request
-      specify the batch window instead.]
+      [OPEN ISSUE: For leader_selected tasks, the Leader is in complete control
+      over which batch a report is included in. For time_interval tasks, the
+      Client has some control, since the timestamp determines which batch window
+      it falls in. Is this desirable from a privacy perspective? If not, it
+      might be simpler to drop the timestamp altogether and have the agg init
+      request specify the batch window instead.]
```
cjpatton, line 1365: This doesn't appear to be desirable from a privacy perspective. I'll note here that having the timestamp in the metadata seems to be helpful for replay protection, since it allows you to shard by batch interval.

I suggest we just delete this text.

draft-ietf-ppm-dap.md
```diff
@@ -1463,10 +1461,9 @@ enum {
   hpke_unknown_config_id(3),
   hpke_decrypt_error(4),
   vdaf_prep_error(5),
-  batch_saturated(6),
-  task_expired(7),
-  invalid_message(8),
-  report_too_early(9),
+  task_expired(6),
+  invalid_message(7),
+  report_too_early(8),
```
cjpatton, line 1466: This is a wire-breaking change. We can avoid it by just deleting `batch_saturated` and not renumbering the remaining variants. However this would leave a gap in the variants defined in our draft, which is a bit odd.

I think we should avoid breaking changes if we can. An alternative would be to leave a note here that if we want a new variant before RFC, then the variant should use `6`.

We could take this one step further by being very explicit about what we want in the end:

```
enum {
  batch_collected(0),
  report_replayed(1),
  report_dropped(2),
  hpke_unknown_config_id(3),
  hpke_decrypt_error(4),
  vdaf_prep_error(5),
  /* [TODO Remove or replace this variant.] */
  todo_remove_me(6),
  task_expired(7),
  invalid_message(8),
  report_too_early(9),
  (256)
} PrepareError;
```

draft-ietf-ppm-dap.md
```diff
@@ -2003,6 +1987,10 @@ struct {
 The body's media type is "application/dap-collection". The `Collection`
 structure includes the following:
 
+* `part_batch_selector`: Information used to bind the aggregate result to the
+  query. For fixed_size tasks, this includes the batch ID assigned to the batch
```
cjpatton, line 1991:
```suggestion
  query. For leader_selected tasks, this includes the batch ID assigned to the batch
```

draft-ietf-ppm-dap.md
cjpatton: I think we should include the batch ID in the AAD. It's not strictly necessary in our threat model, so think of it as defense in depth: successful execution implies agreement on the batch ID between the aggregators and collector; this may have benefits we don't see right now.

As a general rule, binding protocol execution to crypto operations is a good idea whenever feasible. There are tons of examples of this, the most recent of which might be https://eprint.iacr.org/2023/691.

draft-ietf-ppm-dap.md
wangshan: generally speaking larger batch size gives better DP guarantee, however, there can be a case where the DP guarantee is based on a certain sampling rate and allowing too big a batch may break that condition. But in practice, I don't think sampling would be implemented in a way that make this a problem. If we decide to remove max_batch_size, I agree it doesn't make much sense to mention sampling here.

draft-ietf-ppm-dap.md
wangshan: Similar to above, shall we just say whether to control the size of the batch is an implementation specific choice?

wangshan:
@branlwyd I feel `leader-selected` is a bit misleading, given that the batch size is decided by task creator based on some privacy requirements, for eg differential privacy. `leader-selected query` sounds odd to me.
I suggest we change it to something that reflects the batch releasing criteria, how about `batch-threshold`?

cjpatton:
> @branlwyd I feel `leader-selected` is a bit misleading, given that the batch size is decided by task creator based on some privacy requirements, for eg differential privacy. `leader-selected query` sounds odd to me. I suggest we change it to something that reflects the batch releasing criteria, how about `batch-threshold`?

You're correct that the batch size is not chosen by the Leader, but it is true that the Leader decides which reports go in which batch. On the other hand, that may not be the intent of the deployment.

That said, `batch-threshold` sounds generic enough to apply to any query type, no? E.g., isn't the "batch releasing criteria" now the same for time_interval as for fixed_size?

What about something that emphasizes that there are no batch boundary criteria, as there are for time_interval? Perhaps any_batch or something?

draft-ietf-ppm-dap.md
simon-friedberger: I'm sure we've discussed this before so maybe I am missing something but there is an attack where a Leader separately collects reports from a single client and then only aggregates those.

For time intervals the leader is strongly constrained here and that's exactly why we prohibit overlapping time-intervals. We might however have lost this meaning in the wording over time...

> If the report pertains to a batch that was previously collected, then the input share MUST be marked as invalid with error batch_collected.

> Finally, the Aggregator checks that the batch does not contain a report that was included in any previous batch. If this batch overlap check fails, then the Aggregator MUST abort with error of type "batchOverlap". For time_interval tasks, it is sufficient (but not necessary) to check that the batch interval does not overlap with the batch interval of any previous query. If this batch interval check fails, then the Aggregator MAY abort with error of type "batchOverlap".

> For each report in the batch, the time at which that report was generated (see [Section 4.4](https://ietf-wg-ppm.github.io/draft-ietf-ppm-dap/draft-ietf-ppm-dap.html#upload-flow)) MUST fall within the batch interval specified by the Collector.

According to my probably incorrect memory this is supposed to be prevented by the collector deciding which reports to aggregate and the other aggregator checking as well.
We should be explicit about this attack in https://ietf-wg-ppm.github.io/draft-ietf-ppm-dap/draft-ietf-ppm-dap.html#name-batch-parameters

draft-ietf-ppm-dap.md
branlwyd: Fair enough. Even if this AAD doesn't support the explicit threat model, at the very least it will help avoid accidental bugs which e.g. mismatch. (I'm not sure it can protect against any form of malice, as the components of the AAD are public to both  aggregators; and each aggregator can already falsify results of any aggregate shares it can control -- i.e. the Helper can falsify the Helper's aggregate share, the Leader can falsify the Leader or Helper's aggregate share.)

draft-ietf-ppm-dap.md
branlwyd: Done, though now we're referencing "fixed-size" directly before we mention that it is renamed.

draft-ietf-ppm-dap.md
branlwyd: I got rid of this text and merged the next paragraph into it, since that paragraph is also concerned with the range of possible batch sizes. The following paragraph is concerned with the batch sizes that might be generated by an implementation, but IMO the existing text makes it clear the difference between specification & implementation advice (and this section is not normative).

draft-ietf-ppm-dap.md
branlwyd: Saying something is "implementation-specific" is (in my experience) much more common than "deployment-specific". Prior to this document, I am not sure I have seen the use of "deployment" in this fashion; instead, typically "implementation" would be used.

draft-ietf-ppm-dap.md
branlwyd: Deleting the text SGTM because I think `time_interval`'s association of report to batch via timestamp is (plausibly) desirable from a privacy perspective.

draft-ietf-ppm-dap.md
branlwyd: I suppose -- every new draft is a breaking change because it will, at least, change the domain-separation strings. I suppose it could be argued that shifting the numbers around could lead to bugs in implementations that implement multiple DAP drafts & use the wrong "version" of `PrepareError`; but we could run into similar issues if/when we reuse the number with a new semantic meaning.

That said, I'm happy enough kicking the can down the road, as long as we clean things up for the final RFC.

branlwyd:
> > You're correct that the batch size is not chosen by the Leader, but it is true that the Leader decides which reports go in which batch. On the other hand, that may not be the intent of the deployment.
> > That said, `batch-threshold` sounds generic enough to apply to any query type, no? E.g., isn't the "batch releasing criteria" now the same for time_interval as for fixed_size?
> > What about something that emphasizes that there are no batch boundary criteria, as there are for time_interval? Perhaps any_batch or something?
> 
> The batch releasing criteria is still different, `fixed_size` only requires minimum batch size to be met, `time_interval` requires size and time. How about `size_conditioned`? I feel the query type as is only defines the batch creation criteria, not the releasing criteria. If that's what we want query type to do, then something like `any_batch` would work too.

My interpretation is also that the query type determines how batches are determined:

* `time_interval` determines batches by intervals of time.
* `leader_selected` determines batches by the mostly-arbitrary selection of the Leader. 

`any_batch`, or even `unconstrained`/`arbitrary` would also work for `leader_selected`; but IMO these are not strongly better than `leader_selected` -- they might even be considered worse, in the sense that `leader_selected` indicates which protocol participant is doing the batch selection.

draft-ietf-ppm-dap.md
cjpatton: @simon-friedberger I agree that time interval queries are much more constrained, and this definitely has benefits. But I don't think fixed size queries are unconstrained in a way that leads to overlapping batches. The aggregators are supposed to assign reports to batches and make sure each report is mapped to a single batch. At a high level, any query type needs to ensure that queries _partition_ the set of reports for a task.

draft-ietf-ppm-dap.md
cjpatton: We don't actually need to update the domain separation tags with each version. In fact, in the latest version of the VDAF draft, we didn't update the version tag because there were no breaking changes.

branlwyd:
I think a query type effectively:

* determines how reports can be grouped together
* determines how batches are addressed (i.e. what is the identifier for a batch)
* determines the semantics of how aggregate results can be queried by the collector

I added some text to this effect -- I'm not totally sure I understand what is meant by "release criteria". If this doesn't capture what we hoped, please push back.

draft-ietf-ppm-dap.md
branlwyd: I added some text to the security considerations section capturing this concern.

draft-ietf-ppm-dap.md
branlwyd: Thinking about this a little more, at least in my head:

* An "implementation" is a specific software implementation of some DAP protocol actor. Examples would include Janus or Daphne.
* A "deployment" is a practical deployment of DAP; it would use one or more implementations of the DAP protocol actors, as well as defining tasks (or deciding how tasks can be dynamically configured, e.g. taskprov or some other task-provisioning scheme), defining other configuration parameters, deciding how the software implementations are run on machines, and everything else that comes with deploying software. Examples would include Divvi Up.

draft-ietf-ppm-dap.md
```diff
@@ -699,24 +706,28 @@ VDAF types:
...
-Aggregated results are computed based on sets of reports, called "batches". The
-Collector influences which reports are used in a batch via a "query." The
+Aggregated results are computed based on sets of reports, called "batches." The
+Collector requests the aggregated results for a given batch via a "query." The
 Aggregators use this query to carry out the aggregation flow and produce
 aggregate shares encrypted to the Collector.
 
+Each measurement task has a preconfigured "query type." This query type defines
+both how reports may be grouped into batches, as well as how these batches are
+addressed & the semantics of the query used for collection.
```
cjpatton, line 716:
```suggestion
both how reports may be partitions into batches, as well as how these batches are
addressed and the semantics of the query used for collection.
```

cjpatton (approved):
Looks good to me. I lean towards "leader_selected", but I don't much care what we call it at the end of the day.

draft-ietf-ppm-dap.md
```diff
@@ -699,24 +706,28 @@ VDAF types:
...
 ## Queries {#query}
 
-Aggregated results are computed based on sets of reports, called "batches". The
-Collector influences which reports are used in a batch via a "query." The
+Aggregated results are computed based on sets of reports, called "batches." The
+Collector requests the aggregated results for a given batch via a "query." The
 Aggregators use this query to carry out the aggregation flow and produce
 aggregate shares encrypted to the Collector.
 
+Each measurement task has a preconfigured "query type." This query type defines
```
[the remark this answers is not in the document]
branlwyd: (made this change throughout.)

tgeoghegan (approved):

draft-ietf-ppm-dap.md
```diff
@@ -2003,6 +1977,11 @@ struct {
 The body's media type is "application/dap-collection". The `Collection`
 structure includes the following:
 
+* `part_batch_selector`: Information used to bind the aggregate result to the
+  query. For leader_selected tasks, this includes the batch ID assigned to the
```
wangshan, line 1981:
```suggestion
  query. For leader-selected tasks, this includes the batch ID assigned to the
```

draft-ietf-ppm-dap.md
```diff
@@ -2066,7 +2045,7 @@ message contains the following parameters:
 
     * For time_interval tasks, the request specifies the batch interval.
 
-    * For fixed_size tasks, the request specifies the batch ID.
+    * For leader_selected tasks, the request specifies the batch ID.
```
wangshan, line 2048: names with _ are usually backquoted. Consider adding `` to  this and time_interval on above row

wangshan:
@cjpatton @branlwyd I don't have a strong opinion on what it should be called. From the discussion so far, I think we agree the purpose of this `query_type` is how the batch is created, or partitioned/selected. In this context, `leader_selected` makes sense, as long as we explain this in the text. I would also preferred to change `query_type` to `batch_select_type`, but I don't have strong opinion on that either.

cjpatton:
> @cjpatton @branlwyd I don't have a strong opinion on what it should be called. From the discussion so far, I think we agree the purpose of this `query_type` is how the batch is created, or partitioned/selected. In this context, `leader_selected` makes sense, as long as we explain this in the text. I would also preferred to change `query_type` to 
`batch_select_type`, but I don't have strong opinion on that either.

It's a little broader than that, but I think I might know what's missing: a notion of "bucket" (#560). Namely, I've understood "query" to refer to both:

1. how reports may be mapped to batch buckets (e.g., a time window)
2. how batch buckets are merged into batches (e.g., a sequence of time windows, i.e., a time interval)

branlwyd:
> I would also preferred to change `query_type` to `batch_select_type`, but I don't have strong opinion on that either.

What about `batching_mode` (or `batch_mode`)?

wangshan:
> 
> What about `batching_mode` (or `batch_mode`)?

`batch_mode` sounds nice.

branlwyd:
Thanks -- I've updated this PR to rename "query type" to "batch mode" & I like the change. @cjpatton interested in your thoughts here, I can revert (or rename to something else) if this isn't so good.

cjpatton:
> Thanks -- I've updated this PR to rename "query type" to "batch mode" & I like the change. @cjpatton interested in your thoughts here, I can revert (or rename to something else) if this isn't so good.

No objection.

draft-ietf-ppm-dap.md
```diff
@@ -1335,10 +1331,10 @@ Once all the report shares have been initialized, the Leader creates an
...
 struct {
-  QueryType query_type;
-  select (PartialBatchSelector.query_type) {
+  BatchMode batch_mode;
+  select (PartialBatchSelector.batch_mode) {
     case time_interval: Empty;
-    case fixed_size: BatchID batch_id;
+    case leader_selected: BatchID batch_id;
   };
 } PartialBatchSelector;
```
cjpatton, line 1339: While at it: Do we still like the term "partial batch selector"?

cjpatton (approved):

draft-ietf-ppm-dap.md
```diff
@@ -2532,6 +2512,30 @@ different ways to address Sybil attacks, such as:
    {{anon-proxy}}.
 1. Differential privacy ({{dp}}) can help mitigate Sybil attacks to some extent.
 
+## Batch-selection Attacks {#batch-selection}
+
+Depending on the batch mode, the privacy of an individual Client may be
+infringed upon by selection of the batch. For example, in the leader-selected
```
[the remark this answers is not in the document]
branlwyd: I went with a different change: everywhere we are discussing the modes in text, rather than specifying a specific enum variant, I use time-interval/leader-selected instead of `time_interval`/`leader_selected`. I think this reads a little better.

draft-ietf-ppm-dap.md
branlwyd: I do not like "partial batch selector" at all, but I have struggled to come up with a better name (e.g. more descriptive, without being too long). Suggestions for a better name are very welcome IMO

cjpatton:
@branlwyd please squash and merge at will.

PR #565, opened 2024-07-03
Specify aggregation with batched preparation

### LINKED ISSUE #561: Support for batched preparation
There is a [new paper](https://eprint.iacr.org/2024/666) (to appear at IEEE S&P 2024) that shows how to lift an [FLP](https://datatracker.ietf.org/doc/html/draft-irtf-cfrg-vdaf-09#section-7.1) into a proof system that is "silently verifiable". For such a proof system, it's possible to check if _some_ measurement in a set of measurements is invalid, exchanging only a constant number of bits (128). In Prio3, we must exchange a number of bits linear in the number of measurements.

The bandwidth savings are significant. For Prio3SumVec with bits=1, length=1,000, and chunk_length=32 and 100 reports, an `AggregationJobInitReq` is 122KB. If we modified Prio3 by making the FLP silently verifiable, the same message would be roughly 7KB (back-of-a-napkin calculation). Computational cost remains about the same. (See Figure 7 from the paper for details.)

This approach does a have a potential downside: if verification fails, then we learn only that _some_ measurement is invalid, not _which_ measurement is invalid. Thus, in the case that some fraction of the reports in the aggregation job, you either need to throw away the data or exchange more bits in order to find the subset of measurements that are invalid. The communication cost increases as the fraction of malicious Clients increases.

There are many ways we might implement this. Perhaps the simplest would be to treat batched preparation as an alternative aggregation flow that falls back to the existing aggregation flow (i.e., per report validation) if verification fails. Concretely, we would specify an alternative aggregation sub-protocol that is the same as the existing flow except:
1. It is only compatible with 1-round, "silently verifiable" VDAFs.
2. If verification fails, then it needs to be safe to fallback to per-report validation.

cjpatton:
Closes #561.

Adds an alternative aggregation flow for VDAFs that support batched preparation.

In this mode of operation, prep shares for many reports are aggregated into a short message called a "tag". The tags computed by the Leader and Helper are used to decide validity of the reports: if the tags are valid, then every report is deemed valid; otherwise, at least one report is known to be invalid.

The Helper handles invalidity by aborting the request. In this case the Leader is allowed to retry, either with a subset of the reports or with the existing, pep-report preparation code path.

DRAFT STATUS: This PR is not quite feature-complete. In fact, before merging it, we may need to make some changes to the VDAF draft. It is intended primarily to drive conversation about whether there is appetite to support the feature.

### DIFF
--- a/draft-ietf-ppm-dap.md
+++ b/draft-ietf-ppm-dap.md
@@ -1883,6 +1883,224 @@ request, indexed by aggregation job ID and step, and refuse to service a request
 for a given aggregation step unless it matches the previously seen request (if
 any).
 
+### Batched Preparation
+
+Some VDAFs support a mode of operation known as "batched preparation". In this
+mode, prep shares for many reports are aggregated into a short message called a
+"tag". The tags computed by the Leader and Helper are used to decide validity
+of the reports: if the tags are valid, then every report is deemed valid;
+otherwise, at least one report is known to be invalid. This section describes
+an alternative aggregation flow for batched preparation.
+
+> NOTE For context, there are two recent papers that describe alternatives to
+> Prio3 and Poplar1 respectively that support this mode of operation:
+>
+> - Whisper (https://eprint.iacr.org/2024/666, IEEE S&P 2024): Extends the FLP
+>   system underlying Prio3 to make input validation batchable. In more detail,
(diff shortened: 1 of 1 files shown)


cjpatton:
The issue was closed, so closing this PR as well.

PR #566, opened 2024-07-06
Document deviations from RFC 8446 PL

### LINKED ISSUE #472: Document extensions to and/or deviations from RFC 8446 presentation language
In #393, we define a union with an enum discriminant roughly like:

```
enum {
  continue(0),
  finished(1)
  reject(2),
  (255)
} PrepareStepState;

struct {
  PrepareStepState prepare_step_state;
  select (PrepareStep.prepare_step_state) {
    case continue:
      ReportId report_id;
      opaque payload<0..2^32-1>;
    case finished:
      ReportId report_id;
    case reject:
      ReportId report_id;
      ReportShareError report_share_error;
  };
} PrepareStep;
```

Then later in protocol text, we define partial constructions of values like:

```
struct {
  PrepareStepState prepare_step_state = 2; /* reject */
  ReportId report_id;
  ReportShareError report_share_error;
} PrepareStep;
```

This notation is clear and certainly is more compact than a paragraph explaining to implementers what enum variant to use, but it conflicts with [section 3.7 of RFC 8446](https://datatracker.ietf.org/doc/html/rfc8446#section-3.7), which tells us that this is the notation for defining structs that contain fields whose values are fixed (I gather this was useful in TLS 1.3 for backward compatibility with older TLS versions).

We can resolve this by adding some text to "Conventions and Definitions", where we introduce the usage of RFC 8446 presentation language, explaining our extensions and deviations. There may also be some prior art in other TLS WG documents we can reference, but I haven't found it.

tgeoghegan:
Explains in detail how the syntax from RFC 8446 section 3.7 is repurposed for concise descriptions of structure variants. Additionally, the section on message encoding is moved to "Protocol Definition" to make the document flow better.

Closes #472

### DIFF
--- a/draft-ietf-ppm-dap.md
+++ b/draft-ietf-ppm-dap.md
@@ -388,10 +388,6 @@ Report Share:
 
 {:br}
 
-This document uses the presentation language of {{!RFC8446}} to define messages
-in the DAP protocol. Encoding and decoding of these messages as byte strings
-also follows {{RFC8446}}.
-
 # Overview {#overview}
 
 The protocol is executed by a large set of Clients and a pair of servers
@@ -639,6 +635,82 @@ DAP has three major interactions which need to be defined:
 
 Each of these interactions is defined in terms of "resources". In this section
 we define these resources and the messages used to act on them.
+
+## Message Encoding
+
+We use the presentation language defined in {{!RFC8446, Section 3}} to define
+messages in the DAP protocol, with the following deviations.
+
+{{Section 3.7 of !RFC8446}} defines a syntax for structure fields whose values
+are constants. In this document, we do not use that notation, but use something
+similar to describe specific variants of structures containing enumerated types,
+described in {{!RFC8446, Section 3.8}}.
+
+For example, suppose we have an enumeration and a structure defined as follows:
+
+~~~
+enum {
+  number(0),
+  string(1),
+  (255)
+} ExampleEnum;
+
+struct {
+  uint32 always_present;
+  ExampleEnum type;
+  select (ExampleStruct.type) {
+    case number: uint32 a_number;
+    case string: opaque a_string<0..10>;
+  };
+} ExampleStruct;
+~~~
+
+Then we describe the specific variant of `ExampleStruct` where `type == number`
+with a `variant` block like so:
+
+~~~
+variant {
+  /* Field exists regardless of variant */
+  uint32 always_present;
+  ExampleEnum type = number;
+  /* Only fields included in the type == number variant is described */
+  uint32 a_number;
+} ExampleStruct;
+~~~
+
+The protocol text accompanying this would explain how implementations should
+handle the `always_present` and `a_number` fields but not `type`. This does not
+mean that the `type` field of `ExampleStruct` can only ever have value `number`.
+
+This notation can also be used in structures where the enum field does not
+affect what fields are or are not present in the structure. For example:
+
+~~~
+enum {
+  something(0),
+  something_else(1),
+  (255)
+} FailureReason;
+
+struct {
+  FailureReason failure_reason;
+  opaque another_field<0..256>;
+} FailedOperation;
+~~~
+
+The protocol text might include a description like:
+
+~~~
+variant {
+  FailureReason failure_reason = something;
+  opaque another_field<0..256>;
+} FailedOperation;
+~~~
+
+Encoding and decoding of these messages as byte strings also follows
+{{RFC8446}}.
+
+## Basic Type Definitions
 
 The following are some basic type definitions used in other messages:
 
@@ -1492,7 +1564,7 @@ For any report that was rejected, the Helper sets the outbound preparation
 response to
 
 ~~~
-struct {
+variant {
   ReportID report_id;
   PrepareRespState prepare_resp_state = reject;
   PrepareError prepare_error;
@@ -1525,7 +1597,7 @@ initial `outbound` message to send in response to the Leader. If `state` is of
 type `Rejected`, then the Helper responds with
 
 ~~~
-struct {
+variant {
   ReportID report_id;
   PrepareRespState prepare_resp_state = reject;
   PrepareError prepare_error = vdaf_prep_error;
@@ -1535,7 +1607,7 @@ struct {
 Otherwise the Helper responds with
 
 ~~~
-struct {
+variant {
   ReportID report_id;
   PrepareRespState prepare_resp_state = continue;
   opaque payload<0..2^32-1> = outbound;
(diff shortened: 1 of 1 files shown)


cjpatton:
Reviewer note: @suman-ganta's [observed](https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/issues/472#issuecomment-1781933269) that TLS-syntax does not explicitly allow `case` blocks with multiple fields. We currently don't have any cases of this.

cjpatton (approved):

draft-ietf-ppm-dap.md
```diff
@@ -640,6 +636,82 @@ DAP has three major interactions which need to be defined:
...
+like so:
+
+~~~
+struct {
+  /* Field exists regardless of variant */
+  uint32 always_present;
+  ExampleEnum type = number;
+  /* Only fields included in the type == number variant is described */
+  uint32 a_number;
+} ExampleStruct;
```
suman-ganta, line 678: This semantic is defined in [variant](https://datatracker.ietf.org/doc/html/rfc8446#section-3.8) section, right? Are we making it `select` readable here? If so, do we need to do this for every structure? Or there is a special case where illustrating the above way makes it clear?

Seeing same type definition multiple times with different fields could confuse readers. Instead of extending the semantics of `struct`, should this type be called `const` or `type`?

draft-ietf-ppm-dap.md
```diff
@@ -640,6 +636,82 @@ DAP has three major interactions which need to be defined:
...
+
+Then we describe the specific variant of `ExampleStruct` where `type == number`
+like so:
+
+~~~
+struct {
+  /* Field exists regardless of variant */
+  uint32 always_present;
+  ExampleEnum type = number;
+  /* Only fields included in the type == number variant is described */
```
suman-ganta, line 676: How do we express multiple selects and their associated qualified members in this notation? (Such case does not exist in DAP so far I think)

tgeoghegan: Since this doesn't occur in DAP, I don't think we need to define it. Also, I would push back pretty hard on any struct definition that contained multiple selects. That would almost certainly allow construction of messages that aren't semantically valid, which would force extra validation checks on implementations. I think we would always prefer a struct contain a single enumeration where all its variants are valid over multiple enumerated types.

draft-ietf-ppm-dap.md
tgeoghegan: The pertinent examples are various cases of constructing `PrepareResp` as shown [here](https://datatracker.ietf.org/doc/html/draft-ietf-ppm-dap-11#section-4.5.1.2-9) and [here](https://datatracker.ietf.org/doc/html/draft-ietf-ppm-dap-11#section-4.5.1.2-15) or [here](https://datatracker.ietf.org/doc/html/draft-ietf-ppm-dap-11#section-4.5.2.2-13).

I like the idea of annotating such partial constants somehow to make it clear this special syntax is in use. `const` doesn't feel right because only part of the structure is constant. Maybe `struct specialized` to indicate we are describing a particular specialization of the structure:

```
struct specialized {
  ReportID report_id;
  PrepareRespState prepare_resp_state = continue;
  opaque payload<0..2^32-1> = outbound;
} PrepareResp;
```

Or perhaps `struct variant` is better, since we are dealing with enums here, and "specialized" implies generics.

suman-ganta: `struct variant` sounds good to me.

draft-ietf-ppm-dap.md
suman-ganta: From, the semantics of variants, it is possible to have several selects in a structure. Since variants allow dynamic members of structure, it sounds like legitimate definition to me. Am I missing anything here? Since we are coming up with a notation that could probably be followed in future DAP specs too, I'm trying to make sure we define something that lasts.

I'm fine if you would like to defer this to later updates.

Overall, In this notation, I didn't see the distinction between representing `alwaysExsting` fields vs fields added as a result of a specific value of enum. Since the order of fields matter, and these dynamic fields can appear in the middle of two static fields, some visual cue (bold/italicized etc) could help to locate them.

draft-ietf-ppm-dap.md
cjpatton: I'm confused by the need for special syntax. What is a "partial constant"?

draft-ietf-ppm-dap.md
cjpatton: Can you give an example of when multiple `select`s in the same `struct` is ambiguous?

draft-ietf-ppm-dap.md
tgeoghegan: [RFC 8446 3.7](https://datatracker.ietf.org/doc/html/rfc8446#section-3.7) defines constant fields of structs:

```
 struct {
          T1 f1 = 8;  /* T.f1 must always be 8 */
          T2 f2;
      } T;
```

This notation collides with our usage, where `PrepareResp.prepare_resp_state` can have many values, but we want to describe the variant where `prepare_resp_state = continue`.

cjpatton: In the example given here, can't `f1` have many values?

tgeoghegan: No, that is the point of the RFC 8446 constant notation. The comment says "T.f1 must always be 8". If you click through to 8446 s3.7 it says 'Fields and variables may be assigned a fixed value using "="'.

cjpatton: Got it, thanks for clarifying. I think this could be clearer in the PR.

draft-ietf-ppm-dap.md
```diff
@@ -640,6 +636,82 @@ DAP has three major interactions which need to be defined:
...
 we define these resources and the messages used to act on them.
 
+## Message Encoding
+
+We use the presentation language defined in {{!RFC8446, Section 3}} to define
+messages in the DAP protocol, with the following deviations.
+
+{{Section 3.7 of !RFC8446}} defines a syntax for structure fields whose values
+are constants. In this document, we do not use that notation, but use something
+similar to describe specific variants of structures containing enumerated types,
```
wangshan, line 646: @tgeoghegan does this mean the syntax from `{{Section 3.7 of !RFC8446}}` is forbidden in this doc? 
If a structure has the following def:
```
struct {
   uint32 always_present;
   ExampleEnum type;
}
```

Then does the below have ambiguity?
```
struct variant {
   uint32 always_present = 0;
   ExampleEnum type = number;
}
```
Does it mean `always_present` can only be a constant 0 or it just happens to be 0 in this variant?

tgeoghegan: It does not override that section of RFC 8446. The notation in 8446 section 3.7 is `struct { ... } whatever`. The notation we introduce here is `struct variant { ... } whatever` which does not conflict.

branlwyd (approved):

suman-ganta:
`struct variant` is kind of indicating it is deviation from `struct`, but most of the `struct` semantics apply. `variant` alone is not defined anywhere, so it would leave every one for wild interpretations.

cjpatton:
@tgeoghegan please squash and merge at will!

PR #567, opened 2024-07-06
Remove per-task HPKE configurations

### LINKED ISSUE #505: Security considerations: privacy impact of per-task HpkeConfigs

tgeoghegan:
As discussed in #505, putting the task ID into the HPKE configuration request leaks which tasks a client is participating in, which can otherwise be concealed by OHTTP. This commit removes the notion of per-task HPKE keys from DAP, requiring global keys.

Closes #505

### DIFF
--- a/draft-ietf-ppm-dap.md
+++ b/draft-ietf-ppm-dap.md
@@ -603,7 +603,6 @@ standard tokens for use in the "type" field (within the DAP URN namespace
 | batchQueriedMultipleTimes  | A batch was queried with multiple distinct aggregation parameters. |
 | batchMismatch              | Aggregators disagree on the report shares that were aggregated in a batch. |
 | unauthorizedRequest        | Authentication of an HTTP request failed (see {{request-authentication}}). |
-| missingTaskID              | HPKE configuration was requested without specifying a task ID. |
 | stepMismatch               | The Aggregators disagree on the current step of the DAP aggregation protocol. |
 | batchOverlap               | A request's query includes reports that were previously collected in a different batch. |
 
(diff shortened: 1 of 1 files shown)


draft-ietf-ppm-dap.md
```diff
@@ -895,25 +894,14 @@ configuration of each Aggregator. See {{compliance}} for information on HPKE
...
+GET request to `{aggregator}/hpke_config`.
 
 An Aggregator responds to well-formed requests with HTTP status code 200 OK and
 an `HpkeConfigList` value, with media type "application/dap-hpke-config-list".
 The `HpkeConfigList` structure contains one or more `HpkeConfig` structures in
 decreasing order of preference. This allows an Aggregator to support multiple
 HPKE configurations simultaneously.
 
-[TODO: Allow Aggregators to return HTTP status code 403 Forbidden in deployments
-that use authentication to avoid leaking information about which tasks exist.]
```
tgeoghegan, line 915: This TODO is moot:

1. If we only have global HPKE keys, we don't leak which tasks exist
2. This would at best be a security consideration; DAP does not need to give implementations permission to use well-established HTTP semantics.

branlwyd (approved):

cjpatton (approved):
No objection to merging, but since this is making DAP a little less general, we need to make the case (at IETF 120) for this resolution and against alternatives. An alternative resolution is to discuss the side-effect for OHTTP in security considerations.

tgeoghegan:
We discussed this at PPM @ IETF 120 and nobody spoke up in defense of per-task HPKE configurations, so let's merge this!

PR #568, opened 2024-07-08
DAP Version compatibility

### LINKED ISSUE #541: Versioning PPM media types

suman-ganta:
Closes #541.

Made versioning compatibilities explicit and an optional way to express draft versions.

### DIFF
--- a/draft-ietf-ppm-dap.md
+++ b/draft-ietf-ppm-dap.md
@@ -2680,6 +2680,22 @@ in this section for all media types listed above.
 
 [OPEN ISSUE: Solicit review of these allocations from domain experts.]
 
+### Message versioning
+
+Media types for the HTTP requests are specific to this version of DAP. When a
+new major enhancement is proposed that results in newer IETF specification
+for DAP, a new set of media types need to be defined. In other words, newer
+versions of DAP will not be backward compatible with this version of DAP.
+
+(NOTE TO RFC EDITOR: Remove this paragraph.) HTTP requests with DAP media types
+MAY express an optional parameter 'version', following {{Section 8.3 of !RFC9110}}.
+Value of this parameter indicates current draft version of the protocol the
+component is using. This MAY be used as a hint by the receiver of the request
+to do compatibility checks between client and server.
+For example, A report submission to leader from a client that supports
+draft-ietf-ppm-dap-09 could have the header
+`Media-Type: application/dap-report;version=09`.
+
 ### "application/dap-hpke-config-list" media type
 
 Type name:


draft-ietf-ppm-dap.md
```diff
@@ -2654,6 +2654,11 @@ the input shares and defeat privacy.
 
 # IANA Considerations
 
+## Message versioning
+Media types for the HTTP requests are specific to this version of DAP. When a new major enhancement is proposed that results in newer IETF specification for DAP, a new set of media types need to be defined. In other words, newer versions of DAP will not be backward compatible with this version of DAP.
+
+HTTP requests with DAP media types MAY express an optional parameter 'version'. Value of this parameter indicates current draft version of the protocol the component is using. This MAY be used as a hint by the receiver of the request to do compatibility checks between client and server.
```
branlwyd, line 2660: I think this paragraph would not fit into the eventual RFC for DAP, since there will be no draft version to refer to. IMO, add a TODO to remove this before RFC.

branlwyd (approved):
N.B. the cryptographic domain separation string used for encryption of report & aggregate shares contains the current draft number, so implementations of different drafts of DAP will not interoperate. But there's no reason we can't include this as well, it would likely be much easier to debug a version-skew issue using this mechanism.

draft-ietf-ppm-dap.md
```diff
@@ -2654,6 +2654,11 @@ the input shares and defeat privacy.
 
 # IANA Considerations
 
+## Message versioning
```
cjpatton, line 2657:
```suggestion
## Message versioning

```

draft-ietf-ppm-dap.md
```diff
@@ -2654,6 +2654,11 @@ the input shares and defeat privacy.
...
+Media types for the HTTP requests are specific to this version of DAP. When a new major enhancement is proposed that results in newer IETF specification for DAP, a new set of media types need to be defined. In other words, newer versions of DAP will not be backward compatible with this version of DAP.
```
cjpatton, line 2658: Wrap paragraphs at 80 columns.

draft-ietf-ppm-dap.md
cjpatton: Agreed. Concretely, add something like this: 
```suggestion
(NOTE TO RFC EDITOR: Remove this paragraph.) HTTP requests with DAP media types MAY express an optional parameter 'version'. Value of this parameter indicates current draft version of the protocol the component is using. This MAY be used as a hint by the receiver of the request to do compatibility checks between client and server.
```

tgeoghegan: I'd like to see an example of this optional `version` parameter. Would this be something like `Media-Type: application/dap-aggregation-job-init-req;version=11`? A reference to [RFC 9110 section 8.3](https://www.rfc-editor.org/rfc/rfc9110.html#section-8.3.1) (which describes this kind of thing in media types) also would help.

tgeoghegan (commented):
I like the idea of doing this as an optional parameter in the existing `Content-Type`/`Accept` headers. However, for that reason, I'm not sure we need this spelled out in the DAP protocol. DAP is built on HTTP, and RFC 9110 already makes it clear that media types can include "semicolon-delimited parameters  in the form of name/value pairs" ([Section 8.3.1](https://www.rfc-editor.org/rfc/rfc9110.html#section-8.3.1)). We don't need to give DAP implementations permission to use well-defined HTTP semantics, just the same as we don't need to give them permission to use HTTP 3xx redirects.

Since we don't specify what should be in this `version` parameter, nor do we specify how implementations should evaluate it, I'm not sure this text adds value to this protocol.

cjpatton:
> I like the idea of doing this as an optional parameter in the existing `Content-Type`/`Accept` headers. However, for that reason, I'm not sure we need this spelled out in the DAP protocol. DAP is built on HTTP, and RFC 9110 already makes it clear that media types can include "semicolon-delimited parameters in the form of name/value pairs" ([Section 8.3.1](https://www.rfc-editor.org/rfc/rfc9110.html#section-8.3.1)). We don't need to give DAP implementations permission to use well-defined HTTP semantics, just the same as we don't need to give them permission to use HTTP 3xx redirects.
> 
> Since we don't specify what should be in this `version` parameter, nor do we specify how implementations should evaluate it, I'm not sure this text adds value to this protocol.

I think it's a good idea to at least mention this is a possibilty, rather than leaving it to implementers to dig this up themselves. Language like "MAY add a version parameter as specified in {{Section whatever of !RFC9110}}" would be helpful.

draft-ietf-ppm-dap.md
suman-ganta: incorporated both suggestions.

draft-ietf-ppm-dap.md
suman-ganta: done

draft-ietf-ppm-dap.md
```diff
@@ -2654,6 +2654,21 @@ the input shares and defeat privacy.
...
+DAP, a new set of media types need to be defined. In other words, newer 
+versions of DAP will not be backward compatible with this version of DAP.
+
+(NOTE TO RFC EDITOR: Remove this paragraph.) HTTP requests with DAP media types
+MAY express an optional parameter 'version', following {{Section 8.3 of !RFC9110}}.
+Value of this parameter indicates current draft version of the protocol the 
+component is using. This MAY be used as a hint by the receiver of the request 
+to do compatibility checks between client and server.
+For example, A report submission to leader from a client that supports DAP09
+would have the header Media-Type: application/dap-report;version=09.
```
tgeoghegan, line 2670:
```suggestion
For example, a report submission to leader from a client that supports
draft-ietf-ppm-dap-09 could have the header `Media-Type:
application/dap-report;version=09`.
```

Prefer notation `draft-ietf-ppm-dap-09` because that's a bit more formal in IETF parlance. Also use "could" instead of "would", because this text isn't normative: it's completely up to implementations what to put in the `version` field and how to interpret it.

draft-ietf-ppm-dap.md
```diff
@@ -2654,6 +2654,21 @@ the input shares and defeat privacy.
 
 # IANA Considerations
 
+## Message versioning
```
cjpatton, line 2657: Let's move this to a subsection of `## Protocol Message Media Types` immediately below this.

cjpatton (commented):
One more editorial comment from me, then I'm good to go!

draft-ietf-ppm-dap.md
suman-ganta: Moved.

draft-ietf-ppm-dap.md
suman-ganta: Sounds good. Done.

cjpatton (approved):

draft-ietf-ppm-dap.md
```diff
@@ -2680,6 +2680,22 @@ in this section for all media types listed above.
...
+### Message versioning
+
+Media types for the HTTP requests are specific to this version of DAP. When a 
+new major enhancement is proposed that results in newer IETF specification for 
+DAP, a new set of media types need to be defined. In other words, newer 
+versions of DAP will not be backward compatible with this version of DAP.
+
+(NOTE TO RFC EDITOR: Remove this paragraph.) HTTP requests with DAP media types
+MAY express an optional parameter 'version', following {{Section 8.3 of !RFC9110}}.
+Value of this parameter indicates current draft version of the protocol the
```
cjpatton, line 2692:
```suggestion
Value of this parameter indicate current draft version of the protocol the 
```

cjpatton:
@suman-ganta CI is failing due to trailing whitespace:
```
draft-ietf-ppm-dap.md contains trailing whitespace
```

suman-ganta:
Addressed trailing whitespace issue.

draft-ietf-ppm-dap.md
```diff
@@ -2680,6 +2680,22 @@ in this section for all media types listed above.
...
+
+Media types for the HTTP requests are specific to this version of DAP. When a
+new major enhancement is proposed that results in newer IETF specification
+for DAP, a new set of media types need to be defined. In other words, newer
+versions of DAP will not be backward compatible with this version of DAP.
+
+(NOTE TO RFC EDITOR: Remove this paragraph.) HTTP requests with DAP media types
+MAY express an optional parameter 'version', following {{Section 8.3 of !RFC9110}}.
+Value of this parameter indicates current draft version of the protocol the
+component is using. This MAY be used as a hint by the receiver of the request
```
cjpatton, line 2693: This is non-normative IMO
```suggestion
component is using. This might be used as a hint by the receiver of the request
```

cjpatton:
@suman-ganta this is on hold while we get some feedback from HTTP experts. I'm on it!

mnot:
What's being versioned here - the format itself, or the protocol overall?

Regardless, I wouldn't recommend using a media type parameter to version either. Parameters are lost by some types of software, and if there are breaking changes in the version, that can cause compatibility issues.

Best practice is to use a new media type to indicate breaking changes.

cjpatton:
> @mnot What's being versioned here - the format itself, or the protocol overall?
>
> Regardless, I wouldn't recommend using a media type parameter to version either. Parameters are lost by some types of software, and if there are breaking changes in the version, that can cause compatibility issues.
>
> Best practice is to use a new media type to indicate breaking changes.

The protocol specifies the format of each message, so both.

The PR intends only to use the parameter as a hint for debugging purposes, not as a means of indicating the version or draft. We have other ways of handling breaking changes across drafts.

We plan to update the media types for future versions of the specification (i.e., future RFCs).

mnot:
If it's very clearly just a hint and doesn't change behaviour or trigger normative requirements, that's OK. From what I see, it's more typical to do this inside the format itself, although you may have Reasons™️.

PR #564, opened 2024-06-25
Helper-side aggregation job init/continue asynchronous processing

inahga:
Closes https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/issues/557.

Allows helper-side aggregate job preparation to be handled asynchronously,
similarly to how collection jobs work. Briefly, after calls with
`AggregationJobInitReq` and `AggregationJobContinueReq`, the leader polls
the helper at `GET /tasks/{task-id]/aggregation_jobs/{aggregation-job-id}`
until receiving an `AggregationJobResp`.

In discussion in #577 we had soft consensus on making this the only option,
i.e. for spec simplicity we don't need to also allow the old synchronous
behavior.

PR #569, opened 2024-07-17
Allow for aggregation step skew recovery of arbitrary number of steps

kristineguo:
Context: https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/pull/564#discussion_r1674725591

The current draft focuses on only the recovery from `n` to `n+1`. We should generalize this section to allow aggregation step skew recovery of any arbitrary number of previous steps.

### DIFF
--- a/draft-ietf-ppm-dap.md
+++ b/draft-ietf-ppm-dap.md
@@ -1854,24 +1854,26 @@ it can clean up its state.
 
 #### Recovering from Aggregation Step Skew {#aggregation-step-skew-recovery}
 
-`AggregationJobContinueReq` messages contain a `step` field, allowing
-Aggregators to ensure that their peer is on an expected step of the DAP
-aggregation protocol. In particular, the intent is to allow recovery from a
-scenario where the Helper successfully advances from step `n` to `n+1`, but its
-`AggregationJobResp` response to the Leader gets dropped due to something like a
-transient network failure. The Leader could then resend the request to have the
-Helper advance to step `n+1` and the Helper should be able to retransmit the
-`AggregationJobContinueReq` that was previously dropped. To make that kind of
-recovery possible, Aggregator implementations SHOULD checkpoint the most recent
-step's prep state and messages to durable storage such that the Leader can
-re-construct continuation requests and the Helper can re-construct continuation
-responses as needed.
+`AggregationJobContinueReq` messages contain a `step` field to indicate the
+expected step of the DAP aggregation protocol their peer should be on. The
+inclusion of this field also allows the Leader to recover if it falls behind
+the Helper (e.g., Leader drops a response due to a transient network failure).
+In this situation, the Leader can re-send the `AggregationJobContinueReq`
+messages for previously computed steps. The Helper can respond by retransmitting
+the corresponding `AggregationJobResp` messages. By replaying the exact sequence
+of messages, the Leader can re-compute its previously dropped state and resume
+aggregation with the Helper from the most recent step.
+
+To make this kind of recovery possible, Aggregator implementations SHOULD, at
+minimum, checkpoint the most recent step's prep state and messages to durable
+storage such that the Leader can re-construct continuation requests and the
+Helper can re-construct continuation responses as needed.
 
 When implementing an aggregation step skew recovery strategy, the Helper SHOULD
-ensure that the Leader's `AggregationJobContinueReq` message did not change when
-it was re-sent (i.e., the two messages must be identical). This prevents the
-Leader from re-winding an aggregation job and re-running an aggregation step
-with different parameters.
+ensure that the Leader's `AggregationJobContinueReq` message for a given step did
+not change when it was re-sent (i.e., the two messages must be identical). This
+prevents the Leader from re-winding an aggregation job and re-running an aggregation
+step with different parameters.
 
 [[OPEN ISSUE: Allowing the Leader to "rewind" aggregation job state of the
 Helper may allow an attack on privacy. For instance, if the VDAF verification


branlwyd:
I don't object to generalizing the protocol in this way (as long as implementations are still free to only implement `n`/`n+1` skew); but I believe the semantics were written this way originally because the only skew possible is `n`/`n+1` skew, that is, implementing single-step recovery is good enough for all scenarios.

Is that incorrect? Is there a plausible way for the Leader/Helper to skew more than one step apart via transient network failures or other distributed-systems concerns?

kristineguo:
It depends on leader implementation. If the leader is checkpointing its state to durable storage after every aggregation step, then you're correct that it only ever has to implement `n`/`n+1` skew recovery.

However, I don't think it's fair to assume that every leader implementation will be doing this. Some leader implementations may choose to omit this checkpointing (e.g., to reduce I/O or complexity) and perform a job's aggregation all in memory.

So if the job fails or if the leader goes down, it may have to retry the entire aggregation job from step 0 (init). The protocol should allow for this as long as the requests re-sent are identical, effectively allowing an aggregator to replay its state transitions to catch back up.

branlwyd:
That's fair. I think this leads to several different skew-recovery models models, each with fairly different implementation requirements:

* In the "current" model, where only `n`/`n+1` skew recovery is supported, the Helper needs to store the hash of the previous request, and the entirety of the previous response (or enough information to recover this response based on the previous request, i.e. the previous VDAF preparation states, I believe).
* In the model where the Leader always restarts aggregation from the beginning, the Helper needs to store the hash the initial request, but does not need to store any responses (since they can be recomputed on retry).
* In the model where the Leader restarts from anywhere in the aggregation process, the Helper needs to store the hash of every request, and every response (or enough information to recompute it based on the relevant request).

These differences might be an interoperability pain point, e.g. if the Leader expects to be able to restart aggregations from the beginning but the Helper implements only `n`/`n+1`-skew recovery.

Perhaps we can pick just one of these skew-recovery methods? I would be happiest with either the `n`/`n+1`-skew recovery that is specified today, or a recovery method that always restarts from the beginning. However, my implementation experience is with 1-step VDAFs -- where these models are equivalent -- so I would appreciate the perspective of folks who have deployed multi-step VDAFs.

draft-ietf-ppm-dap.md
```diff
@@ -1854,24 +1854,26 @@ it can clean up its state.
...
-responses as needed.
+`AggregationJobContinueReq` messages contain a `step` field to indicate the
+expected step of the DAP aggregation protocol their peer should be on. The
+inclusion of this field also allows the Leader to recover if it falls behind
+the Helper (e.g., Leader drops a response due to a transient network failure).
+In this situation, the Leader can re-send the `AggregationJobContinueReq`
+messages for previously computed steps. The Helper can respond by retransmitting
+the corresponding `AggregationJobResp` messages. By replaying the exact sequence
+of messages, the Leader can re-compute its previously dropped state and resume
+aggregation with the Helper from the most recent step.
```
cjpatton, line 1865: Let's make this a bit more normative.

> If the Leader re-sends an `AggregationJobContinueReq` from a previous step, the Helper SHOULD respond with the corresponding `AggregateShareResp`.

cjpatton:
Because we're not requiring the Helper to recover, I don't think we need to mandate any particular recovery strategy. This is an area where I think flexibility is good. One thing that would be useful is to know what step the Leader expects to recover. @inahga's [option 1](https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/pull/564#discussion_r1683276020) should make this easier.

branlwyd:
> Because we're not requiring the Helper to recover, I don't think we need to mandate any particular recovery strategy.

OK, I disagree here -- the downside of a mismatch in the Leader/Helper's recovery strategies may mean that recovery is impossible, effectively leading to the reports in the "bad" aggregation job being lost. (This would be the case e.g. if the Helper implements `n`/`n+1` recovery, while the Leader expects to be able to restart, and the VDAF is >1 step.) I see this as a step backwards in interoperability.

If a single strategy isn't acceptable, we _could_ protocol-negotiate how recovery works somehow -- but that is likely too much complexity to add to the protocol.

Or perhaps we could support both "restart from beginning" and "restart from previous step", but not "restart from arbitrary step" -- the storage requirements for supporting "restart from beginning" are a single request hash, while the storage requirements for "restart from arbitrary step" are either the entire sequence of responses or the entire sequence of VDAF preparation states. (This could be an implementation detail, of course -- but if this is the way we go, I'd like to avoid normative language that suggests the Helper store every step.)

kristineguo:
> In the model where the Leader restarts from anywhere in the aggregation process, the Helper needs to store the hash of every request, and every response (or enough information to recompute it based on the relevant request).

Is this not already the case? If a Helper supports idempotent init and cont request handling, it should store either state (to recompute the response) or the cached response itself so that it can re-transmit the previously computed AggregationJobResp at any point.

Or are you implying that once we advance to say step `n+2`, any requests to step `n` should be aborted?

> These differences might be an interoperability pain point
I honestly think the exact recovery mechanism can be an implementation detail worked out between aggregators. There's no need to be partial to any one of these strategies in the spec when all have equally valid use cases.

Keeping it purposefully vague (quoting Chris: `If the Leader re-sends an AggregationJobContinueReq from a previous step, the Helper SHOULD respond with the corresponding AggregationJobResp`) and using SHOULD language on how much previous state a Helper should persist should allow for variations.

branlwyd:
> > In the model where the Leader restarts from anywhere in the aggregation process, the Helper needs to store the hash of every request, and every response (or enough information to recompute it based on the relevant request).
> 
> Is this not already the case? If a Helper supports idempotent init and cont request handling, it should store either state (to recompute the response) or the cached response itself so that it can re-transmit the previously computed AggregationJobResp at any point.

Yes -- the existing specification language prior to this PR (and, FWIW, Janus' implementation) require the Helper to store state for only a single "old" step, specifically the one directly prior to the "current" step. This works if the Leader is expected to durably store the result of each step -- the Leader can only be one step "behind" due to transient failures, in this model.

Speaking as an implementor, I would like to avoid storing the state for every step of every aggregation job. These states can be quite large, and 

> Or are you implying that once we advance to say step `n+2`, any requests to step `n` should be aborted?

Yes, that is the current state of things (emphasis mine):

> To make that kind of recovery possible, Aggregator implementations SHOULD checkpoint the **most recent step's** prep state and messages to durable storage such that the Leader can re-construct continuation requests and the Helper can re-construct continuation responses as needed.

kristineguo:
If we want to limit step skew recovery to just `n`/`n+1`, we should make it more explicit in this section that the Leader is expected to store its state at each step.

But it seems that the amount of stored state is overall the same between Aggregators - either we make the Leader store state of each step (currently implied model), or the Helper stores the state of each step (this PR's suggestion). 

I had always assumed the helper's APIs were idempotent - the same request could be repeated independently of other requests that have been sent and still receive the same response as it had. You seem to be disagreeing with this. If a Helper has moved on to step `n+2`,, even if the same request is sent by the leader for step `n`, it will be aborted.

kristineguo:
Discussion was taken offline but to summarize here for posterity's sake:

**Conclusion:** **We will defer this issue.** The current status quo and what's proposed here are functionally the same for 1-round and 2-round VDAfs. Once more multi-round VDAFs are implemented and the async conversations (https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/pull/564) are converged, we can revisit this issue.

**Details:**
There are two options:
1. Stick with leader being the one to store state for every step and only supporting `n` / `n+1` recovery.
2. Move to the helper storing state (i.e., sufficient to reconstruct the same original response) and supporting step arbitrary recovery.

The current draft assumes that the leader takes the responsibility for tracking state. If this is maintained, only option 1 (the status quo) is required.

However, if we want to support aggregation job recovery from either the first step, we will need the helper to maintain state for every step. This is so that we can support idempotency: aggregators should ensure that each step that’s retried is the same as before. So supporting this recovery strategy requires per-step state storage on the helper side. 

**Tangent: Meaning of "Aggregated" Reports**
This discussion on "recovery from the beginning" brought up a relevant tangent on what it means for a report to have been "aggregated." @cjpatton mentioned that Daphne delays the anti-replay validation until right before the output shares are aggregated, at which point the job is "committed". Before this point, the reports could be sent in a brand new aggregation job if the leader wished to start over. This will be explored more in a separate issue by @branlwyd.

**Personal Notes**
I still personally prefer option 2, since it’s ultimately the same amount of state stored in sum between the aggregators and allows for more flexibility in recovery.

simon-friedberger:
@kristineguo Why does the leader have to store all steps anyway? If the leader receives a response for step x, then x can be discarded, right? Basically, we would need to store 3 steps. What the leader thinks the current step is. What the helper thinks the current step is and the previous step on the helper in case the leader hasn't received the response and retries.

branlwyd:
> @kristineguo Why does the leader have to store all steps anyway? If the leader receives a response for step x, then x can be discarded, right? Basically, we would need to store 3 steps. What the leader thinks the current step is. What the helper thinks the current step is and the previous step on the helper in case the leader hasn't received the response and retries.

I think that's correct. I think the summary of the situation discussed so far is as follows:

No matter what skew-recovery scheme is implemented, both the Leader & the Helper must store their view of the current aggregation state to enable "normal" (happy-path) operation of the protocol.

The information that must be additionally stored to enable a skew-recovery scheme differs by scheme:
* In `n`/`n+1` recovery, the Helper stores the hash of the previous request & the previous response (or enough other information to reconstruct the previous response from the previous request).
* In restart-from-beginning skew recovery, the Helper must store the hash of the original request.
* In arbitrary-step skew recovery, the Helper must store the hash of every previous request & every previous response (or enough other information to reconstruct that response from the associated request).

(I don't believe the Leader needs to store additional state for any of the skew-recovery schemes discussed so far.)

branlwyd:
Now that async aggregation has been merged, is there any interest in revisiting this issue? @kristineguo

To summarize the options considered so far:
* Restart-from-previous-step (`n`/`n+1`) recovery: in this scheme, the Helper stores the required information to allow the Leader to replay the last aggregation step (i.e. a hash of the most recent request, and the most recent response). The Leader can recover from failure by replaying the last request it sent. The Leader is expected to durably store the result of each step. This is what is specified in DAP today.
* Restart-from-beginning recovery: in this scheme, the Helper stores the required information to allow the Leader to restart an aggregation job (i.e. the hash of the initial request). The Leader can recover from failure by replaying the initial request, then continuing the aggregation process from the beginning. The Leader is expected to store the initial request until aggregation is complete.
* Restart-from-arbitrary-step recovery: in this scheme, the Helper stores the required information to allow the Leader to replay any aggregation step (i.e. a hash of each received request, and the response to that request). The Leader can recover from failure by replaying any request it can reconstruct.

My personal opinions/observations:

* I object to restart-from-arbitrary-step recovery, since (a) no one has justified why this might is desirable over the other options and (b) it has a much higher storage cost than the other options for multi-round VDAFs. **For the remainder of this message, I consider only restart-from-previous-step & restart-from-beginning recovery strategies.**
* Between the remaining two options, the storage requirements are about the same. The Helper must store a request hash for both strategies; in addition, the Helper must store a response in restart-from-previous-step, or the Leader must store the initial request in restart-from-beginning.
* The largest differentiator, in my opinion, is in the implication to the Leader storage: restart-from-previous-step assumes that the Leader will durably store the result of each aggregation step; restart-from-beginning assumes that the Leader will not do so, only updating durable storage once aggregation is complete.
* On the other hand, restart-from-previous-step requires less repeated work than restart-from-beginning when an error occurs. And the Helper's implementation for restart-from-beginning is likely to be slightly more complicated -- the Helper needs to track if an aggregation job has contributed to a batch's aggregate, so that it does not do so again on replay.

kristineguo:
Catching up on the recent async merges, I see this part in the merged change:
```
For further requests with
the same `AggregationJobInitReq` in the body, the Helper SHOULD respond as it
did for the original `AggregationJobInitReq`, or otherwise fail with an HTTP
client error status code.
```

With this, it seems the spec is now supporting restart-from-beginning, as the leader should be able to effectively restart the job by sending the same requests and receive its previous responses. Similarly, restart-from-previous-step is still using `MAY` language, so the draft seems to position restart-from-beginning as the first principle for recovering a job.

If I am understanding all this correctly, then I believe the current draft is sufficient for recovery, and am okay with dropping recover-from-arbitrary-step.

branlwyd:
Unfortunately, that's not the case -- see the final paragraph in the Helper Initialization section:

> Additionally, it is not possible to rewind or reset the state of an aggregation job. Once an aggregation job has been continued at least once (see [Section 4.7.2](https://ietf-wg-ppm.github.io/draft-ietf-ppm-dap/draft-ietf-ppm-dap.html#agg-continue-flow)), further requests to initialize that aggregation job MUST fail with an HTTP client error status code.

What's specified in DAP today is recover-from-previous-step (`n`/`n+1`) recovery, and only that strategy.

That said, I would personally support switching to restart-from-beginning; I think it is likely simpler to implement, and in particular simplifies the Helper. The requirement that the Leader store the initial request effectively only slightly extends the length of time that it must store reports, i.e. it must now store reports until the end of aggregation rather than the beginning of aggregation.

If you are interested in restart-from-beginning, I think the best next step would be to write text specifying it to allow review & discussion.

branlwyd:
Hi @kristineguo, do you have any interest in specifying restart-from-beginning semantics? We would like to publish DAP-12 soon, and if we take this change, we would like to include it in DAP-12.

kristineguo:
@branlwyd When is the deadline for DAP-12 changes? I can start to work on it but I'll be OOO 10/5 - 10/19

cjpatton:
@kristineguo We'd like to cut draft-12 next week some time so we can start implementation. I think it would be fine to delay this until draft-13, though we intended for that change to be primarily editorial. What do you think @branlwyd ?

branlwyd:
I am fine waiting for DAP-13. While this is not an editorial change, I think it is low-risk & will have a relatively small impact on implementations.

cjpatton:
2024/10/24 interim: We're uncomfortable with supporting recovery from any step, as we'd have to figure out how to deal with rewind-then-fork attacks on VDAF preparation as described [here](https://datatracker.ietf.org/meeting/interim-2024-ppm-02/materials/slides-interim-2024-ppm-02-sessa-dap-and-taskprov-open-issues-00). We'll give this a little more time to discuss, but if there are no new ideas by 2024/10/31, then we'll close this PR.

PR #573, opened 2024-08-28
Add `max_aggregation_job_report_count` task parameter.

branlwyd:
This parameter configures the maximum aggregation job size, allowing a standard way for the Helper to opt out of processing aggregation jobs which are too large.

### DIFF
--- a/draft-ietf-ppm-dap.md
+++ b/draft-ietf-ppm-dap.md
@@ -906,7 +906,9 @@ the following parameters associated with it:
   task. The task is considered completed after this time. Aggregators MAY reject
   reports that have timestamps later than `task_expiration`.
 * `max_aggregation_job_report_count`: The maximum number of reports which can be
-  included in a single aggregation job (see {{aggregate-flow}}).
+  included in a single aggregation job (see {{aggregate-flow}}). This parameter
+  is optional and may not be specified, in which case there is no specified
+  maximum aggregation job size.
 * A unique identifier for the VDAF in use for the task, e.g., one of the VDAFs
   defined in {{Section 10 of !VDAF}}.
 
@@ -1310,7 +1312,8 @@ of the task. The job ID is a 16-byte value, structured as follows:
 opaque AggregationJobID[16];
 ~~~
 
-The number of reports in the candidate set MUST be at most the
+If the task has specified a value for `max_aggregation_job_report_count`, the
+number of reports in the candidate set MUST be at most the
 `max_aggregation_job_report_count` configured for the task.
 
 The Leader can run this process for many sets of candidate reports in parallel
@@ -1500,9 +1503,10 @@ that the Leader will use to continue preparing the report.
 To begin this process, the Helper checks if it recognizes the task ID. If not,
 then it MUST abort with error `unrecognizedTask`.
 
-Next, the Helper checks that the number of `PrepareInit`s in the message is at
-most the `max_aggregation_job_report_count` configured for the task. If not,
-then it SHOULD abort with error `aggregationJobTooLarge`.
+Next, if the task has specified a value for `max_aggregation_job_report_count`,
+the Helper checks that the number of `PrepareInit`s in the message is at most
+the `max_aggregation_job_report_count` configured for the task. If this check
+fails, then the Helper SHOULD abort with error `aggregationJobTooLarge`.
 
 Next, the Helper checks that the report IDs in
 `AggregationJobInitReq.prepare_inits` are all distinct. If two preparation
--- a/draft-ietf-ppm-dap.md
+++ b/draft-ietf-ppm-dap.md
@@ -615,7 +615,7 @@ standard tokens for use in the "type" field (within the DAP URN namespace
 | unauthorizedRequest        | Authentication of an HTTP request failed (see {{request-authentication}}). |
 | stepMismatch               | The Aggregators disagree on the current step of the DAP aggregation protocol. |
 | batchOverlap               | A request's query includes reports that were previously collected in a different batch. |
-| aggregationJobTooLarge     | An aggregation request contained too many report shares. |
+| aggregationJobTooLarge     | An aggregation job contains more reports than the negotiated threshold. |
 
 This list is not exhaustive. The server MAY return errors set to a URI other
 than those defined above. Servers MUST NOT use the DAP URN namespace for errors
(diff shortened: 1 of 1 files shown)


branlwyd:
Some notes:
* The Leader behavior is a `MUST`, while the Helper behavior is a `SHOULD`. Since this parameter is intended to allow the Helper to protect itself from the Leader producing too-large aggregation jobs, I think this is acceptable. (Notably, this parameter does not need to be checked by both aggregators to achieve privacy/robustness properties, so it is at least _permissible_ for the Helper to ignore this check.)
* We could get away without specifying this at all -- the Helper could simply abort on a too-large aggregation job even without this protocol text. However, I think it is useful to specify this so that we have a standardized error for debugging purposes, and so that task-provisioning schemes like taskprov can eventually standardize a way to negotiate this parameter.

draft-ietf-ppm-dap.md
```diff
@@ -615,6 +615,7 @@ standard tokens for use in the "type" field (within the DAP URN namespace
 | unauthorizedRequest        | Authentication of an HTTP request failed (see {{request-authentication}}). |
 | stepMismatch               | The Aggregators disagree on the current step of the DAP aggregation protocol. |
 | batchOverlap               | A request's query includes reports that were previously collected in a different batch. |
+| aggregationJobTooLarge     | An aggregation request contained too many report shares. |
```
cjpatton, line 618: nit: not every aggregation request has report shares (only the first does).
1. If the intent is that we abort only on the initial request, then say "initial" here
2. (my preference): Use more agnostic language, e.g., "An aggregation job contains more reports than the negotiated threshold."

draft-ietf-ppm-dap.md
```diff
@@ -904,6 +905,8 @@ the following parameters associated with it:
 * `task_expiration`: The time up to which Clients are expected to upload to this
   task. The task is considered completed after this time. Aggregators MAY reject
   reports that have timestamps later than `task_expiration`.
+* `max_aggregation_job_size`: The maximum size, measured in number of reports,
```
cjpatton, line 908: suggestion, feel free to ignore: "count" to align with "report_count" elsewhere?
```suggestion
* `max_reports_per_aggregation_job_count`: The maximum size, measured in number of reports,
```

draft-ietf-ppm-dap.md
```diff
@@ -1494,6 +1500,8 @@ that the Leader will use to continue preparing the report.
 To begin this process, the Helper checks if it recognizes the task ID. If not,
 then it MUST abort with error `unrecognizedTask`.
 
+Next, the Helper checks that the number of `PrepareInit`s in the message is at most the `max_aggregation_job_size` configured for the task. If not, then it SHOULD abort with error `aggregationJobTooLarge`.
```
cjpatton, line 1503: nit: break lines at 80 columns.

draft-ietf-ppm-dap.md
branlwyd: Went with `max_aggregation_job_report_count`, only because it's slightly shorter and (IMO) equally comprehensible.

cjpatton (approved):
LGTM. Let's hold off on merging until we have rough consensus.

simon-friedberger:
Being explicit about this error seems like a good idea. Should there be an "unlimited" option? Since task parameter structs are not part of the spec there is no bit-size for this option and so it's not obvious what the value for "maximum" should be.

cjpatton:
> Being explicit about this error seems like a good idea. Should there be an "unlimited" option? Since task parameter structs are not part of the spec there is no bit-size for this option and so it's not obvious what the value for "maximum" should be.

We could treat `max_aggregation_job_report_count == 0` as the unlimited case.

branlwyd:
I agree that this parameter should be optional. Since there is no specified encoding of this parameter (or, more generally, for a task configuration), I think we don't need to worry about sentinel values -- we can just say the parameter is optional & specify what to do if it is not configured, as we used to with `max_batch_size`.

kristineguo:
I’m not sure this is fit for the task config. It makes clients privy to implementation-specific details for the aggregators and does not affect the actual task aggregation. I think it also lacks flexibility, as it pins the max job size for the entire lifetime of the task, preventing the leader from adjusting if needed. All in all, this feels too implementation specific to include into the spec.

branlwyd:
> It makes clients privy to implementation-specific details for the aggregators and does not affect the actual task aggregation.

The clients do not need to know this value -- only the aggregators do. (Also, I don't think "affects task aggregation" is a bar that task configuration values need to meet -- e.g. `collector_hpke_config` doesn't affect aggregation either, nor arguably do `min_batch_size` or `time_precision`.)

FWIW, I think it might be worthwhile for us to write down which parts of the task configuration each protocol participant needs to know. This is slightly complicated -- for example, technically speaking the client does not need to know the task's `min_batch_size` in order to participate in the protocol; but the client may nonetheless wish to know this value to decide whether to participate, as it affects "how private" the aggregations are. We should probably write down which task configuration values are technically required for the protocol to work, and note which other task configuration values each participant might be interested in knowing.

> I think it also lacks flexibility, as it pins the max job size for the entire lifetime of the task, preventing the leader from adjusting if needed.

I think this is a fair concern -- this is intended as a configuration knob to allow the Helper to assert a maximum aggregation job size it is willing to aggregate. The Leader can send smaller jobs if it wishes, but there is no way for the Leader & Helper to negotiate a larger value if they wish to do so outside of creating an entirely new measurement task.

kristineguo:
The way it's framed in the spec, the `collector_hpke_config` and `vdaf_verify_key` are configured by the aggregators separately and are not present in the actual task params. Furthermore, for extensions like task binding and in band task provisioning, clients will be creating the TaskConfig struct. It feels weird that they may be able to dictate the behavior of the aggregators.

Thus, if this `max_aggregation_job_report_count` is to be included, it should be treated similar to the `collector_hpke_config` and `vdaf_verify_key` params and configured by aggregators separately from the task itself.

branlwyd:
> The way it's framed in the spec, the `collector_hpke_config` and `vdaf_verify_key` are configured by the aggregators separately and are not present in the actual task params. Furthermore, for extensions like task binding and in band task provisioning, clients will be creating the TaskConfig struct. It feels weird that they may be able to dictate such specific implementation details of the aggregators.
> 
> Thus, if this `max_aggregation_job_report_count` is to be included, it should be treated similar to the `collector_hpke_config` and `vdaf_verify_key` params and configured by aggregators separately from the task itself.

I think you are correct -- for (eg) `taskprov`, including this parameter in the `TaskConfig` would give the clients too much control over the aggregators' behavior. The parameter could be included in the parameters configured out-of-band (likely parameterized per VDAF configuration), but this would significantly reduce the utility of this parameter.

It might be useful for other task-provisioning schemes that give more control to the aggregators in negotiating task parameters. But I don't think we have any in hand, certainly not any going through standardization.

I have sent #576 which instead adds an operational consideration that maximum aggregation job size should be negotiated by the aggregators somehow, without specifying a parameter -- perhaps that will turn out to be sufficient, and this PR can be withdrawn.

cjpatton:
Thanks @branlwyd, this resolution makes sense to me. Let's close this PR -- we can re-open if anyone feels strongly about it.

PR #574, opened 2024-08-30
Simplify report replay checking.

### LINKED ISSUE #557: Explicit backoff logic for aggregation jobs
As mentioned in #556, replay protection implies a relatively stringent operational requirement for the Helper: many aggregation jobs might make concurrent transactions on the same database, which can easily overwhelm the database if not sufficiently provisioned.

To mitigate this problem, the Helper can cancel an aggregation job and ask the Leader to try again later. One option is to respond with HTTP status 429 and a "retry-after" header indicating when it should be safe to retry.

I'd like to suggest that we spell this out explicitly in the draft.

branlwyd:
Closes #557.

Specifically, we relax the check from the specific report ID, to intersection with intervals of report IDs. This will allow implementations to drop all long-term per-report storage, considerably reducing the storage requirements to implement the protocol. The tradeoff is in increased aggregation latency: typically, aggregation cannot begin until the Leader is sure that no further reports will arrive for a given time window.


branlwyd:
A few thoughts:
* This specifies proposal (2a) from [Bearing the cost of aggregation](https://docs.google.com/document/d/1LNzDEzY5CUlOlOjzk5Z0fkECDF-5gukV-SVwC1gj6Wo/edit#bookmark=id.c0uo88am22oz), extended to treat the leader-selected batch mode uniformly with the time-interval batch mode (as described in [a comment](https://docs.google.com/document/d/1LNzDEzY5CUlOlOjzk5Z0fkECDF-5gukV-SVwC1gj6Wo/edit?disco=AAABUyvf-TY)).
* I am not so happy with the name "report ID range" for the new data structure described in the PR, as the structure is actually a `(timestamp, report ID interval)`. Thoughts on naming welcome.
* I did a small amount of editorial work in the sections I was already touching; please review closely.

draft-ietf-ppm-dap.md
```diff
@@ -1283,8 +1286,11 @@ subsections.
 
 In general, reports cannot be aggregated until the Collector specifies an
 aggregation parameter. However, in some situations it is possible to begin
-aggregation as soon as reports arrive. For example, Prio3 has just one valid
-aggregation parameter (the empty string).
+aggregation before a collection request arrives. For example, Prio3 has just one
+valid aggregation parameter (the empty string); or, perhaps the relevant
+aggregation parameter can be determined in a deployment-specific way. In these
+cases, aggregation can begin as soon as the Leader is confident that no
```
simon-friedberger, line 1292: Maybe be explicit about the requirement instead "aggregation can begin as soon as the leader is willing to discard any further incoming reports for this time window."

draft-ietf-ppm-dap.md
```diff
@@ -1307,6 +1313,10 @@ of the task. The job ID is a 16-byte value, structured as follows:
 opaque AggregationJobID[16];
 ~~~
 
+The set of candidate reports MUST produce a set of report ID ranges which does
```
simon-friedberger, line 1316: I agree that calling the new thing a "report ID range" makes everything hard to read. This sentence looks like it is comparing report ID ranges but it is actually comparing "report ID ranges".

Considering the fact that you're saying "a set of" here. Does that mean the old behavior would still be allowed by making each range have exactly one report ID?

simon-friedberger: To clarify, I am wondering why anybody would want to do this given that the report IDs are randomly assigned. Is the intent to allow people to recover from some amount of late arriving reports?

cjpatton, line 1316: How strict do you think we should be? As we discussed on slack today, at a minimum we want the helper to be able to reject any range that overlaps with an already aggregated range. But the helper could make a smarter decision by "merging" the two ranges and rejecting only the overlap, in which case the MUST here would not be enforced by the helper.

draft-ietf-ppm-dap.md
```diff
@@ -1715,6 +1718,33 @@ following checks:
...
+interval of report IDs (using lexicographic ordering), along with an associated
+timestamp.
+
+A set of reports for a task determines a set of report ID ranges. The report ID
+ranges associated with a given set of reports are determined by the following
+procedure:
+
+1. Round each report's timestamp down to the nearest multiple of
+   `time_precision`, and partition the set of reports into groups based on this
+   rounded timestamp.
```
cjpatton, line 1733: Let's make sure this is harmonized with the rest of the draft. I believe we say "truncate", not "round". However I haven't checked.

draft-ietf-ppm-dap.md
```diff
@@ -2411,6 +2430,37 @@ to the end as needed to produce the final aggregate output.
 Collectors must also maintain state for the lifetime of each task, which
 includes key material associated with the HPKE key configuration.
 
+## Aggregation Jobs & Replay Protection
```
cjpatton, line 2433: Note that we never use "&" in place of "and" in this draft.
```suggestion
## Aggregation Jobs and Replay Protection
```

draft-ietf-ppm-dap.md
```diff
@@ -2411,6 +2430,37 @@ to the end as needed to produce the final aggregate output.
...
+
+### Leader
+
+As part of the aggregation protocol, the Leader is responsible for generating
```
cjpatton, line 2437: nit: consistency. I'd be cool with saying "protocol" instead of "sub-protocol", but let's track in an issue.
```suggestion
As part of the aggregation sub-protocol, the Leader is responsible for generating
```

draft-ietf-ppm-dap.md
```diff
@@ -2411,6 +2430,37 @@ to the end as needed to produce the final aggregate output.
...
+aggregation jobs. The Leader must produce aggregation jobs using reports with
+non-overlapping intervals of report IDs for each time window (see {{agg-init}}).
```
cjpatton, line 2439: nit: consistency. FWIW I weakly prefer "overlap" to "intersect".
```suggestion
non-intersecting intervals of report IDs for each time window (see {{agg-init}}).
```

cjpatton, line 2439: Let's make sure "time window" is defined in the terms section of the doc

draft-ietf-ppm-dap.md
```diff
@@ -2459,23 +2509,12 @@ designed to allow implementations to reduce operational costs in certain cases.
 In general, the Aggregators are required to keep state for tasks and all valid
 reports for as long as collection requests can be made for them. However, it is
 not necessary to store the complete reports. Each Aggregator only needs to store
-an aggregate share for each possible batch interval (for time-interval) or batch
-ID (for leader-selected), along with a flag indicating whether the aggregate
-share has been collected. This is due to the requirement that in the
+an aggregate share for each batch, along with a flag indicating whether the
+aggregate share has been collected. This is due to the requirement that in the
```
cjpatton, line 2513: I believe the old text still applies: a batch interval is a sequence of time windows; and the batch interval is not known at aggregation time.

draft-ietf-ppm-dap.md
```diff
@@ -2459,23 +2509,12 @@ designed to allow implementations to reduce operational costs in certain cases.
...
 
-However, Aggregators are also required to implement several per-report checks
-that require retaining a number of data artifacts. For example, to detect replay
-attacks, it is necessary for each Aggregator to retain the set of report IDs of
-reports that have been aggregated for the task so far. Depending on the task
-lifetime and report upload rate, this can result in high storage costs. To
-alleviate this burden, DAP allows Aggregators to drop this state as needed, so
-long as reports are dropped properly as described in {{input-share-validation}}.
-Aggregators SHOULD take steps to mitigate the risk of dropping reports (e.g., by
-evicting the oldest data first).
```
cjpatton, line 2477: I think the core idea here still applies: it's just that we're storing report ranges rather than IDs.

draft-ietf-ppm-dap.md
branlwyd: We actually use both! I prefer `truncate` so I switched to that.

draft-ietf-ppm-dap.md
branlwyd: aagh, sorry, I keep making that mistake

draft-ietf-ppm-dap.md
branlwyd: 👍🏻 It turned out there were a bunch of places that got this wrong -- I fixed them all up. There were a few places where it was ambiguous; for the most part I made as few changes as possible while still referring to "sub-protocols" when discussing only the upload/aggregation/collection portions of DAP.

draft-ietf-ppm-dap.md
branlwyd: Hmm, I think the old text was slightly inaccurate for the time-interval case: aggregators don't have to store an aggregate share for each possible a batch interval (i.e., sequence of time windows) -- an aggregator needs to store an aggregate share for each _time window_.

I had felt that distinguishing between the batch modes was not so useful here; I restored the text, modified for accuracy.

draft-ietf-ppm-dap.md
branlwyd: OK -- I dropped this because we'll now be storing perhaps a few thousand ID ranges per time window, but no objection to keeping this text around.

draft-ietf-ppm-dap.md
branlwyd: I think a `MUST` is appropriate here for the Leader: the Leader must follow this suggestion to avoid having otherwise-acceptable reports rejected as a replay.

The Helper can indeed make a few different choices about how to deal with an aggregation job which does intersect with a previous aggregation job. The relevant text for the Helper is currently:

> Check if the report has been previously aggregated. If so, the input share MUST be marked as invalid with the error `report_replayed`. A report is considered previously aggregated if it intersects with a report ID range for a completed aggregation job (see {{report-id-ranges}}).

... which implies to me that we have to effectively allow "merging" of ranges if they intersect, in the case that an intersecting aggregation job is received. (A Helper which does something more specific, e.g. marking individual report IDs as used, would work to avoid replays; but I'm not sure a Leader could take advantage of this without knowing the specific strategy used by the Helper.)

I'm personally happy with this level of flexibility, but maybe the Helper should be more be more flexible? If so, we might consider relaxing the Leader's MUST here.

branlwyd: > I agree that calling the new thing a "report ID range" makes everything hard to read. This sentence looks like it is comparing report ID ranges but it is actually comparing "report ID ranges".

Indeed; I am careful to use the term "interval" where an "interval" is truly needed. But "report ID range" does not capture that a "report ID range" includes a time window, too.

> Considering the fact that you're saying "a set of" here. Does that mean the old behavior would still be allowed by making each range have exactly one report ID?

Not quite -- an aggregation job may include reports in multiple time windows. In this case, there would be one "report ID range" for each time window represented in the aggregation job.

> To clarify, I am wondering why anybody would want to do this given that the report IDs are randomly assigned. Is the intent to allow people to recover from some amount of late arriving reports?

Nope, I think late-arriving reports are effectively lost (with high probability) in this scheme. The "report ID ranges" in use are all "maximal" in the sense that they stretch from the smallest to the largest available report ID.

cjpatton: I prefer MUST here as well.

draft-ietf-ppm-dap.md
cjpatton: I'm just suggesting we update the text.

draft-ietf-ppm-dap.md
simon-friedberger: I think if the "set" only covers the "different time intervals" part you need to clarify. The way it's written the ranges don't need to be maximal. Somebody could send 3 ranges for t1 and 7 ranges for t2 as longs as they in total cover all the reports. And in the limit, send one range per report and everything turns into linear search again.

draft-ietf-ppm-dap.md
```diff
@@ -399,6 +402,11 @@ Report:
 Report Share:
 : An encrypted input share comprising a piece of a report.
 
+Time Window:
+: An interval of time associated with a measurement task, whose duration is
+  equal to the task's time precision and whose start has a UNIX timestamp value
+  which is a multiple of the task's time precision.
```
cjpatton, line 408: I might also mention briefly what this is used for (replay protection for one!)

draft-ietf-ppm-dap.md
```diff
@@ -1715,6 +1723,33 @@ following checks:
 
 If all of the above checks succeed, the input share is not marked as invalid.
 
+#### Report ID ranges {#report-id-ranges}
+
+Replay checking is implemented using "report ID ranges". A report ID range is an
```
cjpatton, line 1728:
```suggestion
Replay protection is implemented using "report ID ranges". A report ID range is an
```

cjpatton (approved):

branlwyd:
I renamed "report ID range" to "report ID commitment", in an attempt to alleviate confusion around the fact that a "report ID range" contained an "interval of report IDs" (but also a timestamp). Along with this change, a report ID commitment "conflicts" (rather than "intersects") with another report ID commitment.

draft-ietf-ppm-dap.md
branlwyd: I think the ranges do need to be maximal; there actually isn't much wiggle room here. 

The relevant Leader text is:

> The set of candidate reports MUST produce a set of report ID commitments which
does not conflict with the report ID commitments for any other aggregation job
associated with the task (see {{report-id-commitments}}).

[note that "report ID range" got renamed to "report ID commitment" here]

The relevant Helper text is:

> 1. Check if the report has been previously aggregated. If so, the input share
   MUST be marked as invalid with the error `report_replayed`. A report is
   considered previously aggregated if it conflicts with a report ID commitment
   for a completed aggregation job (see {{report-id-commitments}}).

That is, both fall back to the report-id-commitments section; this section specifies a maximal range:

> The report ID commitments associated with a given set of reports are determined by the following procedure:
> 
> 1. Partition the set of reports into groups based on the time window their
>    client timestamp falls into; this might be achieved by partitioning the
>    reports on a key generated by truncating each report's timestamp down to the
>    nearest multiple of the task's `time_precision`.
> 2. For each partitioned group of reports, generate an interval from the smallest
>    report ID in the group to the largest report ID in the group, inclusive. The
>    report ID commitment for this partition is this interval, associated with the
>    time window that generated this partition.

I had previously believed there was some wiggle room for what the Helper commits if the Leader fails in its `MUST` & sends an aggregation job containing a report which falls into an existing report ID commitment. However, now I think the Helper's behavior is fully specified here with no wiggle room:

* Any reports conflicting with an existing report ID commitment must be marked as `report_replayed`.
* The new commitment associated with the aggregation job is still the maximal report ID commitments suggested by the report-id-commitments section -- this commitment will conflict with an already-existing commitment, but that's OK. (A clever implementation might merge conflicting commitments without changing its externally-observable behavior, but this isn't required.)

branlwyd: By the way -- I'm interested in your thoughts on the new text in the report-id-commitments section @simon-friedberger. Other than renaming "report ID ranges" to "report ID commitments", I tried to make it clearer that the set of report ID commitments is one-per-time-window, rather than allowing the aggregators to generate a larger count of smaller commitment intervals.

mendess:
I think this change is very positive to the protocol, provided the leader is okay with the latency cost it will pay when using leader-selected. But I think one final tweak should be made

### Problem

Lets assume the storage solution for replay protection is sharding on time the time window of reports. Meaning storage transactions only contend when they are trying to check for replays of two time windows.

If two Aggregation Jobs have reports belonging to the same time window then they will contend on the storage shard used. This gets even worse if the aggregation jobs have a "mixed bag" of time windows, which increases the chance that two jobs might contend on the time window.

### Suggested change

I think the protocol should encourage leaders (non normatively) to avoid having multiple aggregation jobs "share time windows". Put another way: It might create higher resource contention when N aggregation jobs contain reports belonging to the same time window.

draft-ietf-ppm-dap.md
```diff
@@ -613,7 +622,7 @@ standard tokens for use in the "type" field (within the DAP URN namespace
 | batchQueriedMultipleTimes  | A batch was queried with multiple distinct aggregation parameters. |
 | batchMismatch              | Aggregators disagree on the report shares that were aggregated in a batch. |
 | unauthorizedRequest        | Authentication of an HTTP request failed (see {{request-authentication}}). |
-| stepMismatch               | The Aggregators disagree on the current step of the DAP aggregation protocol. |
+| stepMismatch               | The Aggregators disagree on the current step of the DAP aggregation sub-protocol. |
```
tgeoghegan, line 625: Orthongonal nit: back when we got a review from HTTPDIR, we got a recommendation to use the term "interaction" or "sub-interaction" instead of "sub-protocol". Mostly this was cleaned up in #450, but I appear to have missed this one.

draft-ietf-ppm-dap.md
```diff
@@ -910,8 +919,8 @@ the following parameters associated with it:
 Note that the `leader_aggregator_url` and `helper_aggregator_url` values may
 include arbitrary path components.
 
-In addition, in order to facilitate the aggregation and collect protocols, each
-of the Aggregators is configured with following parameters:
+In addition, in order to facilitate the aggregation and collection
+sub-protocols, each of the Aggregators is configured with following parameters:
```
tgeoghegan, line 923: As above: prefer "interaction" to "sub-protocol", per HTTPDIR's advice.

draft-ietf-ppm-dap.md
```diff
@@ -1049,7 +1058,7 @@ struct {
       number generator.
 
     * `time` is the time at which the report was generated. The Client SHOULD
-      round this value down to the nearest multiple of the task's
+      truncate this value down to the nearest multiple of the task's
```
tgeoghegan, line 1061: I don't object to using the verb "truncate" here (is the precedent [Go's `time.Truncate`](https://pkg.go.dev/time#Time.Truncate)?), but "truncate [...] down" is awkward. There's no such thing as "truncating up" so why say "truncate down"?
```suggestion
      truncate this value to the nearest multiple of the task's
```

draft-ietf-ppm-dap.md
```diff
@@ -1147,25 +1156,25 @@ ignore it. In addition, it MAY alert the Client with error `reportRejected`. See
 the implementation note in {{input-share-validation}}.
 
 The Leader MUST ignore any report pertaining to a batch that has already been
-collected (see {{input-share-validation}} for details). Otherwise, comparing
-the aggregate result to the previous aggregate result may result in a privacy
+collected (see {{input-share-validation}} for details). Otherwise, comparing the
+aggregate result to the previous aggregate result may result in a privacy
 violation. Note that this is also enforced by the Helper during the aggregation
-interaction. The Leader MAY also abort the upload protocol and alert the
+interaction. The Leader MAY also abort the upload sub-protocol and alert the
```
tgeoghegan, line 1162: As above: prefer "interaction" to "sub-protocol", per HTTPDIR's advice.

draft-ietf-ppm-dap.md
```diff
@@ -1147,25 +1156,25 @@ ignore it. In addition, it MAY alert the Client with error `reportRejected`. See
...
 Client with error `reportRejected`.
 
 The Leader MAY ignore any report whose timestamp is past the task's
-`task_expiration`. When it does so, it SHOULD also abort the upload protocol and
-alert the Client with error `reportRejected`. Client MAY choose to opt out of
-the task if its own clock has passed `task_expiration`.
+`task_expiration`. When it does so, it SHOULD also abort the upload sub-protocol
```
tgeoghegan, line 1166: As above: prefer "interaction" to "sub-protocol", per HTTPDIR's advice.

draft-ietf-ppm-dap.md
```diff
@@ -1147,25 +1156,25 @@ ignore it. In addition, it MAY alert the Client with error `reportRejected`. See
...
 
 The Leader may need to buffer reports while waiting to aggregate them (e.g.,
 while waiting for an aggregation parameter from the Collector; see
 {{collect-flow}}). The Leader SHOULD NOT accept reports whose timestamps are too
 far in the future. Implementors MAY provide for some small leeway, usually no
 more than a few minutes, to account for clock skew. If the Leader rejects a
-report for this reason, it SHOULD abort the upload protocol and alert the Client
-with error `reportTooEarly`. In this situation, the Client MAY re-upload the
-report later on.
+report for this reason, it SHOULD abort the upload sub-protocol and alert the
```
tgeoghegan, line 1175: As above: prefer "interaction" to "sub-protocol", per HTTPDIR's advice.

draft-ietf-ppm-dap.md
```diff
@@ -1236,12 +1245,12 @@ themselves: if preparation succeeds, then the resulting output shares are
 guaranteed to combine into a valid, refined measurement.
 
 VDAF preparation is mapped onto an aggregation job as illustrated in
-{{agg-flow}}. The protocol is comprised of a sequence of HTTP requests from the
-Leader to the Helper, the first of which includes the aggregation parameter, the
-Helper's report share for each report in the job, and for each report the
-initialization step for preparation. The Helper's response, along with each
-subsequent request and response, carries the remaining messages exchanged during
-preparation.
+{{agg-flow}}. The aggregation sub-protocol is comprised of a sequence of HTTP
```
tgeoghegan, line 1248: As above: prefer "interaction" to "sub-protocol", per HTTPDIR's advice.

draft-ietf-ppm-dap.md
```diff
@@ -1715,6 +1724,39 @@ following checks:
 
 If all of the above checks succeed, the input share is not marked as invalid.
 
+#### Report ID commitments {#report-id-commitments}
+
+Replay protection is implemented using "report ID commitments". A report ID
+commitment is an interval of report IDs (using lexicographic ordering),
```
tgeoghegan, line 1730: An interval of time makes sense, but an interval of report IDs is a strange notion.
```suggestion
commitment is an ordered set of report IDs (using lexicographic ordering),
```

draft-ietf-ppm-dap.md
```diff
@@ -1715,6 +1724,39 @@ following checks:
...
+which is a multiple of the task's `time_precision`).
+
+A set of reports for a task determines a set of report ID commitments, producing
+one commitment per time window represented in the set of reports. The report ID
+commitments associated with a given set of reports are determined by the
+following procedure:
+
+1. Partition the set of reports into groups based on the time window their
+   client timestamp falls into; this might be achieved by partitioning the
+   reports on a key generated by truncating each report's timestamp down to the
```
tgeoghegan, line 1742:
```suggestion
   reports on a key generated by truncating each report's timestamp to the
```

draft-ietf-ppm-dap.md
```diff
@@ -1898,7 +1940,7 @@ it can clean up its state.
 
 `AggregationJobContinueReq` messages contain a `step` field, allowing
 Aggregators to ensure that their peer is on an expected step of the DAP
-aggregation protocol. In particular, the intent is to allow recovery from a
+aggregation sub-protocol. In particular, the intent is to allow recovery from a
```
tgeoghegan, line 1943: As above: prefer "interaction" to "sub-protocol", per HTTPDIR's advice.

draft-ietf-ppm-dap.md
```diff
@@ -2378,39 +2421,71 @@ Helpers are generally required to:
...
 - Publish and manage an HPKE configuration that can be used for the upload
-  protocol.
-
-In addition, for each DAP task, the Helper is required to:
-
-- Implement some form of batch-to-report index, as well as inter- and
-  intra-batch replay mitigation storage, which includes some way of tracking
-  batch report size. Some of this state may be used for replay attack
-  mitigation. The replay mitigation strategy is described in
+  sub-protocol; and
```
tgeoghegan, line 2424: As above: prefer "interaction" to "sub-protocol", per HTTPDIR's advice.

draft-ietf-ppm-dap.md
```diff
@@ -2378,39 +2421,71 @@ Helpers are generally required to:
...
+- Implement and store state for the replay mitigation in
   {{input-share-validation}}.
 
 Beyond the minimal capabilities required of Helpers, Leaders are generally
 required to:
 
-- Support the upload protocol and store reports; and
+- Support the upload sub-protocol and store reports; and
```
tgeoghegan, line 2431: As above: prefer "interaction" to "sub-protocol", per HTTPDIR's advice.

draft-ietf-ppm-dap.md
```diff
@@ -2378,39 +2421,71 @@ Helpers are generally required to:
...
 ### Collector Capabilities
 
 Collectors statefully interact with Aggregators to produce an aggregate output.
 Their input to the protocol is the task parameters, configured out of band,
-which include the corresponding batch window and size. For each collect
-invocation, Collectors are required to keep state from the start of the protocol
-to the end as needed to produce the final aggregate output.
+which include the corresponding batch window and size. For each collection
+invocation, Collectors are required to keep state from the start of the
+sub-protocol to the end as needed to produce the final aggregate output.
```
tgeoghegan, line 2441: As above: prefer "interaction" to "sub-protocol", per HTTPDIR's advice.

draft-ietf-ppm-dap.md
```diff
@@ -2691,7 +2766,7 @@ chosen independently of the generated reports.
 An important parameter of a DAP deployment is the minimum batch size. If a batch
 includes too few reports, then the aggregate result can reveal information about
 individual measurements. Aggregators enforce the agreed-upon minimum batch size
-during the collection protocol, but implementations SHOULD also opt out of
+during the collection sub-protocol, but implementations SHOULD also opt out of
```
tgeoghegan, line 2769: As above: prefer "interaction" to "sub-protocol", per HTTPDIR's advice.

tgeoghegan (commented):
This is an editorial review, and I will defer to those who are knee-deep in implementation on whether the change is a good idea.

cjpatton:
Closed in favor of #564. The Leader must wait until the current time window has elapsed before it can begin aggregating. This means the latency of collecting the aggregate result depends on the `time_precision` parameter for the task, which maybe something like an hour. See https://mailarchive.ietf.org/arch/msg/ppm/mPYFQ9-AjfbSZ5r67gJyJ4cSaBw/ for details.

PR #578, opened 2024-09-17
tls-presentation: Fix limits for length-prefixed fields

### LINKED ISSUE #110: TODO: Fix the bounds for length-prefixed parameters in protocol messages
.(E.g., `<23..479>` instead of `<1..2^16-1>`.)]

cjpatton:
Closes #110.

Assume 10 bytes for the smallest VDAF ping pong message: https://github.com/cfrg/draft-irtf-cfrg-vdaf/pull/428

* HpkeConfigList: The smallest HpkeConfig is 10 bytes: one byte for the
  config ID, two bytes for each algorithm ID, 2 bytes for the public key
  length prefix, and assume one byte for the public key (the actual
  value depends on which HPKE KEM algorithms are supported).

* PlaintextInputShare.payload: Assume 1 byte as the smallest input share
  (the actual value depends on which VDAFs are supported).

* PrepareInit.payload: 5 bytes lest ping pong message (see above)

* AggregationJobInitReq.prepare_inits: The smallest PrepareInit is 45
  bytes:

   * PrepareInit.report_share: 16 bytes for the report ID, 8 bytes for
     the timestamp, 0 bytes for the public share (this is empty for some
     Prio3 variants), 2 bytes for the public share length prefix, and
     for the HPKE ciphertext:

      * 1 for the config ID

      * assume 1 byte for the encapsulated key (depends on the supported
        KEMs) and 2 bytes for the length prefix

      * assume 1 byte for the ciphertext (depends on the supported
        AEADs) and 4 bytes for the length prefix

   * PrepareInit.payload: 5 bytes (smallest ping pong message) and 4
     bytes for the length prefix

* PrepareResp.payload: 5 bytes (smallest ping poing message)

* AggregationJobJesp.prepare_resps: The smallest PrepareResp has 26 bytes:

    * 16 bytes for the report ID, 1 byte for the variant discriminator,
      and 5 + 4 bytes length prefix for the ping pong message.

* PrepareContinue.payload: 5 bytes (smallest ping pong message)

* AggregationJobContinueReq.prepare_continues: The smallest
  PRepareContinue is 16 bytes for the report ID, and 5 + 4 bytes for
  the payload.

### DIFF
--- a/draft-ietf-ppm-dap.md
+++ b/draft-ietf-ppm-dap.md
@@ -975,7 +975,7 @@ decreasing order of preference. This allows an Aggregator to support multiple
 HPKE configurations simultaneously.
 
 ~~~
-HpkeConfig HpkeConfigList<1..2^16-1>;
+HpkeConfig HpkeConfigList<10..2^16-1>;
 
 struct {
   HpkeConfigId id;
@@ -1093,7 +1093,7 @@ The Client then wraps each input share in the following structure:
 ~~~
 struct {
   Extension extensions<0..2^16-1>;
-  opaque payload<0..2^32-1>;
+  opaque payload<1..2^32-1>;
 } PlaintextInputShare;
 ~~~
 
@@ -1370,7 +1370,7 @@ struct {
 
 struct {
   ReportShare report_share;
-  opaque payload<0..2^32-1>;
+  opaque payload<5..2^32-1>;
 } PrepareInit;
 ~~~
 
@@ -1405,7 +1405,7 @@ struct {
 struct {
   opaque agg_param<0..2^32-1>;
   PartialBatchSelector part_batch_selector;
-  PrepareInit prepare_inits<1..2^32-1>;
+  PrepareInit prepare_inits<44..2^32-1>;
 } AggregationJobInitReq;
 ~~~
 
@@ -1528,7 +1528,7 @@ struct {
   ReportID report_id;
   PrepareRespState prepare_resp_state;
   select (PrepareResp.prepare_resp_state) {
-    case continue: opaque payload<0..2^32-1>;
+    case continue: opaque payload<5..2^32-1>;
     case finished: Empty;
     case reject:   PrepareError prepare_error;
   };
@@ -1592,7 +1592,7 @@ Otherwise the Helper responds with
 variant {
   ReportID report_id;
   PrepareRespState prepare_resp_state = continue;
-  opaque payload<0..2^32-1> = outbound;
+  opaque payload<5..2^32-1> = outbound;
 } PrepareResp;
 ~~~
 
@@ -1601,7 +1601,7 @@ message is structured as follows:
 
 ~~~
 struct {
-  PrepareResp prepare_resps<1..2^32-1>;
+  PrepareResp prepare_resps<26..2^32-1>;
 } AggregationJobResp;
 ~~~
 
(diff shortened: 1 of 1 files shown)


draft-ietf-ppm-dap.md
```diff
@@ -1370,7 +1370,7 @@ struct {
 
 struct {
   ReportShare report_share;
-  opaque payload<0..2^32-1>;
+  opaque payload<10..2^32-1>;
```
branlwyd, line 1373: I think the 10-byte minimum is incorrect; see https://github.com/cfrg/draft-irtf-cfrg-vdaf/pull/428/files#r1765155695.

branlwyd (commented):
I think we might be better served using a more lax set of ranges, perhaps `<0..MAX>` for fields which can be empty & `<1..MAX>` for fields which cannot be empty.

I think the benefit of specifying to-the-byte minimums for fields is pretty marginal: I think malformed messages will fail to parse anyway. And the risk is that we get the minimum wrong, and risk rejecting messages that would otherwise be accepted.

draft-ietf-ppm-dap.md
cjpatton: Done, good catch.

cjpatton:
> @branlwyd  I think we might be better served using a more lax set of ranges, perhaps `<0..MAX>` for fields which can be empty & `<1..MAX>` for fields which cannot be empty.
> 
> I think the benefit of specifying to-the-byte minimums for fields is pretty marginal: I think malformed messages will fail to parse anyway. And the risk is that we get the minimum wrong, and risk rejecting messages that would otherwise be accepted.

I agree with you, but I think our hands are tied here. The intent of [RFC 8446, Section 3.4](https://datatracker.ietf.org/doc/html/rfc8446#section-3.4) is pretty clear that the lower limit is supposed to be the size of the smallest vector. There are examples of this all over RFC 8446.

tgeoghegan:
I agree with Brandon that sprinkling constants like this throughout the document isn't ideal. Even in the context of your reading of RFC 8446, I think we can get away with using `0` as the minimum size in all cases in this PR.

I think we can imagine a degenerate testing VDAF for which things like input or output shares could have length 0. Imagine a dummy VDAF where the leader gets the entire input and the helper gets an empty byte string as its input share. On that basis, I think we can justify making the minimum length of various `payload` fields 0.

Any list of objects (i.e., `HpkeConfigList` or the `prepare_inits` in an aggregation job) _could_ be empty. Those messages are useless in the protocol, of course, but I think they're valid encodings of messages on the wire. In particular, I worry about nudging implementations into checking things at the wrong layer.

If the protocol text asserts that field `prepare_inits` of structure `AggregationJobInitReq` will always be at least 44 bytes, then implementations will want to verify this in their message decoding routines at the transport layer. When the resulting decoded structure makes it to the DAP application logic, it probably looks like a `Vec<AggregationJobInitReq>` (or `ArrayList<AggregationJobInitReq>` or `[]AggregationJobInitReq` or whatever your preferred language's syntax is). In a robust implementation, the application logic receiving that list needs to be ready for the case where the incoming list/vector/slice is empty, regardless of what guarantees the transport or message decode layer might provide. So I think we might as well allow empty lists in the message syntax, such that implementations are guided toward handling this error/edge case in only one place, which is easier to test and maintain.

I think what I'm describing is essentially consistent with what MLS, which also uses an extended RFC 8446 syntax, already does. That document extensively uses a notation [`opaque variable<V>`](https://datatracker.ietf.org/doc/html/rfc9420#section-2.1.2), which allows a vector of 0..2^30 bytes. However I'm not suggesting we adopt their variable length vector encoding, just citing it as prior art.

Long story short: I think we should use 0 as the minimum bound for all messages.

cjpatton:
Alright, we have some disagreement here, so I'll raise this on the list. Hopefully we can settle it at the interim.

Issue #385, opened 2022-11-29
Pre-Aggregation into buckets


simon-friedberger:
We are indicating that batches can already start aggregating before the collector sends a query at least in three places:

2.1
> The leader distributes the shares to the helpers and orchestrates the process of verifying them (see [Section 2.2](https://ietf-wg-ppm.github.io/draft-ietf-ppm-dap/draft-ietf-ppm-dap.html#validating-inputs)) and assembling them into a final measurement for the collector. Depending on the VDAF, it may be possible to incrementally process each report as it comes in, or may be necessary to wait until the entire batch of reports is received.

4.4
> Once a set of clients have uploaded their reports to the leader, the leader can send them to the helpers to be verified and aggregated. In order to enable the system to handle very large batches of reports, this process can be performed incrementally.

4.5.2
> Note that for most VDAFs, it is possible to aggregate output shares as they arrive rather than wait until the batch is collected. To do so however, it is necessary to enforce the batch parameters as described in [Section 4.5.6](https://ietf-wg-ppm.github.io/draft-ietf-ppm-dap/draft-ietf-ppm-dap.html#batch-validation) so that the aggregator knows which aggregate share to update.

Optional, but we might want to expand 4.5.2 and give this process a name and clarify what it means to validate a _bucket_ (is this the name we're using?) because some of the point in 4.5.6 don't really apply. For example, a bucket may, by definition, contain too few reports.

wangshan:
Perhaps we should remove the mentioning of 4.5.6, since the "bucket" isn't a final batch and doesn't have to be valid at pre-aggregation stage.

cjpatton:
Blocked on resolving https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/pulls?q=is%3Apr+is%3Aopen+label%3A%22aggregation+cost%22, as this will conflict with text changes.

cjpatton:
Just wanted to note an observation here: to specify this normatively, we might need a syntactic change to VDAFs: https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/pull/580#discussion_r1765530675

Of course, there is a non-normative way to resolve this PR, which is to define "batch bucket", say there is an aggregate share per bucket, and say that each output share is "added" into the aggregate share of its bucket (without defining "added").

branlwyd:
Closing this issue, as it was addressed in #623.

PR #580, opened 2024-09-18
Define streaming aggregation in overview

### LINKED ISSUE #571: Clarification request: Does aggregation happen in the AggregationJob?
Section 4.5 says that an Aggregation Job has three phases, the last of which is

>    *  Completion: Finish the aggregate flow, yielding an output share
      corresponding to each report share in the aggregation job.

That makes it sound like Aggregation Jobs don't perform any aggregation.  If that's true, maybe they should be called Preparation Jobs.  If it's false, maybe it can be corrected.

cjpatton:
Closes #571.

Also, clarify that streaming aggregation is a feature of most, but not all, combinations of VDAFs and batch modes.

### DIFF
--- a/draft-ietf-ppm-dap.md
+++ b/draft-ietf-ppm-dap.md
@@ -1286,16 +1286,23 @@ aggregation parameter. However, in some situations it is possible to begin
 aggregation as soon as reports arrive. For example, Prio3 has just one valid
 aggregation parameter (the empty string).
 
-An aggregation job can be thought of as having three phases:
+An aggregation job can be thought of as having three phases, which are
+described in the remaining subsections:
 
 - Initialization: Begin the aggregation flow by disseminating report shares and
   initializing the VDAF prep state for each report.
 - Continuation: Continue the aggregation flow by exchanging prep shares and
   messages until preparation completes or an error occurs.
-- Completion: Finish the aggregate flow, yielding an output share corresponding
-  to each report share in the aggregation job.
+- Completion: Finish the aggregation flow, yielding an output share
+  corresponding to each report share in the aggregation job.
 
-These phases are described in the following subsections.
+After an aggregation job is completed, each Aggregator stores the output shares
+until the aggregate share is collected as described in {{collect-flow}}. Note
+that it is usually not necessary to store output shares individually: depending
+on the batch mode and VDAF, the output shares can be merged into existing
+aggregate shares that are updated as aggregation jobs complete. This streaming
+aggregation is compatible with Prio3 and all batch modes specified in this
+document.
 
 ### Aggregate Initialization {#agg-init}
 
@@ -2150,11 +2157,11 @@ output shares, denoted `out_shares`, for the batch interval, as follows:
 agg_share = Vdaf.aggregate(agg_param, out_shares)
 ~~~
 
-Implementation note: For most VDAFs, it is possible to aggregate output shares
-as they arrive rather than wait until the batch is collected. To do so however,
-it is necessary to enforce the batch parameters as described in
-{{batch-validation}} so that the Aggregator knows which aggregate share to
-update.
+Implementation note: For most VDAFs, including Prio3, it is possible to
+aggregate output shares as they arrive rather than wait until the batch is
+collected. For the batch modes specified in this document, it is necessary to
+enforce the batch parameters as described in {{batch-validation}} so that the
+Aggregator knows which aggregate share to update.
 
 The Helper then encrypts `agg_share` under the Collector's HPKE public key as
 described in {{aggregate-share-encrypt}}, yielding `encrypted_agg_share`.


draft-ietf-ppm-dap.md
```diff
@@ -2150,11 +2157,11 @@ output shares, denoted `out_shares`, for the batch interval, as follows:
 agg_share = Vdaf.aggregate(agg_param, out_shares)
 ~~~
 
-Implementation note: For most VDAFs, it is possible to aggregate output shares
-as they arrive rather than wait until the batch is collected. To do so however,
-it is necessary to enforce the batch parameters as described in
-{{batch-validation}} so that the Aggregator knows which aggregate share to
-update.
+Implementation note: For most VDAFs, including Prioe, it is possible to
```
branlwyd, line 2160:
```suggestion
Implementation note: For most VDAFs, including Prio3, it is possible to
```

(typo)

branlwyd (approved):
LGTM w/ one typo nit.

draft-ietf-ppm-dap.md
```diff
@@ -1295,7 +1296,13 @@ An aggregation job can be thought of as having three phases:
 - Completion: Finish the aggregate flow, yielding an output share corresponding
   to each report share in the aggregation job.
```
bemasc, line 1297: This is still weird.  If completing the aggregation job doesn't involve any aggregation, then it doesn't seem to be an aggregation job.  Maybe it's a preparation job, or a sub-aggregation job, or a pre-aggregation job.  Or maybe this phase isn't "completion"; maybe it's "collection".

draft-ietf-ppm-dap.md
```diff
@@ -1295,7 +1296,13 @@ An aggregation job can be thought of as having three phases:
 - Completion: Finish the aggregate flow, yielding an output share corresponding
```
bemasc, line 1296: "aggregate flow" or "aggregation flow"?  "aggregation flow" or "aggregation job"?

draft-ietf-ppm-dap.md
```diff
@@ -1295,7 +1296,13 @@ An aggregation job can be thought of as having three phases:
...
   to each report share in the aggregation job.
 
-These phases are described in the following subsections.
+After an aggregation job is completed, each Aggregator stores the output shares
+until the aggregate share is collected as described in {{collect-flow}}. Note
+that it is usually not necessary to store output shares individually: depending
+on the batch mode and VDAF, the output shares can be merged into existing
+aggregate shares that are updated as aggregation jobs complete. This streaming
+aggregation is compatible with Prio3 and all batch modes specified in this
+document.
```
bemasc, line 1305: I wonder if it would be clearer to phrase this the other way: aggregation is part of the completion phase of each aggregation job, _but_ some future VDAF might require "delayed aggregation" (but probably not).

draft-ietf-ppm-dap.md
branlwyd: I agree with this: while implementations _can_ (and, in practice, often _do_) aggregate the output shares into a running-total aggregate share at the end of processing an aggregation job, this is not required. And depending on the VDAF/batch mode in use, this aggregation may not be possible at this time.

Since performing aggregation at this stage is an optional performance optimization, I agree that "aggregation job" is not the best name.

I would support "preparation job", since the process that is occurring is most correctly called "VDAF preparation". I think "sub-aggregation job"/"pre-aggregation job" are questionable for the same reason that "aggregation job" is.

draft-ietf-ppm-dap.md
cjpatton: Changed to "aggregation flow" for consistency with above. We also refer to this as the "aggregation sub-protocol".

draft-ietf-ppm-dap.md
cjpatton: Personally, I think the likelihood of any VDAF using delayed aggregation as a feature is remote enough that we don't have too twist too hard to anticipate it.

bemasc: Right.  So I'm imagining that we say "Completion of an aggregation job includes aggregation" and then add some fine print about delayed aggregation.  That's the opposite of this proposal, which says "Completion actually only means VDAF preparation", and then notes that everyone will probably do streaming aggregation.

draft-ietf-ppm-dap.md
cjpatton: "Collection" is a term used for the "collection sub-protocol", in which the batch is selected and the aggregate shares are computed. Let's not overload it.

I'm open to renaming, but the terms  "aggregation job" and "collection job" are pretty baked in at this point, into code and into the way we talk about the protocol. In any case, let's please not let this discussion block this PR, as I think it's a positive improvement in any case. I"ll even volunteer to file an issue and lead the discussion.

draft-ietf-ppm-dap.md
cjpatton: I see what you're getting at. Let me give it a try.

cjpatton: Argh, there's a problem. To specify this rigorously, we need two things:

1. The concept of a "batch bucket" (#385). The definition of "batch bucket" depends on the batch mode: for time-interval, it's a window of time; for leader-selected, it's simply a batch ID. Each bucket has an aggregate share. During the "completion phase" of an aggregation job, we would add each output share into the aggregate share of the bucket it corresponds to.

2. We need to be able to express syntactically how to update an aggregate share with an output share. This is precluded by the current VDAF syntax, [which only describes aggregating output shares into an aggregate share](https://cfrg.github.io/draft-irtf-cfrg-vdaf/draft-irtf-cfrg-vdaf.html#section-5.4). We would therefore need to modify the syntax.

We want (1.) in any case, however I think (2.) requires some discussion. For what it's worth I'd before modifying the syntax to account for this, and I even argued for doing this in the past, but at the time there was a strong preference to keep the abstraction boundary where it is.

I propose we merge this PR without this suggestion and kick off a discussion.

draft-ietf-ppm-dap.md
branlwyd: FWIW, I agree that the renaming process would be somewhat painful in terms of the work it induces in existing implementations. I'm not sure that justifies avoiding a rename, however.

cjpatton: Again, I'm not rejecting to renaming. I'm just trying to make incremental progress on the issues.

draft-ietf-ppm-dap.md
bemasc: Fine with me, but you might want to adjust the description so it doesn't close the Github issue.

bemasc (approved):

draft-ietf-ppm-dap.md
cjpatton: I think the discussion is broader than the text of the issue, so I'd prefer to close and open a new issue to discuss making per-bucket aggregation normative.

PR #582, opened 2024-09-19
Tighten language around replay protection

### LINKED ISSUE #442: Replay attack requirements could be tighter
It looks like the entirely of the replay protection requirements are buried in an implementation note in [Section 4.4.1.4](https://ietf-wg-ppm.github.io/draft-ietf-ppm-dap/draft-ietf-ppm-dap.html#section-4.4.1.4-3.6.4.1).  The text here is pretty terse, so even though it is probably a good cue to someone who understands the problem pretty well, there are lots of bits missing:

1. Normative language.  (Maybe this is covered by [this paragraph](https://ietf-wg-ppm.github.io/draft-ietf-ppm-dap/draft-ietf-ppm-dap.html#section-4.4.1.2-2), but that is also buried.)
2. An explanation of the attack that this covers.[*]
3. Discussion of the use of timestamps, clock skew, and the associated trade-offs.

This probably justifies a section of its own.

[*] Replay attack mitigation wouldn't be necessary if you submit the data directly (and don't retry failed uploads ... and avoid using TLS early data).  Therefore, honest clients that submit data directly to the leader are only exposed to the possibility that the leader defects and replays the submission.  So maybe only helpers need to perform validation under this architecture.

cjpatton:
Closes #442.

At the moment, we have very little normative/informative language around replay protection. The only normative language is buried in the "input share validation" subsection:

> 1. Check if the report has been previously aggregated. If so, the input share
>    MUST be marked as invalid with the error `report_replayed`. A report is
>    considered aggregated if its contribution would be included in a relevant
>    collection job.

Another problem is that we never say explicitly to record the IDs of aggregated reports. Here the Aggregator has a choice: it can avoid excess computation by storing the ID of every report it sees in an aggregation job, thereby ensuring reports that are rejected for some other reason aren't processed again; or it can avoid excess storage by only storing the IDs of output shares it wants to aggregate.

This change has several parts to it:

1. Add a section to the protocol overview about replay protection (why/how) and a paragraph to the aggregation sub-protocol overview.

1. Remove the replay check from the "input share validation" section.

1. Have the Leader check for replays right after it has picked a candidate set of reports. Have it to store the IDs either before initialization or after completion of the aggregation job.

1. Have the Helper resolve replays (reject replays and update the stored set) either at the beginning of an aggregation job or just before completing it.

3. Be explicit about the use of report IDs for replay protection. This only impacts an implementation note, which envisions multiple schemes for replay protection, like hashing the report. This note has been moved from the "input share validation" section to operational considerations.

### DIFF
--- a/draft-ietf-ppm-dap.md
+++ b/draft-ietf-ppm-dap.md
@@ -523,6 +523,24 @@ instance, the user might have spent 30s on a task but the Client might report
 address it; it merely ensures that the data is within acceptable limits, so the
 Client could not report 10^6s or -20s.
 
+## Replay Protection {#replay-protection}
+
+Another goal of DAP is to mitigate replay attacks in which a report is
+aggregated multiple times within a batch or across multiple batches. This would
+allow the attacker to learn more information about the underlying measurement
+than it would otherwise.
+
+When a Client generates a report, it also generates a random nonce, called the
+"report ID". Each Aggregator is responsible for storing the IDs of reports it
+has aggregated for a given task. To check whether a report has been replayed,
+it checks whether the report's ID is in the set of stored IDs.
+
+Note that IDs do not need to be stored indefinitely. The protocol allows
+Aggregators to enforce replay only for a sliding time window --- say, within
+the last two weeks of the current time --- and reject reports that fall outside
+of the replay window. This allows implementation to save resources by
+forgetting old report IDs.
+
 # Message Transport {#message-transport}
 
 Communications between DAP participants are carried over HTTP {{!RFC9110}}. Use
@@ -1014,10 +1032,9 @@ struct {
 
 * `report_metadata` is public metadata describing the report.
 
-    * `report_id` is used by the Aggregators to ensure the report appears in at
-      most one batch (see {{input-share-validation}}). The Client MUST generate
-      this by generating 16 random bytes using a cryptographically secure random
-      number generator.
+   * `report_id` is used by the Aggregators to ensure the report is not
+      replayed ({{agg-flow}}). The Client MUST generate this by generating 16
+      random bytes using a cryptographically secure random number generator.
 
     * `time` is the time at which the report was generated. The Client SHOULD
       round this value down to the nearest multiple of the task's
@@ -1114,8 +1131,7 @@ HpkeConfigList and retry with a freshly generated Report. If this retried upload
 does not succeed, the Client SHOULD abort and discontinue retrying.
 
 If a report's ID matches that of a previously uploaded report, the Leader MUST
-ignore it. In addition, it MAY alert the Client with error `reportRejected`. See
-the implementation note in {{input-share-validation}}.
+ignore it. In addition, it MAY alert the Client with error `reportRejected`.
 
 The Leader MUST ignore any report pertaining to a batch that has already been
 collected (see {{input-share-validation}} for details). Otherwise, comparing
@@ -1287,6 +1303,12 @@ aggregate shares that are updated as aggregation jobs complete. This streaming
 aggregation is compatible with Prio3 and all batch modes specified in this
 document.
 
+Apart from VDAF preparation, another important task of the aggregation
+interaction is to provide replay protection ({{replay-protection}}). Along with
+the output shares, each Aggregator records the IDs of all reports it is has
+aggregated for a given task: before committing to an output share, it checks
+whether the corresponding report ID is in the set of stored IDs.
+
 ### Aggregate Initialization {#agg-init}
 
 The Leader begins an aggregation job by choosing a set of candidate reports that
@@ -1310,17 +1332,17 @@ The Leader and Helper initialization behavior is detailed below.
 
 #### Leader Initialization {#leader-init}
 
-The Leader begins the aggregate initialization phase with the set of candidate
-reports as follows:
+The Leader begins the aggregate initialization by sampling a fresh
+AggregationJobID.
 
-1. Generate a fresh AggregationJobID.
-1. Decrypt the input share for each report share as described in
-   {{input-share-decryption}}.
-1. Check that the resulting input share is valid as described in
-   {{input-share-validation}}.
+Next, for each report in the candidate set, it checks if the report ID
+corresponds to a report ID it has previously stored for this task. If so, it
+marks the report as invalid and removes it from the candidate set.
 
-If any step invalidates the report, the Leader rejects the report and removes
-it from the set of candidate reports.
+Next, the Leader decrypts each of its report shares as described in
+{{input-share-decryption}}, then checks input share validity as described in
+{{input-share-validation}}. If either step invalidates the report, the Leader
+rejects the report and removes it from the set of candidate reports.
 
 Next, for each report the Leader executes the following procedure:
 
@@ -1475,6 +1497,9 @@ Otherwise, the Leader proceeds as follows with each report:
 1. Else the type is invalid, in which case the Leader MUST abort the
    aggregation job.
 
+When the Leader stores the `out_share`, it MUST also store the report ID for
+replay protection.
+
 (Note: Since VDAF preparation completes in a constant number of rounds, it will
 never be the case that some reports are completed and others are not.)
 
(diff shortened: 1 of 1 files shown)


draft-ietf-ppm-dap.md
```diff
@@ -1542,15 +1568,27 @@ struct {
 } PrepareResp;
 ~~~
 
-First the Helper preprocesses each report as follows:
+To begin, the Helper MAY resolve replays with the following procedure. It MUST
+perform this procedure prior to completing the aggregation job. For each report
```
tgeoghegan, line 1572: It's a bit confusing that this is a MAY in one sentence and a MUST in the next. I guess the idea is that the helper is allowed to defer replay checks if it wants to?

draft-ietf-ppm-dap.md
```diff
@@ -1796,18 +1847,21 @@ Otherwise, the Leader proceeds as follows with each report:
    `state == Rejected()`, in which case the Leader rejects the report and
    removes it from the candidate set; or `state == Finished(out_share)`, in
    which case preparation is complete and the Leader stores the output share for
-   use in the collection interaction {{collect-flow}}.
+   use in the collection sub-protocol {{collect-flow}}.
```
tgeoghegan, line 1850: As above, prefer "interaction" to "sub-protocol" per HTTPDIR review.

draft-ietf-ppm-dap.md
```diff
@@ -1796,18 +1847,21 @@ Otherwise, the Leader proceeds as follows with each report:
...
 
 1. Else if the type is "finished" and `state == Finished(out_share)`, then
    preparation is complete and the Leader stores the output share for use in
-   the collection flow ({{collect-flow}}).
+   the collection sub-protocol ({{collect-flow}}).
```
tgeoghegan, line 1854: As above, prefer "interaction" to "sub-protocol" per HTTPDIR review.

draft-ietf-ppm-dap.md
```diff
@@ -2405,7 +2473,10 @@ required to:
 In addition, for each DAP task, the Leader is required to:
 
 - Implement and store state for the form of inter- and intra-batch replay
-  mitigation in {{input-share-validation}}.
+  mitigation in {{agg-flow}}. This requires storing the report IDs of all
+  reports processed for a given task. Implementations may find it helpfu to
```
tgeoghegan, line 2477:
```suggestion
  reports processed for a given task. Implementations may find it helpful to
```

tgeoghegan (approved):

draft-ietf-ppm-dap.md
```diff
@@ -1304,6 +1320,12 @@ aggregate shares that are updated as aggregation jobs complete. This streaming
 aggregation is compatible with Prio3 and all batch modes specified in this
 document.
 
+Apart from VDAF preparation, another important task of the aggregation
+sub-protocol is to prevent replay protection ({{replay-protection}}). Along
```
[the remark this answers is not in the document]
cjpatton: Thanks for catching this, will fix throughout this PR. I'll follow up with a more complete fix in another PR.

draft-ietf-ppm-dap.md
cjpatton: Yes, that's the idea.

cjpatton:
Took Tim's suggestions, rebased, and squshed.

draft-ietf-ppm-dap.md
tgeoghegan: Alright. This a little more laborious but I think it's less ambiguous:

> The Helper MUST resolve replays with the following procedure. It MAY do so immediately upon receiving the `AggregationJobInitReq`, but it MUST do so prior to completing the aggregation job. For each report...

draft-ietf-ppm-dap.md
```diff
@@ -1765,14 +1817,17 @@ Otherwise, the Leader proceeds as follows with each report:
...
-   the collection flow ({{collect-flow}}).
+   the collection interaction ({{collect-flow}}).
 
 1. Else if the type is "reject", then the Leader rejects the report and removes
    it from the candidate set.
 
 1. Else the type is invalid, in which case the Leader MUST abort the
    aggregation job.
 
+When the Leader stores the `out_share`, it MUST also store the report ID for
```
branlwyd, line 1828: I think we want to both check & store here, rather than only storing here. (this may have been the intent)

Otherwise, a Leader which checks at time of creation but only stores at time of completion could:
1) Generate two distinct aggregation jobs, both containing some report R. Both will check R against the report ID set, but will pass since R has not yet been stored in the set.
2) Concurrently step both aggregation jobs to completion. The Helper _should_ reject at least one of the copies of R as a replay; but if it doesn't, the Leader would store the report ID twice, rejecting neither copy of R.

That is, we are potentially left open to replay if the Helper's check is mis-implemented.

branlwyd, line 1828: editorial: I think this text would fit better in the numbered list handling the `Finished(out_share)` case.

draft-ietf-ppm-dap.md
```diff
@@ -1298,17 +1320,21 @@ The Leader and Helper initialization behavior is detailed below.
...
+At this point, the Leader MAY store the ID of each report in the candidate set.
+It MUST do so at the end of the continuation phase
+({{aggregation-leader-continuation}}).
 
-If any step invalidates the report, the Leader rejects the report and removes
-it from the set of candidate reports.
+Next, the Leader decrypts each of its report shares as described in
+{{input-share-decryption}}, then checks input share validity as described in
+{{input-share-validation}}. If either step invalidates the report, the Leader
+rejects the report and removes it from the set of candidate reports.
```
branlwyd, line 1337: I think we should have some text suggesting that the `Finished(out_share)` case might require replay-checking, around L1473 of this snaphost, similarly to what is written in the Leader Continuation section for the `Finished` state handling.

draft-ietf-ppm-dap.md
cjpatton: ack, won't fix. (I don't think that flows works, but we can revisit once the semantics is right.)

draft-ietf-ppm-dap.md
```diff
@@ -1298,17 +1320,21 @@ The Leader and Helper initialization behavior is detailed below.
...
+Next, for each report in the candidate set, it checks if the report ID
+corresponds to a report ID it has previously stored for this task. If so, it
+marks the report as invalid and removes it from the candidate set.
 
-1. Generate a fresh AggregationJobID.
-1. Decrypt the input share for each report share as described in
-   {{input-share-decryption}}.
-1. Check that the resulting input share is valid as described in
-   {{input-share-validation}}.
+At this point, the Leader MAY store the ID of each report in the candidate set.
```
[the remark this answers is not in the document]
cjpatton: Alright, to resolve this I'm going to take away the choice of when to store.

draft-ietf-ppm-dap.md
cjpatton: Done.

draft-ietf-ppm-dap.md
cjpatton: The intent was only to store here. The only time the Leader checks is just after creating the aggregation job. It's therefore the Leader's responsibility to make sure it doesn't accidentally replay a report.

On the other hand, it's also the responsibility of the Helper to make sure a report isn't replayed. In other words, the scenario you're describing here is handled the same as an attack by a malicious Leader. I don't think we need to split hairs here.

cjpatton: Ok, having fixed everything else, I think what you're describing works but is repetitive. I'd rather not have to repeat "MUST store the report ID for replay protection" more times than we have to. But if you feel strongly about this I can fix.

branlwyd: Yeah -- aggregation job creation & replay checking are similar, but have slightly different requirements.

A Leader doesn't want to put a report into more than one aggregation job, since that could lead to wasted work & more aggregation jobs for the same number of reports. But both the Leader's & Helper's replay checking should ensure that if a report does end up in two aggregation jobs, it will still only be aggregated once. So a Leader will want to track a couple of bits of state for each report: "is currently undergoing aggregation" (to control whether it is might be added to a new aggregation job) and "is aggregated into some batch" (to control whether it is considered a replay). These two bits are different since "is currently undergoing aggregation" might go back to false after being true (e.g. if it is `report_too_early`), but "is aggregated into some batch" is a one-way transition.

branlwyd (approved):

cjpatton:
Rebased and squashed.

PR #591, opened 2024-09-25
Replace "collection job URI" with its definition

### LINKED ISSUE #586: Define "collection job URI" and the media type of collection polling requests
In https://datatracker.ietf.org/doc/html/draft-ietf-ppm-dap-09#section-4.6.1-15 the draft says:
> After receiving the response to its CollectionReq, the Collector makes an HTTP POST request to the **collection job URI** to check on the status of the collect job and eventually obtain the result.

But nowhere else in the draft does it define what the `collection job URI` is. It could be reasonably inferred that it's the same URI used to initialize the collection request (`/tasks/{tasks}/collection_jobs/{collection-job-id}`) but I believe this should be better defined.

Another detail that's left unspecified is what is the media type of this request. The draft defines a media type for every request except this one.

cjpatton:
Closes #586.

### DIFF
--- a/draft-ietf-ppm-dap.md
+++ b/draft-ietf-ppm-dap.md
@@ -2104,11 +2104,13 @@ Changing a collection job's parameters is illegal, so further requests to
 `collection-job-id` but with a different `CollectionReq` in the body MUST fail
 with an HTTP client error status code.
 
-After receiving the response to its `CollectionReq`, the Collector makes an HTTP
-GET request to the collection job URI to check on the status of the collect job
-and eventually obtain the result. If the collection job is not finished yet, the
-Leader responds with HTTP status 202 Accepted. The response MAY include a
-Retry-After header field to suggest a polling interval to the Collector.
+After receiving the response to its `CollectionReq`, the Collector makes an
+HTTP GET request to
+`{leader}/tasks/{task-id}/collection_jobs/{collection-job-id}` to check on the
+status of the collect job and eventually obtain the result. If the collection job is not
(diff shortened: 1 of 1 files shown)


draft-ietf-ppm-dap.md
```diff
@@ -2104,11 +2104,14 @@ Changing a collection job's parameters is illegal, so further requests to
...
-After receiving the response to its `CollectionReq`, the Collector makes an HTTP
-GET request to the collection job URI to check on the status of the collect job
-and eventually obtain the result. If the collection job is not finished yet, the
-Leader responds with HTTP status 202 Accepted. The response MAY include a
-Retry-After header field to suggest a polling interval to the Collector.
+After receiving the response to its `CollectionReq`, the Collector makes an
+HTTP GET request to
+`{leader}/tasks/{task-id}/collection_jobs/{collection-job-id}` to check on the
+status of the collect job and eventually obtain the result. Note that since the
+request has an empty body, no media type is set. If the collection job is not
```
tgeoghegan, line 2111: I don't think it's necessary to explain that the request has no body or content type. Both are implied by the GET method. We don't bother to explain this for [GETting HPKE configs](https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/blob/main/draft-ietf-ppm-dap.md#hpke-configuration-request-hpke-config) so I don't think we should do it here. But there's no harm in including the sentence.
 
```suggestion
status of the collect job and eventually obtain the result. If the collection job is not
```

PR #595, opened 2024-09-26
Add synchronization concerns to operational considerations.

### LINKED ISSUE #556: Discuss distributed systems-related concerns in Operational Considerations section
I think it would be helpful to explicitly list what sort of synchronization guarantees the aggregators need to uphold. Some of these are implicit in the text elsewhere, and they would be important to the architecture of a distributed aggregator. Here's what I have so far:

- Leader
  - The leader has to perform anti-replay checks between receiving a report and sending it in an aggregation job (i.e. deduplicating by ReportMetadata). This is easily amenable to approaches that only provide eventual consistency.
  - The leader needs some synchronization between aggregate share requests and aggregation job requests to make sure it doesn't aggregate any new reports into a batch that has already been collected. This requirement is significantly different between time interval queries, where the client metadata determines the batch, and fixed size query, where the leader has full control of batches.
  - The leader needs to synchronize between sending aggregation job requests and sending aggregate share requests, to ensure that it never has both an aggregate share request collecting a batch and an aggregation job that affects the same batch outstanding at the same time. Note that with time interval queries, there is a many-to-many mapping between aggregation jobs and batches, while with fixed size queries, each aggregation job impacts only one batch.
- Helper
  - The helper needs to perform duplicate report detection across aggregation job requests.
  - The helper needs strong consistency between aggregation job requests and subsequent aggregate share requests, so that it includes every eligible output share in its aggregate share.

branlwyd:
Closes #556.

This is intended to list the parts of a DAP deployment where synchronization is required between different componenets of the system. This will hopefully be useful both as hints to implementers, as well as providing a guide to where we might hope to introduce the opportunity for eventual consistency.

### DIFF
--- a/draft-ietf-ppm-dap.md
+++ b/draft-ietf-ppm-dap.md
@@ -1394,6 +1394,14 @@ and Helper then run the aggregate initialization flow to accomplish two tasks:
 
 The Leader and Helper initialization behavior is detailed below.
 
+Implementation note: the Leader will generally want to associate each report
+with a single aggregation job, as otherwise the duplicated reports will
+eventually be discarded as a replay. However, it is likely not appropriate to
+directly use the used-ID storage used for replay protection to determine which
+reports can be added to an aggregation job: certain errors (e.g.
+`report_too_early`) allow the report to be added to another aggregation job in
+the future; but storage into the used-ID storage is permanent.
+
 #### Leader Initialization {#leader-init}
 
 The Leader begins the aggregate initialization by sampling a fresh
@@ -1833,7 +1841,8 @@ following checks:
       determined by query ({{batch-mode}}) conveyed in these messages. Queries
       must satisfy the criteria covered in {{batch-validation}}. These criteria
       are meant to restrict queries in a way that makes it easy to determine
-      whether a report pertains to a batch that was collected.
+      whether a report pertains to a batch that was collected. See
+      {{distributed-systems}} for more information.
 
 1. Finally, if an Aggregator cannot determine if an input share is valid, it
    MUST mark the input share as invalid with error `report_dropped`. For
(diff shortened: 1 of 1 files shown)


branlwyd:
Closes #556.

draft-ietf-ppm-dap.md
```diff
@@ -2587,6 +2587,62 @@ the task and all data pertaining to this task after `task_expiration`.
...
+  report is accepted.
+
+* The Leader is responsible for generating aggregation jobs, and will generally
+  want to place each report in exactly one aggregation job. This may require
+  synchronization between different components of the system which are
+  generating aggregation jobs. Note that placing a report into more than one
+  aggregation job will result in a loss of throughput, rather than a loss of
+  correctness, privacy, or robustness, so it is acceptable for implementations
+  to use an eventually-consistent scheme which may rarely place a report into
+  multiple aggregation jobs.
```
divergentdave, line 2611: On its own, placing a report in multiple aggregation jobs would be bad for privacy, as it may allow a malicious helper to execute a replay attack by simply ignoring its own replay checks. (though it wouldn't be able to control what is replayed) Is the idea that the leader can use lax coordination to assemble aggregation jobs, and then catch the duplicates before it sends out the aggregation job initialization request, or before it finalizes the aggregation job and accumulates its output share?

branlwyd: Yes -- grouping reports into aggregation jobs is separate from report-replay checking during aggregation [1]. The Leader could/should perform replay checking separately from aggregation job grouping, which would catch duplicates.

[1] Specifically, once a report ID is stored in the used-report storage used for replay checking, it can never be aggregated again. However, a report which is placed in an aggregation job, and then failed due to `report_too_early`, can be placed in another aggregation job later.

branlwyd: > [1] Specifically, once a report ID is stored in the used-report storage used for replay checking, it can never be aggregated again. However, a report which is placed in an aggregation job, and then failed due to report_too_early, can be placed in another aggregation job later.

I've touched on this point with a few different folks at this point, WDYT about adding some explicit text indicating that used-report storage/aggregation replay checking should not be conflated with storage used to avoid placing the same report in multiple aggregation jobs? It might be a somewhat subtle point. (The only place this practically matters is due to `report_too_early`, but I don't think we want to remove those semantics in the name of simplicity.)

divergentdave: Sure, that sounds good. Though, the two functions could be done by either one store with multiple states for each report, or by separate stores, depending on implementation choices. It seems like grouping reports into aggregation jobs should be easier to horizontally scale than the "used-report" part, without having to accept spurious duplication, by using  a distributed queue.

branlwyd: Added an implementation note around this to the aggregation section; I think the phrasing doesn't contradict using the same storage system to store both bits of per-report state, it only notes that the storage used for replay protection shouldn't be used "directly" when generating new aggregation jobs.

divergentdave (approved):

tgeoghegan (approved):

draft-ietf-ppm-dap.md
```diff
@@ -2587,6 +2595,62 @@ the task and all data pertaining to this task after `task_expiration`.
...
+
+* The upload interaction requires the Leader to ignore uploaded reports with a
+  duplicated ID, including concurrently-uploaded reports. This might be
+  implemented by synchronization or via an eventually-consistent process. If the
+  Leader wishes to alert the Client with a `reportRejected` error,
+  synchronization will be necessary to ensure all but one concurrent request
+  receive the error.
+
+* The Leader is responsible for generating aggregation jobs, and will generally
+  want to place each report in exactly one aggregation job. This may require
```
cjpatton, line 2613: This suggests to me there are situations where a Leader would stick a report in two jobs. When would this happen?

draft-ietf-ppm-dap.md
```diff
@@ -2587,6 +2595,62 @@ the task and all data pertaining to this task after `task_expiration`.
...
+  aggregation requests.
+
+* Aggregation requires checking and updating used-report storage as part of
+  implementing replay protection. This must be done while processing the
+  aggregation job, though which steps the checks are performed at is up to the
+  implementation. The checks and storage require synchronization, so that if two
+  aggregation jobs contianing the same report are processed, at most one
+  instance of the report will be aggregated. However, the interaction with the
+  used-report storage does not necessarily have to be synchronized with the
+  processing and storage for the remainder of the aggregation process -- for
```
cjpatton, line 2636:
```suggestion
  processing and storage for the remainder of the aggregation process. For
```

draft-ietf-ppm-dap.md
```diff
@@ -2587,6 +2595,62 @@ the task and all data pertaining to this task after `task_expiration`.
...
+  example, used-report storage could be implemented in a separate datastore than
+  is used for the remainder of data storage, without any transactionality
+  between updates to the two datastores.
+
+* The aggregation and collection interactions require synchronization to avoid
```
cjpatton, line 2641: I believe we point this out elsewhere in operational considerations. If so, perhaps we should remove that text and replace with a reference to this section.

cjpatton (commented):
This mostly looks good ... there's one point I think we should try to make clearer.

draft-ietf-ppm-dap.md
branlwyd: * If a report is rejected due to `report_too_early`, the Leader can place it in a later aggregation job. (this is touched on by the implementation note added in a late commit to this PR)
* In general, it is _acceptable_ to place a report in multiple aggregation jobs, as long as the Leader also performs report-replay checking later. This results in a loss of throughput but no correctness/privacy/robustness issues; it may be the right tradeoff to allow aggregation job creation to generate aggregation jobs with duplicated reports in some corner cases. (this is touched on by the remainder of this paragraph)

cjpatton: > * If a report is rejected due to `report_too_early`, the Leader can place it in a later aggregation job. (this is touched on by the implementation note added in a late commit to this PR)

Let's make a note of this here.


>     * In general, it is _acceptable_ to place a report in multiple aggregation jobs, as long as the Leader also performs report-replay checking later. This results in a loss of throughput but no correctness/privacy/robustness issues; it may be the right tradeoff to allow aggregation job creation to generate aggregation jobs with duplicated reports in some corner cases. (this is touched on by the remainder of this paragraph)

Does this happen in your experience, or is this just something you imagine could happen? This sounds like an extreme edge case to me, and I worry that trying to make room for it here might mislead readers.

draft-ietf-ppm-dap.md
branlwyd: I thought so too, but I can't find any text to this effect. The closest I can find is some text in an implementation note in the Input Share Validation section, though it doesn't mention synchronization explicitly. I added a reference to this section to that implementation note.

draft-ietf-ppm-dap.md
branlwyd: > Let's make a note of this here.

👍🏻 

> Does this happen in your experience, or is this just something you imagine could happen? This sounds like an extreme edge case to me, and I worry that trying to make room for it here might mislead readers.

Janus doesn't currently take advantage of this; the Leader implementation will generate aggregation jobs which partition the space of reports. I do think it's useful to note that eventual consistency is acceptable here; for example, if aggregation jobs are generated by pulling reports from a distributed queue of pending reports, any time the distributed queue accidentally delivered a message multiple times we would see the same report showing up in multiple aggregation jobs -- and that's OK.

Issue #588, opened 2024-09-25
Align collection job polling semantics with aggregation job changes from #564


inahga:
From https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/pull/564#discussion_r1775692873. After #564 is merged, we should fix this.

In particular:
- Job status is communicated through HTTP status, should instead use explicit fields in the response.
- Polling endpoint should be communicated with Location header

inahga:
(This can be assigned to me.)

PR #596, opened 2024-09-27
Align collection job with new aggregation job semantics

inahga:
Closes https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/issues/588.

This does the following:
- Communicate job status through field in the body, while always returning 200 OK for polls.
- Polling endpoint is communicated through Location header.
- Shuffling around of paragraphs for readability.
- Align structure names and media types to look like aggregation job ones.

This is a fair bit more invasive than I would have hoped, LMK if anything can be stripped out.

### DIFF
--- a/draft-ietf-ppm-dap.md
+++ b/draft-ietf-ppm-dap.md
@@ -1832,7 +1832,7 @@ following checks:
    input share MUST be marked as invalid with error `batch_collected`.
 
     * Implementation note: The Leader considers a batch to be collected once it
-      has completed a collection job for a CollectionReq message from the
+      has completed a collection job for a CollectionJobReq message from the
       Collector; the Helper considers a batch to be collected once it has
       responded to an `AggregateShareReq` message from the Leader. A batch is
       determined by query ({{batch-mode}}) conveyed in these messages. Queries
@@ -2128,14 +2128,14 @@ This ID value MUST be unique within the scope of the corresponding DAP task.
 
 To initiate the collection job, the collector issues a PUT request to
 `{leader}/tasks/{task-id}/collection_jobs/{collection-job-id}`. The body of the
-request has media type "application/dap-collect-req", and it is structured as
-follows:
+request has media type "application/dap-collection-job-req", and it is structured
+as follows:
 
 ~~~ tls-presentation
 struct {
   Query query;
   opaque agg_param<0..2^32-1>; /* VDAF aggregation parameter */
-} CollectionReq;
+} CollectionJobReq;
 ~~~
 
 The named parameters are:
@@ -2159,54 +2159,30 @@ aggregation parameter in advance. For example, for Prio3 the only valid
 aggregation parameter is the empty string. For these reasons, the collection
 job is handled asynchronously.
 
-Upon receipt of a `CollectionReq`, the Leader begins by checking that it
+Upon receipt of a `CollectionJobReq`, the Leader begins by checking that it
 recognizes the task ID in the request path. If not, it MUST abort with error
 `unrecognizedTask`.
 
 The Leader MAY further validate the request according to the requirements in
 {{batch-validation}} and abort with the indicated error, though some conditions
 such as the number of valid reports may not be verifiable while handling the
-CollectionReq message, and the batch will have to be re-validated later on
+`CollectionJobReq` message, and the batch will have to be re-validated later on
 regardless.
 
-If the Leader finds the CollectionReq to be valid, it immediately responds with
-HTTP status 201.
-
-The Leader then begins working with the Helper to aggregate the reports
-satisfying the query (or continues this process, depending on the VDAF) as
-described in {{aggregate-flow}}.
-
 Changing a collection job's parameters is illegal, so further requests to
-`PUT /tasks/{tasks}/collection_jobs/{collection-job-id}` for the same
-`collection-job-id` but with a different `CollectionReq` in the body MUST fail
-with an HTTP client error status code.
+`PUT /tasks/{task-id}/collection_jobs/{collection-job-id}` for the same
+`collection-job-id` but with a different `CollectionJobReq` in the body MUST
+fail with an HTTP client error status code.
 
-After receiving the response to its `CollectionReq`, the Collector makes an
-HTTP GET request to
-`{leader}/tasks/{task-id}/collection_jobs/{collection-job-id}` to check on the
-status of the collect job and eventually obtain the result. If the collection job is not
-finished yet, the Leader responds with HTTP status 202 Accepted. The response
-MAY include a Retry-After header field to suggest a polling interval to the
-Collector.
-
-Asynchronously from any request from the Collector, the Leader attempts to run
-the collection job. It first checks whether it can construct a batch for the
-collection job by applying the requirements in {{batch-validation}}. If so, then
-the Leader obtains the Helper's aggregate share following the aggregate-share
-request flow described in {{collect-aggregate}}. If not, it either aborts the
-collection job or tries again later, depending on which requirement in
-{{batch-validation}} was not met. If the Leader has a pending aggregation job
-that overlaps with the batch and aggregation parameter for the collection job,
-the Leader MUST first complete the aggregation job before proceeding and
-requesting an aggregate share from the Helper. This avoids a race condition
-between aggregation and collection jobs that can yield trivial batch mismatch
-errors.
-
-Once both aggregate shares are successfully obtained, the Leader responds to
-subsequent HTTP GET requests to the collection job with HTTP status code 200 OK
-and a body consisting of a `Collection`:
+The Leader responds to `CollectionJobReq`s with a `CollectionJobResp`, which is
+structured as follows:
 
 ~~~ tls-presentation
+enum {
+  processing(0),
+  finished(1),
+} CollectionJobStatus;
+
 struct {
   PartialBatchSelector part_batch_selector;
   uint64 report_count;
@@ -2214,9 +2190,17 @@ struct {
   HpkeCiphertext leader_encrypted_agg_share;
   HpkeCiphertext helper_encrypted_agg_share;
 } Collection;
+
+struct {
+  CollectionJobStatus status;
+  select (CollectionJob.status) {
+    case processing: Empty;
+    case finished:   Collection;
+  }
+} CollectionJobResp;
 ~~~
 
-The body's media type is "application/dap-collection". The `Collection`
+The body's media type is "application/dap-collection-job-resp". The `Collection`
 structure includes the following:
 
 * `part_batch_selector`: Information used to bind the aggregate result to the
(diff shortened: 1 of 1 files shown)


inahga:
It raises the question of whether we should support both sync/async like with aggregation. (n.b. I preserved async-only in this PR).

The only technical rationale I have for it is that since most implementations are doing VDAFs without aggregation parameters, and thus can eagerly aggregate, collection jobs can be fulfilled remarkably quickly. We find that a lot of delay in processing collection jobs happens to just be with polling. If the collection job could be fulfilled instantly, that would be nice.

I don't feel strongly about this though, nor do I think collection jobs are a particularly expensive part of the system, so I'm content to leave it be.

draft-ietf-ppm-dap.md
```diff
@@ -2084,64 +2084,47 @@ aggregation parameter in advance. For example, for Prio3 the only valid
...
 
-If the Leader finds the CollectionReq to be valid, it immediately responds with
-HTTP status 201.
-
-The Leader then begins working with the Helper to aggregate the reports
-satisfying the query (or continues this process, depending on the VDAF) as
-described in {{aggregate-flow}}.
-
 Changing a collection job's parameters is illegal, so further requests to
 `PUT /tasks/{tasks}/collection_jobs/{collection-job-id}` for the same
```
branlwyd, line 2098:
```suggestion
`PUT /tasks/{task-id}/collection_jobs/{collection-job-id}` for the same
```

(not caused by this PR -- aligns with other uses of parameterized URL path fragments)

draft-ietf-ppm-dap.md
```diff
@@ -2155,14 +2138,50 @@ structure includes the following:
...
   Collector.
 
 * `helper_encrypted_agg_share`: The Helper's aggregate share, encrypted to the
   Collector.
 
+If the Leader finds the `CollectionJobReq` to be valid, it immediately responds
+with HTTP status 201 Created with a body containing a `CollectionJobResp`. The
+`status` field is set to `processing`, and the response contains a Location
+header containing the relative reference
+`/tasks/{task-id}/collection_jobs/{collection-job-id}`. The Leader SHOULD
```
branlwyd, line 2153: Q. What's the justification for including a Location field, here or in the aggregation interaction? In both cases, the URI included in the Location header field is entirely determined by the specification, and the HTTP client has enough information to construct the URI themselves.

draft-ietf-ppm-dap.md
```diff
@@ -2155,14 +2138,50 @@ structure includes the following:
...
+include a Retry-After header field to suggest a polling interval to the
+Collector.
+
+After receiving the response to its `CollectionJobReq`, the Collector makes an
+HTTP GET request to the aforementioned Location to check on the status of the
```
branlwyd, line 2158:
```suggestion
After receiving the response to its `CollectionJobReq`, the Collector periodically makes
HTTP GET requests to the aforementioned Location to check on the status of the
```

I think we should highlight that this is a repeated polling operation -- as currently written, the start of this sentence reads as if only a single HTTP GET is sent.

branlwyd:
> It raises the question of whether we should support both sync/async like with aggregation. (n.b. I preserved async-only in this PR).
> 
> The only technical rationale I have for it is that since most implementations are doing VDAFs without aggregation parameters, and thus can eagerly aggregate, collection jobs can be fulfilled remarkably quickly. We find that a lot of delay in processing collection jobs happens to just be with polling. If the collection job could be fulfilled instantly, that would be nice.
> 
> I don't feel strongly about this though, nor do I think collection jobs are a particularly expensive part of the system, so I'm content to leave it be.

IMO, we shouldn't consider specifying sync collection unless someone wants to implement it. It would be easy enough to specify, but would lead to further implementation complexity, similarly to the complexity induced by supporting both sync & async aggregation.

draft-ietf-ppm-dap.md
cjpatton: Also, the response might have had status "finished", in which case there's no need to poll?

draft-ietf-ppm-dap.md
```diff
@@ -2856,8 +2875,8 @@ corresponding media types types:
 - AggregationJobContinueReq {{aggregation-leader-continuation}}: "application/dap-aggregation-job-continue-req"
 - AggregateShareReq {{collect-flow}}: "application/dap-aggregate-share-req"
 - AggregateShare {{collect-flow}}: "application/dap-aggregate-share"
-- CollectionReq {{collect-flow}}: "application/dap-collect-req"
-- Collection {{collect-flow}}: "application/dap-collection"
+- CollectionJobReq {{collect-flow}}: "application/dap-collection-job-req"
+- CollectionJobResp {{collect-flow}}: "application/dap-collection-job-resp"
```
cjpatton, line 2879: Likewise here: my preference would be to not change the media type, even if we change the name of the message (which in itself is not a wire breaking change).

cjpatton (commented):
> IMO, we shouldn't consider specifying sync collection unless someone wants to implement it. It would be easy enough to specify, but would lead to further implementation complexity, similarly to the complexity induced by supporting both sync & async aggregation.

When we discussed this yesterday (2024/10/2), I was in agreement here, but after reading this PR, I actually think the symmetry with the aggregation job section does a lot to improve readability. I don't have a strong opinion either way, but if we think this won't add too much complexity to the specification, then I think we should keep it.

draft-ietf-ppm-dap.md
cjpatton: Ignore this comment (see above)

branlwyd:
> When we discussed this yesterday (2024/10/2), I was in agreement here, but after reading this PR, I actually think the symmetry with the aggregation job section does a lot to improve readability. I don't have a strong opinion either way, but if we think this won't add too much complexity to the specification, then I think we should keep it.

I've been thinking about this some more.

I still think async-only is the right choice for collection, as (a) no one has expressed a desire to implement sync collection & (b) processing a collection job may require aggregating an arbitrary number of reports, with the number of reports to aggregate at least partially out of control of the aggregators -- so sync collection is unlikely to work well in many cases.

If we want to resolve the difference between the aggregation interaction with the collection interaction, I think we should try to attain enough implementation experience with async aggregation to show that it is at least as good as sync aggregation for all use cases, then remove sync aggregation. This would simplify the protocol as well as resolve the disparity between aggregation and collection.

That said, I don't feel strongly enough about this to insist on it. If this argument is not convincing, we can specify both -- there is nothing stopping us from dropping sync aggregation/collection later, if implementation experience shows this is possible.

draft-ietf-ppm-dap.md
inahga: The rationale is (unfortunately) subtle.

https://www.rfc-editor.org/rfc/rfc9110#name-201-created

> The 201 (Created) status code indicates that the request has been fulfilled and has resulted in one or more new resources being created. The primary resource created by the request is identified by either a [Location](https://www.rfc-editor.org/rfc/rfc9110#field.location) header field in the response or, if no [Location](https://www.rfc-editor.org/rfc/rfc9110#field.location) header field is received, by the target URI.
>
> The 201 response content typically describes and links to the resource(s) created. Any validator fields ([Section 8.8](https://www.rfc-editor.org/rfc/rfc9110#response.validator)) sent in the response convey the current validators for a new representation created by the request. Note that the PUT method ([Section 9.3.4](https://www.rfc-editor.org/rfc/rfc9110#PUT)) has additional requirements that might preclude sending such validators.

Consider aggregation: we PUT at `{helper}/tasks/{task-id}/aggregation_jobs/{aggregation-job-id}`, BUT we expect the Leader to poll the job at `{helper}/tasks/{task-id}/aggregation_jobs/{aggregation-job-id}?step=0`. That is, the resource we're polling is at a different URI than the one that was provided in the PUT.

However, this is not true for collection. So I think we can (and should) eschew the Location header for collection. I will edit.

draft-ietf-ppm-dap.md
inahga: > Also, the response might have had status "finished", in which case there's no need to poll?

I added a blurb to the paragraph `Once both aggregate shares are successfully obtained`. Let me know if this reads clearer.

draft-ietf-ppm-dap.md
```diff
@@ -2155,14 +2138,50 @@ structure includes the following:
...
   both multiples of the task's `time_precision` parameter. Note that in the case
   of a time-interval query (see {{batch-mode}}), this interval can be smaller
-  than the one in the corresponding `CollectionReq.query`.
+  than the one in the corresponding `CollectionJobReq.query`.
 
 * `leader_encrypted_agg_share`: The Leader's aggregate share, encrypted to the
   Collector.
 
 * `helper_encrypted_agg_share`: The Helper's aggregate share, encrypted to the
   Collector.
```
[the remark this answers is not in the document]
inahga: I'll bring it to a separate PR.

draft-ietf-ppm-dap.md
```diff
@@ -2155,64 +2155,48 @@ aggregation parameter in advance. For example, for Prio3 the only valid
...
-requesting an aggregate share from the Helper. This avoids a race condition
-between aggregation and collection jobs that can yield trivial batch mismatch
-errors.
-
-Once both aggregate shares are successfully obtained, the Leader responds to
-subsequent HTTP GET requests to the collection job with HTTP status code 200 OK
-and a body consisting of a `Collection`:
+`PUT /tasks/{task-id}/collection_jobs/{collection-job-id}` for the same
+`collection-job-id` but with a different `CollectionJobReq` in the body MUST
+fail with an HTTP client error status code.
```
cjpatton, line 2171: Typically we would write something like: "the Leader MUST abort with error blah", where "blah" is an error listed in https://ietf-wg-ppm.github.io/draft-ietf-ppm-dap/draft-ietf-ppm-dap.html#table-1. Why not do the same here?

draft-ietf-ppm-dap.md
```diff
@@ -2155,64 +2155,48 @@ aggregation parameter in advance. For example, for Prio3 the only valid
...
 
 ~~~ tls-presentation
+The Leader responds to `CollectionJobReq`s with a `CollectionJobResp`, which is
+structured as follows:
+~~~
```
cjpatton, line 2176:
```suggestion
structured as follows:

~~~
```

draft-ietf-ppm-dap.md
```diff
@@ -2226,14 +2210,48 @@ structure includes the following:
...
 
 * `leader_encrypted_agg_share`: The Leader's aggregate share, encrypted to the
   Collector.
 
 * `helper_encrypted_agg_share`: The Helper's aggregate share, encrypted to the
   Collector.
 
+If the Leader finds the `CollectionJobReq` to be valid, it immediately responds
+with HTTP status 201 Created with a body containing a `CollectionJobResp`. The
+`status` field is set to `processing`. The Leader SHOULD include a Retry-After
```
cjpatton, line 2223:
```suggestion
with HTTP status 201 Created with a body containing a `CollectionJobResp` with the
`status` field set to `processing`. The Leader SHOULD include a Retry-After
```

draft-ietf-ppm-dap.md
```diff
@@ -2226,14 +2210,48 @@ structure includes the following:
...
+header field to suggest a polling interval to the Collector.
+
+After receiving the response to its `CollectionJobReq`, the Collector
+periodically makes HTTP GET requests to the same collection job URI to check on
```
cjpatton, line 2227:
```suggestion
periodically makes HTTP GET requests to `/tasks/{task-id}/collection_jobs/{collection-job-id}` to check on
```

draft-ietf-ppm-dap.md
```diff
@@ -2226,14 +2210,48 @@ structure includes the following:
...
+{{batch-validation}} was not met. If the Leader has a pending aggregation job
+that overlaps with the batch and aggregation parameter for the collection job,
+the Leader MUST first complete the aggregation job before proceeding and
+requesting an aggregate share from the Helper. This avoids a race condition
+between aggregation and collection jobs that can yield trivial batch mismatch
+errors.
+
+Once both aggregate shares are successfully obtained, the Leader responds to
+subsequent HTTP GET requests with the `status` field set to `finished` and the
+`Collection` field populated accordingly. The Collector stops polling once
```
cjpatton, line 2252: Replace `XXX` with the relevant section (the Leader has the Helper's encrypted agg share and produces its own encrypted agg share)
```suggestion
`Collection` field populated as constructed as specified in {{XXX}}. The Collector stops polling once
```

cjpatton (approved):

draft-ietf-ppm-dap.md
inahga: I think at some point we decided to leave it unspecified what to return in case of conflict. Aggregation is similar in this respect https://ietf-wg-ppm.github.io/draft-ietf-ppm-dap/draft-ietf-ppm-dap.html#section-4.6.1.2-28.

I don't have an immediate opinion on whether we're right to do this or not. Maybe we should, in case a Leader/Collector needs to take special action based on a conflict response (I think in Janus we just immediately abandon the job).

cjpatton: Ack, disregard.

draft-ietf-ppm-dap.md
inahga: Sorry, I'm not seeing a section that best fits here. I suppose I can link the reader to the paragraph `The `Collection` structure includes the following` but that's midway through the section. (Maybe this section needs to be broken up a bit?)

cjpatton: You're right. Maybe "with the `Collection` field populated with the encrypted aggregate shares".

inahga: How does the latest look? I broke up the `tls-presentation` definition into two pieces, so the narrative flows a bit better generally. (i.e. "here's what it looks like when processing, now here's what it looks like when finished")

cjpatton: It's better if the definition of `Collection` comes before `CollectionJobResp`, since the latter depends on the former.

inahga: > You're right. Maybe "with the Collection field populated with the encrypted aggregate shares".

I went with this.

branlwyd (approved):

PR #602, opened 2024-10-04
Remove replay check from "input share validation" section.

branlwyd:
The related implementation note is removed, too -- it is nowadays duplicative of the text in the Replay Protection section.

I think this was an intended part of #582, based on the commit text of that PR.


branlwyd:
(Also: I believe the removed normative text is duplicative of the text added by #582, but please check that this is the case.)

cjpatton (approved):

PR #607, opened 2024-10-09
Say the job is "ready" rather than "finished"

cjpatton:
"finished" is not a meaningful name for the state of an aggregation job it describes. This is because the job reaches the state multiple times, after each step of the ping-pong flow. The job is only "finished" once the Helper has aggregated its output shares.

Another problem is that we use the term "finished" to describe the state of preparation for a single report.

To resolve these issues, rename the "finished" state to "ready". Likewise for collection jobs for consistency.


inahga (approved):
SGTM!

branlwyd (approved):

PR #612, opened 2024-10-09
Improve IANA considerations

### LINKED ISSUE #583: Complete IANA considerations
- [x]  Go through this section and make sure it's aligned with the body of the draft.
- [x] Define the "Batch Mode Registry"
- [x] Define the "Upload Extension Registry"
- [ ] Solicit review of media type allocations from domain experts (HTTPDIR review?)

cjpatton:
Closes #583.

* Set an administration policy for our codepoint registries (Specification Required).
* Make the initial contents of the URN namespace self-contained by defining the "errors" sub-namespace in the caption.
* Define the batch mode codepoints for time_interval and leader_selected in their respective sections so that we can use the section as the codepoint's reference.
* Whenever an IANA request refers to "this specification", replace the reference with "RFC XXXX" and ask the RFC Editor to replace "XXXX" with the RFC number assigned to the doc.
* Add a summary at the top of what types of requests are made to IANA (there are three).
* Update Security considerations references for each of the media types to refer to the relevant protocol interaction (upload, aggregation, or collection).
* Update the section reference in the non-normative list of messages and their media types. (Refer to the section in which the message is defined.)
* Merge "Message versioning" section into surrounding text.
* Harmonize request language and simplify instructions for the RFC Editor.

### DIFF
--- a/draft-ietf-ppm-dap.md
+++ b/draft-ietf-ppm-dap.md
@@ -725,8 +725,7 @@ a problem detail object, the HTTP status code MUST indicate a client or server
 error (the 4xx or 5xx classes, respectively, from {{Section 15 of !RFC9110}}).
 
 To facilitate automatic response to errors, this document defines the following
-standard tokens for use in the "type" field (within the DAP URN namespace
-"urn:ietf:params:ppm:dap:error:"):
+standard tokens for use in the "type" field:
 
 | Type                       | Description                                                                                  |
 |:---------------------------|:---------------------------------------------------------------------------------------------|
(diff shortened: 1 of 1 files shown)


draft-ietf-ppm-dap.md
```diff
@@ -898,6 +896,13 @@ The parameters pertaining to specific batch modes are described in
 
 ### Time-interval Batch Mode {#time-interval-batch-mode}
 
+~~~ tls-presentation
```
branlwyd, line 899: Do we really need to split each value out into its own section? IMO, this is harder to follow than what we had before.

I think we could move these back to where they were (e.g. centralize the definition of hte `BatchMode` enum), while still using these sections as the reference point for the Batch Modes Registry section, without loss of readability. (Is there a convention / requirement that the numeric code point be defined in the section referenced by the IANA table?) Alternatively, we could move these back to where they were & have the Batch Modes Registry section refer back to the Batch Modes section for all defined batch modes.

cjpatton: Part of my goal is to make these sections self-contained in a way that makes life a little easier for folks writing documents that specify new batch modes. Such documents will always have a line like this:

```
enum {
   cool_guy(42), (255)
} BatchMode;
```

and now that line can be copy-pasted from our RFC.

I also think this is a little simpler for whoever at IANA is making sure they get the registry right.

To answer your questiona bout conventions: I'm not aware of a hard and fast rule here. However this is a common pattern for example in TLS: https://tlswg.org/draft-ietf-tls-esni/draft-ietf-tls-esni.html#section-5-2

branlwyd (approved):

PR #615, opened 2024-10-16
Pass of abstract, intro, and references

cjpatton:
Make the terminology in the intro more consistent with the rest of the document, e.g., say "measurement" instead of "input".

For the references (cf.
https://github.com/cfrg/draft-irtf-cfrg-vdaf/pull/508):

* Use "refcontent" instead of "seriesinfo" for conferences
* Use "name" instead of "ins" with the authors' full names as they appear in the references
* Harmonize ordering and date format

Tidy up the presentation language section:

* We want to say "encoding and encoding of messages as bytes strings are as in RFC8446", but for some reason we say this into the second to last paragraph, between two exceptions to RFC8446 we want to document. Move this language up to the first paragraph.


draft-ietf-ppm-dap.md
```diff
@@ -387,28 +391,27 @@ input is ever seen in the clear by any aggregator.
...
 Aggregate share:
-: A share of the aggregate result emitted by an Aggregator. Aggregate shares are
-  reassembled by the Collector into the aggregate result, which is the final
-  output of the aggregation function. As defined in {{!VDAF}}.
+: A secret share of the aggregate result computed by each Aggregator and
+  transmitted to the Collector. As defined in {{!VDAF}}.
 
 Aggregation function:
-: The function computed over the Clients' measurements. As defined in {{!VDAF}}.
+: The function computed over the measurements generated by the Client and the
```
branlwyd, line 401:
```suggestion
: The function computed over the measurements generated by the Clients and the
```

draft-ietf-ppm-dap.md
```diff
@@ -451,25 +454,28 @@ Measurement:
...
+  collected.
 
 Public share:
-: The output of the VDAF sharding algorithm broadcast to each of the
+: The output of the VDAF sharding algorithm transmitted to each of the
   Aggregators. As defined in {{!VDAF}}.
 
 Report:
 : A cryptographically protected measurement uploaded to the Leader by a Client.
   Comprised of a set of report shares.
```
branlwyd, line 466:
```suggestion
  Includes a pair of report shares, one for each Aggregator.
```

(the switch from "comprise" to "include" is not for the normal pedantic reason, but because a report contains a few additional pieces of information, e.g. a report ID, client timestamp, public share. We could mention them all but I think that's too much detail, better to just shift language to note that the pair of report shares is only part of a Report.)

draft-ietf-ppm-dap.md
```diff
@@ -451,25 +454,28 @@ Measurement:
...
 
-Report Share:
-: An encrypted input share comprising a piece of a report.
+Report share:
+: An input share encrypted under the HPKE public key of an Aggregator.
```
branlwyd, line 469: Pedantically, the report share includes a few additional pieces of information -- report metadata & public share.

branlwyd (approved):
Just a couple of typographical issues, and some pedantry over the definitions list.

Issue #604, opened 2024-10-08
Determine aggregation skew recovery strategy


branlwyd:
Due to transient network issues, inopportunely-timed process shutdown, or implementation bugs, it is possible for the Leader & Helper aggregators to have a "skewed" view of the aggregation process. Specifically, the Helper might consider the aggregation process to one round further than the Leader, if the Leader "loses" the response from the Helper.

This is currently addressed by a skew-recovery scheme where, if the Helper receives a request corresponding to the _previous_ round, it replays its previous response. This allows a Leader who has skewed to catch up to the Helper. This was specified in https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/pull/400, and has been implemented and deployed.

But maybe this isn't the best recovery mechanism. We have also considered:
* Restart-from-beginning recovery: in this scheme, recovering from round skew requires the Leader to restart the aggregation process by re-sending the original request. The Helper stores the required information to allow the Leader to restart an aggregation job (i.e. the hash of the initial request). The Leader is expected to store the initial request until aggregation is complete.
* Restart-from-arbitrary-step recovery: in this scheme, the Leader is allowed to re-send any of the aggregation requests, and the Helper will replay the aggregation process from that point. The Helper stores the required information to allow the Leader to replay any aggregation step (i.e. a hash of each received request, and the response to that request). The Leader can recover from failure by replaying any request it can reconstruct.

Restart-from-beginning may be better because it is simpler for the Helper to implement, and in practice all of the VDAFs we have considered so far are one-round. Restart-from-arbitrary-step is more flexible than any of the other considered options, but may require storage of too much information.

We should decide the best aggregation skew-recovery scheme, and specify it.

branlwyd:
During an off-list discussion around https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/pull/616#discussion_r1807087612 today, we discovered that things might be more complicated than previously believed.

Specifically, when performing VDAF preparation, if either aggregator is able to "replay" the preparation interaction for a given report while changing some of the messages it sends to its peer, being able to observe their peer's differing response may let them break the privacy properties of the VDAF. (cc @cjpatton / @divergentdave in case I have misstated things.)

This means that a replay-protection scheme that allows either aggregator to change any of the messages on replay, and observe differing behavior from their peer aggregator, has a problem.

Specifically:
[...]

PR #616, opened 2024-10-17
Aggregation step skew recovery: restart from beginning.

branlwyd:
This replaces the existing aggregation step skew recovery scheme, which restarts from the previous step.

This likely simplifies the Helper, which no longer needs to remember the response it sends to each aggregation request. On the other hand, this somewhat complicates the Leader, which must now be able to replay the initial request, and which must understand when it needs to restart. The Leader is no longer expected to durably store each step of the aggregation process, however.

### DIFF
--- a/draft-ietf-ppm-dap.md
+++ b/draft-ietf-ppm-dap.md
@@ -1813,12 +1813,11 @@ same `aggregation-job-id` but with a different `AggregationJobInitReq` in the
 body MUST fail with an HTTP client error status code. For further requests with
 the same `AggregationJobInitReq` in the body, the Helper SHOULD respond as it
 did for the original `AggregationJobInitReq`, or otherwise fail with an HTTP
-client error status code.
-
-Additionally, it is not possible to rewind or reset the state of an aggregation
-job. Once an aggregation job has been continued at least once (see
-{{agg-continue-flow}}), further requests to initialize that aggregation job MUST
-fail with an HTTP client error status code.
+client error status code. The Helper SHOULD allow the aggregation job to be
+replayed even if the aggregation job has been continued (see
+{{agg-continue-flow}}), to allow for a Leader to recover from state skew between
+the Leader and Helper by replaying the aggregation; see
+{{aggregation-step-skew-recovery}} for more detail on the recovery process.
 
 #### Input Share Decryption {#input-share-decryption}
 
@@ -2019,14 +2018,8 @@ to the previous request, then the Helper MAY abort with error `invalidMessage`.
 Leader rejected it.)
 
 Next, the Helper checks if the continuation step indicated by the request is
-correct. (For the first `AggregationJobContinueReq` the value should be `1`;
-for the second the value should be `2`; and so on.) If the Leader is one step
-behind (e.g., the Leader has resent the previous HTTP request), then the Helper
-MAY attempt to recover by sending the same response as it did for the previous
-`AggregationJobContinueReq`, without performing any additional work on the
-aggregation job. In this case it SHOULD verify that the contents of the
-`AggregationJobContinueReq` are identical to the previous message (see
-{{aggregation-step-skew-recovery}}). Otherwise, if the step is incorrect, the
+correct. (For the first `AggregationJobContinueReq` the value should be `1`; for
+the second the value should be `2`; and so on.) If the step is incorrect, the
 Helper MUST abort with error `stepMismatch`.
 
 Let `inbound` denote the payload of the prep step. For each report, the Helper
@@ -2118,28 +2111,30 @@ Helper knows it can clean up its state.
 #### Recovering from Aggregation Step Skew {#aggregation-step-skew-recovery}
 
 `AggregationJobContinueReq` messages contain a `step` field, allowing
-Aggregators to ensure that their peer is on an expected step of DAP
-aggregation. In particular, the intent is to allow recovery from a scenario
-where the Helper successfully advances from step `n` to `n+1`, but its
-`AggregationJobResp` response to the Leader gets dropped due to something like
-a transient network failure. The Leader could then resend the request to have
-the Helper advance to step `n+1` and the Helper should be able to retransmit
-the `AggregationJobResp` that was previously dropped. To make that kind of
-recovery possible, Aggregator implementations SHOULD checkpoint the most recent
-step's prep state and messages to durable storage such that the Leader can
-re-construct continuation requests and the Helper can re-construct continuation
-responses as needed.
+Aggregators to ensure that their peer is on the expected step of DAP
+aggregation. In particular, the intent is to allow detection and recovery from a
+scenario where the Helper and the Leader have differing views of the current
+aggregation step, which may occur if an `AggregationJobResp` response to the
+Leader is dropped due to a transient network failure. The Leader could then
+resend the original `AggregationJobInitReq` message to restart the aggregation
+process, allowing aggregation to complete successfully. To make this kind of
+recovery possible, Leader implementations SHOULD retain enough information to
+reconstruct the `AggregationJobInitReq` message in the case that the aggregation
+process becomes desynchronized.
 
 When implementing an aggregation step skew recovery strategy, the Helper SHOULD
-ensure that the Leader's `AggregationJobContinueReq` message did not change when
-it was re-sent (i.e., the two messages must be identical). This prevents the
-Leader from re-winding an aggregation job and re-running an aggregation step
-with different parameters.
+ensure that the Leader's `AggregationJobInitReq` message did not change when it
+was re-sent (i.e., the two messages must be identical). This prevents the Leader
+from re-winding an aggregation job and re-running aggregation with different
+parameters.
 
-One way the Helper could address this would be to store a digest of the Leader's
-request, indexed by aggregation job ID and step, and refuse to service a request
-for a given aggregation step unless it matches the previously seen request (if
-any).
+One way the Helper could implement this would be to store a digest of the
+Leader's request, indexed by aggregation job ID, and refuse to service a re-sent
+request unless it matches the previously seen request.
+
+Implementation note: in the case that the Leader replays an aggregation which
+the Helper considers to be complete, the Helper should ensure it does not commit
+the results of the aggregation to its aggregate shares more than once.
 
 ## Collecting Results {#collect-flow}
 


branlwyd:
This is a prospective change to be discussed at the upcoming PPM interim meeting.

See #604.

draft-ietf-ppm-dap.md
```diff
@@ -2118,28 +2111,30 @@ Helper knows it can clean up its state.
...
+scenario where the Helper and the Leader have differing views of the current
+aggregation step, which may occur if an `AggregationJobResp` response to the
+Leader is dropped due to a transient network failure. The Leader could then
+resend the original `AggregationJobInitReq` message to restart the aggregation
+process, allowing aggregation to complete successfully. To make this kind of
+recovery possible, Leader implementations SHOULD retain enough information to
+reconstruct the `AggregationJobInitReq` message in the case that the aggregation
+process becomes desynchronized.
 
 When implementing an aggregation step skew recovery strategy, the Helper SHOULD
```
cjpatton, line 2125: Here we should repeat the line about "changing an aggregation job's parameters is illegal".

draft-ietf-ppm-dap.md
```diff
@@ -2118,28 +2111,30 @@ Helper knows it can clean up its state.
...
+was re-sent (i.e., the two messages must be identical). This prevents the Leader
+from re-winding an aggregation job and re-running aggregation with different
+parameters.
+
+One way the Helper could implement this would be to store a digest of the
+Leader's request, indexed by aggregation job ID, and refuse to service a re-sent
+request unless it matches the previously seen request.
+
+Implementation note: in the case that the Leader replays an aggregation which
+the Helper considers to be complete, the Helper should ensure it does not commit
```
cjpatton, line 2136:
```suggestion
the Helper considers to be complete, the Helper must ensure it does not commit
```

draft-ietf-ppm-dap.md
```diff
@@ -1813,12 +1813,11 @@ same `aggregation-job-id` but with a different `AggregationJobInitReq` in the
...
-
-Additionally, it is not possible to rewind or reset the state of an aggregation
-job. Once an aggregation job has been continued at least once (see
-{{agg-continue-flow}}), further requests to initialize that aggregation job MUST
-fail with an HTTP client error status code.
+client error status code. The Helper SHOULD allow the aggregation job to be
+replayed even if the aggregation job has been continued (see
+{{agg-continue-flow}}), to allow for a Leader to recover from state skew between
+the Leader and Helper by replaying the aggregation; see
+{{aggregation-step-skew-recovery}} for more detail on the recovery process.
```
cjpatton, line 1820: I think there's terminology here that we need to define more explicitly:
* "skew": Perhaps add a sentence like `The state of an aggregation job is said to be "skewed" if ...`
* "replay": I actually think we should try and avoid overloading this term (it's already used in the context of overexposing reports).

cjpatton (commented):
I think it would be a good idea to summarize the intent of this mechanism somewhere:

> An aggregation job takes requires many round trips over the network to complete. If one of the Aggregators fails during this process in a manner that results in it losing its state, then the Aggregators risk losing the data associated with the aggregation job, or the batch altogether (if one of the Aggregators fails to aggregate, resulting in a batch mismatch). DAP is designed to deal with some common failure scenarios:
>
> * <List the scenarios>

Then describe the mechanism at a high level.

Stepping back a bit: I think we need to build up some language/vocabulary for how to talk about this.

draft-ietf-ppm-dap.md
branlwyd: OK. Perhaps more controversially, I changed the `SHOULD` to a `MUST` to match the `MUST` around "changing an aggregation job's parameters is illegal" -- I think these were effectively in conflict in terms of how required this was.

(Historically, I think we specified a `SHOULD` here to permit an implementation to ignore the request body on a re-send -- other than the `step` parameter, I suppose -- and blindly reply with the previous response. I'm OK dropping this because (a) no one has expressed interest in implementing this and (b) boy howdy, those semantics would be challenging to debug -- the response would potentially be completely independent of the request.)

draft-ietf-ppm-dap.md
branlwyd: * "skew": Done. While wordsmithing I also added the exact condition that the Leader would re-send the original request, i.e. on receiving a `stepMismatch` error.
* "replay": makes sense, I replaced these usages of "replay" with "restart".

draft-ietf-ppm-dap.md
```diff
@@ -1813,12 +1813,11 @@ same `aggregation-job-id` but with a different `AggregationJobInitReq` in the
...
 did for the original `AggregationJobInitReq`, or otherwise fail with an HTTP
-client error status code.
-
-Additionally, it is not possible to rewind or reset the state of an aggregation
-job. Once an aggregation job has been continued at least once (see
-{{agg-continue-flow}}), further requests to initialize that aggregation job MUST
-fail with an HTTP client error status code.
+client error status code. The Helper SHOULD allow the aggregation job to be
+restarted even if the aggregation job has been continued (see
+{{agg-continue-flow}}), to allow for a Leader to recover from state skew between
```
cjpatton, line 1818: Restarting entails re-sending the `AggregationJobInitReq` _unchanged_, correct?

draft-ietf-ppm-dap.md
```diff
@@ -2117,29 +2110,43 @@ Helper knows it can clean up its state.
...
 #### Recovering from Aggregation Step Skew {#aggregation-step-skew-recovery}
 
+An aggregation job may require many round trips over the network to complete. If
+one of the Aggregators fails during this process in a manner that results in it
+losing its state, then the Aggregators risk losing the data associated with the
+aggregation job, or in some cases all of the batches associated with the
+aggregation job (if one of the Aggregators considers the aggregation to be
+complete but the other does not, the Aggregators will encounter a
+`batchMismatch` error on attempting to collect the batches associated with the
+aggregation job). This section describes how Aggregators can recover from this
```
cjpatton, line 2120:
```suggestion
aggregation job. (For example, if one of the Aggregators considers the aggregation to be
complete but the other does not, the Aggregators will encounter a
`batchMismatch` error on attempting to collect the batches associated with the
aggregation job.) This section describes how Aggregators can recover from this
```

draft-ietf-ppm-dap.md
```diff
@@ -2117,29 +2110,43 @@ Helper knows it can clean up its state.
...
+scenario.
```
cjpatton, line 2121:
```suggestion
aggregation job). This section describes how Aggregators can recover from such
failures.
```

draft-ietf-ppm-dap.md
```diff
@@ -2117,29 +2110,43 @@ Helper knows it can clean up its state.
...
-with different parameters.
-
-One way the Helper could address this would be to store a digest of the Leader's
-request, indexed by aggregation job ID and step, and refuse to service a request
-for a given aggregation step unless it matches the previously seen request (if
-any).
+Aggregators to ensure that their peer is on the expected step of DAP
+aggregation. An aggregation job is said to be "skewed" if the Leader and Helper
+have a different view of the last completed step of aggregation; for example, if
+the Leader does not receive or successfully process an `AggregationJobResp` from
```
cjpatton, line 2127: "say, due to a failure of the underlying transport"

draft-ietf-ppm-dap.md
```diff
@@ -2117,29 +2110,43 @@ Helper knows it can clean up its state.
...
+the Helper, the Leader will believe that the last completed step is one less
+than the Helper. The intent of the `step` field is to allow detection and
+recovery from a skewed aggregation job. When the Leader receives a
+`stepMismatch` error, it SHOULD re-send the original `AggregationJobInitReq`
+message to restart the aggregation process and allow aggregation to complete
+successfully. To make this kind of recovery possible, Leader implementations
+SHOULD retain enough information to reconstruct the `AggregationJobInitReq`
+message in the case that the aggregation process becomes desynchronized.
```
cjpatton, line 2135: I think this SHOULD is redundant, and maybe can be cut completely? What's the intent of this sentence?
```suggestion
successfully.
```

draft-ietf-ppm-dap.md
```diff
@@ -2117,29 +2110,43 @@ Helper knows it can clean up its state.
...
+
+Changing an aggregation job's parameters is illegal, so the Helper MUST ensure
+that the Leader's `AggregationJobInitReq` message did not change when it was
+re-sent (i.e., the two messages must be identical). This prevents the Leader
+from re-winding an aggregation job and re-running aggregation with different
+parameters.
+
+One way the Helper could implement this would be to store a digest of the
+Leader's request, indexed by aggregation job ID, and refuse to service a re-sent
+request unless it matches the previously seen request.
```
cjpatton, line 2145: FWIW each AggreagtionJobContinueReq needs to be the same as well. Though I believe the protocol logic should ensure this? It's still worth mentioning in any case.

Also the responses shouldn't change either.

draft-ietf-ppm-dap.md
```diff
@@ -2117,29 +2110,43 @@ Helper knows it can clean up its state.
...
+
+Implementation note: in the case that the Leader restarts an aggregation which
+the Helper considers to be complete, the Helper must ensure it does not commit
```
cjpatton, line 2148:
```suggestion
the Helper considers to be complete, the Helper MUST NOT commit
```

draft-ietf-ppm-dap.md
```diff
@@ -1813,12 +1813,11 @@ same `aggregation-job-id` but with a different `AggregationJobInitReq` in the
...
-client error status code.
-
-Additionally, it is not possible to rewind or reset the state of an aggregation
-job. Once an aggregation job has been continued at least once (see
-{{agg-continue-flow}}), further requests to initialize that aggregation job MUST
-fail with an HTTP client error status code.
+client error status code. The Helper SHOULD allow the aggregation job to be
+restarted even if the aggregation job has been continued (see
+{{agg-continue-flow}}), to allow for a Leader to recover from state skew between
+the Leader and Helper by restarting the aggregation; see
```
kristineguo, line 1819: Mentioning restarting the aggregation twice is redundant.

```suggestion
client error status code. The Helper SHOULD allow the aggregation job to be
restarted even if the aggregation job has been continued (see
{{agg-continue-flow}}) to allow a Leader to recover from state skew; see
```

draft-ietf-ppm-dap.md
```diff
@@ -2117,29 +2110,43 @@ Helper knows it can clean up its state.
 
 #### Recovering from Aggregation Step Skew {#aggregation-step-skew-recovery}
 
+An aggregation job may require many round trips over the network to complete. If
+one of the Aggregators fails during this process in a manner that results in it
+losing its state, then the Aggregators risk losing the data associated with the
+aggregation job, or in some cases all of the batches associated with the
```
kristineguo, line 2116:
```suggestion
aggregation job, or in some cases, all of the batches associated with the
```

draft-ietf-ppm-dap.md
kristineguo: +1. The Leader could restart an aggregation job and arbitrarily remove valid report shares from subsequent AggregationJobContinueReqs. In that case, the Helper may need to ensure the request digest hasn't changed for every single step involved.

It could do this by storing the digest indexed by aggregation job ID and step. But, this increases the amount of state required from one digest (in the -1 recovery) to n digests (in from-the-beginning recovery).

draft-ietf-ppm-dap.md
branlwyd: Yes, this is spelled out in the "Recovering from Aggregation Step Skew" section; I changed the wording to spell it out here, too.

draft-ietf-ppm-dap.md
branlwyd: I'm explicitly suggesting/hinting to Leader implementations that they need to keep the original request around (or enough information to reconstruct it byte-for-byte), if they want to implement skew recovery.

I suppose this is implied well enough by the previous sentence, and the following description of the Helper behavior -- I removed this sentence.

draft-ietf-ppm-dap.md
branlwyd: Can y'all spell out the specific problem or attack you have in mind?

I don't think we need to ensure the later requests are the same. My reasoning is:
* If the Helper did not consider the aggregation process complete when the retry was initiated, no state outside of the aggregation job has been modified -- so if the Leader ends up removing a different set of reports on retry, there is no problem.
* If the Helper did consider the aggregation process complete when the retry was initiated, it will have already committed the recovered output shares & noted that it has done so, and thus will not commit a second time. If the Leader aggregates a different set of reports, this will result in a `batchMismatch` error, but the Leader has no ability to e.g. aggregate a slightly different set of reports to infringe on a particular report's privacy.

OTOH, the Leader _accidentally_ aggregating a different set of reports against an aggregation job the Helper considers to be complete is likely a downside of this approach -- it would be hard to avoid in corner cases. The most likely specified corner cases I can find here would be (a) the HPKE key rotating or (b) the replay-checking state being dropped, both of which would result in the report becoming un-aggregatable on retry. (Tracking the exact request/responses at each step wouldn't help -- we'd fail earlier & maybe with a better error code, but would still eventually encounter a `batchMismatch` error on collection, since the Leader would never be able to complete the aggregation job. I think a truly robust implementation of the restart-from-beginning idea would require the Helper to have the ability to un-commit an aggregation job from the running-total aggregate shares when the aggregation job is restarted, which I suppose would require tracking the aggregate shares for each aggregation job -- i.e. we'd lose the storage savings of the running-total aggregation strategy.)

draft-ietf-ppm-dap.md
branlwyd: Done. Is it appropriate to include a `MUST` in an implementation note? (If not, I think this behavior is important enough to promote from an implementation note into regular text.)

cjpatton: Yes, I think it's appropriate.

draft-ietf-ppm-dap.md
cjpatton: Good point. I was originally thinking of something like this: the Helper remembers a hash of each request, and if the request doesn't match the hash, then it aborts the aggregation job. But you're right that we need to be robust to either the Leader or Helper making different rejection decisions.

Another wrinkle pointed out by @divergentdave on the 2024/10/23 on the call: we need to be careful about VDAF preparation. Existing security analysis assumes the attacker can't "rewind" the computation and retry with different inputs. In particular, we're worried that reusing the Prio3 proof shares may result in a privacy violation. (Likewise for the Beaver triples in Poplar1.) In particular we need to ensure the Leader doesn't change the prep share it sends for each report.

branlwyd: There was some off-PR discussion about this today, written up in https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/issues/604#issuecomment-2433065034.

The summary is that not only do we need the Helper to verify that the entire sequence of requests is the same on retry, we _also_ need the Leader to verify that the entire sequence of responses is the same on retry.

This is achieved by the restart-from-previous-step currently specified in DAP-12, but is not achieved by the restart-from-beginning (this PR) or restart-from-anywhere (#569) strategies as currently written. We could certainly augment this PR or #569 to specify the requirement that the Leader record the sequence of responses it has successfully processed; but AIUI the original motivation for changing the retry strategy was to permit a Leader implementation which did not need to record anything to durable storage between aggregation steps. If the Leader now has to store response hashes to durable storage between aggregation steps, we no longer meet the original motivation for modifying the retry strategy.

kristineguo: Thank you for the detailed write up and I agree. Although this strategy would work for two happy Aggregators, it cannot protect against scenarios where one is malicious, compromising our security guarantees. If the Leader and Helper would need to remember and enforce the exact digests for each step anyways, we may as well just restart from the last "committed" step between the two, which falls into line with the current status quo.

cjpatton:
2024/10/24 interim: We're uncomfortable with supporting recovery from any sstart, as we'd have to figure out how to deal with rewind-then-fork attacks on VDAF preparation as described [here](https://datatracker.ietf.org/meeting/interim-2024-ppm-02/materials/slides-interim-2024-ppm-02-sessa-dap-and-taskprov-open-issues-00). We'll give this a little more time to discuss, but if there are no new ideas by 2024/10/31, then we'll close this PR.

Issue #581, opened 2024-09-18
Make streaming aggregation normative


cjpatton:
An issue that's been raised several times (most recently in #571) is that the term "aggregation job" is a misnomer since it doesn't actually say anything about aggregation. All it says is that, once an aggregation job is complete, each aggregator commits to the sequence of output shares they recovered. It's not until collection time that we call [`Vdaf.aggregate()`](https://datatracker.ietf.org/doc/html/draft-irtf-cfrg-vdaf-11#section-5.4) on the set of output shares to get our aggregate share for the batch.

Given this fact, it's been suggested we rename "aggregation job" to something like "preparation job", since the main thing that happens during this protocol is [VDAF preparation](https://datatracker.ietf.org/doc/html/draft-irtf-cfrg-vdaf-11#section-5.2).

I don't recall precisely how this term entered our vocabulary, but part of the reason it stuck is that, in practice, we expect implementations to do some, if not most, of the aggregation during this phase. From [Section 4.7](https://ietf-wg-ppm.github.io/draft-ietf-ppm-dap/draft-ietf-ppm-dap.html#section-4.7-11):

> After an aggregation job is completed, each Aggregator stores the output shares until the aggregate share is collected as described in [Section 4.8](https://ietf-wg-ppm.github.io/draft-ietf-ppm-dap/draft-ietf-ppm-dap.html#collect-flow). Note that it is usually not necessary to store output shares individually: depending on the batch mode and VDAF, the output shares can be merged into existing aggregate shares that are updated as aggregation jobs complete. This streaming aggregation is compatible with Prio3 and all batch modes specified in this document.

In other words, we expect "streaming aggregation" to be the norm for DAP implementations and "delayed aggregation" during the collection (credit @bemasc for this terminology) to be the rare exception. In theory, you could have a VDAF for which aggregation is sensitive to the order of output shares; or a batch mode may require us to store output shares individually until collection. Neither is true today, and to me both seem pretty unlikely.

Rather rename the aggregation sub-protocol, we could make this streaming aggregation feature normative. This would require two things:

1. The concept of a "batch bucket" (#385). The definition of "batch bucket" depends on the batch mode: for time-interval, it's a window of time (determined by the `time_precision` task parameter); for leader-selected, it's simply a batch ID. Each bucket has an aggregate share. At the end of an aggregation job, we would add each output share into the aggregate share of the bucket it corresponds to.

[...]

PR #623, opened 2024-10-22
Specify "streaming" aggregation as the norm.

branlwyd:
Previously, aggregation was specified as if all output shares were aggregated into an aggregate share in one fell swoop, at time of collection. However, all implementations instead used an incremental aggregation process to keep a "running total" aggregate share which was updated as aggregation jobs completed, for performance reasons. This commit switches the specification to match the implementations.

Phrasing things this way also makes the name "aggregation job" make sense; as previously specified, there was no aggregation occurring inside of an aggregation job!

### DIFF
--- a/draft-ietf-ppm-dap.md
+++ b/draft-ietf-ppm-dap.md
@@ -413,6 +413,11 @@ Batch:
 : A set of reports (i.e., measurements) that are aggregated into an aggregate
   result. As defined in {{!VDAF}}.
 
+Batch bucket:
+: State associated with a given batch allowing the aggregators to perform
+  incremental aggregation. Depending on the batch mode, there may be many batch
+  buckets tracking the state of a single batch.
+
 Batch duration:
 : The time difference between the oldest and newest report in a batch.
 
@@ -663,7 +668,10 @@ time.
 
 The Leader distributes the shares to the Helper and orchestrates the process of
 verifying them (see {{validating-inputs}}) and assembling them into a final
-aggregate result for the Collector.
+aggregate result for the Collector. Depending on the VDAF, it may be possible to
+incrementally process each report as it is uploaded, or it may be necessary to
+wait until the Collector transmits a collection request before processing can
+begin.
 
 ## Validating Measurements {#validating-inputs}
 
@@ -1406,10 +1414,13 @@ across vast numbers of reports and preparation of large histograms, may be
 better suited for the asynchronous model. For use cases where datastore
 performance is a concern, the synchronous model may be better suited.
 
-In general, reports cannot be aggregated until the Collector specifies an
-aggregation parameter. However, in some situations it is possible to begin
-aggregation as soon as reports arrive. For example, Prio3 has just one valid
-aggregation parameter (the empty string).
+In general, aggregation cannot begin until the Collector specifies a query and
+an aggregation parameter. However, depending on the VDAF and batch mode in use,
+it is often possible to begin aggregation as soon as reports arrive. For
+example, Prio3 has just one valid aggregation parameter (the empty string), and
+thus allows for eager aggregation; and both the time-interval and
+leader-selected batch modes defined in this document allow for eager
+aggregation.
 
 An aggregation job can be thought of as having three phases, which are
 described in the remaining subsections:
@@ -1419,15 +1430,13 @@ described in the remaining subsections:
 - Continuation: Continue the aggregation flow by exchanging prep shares and
   messages until preparation completes or an error occurs.
 - Completion: Finish the aggregation flow, yielding an output share
-  corresponding to each report share in the aggregation job.
+  corresponding to each report share in the aggregation job, and update the
+  appropriate aggregate shares with these output shares.
 
-After an aggregation job is completed, each Aggregator stores the output shares
-until the aggregate share is collected as described in {{collect-flow}}. Note
-that it is usually not necessary to store output shares individually: depending
-on the batch mode and VDAF, the output shares can be merged into existing
-aggregate shares that are updated as aggregation jobs complete. This streaming
-aggregation is compatible with Prio3 and all batch modes specified in this
-document.
+After an aggregation job is completed, each Aggregator updates running-total
+aggregate shares and other values for each "batch bucket" associated with a
+recovered output share, as described in {{batch-buckets}}. These values are
+stored until the batch is collected as described in {{collect-flow}}.
 
 Apart from VDAF preparation, another important task of the aggregation
 interaction is to provide replay protection ({{replay-protection}}). Along with
@@ -1626,8 +1635,8 @@ Otherwise, the Leader proceeds as follows with each report:
    the preparation process is complete: either `state == Rejected()`, in which
    case the Leader rejects the report and removes it from the candidate set; or
    `state == Finished(out_share)`, in which case preparation is complete and the
-   Leader stores the output share for use in the collection interaction
-   {{collect-flow}}.
+   Leader updates the relevant batch bucket with `out_share` as described in
+   {{batch-buckets}}.
 
 1. Else if the type is "reject", then the Leader rejects the report and removes
    it from the candidate set. The Leader MUST NOT include the report in a
@@ -1637,8 +1646,8 @@ Otherwise, the Leader proceeds as follows with each report:
 1. Else the type is invalid, in which case the Leader MUST abort the
    aggregation job.
 
-When the Leader stores the `out_share`, it MUST also store the report ID for
-replay protection.
+When the Leader updates a batch bucket based on `out_share`, it MUST also store
+the report ID for replay protection.
 
 (Note: Since VDAF preparation completes in a constant number of rounds, it will
 never be the case that some reports are completed and others are not.)
@@ -1785,8 +1794,8 @@ variant {
 } PrepareResp;
 ~~~
 
-Otherwise it stores the report ID for replay protection and `out_share` for use
-in the collection interaction ({{collect-flow}}).
+Otherwise it stores the report ID for replay protection and updates the relevant
+batch bucket with `out_share` as described in {{batch-buckets}}.
 
 Finally, the Helper creates an `AggregationJobResp` to send to the Leader. This
 message is structured as follows:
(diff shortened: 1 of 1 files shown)


branlwyd:
Closes #581.
Closes #385.

draft-ietf-ppm-dap.md
```diff
@@ -413,6 +413,10 @@ Batch:
 : A set of reports (i.e., measurements) that are aggregated into an aggregate
   result. As defined in {{!VDAF}}.
 
+Batch bucket:
+: Data associated with a given batch allowing the aggregators to perform
```
cjpatton, line 417: * We should try to make it clear here that there is a many-to-one relationship between buckets and batches.

draft-ietf-ppm-dap.md
```diff
@@ -1395,10 +1400,13 @@ across vast numbers of reports and preparation of large histograms, may be
...
 
-In general, reports cannot be aggregated until the Collector specifies an
-aggregation parameter. However, in some situations it is possible to begin
-aggregation as soon as reports arrive. For example, Prio3 has just one valid
-aggregation parameter (the empty string).
+In general, reports cannot be aggregated until the Collector specifies a query
+and an aggregation parameter. However, depending on the VDAF and batch mode in
+use, it is often possible to begin aggregation as soon as reports arrive. For
+example, Prio3 has just one valid aggregation parameter (the empty string), and
+thus allows for eager aggregation; and both the time-interval and
```
cjpatton, line 1407: Aligns better with discussion on issue.

```suggestion
thus allows for streaming aggregation; and both the time-interval and
```

draft-ietf-ppm-dap.md
```diff
@@ -2121,6 +2128,54 @@ send an HTTP DELETE request to
...
+mode-specific "batch bucket identifier":
+
+* For the time-interval batch mode, the batch bucket identifier is an interval
+  of time whose beginning is a multiple of the task's `time_precision` and whose
+  duration is equal to the task's `time_precision`. The identifier associated
+  with a given output share is the unique such interval containing the client
+  timestamp of the report the output share was recovered from.
+* For the leader-selected batch mode, the batch bucket identifier is a batch ID.
+  The identifier associated with a given output share is the batch ID of the
+  related aggregation.
```
cjpatton, line 2147: What do you mean by "aggregation"?

draft-ietf-ppm-dap.md
```diff
@@ -2121,6 +2128,54 @@ send an HTTP DELETE request to
 `{helper}/tasks/{task-id}/aggregation_jobs/{aggregation-job-id}` so that the
 Helper knows it can clean up its state.
 
+#### Batch Buckets {#batch-buckets}
+
+When aggregation recovers an output share, it must be stored into an appropriate
+"batch bucket", which is defined in this section. The data stored in a batch
+bucket is kept for eventual use in the {{collect-flow}}.
+
+Batch buckets are indexed by an aggregation parameter, as well as a batch
```
cjpatton, line 2137: Since we no longer support multicollection, perhaps it makes sense to remove the aggregation parameter from the index.

draft-ietf-ppm-dap.md
```diff
@@ -2121,6 +2128,54 @@ send an HTTP DELETE request to
...
+mode-specific "batch bucket identifier":
+
+* For the time-interval batch mode, the batch bucket identifier is an interval
+  of time whose beginning is a multiple of the task's `time_precision` and whose
+  duration is equal to the task's `time_precision`. The identifier associated
+  with a given output share is the unique such interval containing the client
+  timestamp of the report the output share was recovered from.
```
cjpatton, line 2144: It might be helpful to give an example:
For example if `time_precision` is 3,600 seconds and the report's `timestamp` is `1729629081`, then the report's batch bucket is `...`

draft-ietf-ppm-dap.md
```diff
@@ -2121,6 +2128,54 @@ send an HTTP DELETE request to
...
+* For the leader-selected batch mode, the batch bucket identifier is a batch ID.
+  The identifier associated with a given output share is the batch ID of the
+  related aggregation.
+
+A few different pieces of information are associated with each batch bucket:
+
+* An VDAF-specific aggregate share, as defined in {{!VDAF}}.
+* A count of a number of reports included in the batch bucket.
+* A 32-byte checksum value.
```
cjpatton, line 2153:
```suggestion
* A 32-byte checksum value, as defined below.
```

draft-ietf-ppm-dap.md
```diff
@@ -2121,6 +2128,54 @@ send an HTTP DELETE request to
...
+    aggregate share value is computed as `Vdaf.agg_init(agg_param)`; the initial
+    count is `0`; and the initial checksum is 32 zero bytes.
+* Update the aggregate share `agg_share` to `Vdaf.agg_update(agg_param,
+  agg_share, out_share)`.
+* Increment the count by 1.
+* Update the checksum value to the bitwise-XOR of the checksum value with the
+  SHA256 {{!SHS=DOI.10.6028/NIST.FIPS.180-4}} hash of the report ID associated
+  with the output share.
+
+Implementation note: this text describes a single set of values associated with
```
cjpatton, line 2171:
```suggestion
Implementation note: this section describes a single set of values associated with
```

draft-ietf-ppm-dap.md
```diff
@@ -2121,6 +2128,54 @@ send an HTTP DELETE request to
...
+* A 32-byte checksum value.
+
+When processing an output share `out_share`, the following procedure is used to
+update the batch buckets:
+
+* Look up the existing batch bucket for the aggregation parameter `agg_param`
+  associated with the aggregation job, and the batch bucket identifier
+  associated with the aggregation job and output share.
+  * If there is no existing batch bucket, initialize a new one. The initial
+    aggregate share value is computed as `Vdaf.agg_init(agg_param)`; the initial
```
cjpatton, line 2162: Add reference to section in the VDAF draft that will define this. Draft 13 isn't out yet, but the section number is stable at this point (https://cfrg.github.io/draft-irtf-cfrg-vdaf/draft-irtf-cfrg-vdaf.html#section-5.4).

draft-ietf-ppm-dap.md
```diff
@@ -2121,6 +2128,54 @@ send an HTTP DELETE request to
...
+* Update the checksum value to the bitwise-XOR of the checksum value with the
+  SHA256 {{!SHS=DOI.10.6028/NIST.FIPS.180-4}} hash of the report ID associated
+  with the output share.
+
+Implementation note: this text describes a single set of values associated with
+each batch bucket. However, implementations are free to shard the aggregate
+share/count/checksum values associated with the batch bucket, combining them
+back into a single set of values when reading the batch bucket. This may be
+useful to avoid write contention. The aggregate share values are combined using
+`Vdaf.merge(agg_param, agg_shares)`, the count values are combined by summing,
```
cjpatton, line 2176: Add reference to VDAF spec

draft-ietf-ppm-dap.md
```diff
@@ -2321,12 +2376,29 @@ state related to it.
...
+
+* For time-interval tasks, this is all batch buckets associated with the
+  aggregation parameter provided in the `CollectionJobReq`, and whose batch
+  bucket identifiers are contained within the batch interval specified in the
+  `CollectionJobReq`'s query.
+* For leader-selected tasks, this is the batch bucket associated with the
+  aggregation parameter provided in the `CollectionJobReq`, and the batch ID the
+  Leader has chosen for this collection job.
+
+The Leader then combines the values inside the batch bucket, as so:
```
cjpatton, line 2394:
```suggestion
The Leader then combines the values inside the batch bucket as follows:
```

draft-ietf-ppm-dap.md
```diff
@@ -2321,12 +2376,29 @@ state related to it.
...
+
+* Aggregate shares are combined via `Vdaf.merge(agg_param, agg_shares)`, where
```
cjpatton, line 2396: Reference to VDAF draft

draft-ietf-ppm-dap.md
```diff
@@ -2381,23 +2454,11 @@ task ID in the request path. If not, it MUST abort with error
...
-agg_share = Vdaf.aggregate(agg_param, out_shares)
-~~~
-
-Implementation note: For most VDAFs, including Prio3, it is possible to
-aggregate output shares as they arrive rather than wait until the batch is
-collected. For the batch modes specified in this document, it is necessary to
-enforce the batch parameters as described in {{batch-validation}} so that the
-Aggregator knows which aggregate share to update.
+Next, the Helper retrieves and combines the batch buckets associated with the
+request using the same process used by the Leader, arriving at its final
```
cjpatton, line 2458: Backwards reference
```suggestion
request using the same process used by the Leader ({{XXX}}), arriving at its final
```

cjpatton (commented):
Excellent work! Just nits.

draft-ietf-ppm-dap.md
branlwyd: Good point!

As written, we do still allow aggregators to aggregate over an arbitrary number of aggregation parameters; the requirement is that we only allow collection of a batch over a single aggregation parameter.

I removed the index & added a few sentences requiring the Leader & Helper to validate that the aggregation parameter in the `CollectionJobReq` & `AggregateShareReq` match what was used for aggregation of the relevant batch.

But this might still be tricky: a naive Helper that permits aggregation of a batch with multiple aggregation parameters, and doesn't account for there being multiple aggregation parameters in its batch bucket implementation, might aggregate into batch buckets "cross-aggregation-parameter." We might want to block this, but I'm not sure this PR is the right place to do so. Or maybe an implementation note is sufficient?

draft-ietf-ppm-dap.md
branlwyd: Hmm, I'm not so sure. I think there are two concepts at play:

* "All-at-once aggregation" vs "streaming aggregation": this refers to whether aggregation happens "all at once", after recovering all of the relevant output shares (as was specified prior to this PR); or if aggregation happens in a "streaming" fashion as output shares are recovered (as is specified in this PR, and implemented in all DAP implementations).
* "Eager aggregation" vs "lazy aggregation": this refers to whether the aggregation process occurs "eagerly" as reports arrive, or if it is performed "lazily" once a collection request arrives. (Lazy aggregation is required if the Collector needs to choose an aggregation parameter.) Which of these is used is dependent on the implementation, both before & after this PR.

branlwyd:
(rebased on current `main`)

draft-ietf-ppm-dap.md
cjpatton: You're right, thanks for clarifying.

cjpatton (approved):

draft-ietf-ppm-dap.md
```diff
@@ -1398,10 +1406,13 @@ across vast numbers of reports and preparation of large histograms, may be
 better suited for the asynchronous model. For use cases where datastore
 performance is a concern, the synchronous model may be better suited.
 
-In general, reports cannot be aggregated until the Collector specifies an
-aggregation parameter. However, in some situations it is possible to begin
-aggregation as soon as reports arrive. For example, Prio3 has just one valid
-aggregation parameter (the empty string).
+In general, reports cannot be aggregated until the Collector specifies a query
```
bemasc, line 1409: Is this what is meant here?

```suggestion
In general, aggregation cannot complete until the Collector specifies a query
```

cjpatton:
interim 2024/10/24: @branlwyd this is good to merge once you've addressed the last comments.

draft-ietf-ppm-dap.md
```diff
@@ -2124,6 +2134,59 @@ send an HTTP DELETE request to
...
+  1729629081, the relevant batch bucket identifier is `(1729629000, 1000)`. (The
+  first element of the pair denotes the start of the batch interval and the
+  second denotes the duration.)
+* For the leader-selected batch mode, the batch bucket identifier is a batch ID.
+  The identifier associated with a given output share is the batch ID of the
+  related aggregation job.
+
+A few different pieces of information are associated with each batch bucket:
+
+* An VDAF-specific aggregate share, as defined in {{!VDAF}}.
```
divergentdave, line 2160: nit: I figure this used to say "an aggregate share ..."
```suggestion
* A VDAF-specific aggregate share, as defined in {{!VDAF}}.
```

divergentdave (approved):

PR #624, opened 2024-10-22
Address issue 606 - add task start time

### LINKED ISSUE #606: Consider adding task start time
Initially brought up by @cjpatton in https://github.com/wangshan/draft-wang-ppm-dap-taskprov/issues/62.

It may make sense to introduce the concept of it in the base DAP draft as well. We can use this parameter to enforce things like:
- Task lifetime is not too long.
- Report timestamp should not be before the task start time.

junyechen1996:
Closes #606 .

### DIFF
--- a/draft-ietf-ppm-dap.md
+++ b/draft-ietf-ppm-dap.md
@@ -1009,9 +1009,14 @@ the following parameters associated with it:
   are grouped into batches and the properties that all batches for this task
   must have. The party MUST NOT configure the task if it does not recognize the
   batch mode.
-* `task_expiration`: The time up to which Clients are expected to upload to this
-  task. The task is considered completed after this time. Aggregators MAY reject
-  reports that have timestamps later than `task_expiration`.
+* `task_start`: The time from which the Clients will start uploading reports
+  to a task. The Client is not expected to upload its report if the Client's
+  clock is earlier than `task_start`. Aggregators MAY reject reports with
+  timestamps earlier than `task_start`.
+* `task_duration`: The duration of a task. The task is considered completed
+  after the end time `task_start + task_duration`. Aggregators MAY reject
+  reports that have timestamps later than the end time, and MAY reject the task
+  if `task_duration` is too long.
 * A unique identifier for the VDAF in use for the task, e.g., one of the VDAFs
   defined in {{Section 10 of !VDAF}}.
 
@@ -1258,10 +1263,11 @@ violation. Note that this is also enforced by the Helper during the aggregation
 interaction. The Leader MAY also abort the upload interaction and alert the
 Client with error `reportRejected`.
 
-The Leader MAY ignore any report whose timestamp is past the task's
-`task_expiration`. When it does so, it SHOULD also abort the upload interaction
-and alert the Client with error `reportRejected`. Client MAY choose to opt out
-of the task if its own clock has passed `task_expiration`.
+The Leader MAY ignore any report whose timestamp is before the task's
+`task_start`, or is past the end time `task_start + task_duration`. When it does
+so, it SHOULD also abort the upload interaction and alert the Client with error
+`reportRejected`. Client MAY choose to opt out of the task if its own clock is
+earlier than `task_start`, or is past `task_start + task_duration`.
 
 The Leader may need to buffer reports while waiting to aggregate them (e.g.,
 while waiting for an aggregation parameter from the Collector; see
@@ -1673,6 +1679,7 @@ enum {
   task_expired(7),
   invalid_message(8),
   report_too_early(9),
+  task_not_started(10),
   (255)
 } ReportError;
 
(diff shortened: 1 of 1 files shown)


junyechen1996:
cc @cjpatton. I cannot add reviewers.

draft-ietf-ppm-dap.md
```diff
@@ -1006,9 +1006,14 @@ the following parameters associated with it:
   are grouped into batches and the properties that all batches for this task
   must have. The party MUST NOT configure the task if it does not recognize the
   batch mode.
+* `task_start`: The start timestamp of a task. Clients MAY opt out of the task
```
cjpatton, line 1009: "opt out" is a taskprov term that we probably shouldn't overload. I'd suggest "don't upload reports" instead of "opt out". Likewise below.

draft-ietf-ppm-dap.md
```diff
@@ -1670,6 +1676,7 @@ enum {
   task_expired(7),
   invalid_message(8),
   report_too_early(9),
+  task_not_started(10),
```
cjpatton, line 1679: My initial though was that we may be able to use `report_too_early` for this case, but looking at how that error currently used, I think I'm in favor of adding `task_not_started`. The reason is that rejecting a report rejected with `report_too_early` can be retried, but if the report's timestamp precedes the task start time, then there is no way to process it.

WDYT @branlwyd?

draft-ietf-ppm-dap.md
```diff
@@ -1867,6 +1874,10 @@ following checks:
    skew. If a report is rejected for this reason, the Aggregator SHOULD mark the
    input share as invalid with the error `report_too_early`.
 
+1. Check if the report's timestamp is before the task's `task_start` time. If
```
cjpatton, line 1877: I think this makes it slightly more clear that you don't mean "smaller than or equal to".
```suggestion
1. Check if the report's timestamp is smaller than the task's `task_start` time. If
```

draft-ietf-ppm-dap.md
branlwyd, line 1009: semantic nit: `task_start` and `task_expiration` are effectively the start and end of the time interval in which the task is accepting report uploads, but "start" and "expiration" are not opposites.

Maybe `task_start_time` and `task_end_time`? (or `task_start` & `task_end`?)

draft-ietf-ppm-dap.md
branlwyd: I agree; as you note, there are semantics around `report_too_early` that are not appropriate to apply to the cases we will be using `task_not_started`.

draft-ietf-ppm-dap.md
cjpatton: Works for me.

junyechen1996: I added `MAY not upload its report`. Let me know if that's the right use of MAY..

junyechen1996: Fixed. Made a pass to change `task_expiration` to `task_end`, and `expired` to `ended`.

draft-ietf-ppm-dap.md
junyechen1996: I changed it to `earlier than`.

draft-ietf-ppm-dap.md
```diff
@@ -1006,9 +1006,14 @@ the following parameters associated with it:
   are grouped into batches and the properties that all batches for this task
   must have. The party MUST NOT configure the task if it does not recognize the
   batch mode.
-* `task_expiration`: The time up to which Clients are expected to upload to this
+* `task_start`: The time from which the Clients will start uploading reports
+  to a task. The Client MAY not upload its report if the Client's clock is
```
cjpatton, line 1010: Let's try and keep the normative language the same.
```suggestion
  to a task. The Client is not expected to upload its report if the Client's clock is
```

branlwyd (approved):

junyechen1996:
Squashed.

draft-ietf-ppm-dap.md
```diff
@@ -1667,9 +1673,10 @@ enum {
   hpke_unknown_config_id(4),
   hpke_decrypt_error(5),
   vdaf_prep_error(6),
-  task_expired(7),
+  task_ended(7),
```
cjpatton, line 1676: nit: I think we should just keep this as `task_expired`. Its meaning is clear enough, and there would be one less thing that implementations need to change.

cjpatton (approved):

cjpatton:
2024/10/24 interim: @junyechen1996 instead of `Time task_start` and `Time task_end`, we will have `Time task_start` and `Duration task_duration`. That way we avoid encoding invalid task validity ranges.

junyechen1996:
> 2024/10/24 interim: @junyechen1996 instead of `Time task_start` and `Time task_end`, we will have `Time task_start` and `Duration task_duration`. That way we avoid encoding invalid task validity ranges.

Updated the PR to reflect this^.

cjpatton (approved):

PR #627, opened 2024-10-24
Improve extension points for new batch modes (*)

### LINKED ISSUE #622: Guidance for defining new batch modes
The contents of the `Query` and `PartialBatchSelector` structures depend on the task's batch mode. Future documents are allowed to add new batch modes, and they'll have to extend these structures some how. This document needs to provide some guidance on how exactly to do this.

The taskprov draft has a related problem, which we try to solve with a section like this (https://ietf-wg-ppm.github.io/draft-ietf-ppm-dap-taskprov/draft-ietf-ppm-dap-taskprov.html#section-8):

> # Extending this Document
> 
> The behavior of the `taskbind` extension may be extended by future documents
> that define:
> 
> 1. A new DAP batch mode
> 1. A new VDAF
> 1. A new DP mechanisms
> 
> Such documents SHOULD include a section titled "Taskbind Considerations" that
> specifies an encoding of any configuration parameters associated with the DAP
> batch mode, VDAF, or DP mechanism. This encoding extends the `BatchConfig`
> structure, the `VdafConfig` structure, or the `DpConfig` structure
> respectively.

This may or may not be the best solution.

cjpatton:
Closes #622.

Improve extension points for new batch modes (*)

Future documents may define new batch modes. We don't make this easy, as
the semantics of batch modes are currently tightly integrated into the
document. We would instead like all of the language pertaining to
*specific* to be self-contained so that it's easy to specify a batch
mode that is sound and complete.

Each batch mode specifies the following things:

* The contents of the `Query`, `BatchSelector`, and
  `PartialBatchSelector`, which are effectively extension points for
  hooking batch modes into aggregation.

* Batch buckets, including how they're identified, how reports are
  mapped to batch buckets, and the buckets specified by a query.

This commit moves the specifics of leader-selected and time-interval to
a new section. It also adds a new section with guidelines for extending
the DAP document, including by adding new batch modes.

Finally, to help reduce implementation complexity, add a length prefix
to the extension points so that collection and aggregation requests can
be decoded even if the indicated batch mode is not recognized. This is
not strictly necessary, as the current text requires abort in this case;
but it does help separate parsing logic (how do I read this message)
from protocol logic (what batch modes do I support).

Other changes:

* Simplify the "batchOverlap" condition: ""the batch buckets of the
  query MUST NOT overlap with the previously aggregated buckets.
  Accordingly remove the following redundant language:

  > For time-interval tasks, it is sufficient (but not necessary) to
  > check that the batch interval does not overlap with the batch
  > interval of any previous query. If this batch interval check fails,
  > then the Aggregator MAY abort with error of type "batchOverlap".

### DIFF
--- a/draft-ietf-ppm-dap.md
+++ b/draft-ietf-ppm-dap.md
@@ -889,116 +889,42 @@ VDAF types:
 
 ## Batch Modes, Batches, and Queries {#batch-mode}
 
-Aggregated results are computed based on sets of reports, called "batches". The
-Collector requests the aggregated results for a given batch via a "query". The
-Aggregators use this query to carry out the aggregation flow and produce
-aggregate shares encrypted to the Collector.
+An aggregate result is computed from a set of reports, called a "batch". The
+Collector requests the aggregate result by making a "query"; the Aggregators
+use this query to select a batch for aggregation.
 
 Each measurement task has a preconfigured "batch mode". The batch mode defines
 both how reports may be partitioned into batches, as well as how these batches
 are addressed and the semantics of the query used for collection.
 
-This document defines the time-interval and leader-selected batch modes.
+This document defines two batch modes:
+
+* time-interval ({{time-interval-batch-mode}}) in which the query specifies a
+  time interval, and the batch consists of all reports with a timestamp in that
+  interval.
+
+* leader-selected ({{leader-selected-batch-mode}}) in which the Leader assigns
+  reports to batches itself, and the Collector's query simply requests the
+  aggregate result for the "next" batch of reports.
+
+Future documents may define additional batch modes for DAP; see
+{{extending-this-doc}}. Implementations are free to implement only a subset of
+the available batch modes.
+
 Batch modes are identified with a single byte in serialized messages, as
 follows:
 
 ~~~ tls-presentation
 enum {
-  reserved(0), /* Reserved for testing purposes */
+  reserved(0),
   (255)
 } BatchMode;
 ~~~
 
-The time-interval batch mode is described in {{time-interval-batch-mode}}; the
-leader-selected batch mode is described in {{leader-selected-batch-mode}}.
-Future specifications may introduce new batch modes as needed (see
-{{batch-mode-reg}}). Implementations are free to implement only a subset of the
-available batch modes.
-
-A query includes parameters used by the Aggregators to select a batch of
-reports specific to the given batch mode. A query is defined as follows:
-
-~~~ tls-presentation
-struct {
-  BatchMode batch_mode;
-  select (Query.batch_mode) {
-    case time_interval: Interval batch_interval;
-    case leader_selected: Empty;
-  }
-} Query;
-~~~
-
-The query is issued in-band as part of the collect interaction
-({{collect-flow}}). Its content is determined by the batch mode, which in turn
-is determined by the measurement task, which is configured out-of-band. All
-batch modes have the following configuration parameters in common:
-
-- `min_batch_size` - The smallest number of reports the batch is allowed to
-  include. In a sense, this parameter controls the degree of privacy that will
-  be obtained: the larger the minimum batch size, the higher degree of privacy.
-  However, this ultimately depends on the application and the nature of the
-  measurements and aggregation function.
-
-- `time_precision` - Clients use this value to truncate their report timestamps;
-  see {{upload-flow}}. Additional semantics may apply, depending on the batch
-  mode. (See {{batch-validation}} for details.)
-
-The parameters pertaining to specific batch modes are described in
-{{time-interval-batch-mode}} and {{leader-selected-batch-mode}}.
-
-### Time-interval Batch Mode {#time-interval-batch-mode}
-
-~~~ tls-presentation
-enum {
-  time_interval(1),
-  (255)
-} BatchMode;
-~~~
-
-The time-interval batch mode is designed to support applications in which
-reports are collected into batches grouped by an interval of time. The Collector
-specifies a "batch interval" that determines the time range for reports included
-in the batch. For each report in the batch, the time at which that report was
-generated (see {{upload-flow}}) MUST fall within the batch interval specified by
-the Collector.
-
-Typically the Collector issues queries for which the batch intervals are
-continuous, monotonically increasing, and have the same duration. For example,
-the sequence of batch intervals `(1659544000, 1000)`, `(1659545000, 1000)`,
-`(1659546000, 1000)`, `(1659547000, 1000)` satisfies these conditions. (The
-first element of the pair denotes the start of the batch interval and the second
-denotes the duration.) However, this is not a requirement--the Collector may
-decide to issue queries out-of-order. In addition, the Collector may need to
-vary the duration to adjust to changing report upload rates.
-
-### Leader-selected Batch Mode {#leader-selected-batch-mode}
-
-~~~ tls-presentation
-enum {
-  leader_selected(2),
-  (255)
-} BatchMode;
-~~~
-
-The leader-selected batch mode is used to support applications in which it is
-acceptable for reports to be batched in an arbitrary fashion by the Leader. Each
-batch of reports is identified by an opaque "batch ID". Both the reports
-included in each batch and the ID for each batch are allocated in an arbitrary
-fashion by the Leader.
-
-The Collector will not know the set of batch IDs available for collection. To
-get the aggregate of a batch, the Collector issues a query, which does not
-include any information specifying a particular batch (see {{batch-mode}}). The
-Leader selects a recent batch to aggregate. The Leader MUST select a batch that
-has not yet been associated with a collection job.
-
-The Aggregators can output batches of any size that is larger than or equal to
-`min_batch_size`. The target batch size, if any, is implementation-specific, and
-may be equal to or greater than the minimum batch size. Deciding how soon
-batches should be output is also implementation-specific. Exactly sizing batches
-may be challenging for Leader deployments in which multiple, independent nodes
-running the aggregate interaction (see {{aggregate-flow}}) need to be
-coordinated.
+The query is issued to the Leader by the Collector during the collection
+interaction ({{collect-flow}}). In addition, information used to guide batch
+selection is conveyed from the Leader to the Helper when initializing
+aggregation ({{aggregate-flow}}) and finalizing the aggregate shares.
 
 ## Task Configuration {#task-configuration}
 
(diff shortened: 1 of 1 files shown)


tgeoghegan (approved):

draft-ietf-ppm-dap.md
```diff
@@ -1544,7 +1544,7 @@ struct {
 
 struct {
   opaque agg_param<0..2^32-1>;
-  PartialBatchSelector part_batch_selector;
+  opaque part_batch_selector<0..2^16-1>;
```
branlwyd, line 1547: The justification for making this change is to allow unrecognized variants to be decoded, and to separate serialization logic from protocol logic.

However, I think this change doesn't meet those goals: before the change, we would fail to deserialize `AggregationJobInitReq` if there was an unknown batch mode variant; with the change, we will fail to deserialize `PartialBatchSelector` if there was an unknown batch mode variant.

I think the right way to address this would be to make the batch-mode-specific part of the message (i.e. the `select`) opaque instead, e.g:

```
struct {
  BatchMode batch_mode;
  opaque encoded_partial_batch_selector<0..2^16-1>;
} PartialBatchSelector;
```

With this change, we can decode the entire `PartialBatchSelector` message (or any message that contains it), then do application-level checks that the batch mode variant is known before we attempt to decode the batch-mode-specific data.

===

But stepping back a bit, given that the correct behavior when parsing an unknown batch mode is "fail immediately", I think introducing a new length field overcomplicates things. An implementation can arrange for its deserialization code to return a specific error code when encountering an unknown enum variant, which we could then use to drive the corresponding DAP behavior. Since the enum always comes before the values whose type is determined by the enum value, this will always work. Perhaps this does not separate serialization logic quite as cleanly from application logic, but I think it's worth it in the sense that we aren't spending an extra few bytes per message on the wire.

cjpatton: > ```
> struct {
>   BatchMode batch_mode;
>   opaque encoded_partial_batch_selector<0..2^16-1>;
> } PartialBatchSelector;
> ```
> 
> With this change, we can decode the entire `PartialBatchSelector` message (or any message that contains it), then do application-level checks that the batch mode variant is known before we attempt to decode the batch-mode-specific data.

Actually, I think this accomplishes the same thing. It's just a matter of where we put the length prefix. If we put the length prefix before the batch mode, we know we have one more byte to parse for the batch mode, then the rest is specific to the batch mode; if we put the length before the batch mode, then we know the rest is specific to the batch mode.

The reason I spelled it this way is to be consistent with the extension points in taskprov for VDAF variants. We can take your approach if you strongly prefer it, but I think it might complicate the "Extending this Document" text slightly.

> But stepping back a bit, given that the correct behavior when parsing an unknown batch mode is "fail immediately", I think introducing a new length field overcomplicates things. An implementation can arrange for its deserialization code to return a specific error code when encountering an unknown enum variant, which we could then use to drive the corresponding DAP behavior. Since the enum always comes before the values whose type is determined by the enum value, this will always work. Perhaps this does not separate serialization logic quite as cleanly from application logic, but I think it's worth it in the sense that we aren't spending an extra few bytes per message on the wire.

I agree it's not strictly necessary. The thing I'm trying to accomplish is allowing implementations to separate parsing logic from protocol logic. With this PR, wether I recognize a batch mode is irrelevant to whether I can parse the AggregationJobInitReq.

cjpatton: I was thinking of this over the weekend, and it occurred to me that we can spell this much nicer if we put the length tag first as you suggest. Will update the PR today or tomorrow.

draft-ietf-ppm-dap.md
```diff
@@ -888,116 +888,42 @@ VDAF types:
...
-This document defines the time-interval and leader-selected batch modes.
-Batch modes are identified with a single byte in serialized messages, as
-follows:
-
-~~~ tls-presentation
-enum {
-  reserved(0), /* Reserved for testing purposes */
-  (255)
-} BatchMode;
-~~~
```
cjpatton, line 909: This is still defined below.

draft-ietf-ppm-dap.md
```diff
@@ -888,116 +888,42 @@ VDAF types:
...
-The query is issued in-band as part of the collect interaction
-({{collect-flow}}). Its content is determined by the batch mode, which in turn
-is determined by the measurement task, which is configured out-of-band. All
-batch modes have the following configuration parameters in common:
-
-- `min_batch_size` - The smallest number of reports the batch is allowed to
-  include. In a sense, this parameter controls the degree of privacy that will
-  be obtained: the larger the minimum batch size, the higher degree of privacy.
-  However, this ultimately depends on the application and the nature of the
-  measurements and aggregation function.
```
cjpatton, line 939: Moved to task parameters below. Likewise for `time_precision`.

draft-ietf-ppm-dap.md
```diff
@@ -888,116 +888,42 @@ VDAF types:
...
+  interval.
 
-### Time-interval Batch Mode {#time-interval-batch-mode}
-
-~~~ tls-presentation
-enum {
-  time_interval(1),
-  (255)
-} BatchMode;
-~~~
```
cjpatton, line 955: Moved to a section towards the end of the doc. Likewise for leader-selected.

draft-ietf-ppm-dap.md
```diff
@@ -2153,20 +2088,13 @@ When aggregation recovers an output share, it must be stored into an appropriate
...
-  duration is equal to the task's `time_precision`. The identifier associated
-  with a given output share is the unique such interval containing the client
-  timestamp of the report the output share was recovered from; for example, if
-  the task's `time_precision` is 1000 seconds and the report's timestamp is
-  1729629081, the relevant batch bucket identifier is `(1729629000, 1000)`. (The
-  first element of the pair denotes the start of the batch interval and the
-  second denotes the duration.)
-* For the leader-selected batch mode, the batch bucket identifier is a batch ID.
-  The identifier associated with a given output share is the batch ID of the
-  related aggregation job.
```
cjpatton, line 2169: Moved to the new sections below; we'll just leave a high level summary below. (Note that this is a point where the batch mode needs to define semantics.)

draft-ietf-ppm-dap.md
```diff
@@ -888,116 +888,42 @@ VDAF types:
...
-
-~~~ tls-presentation
-struct {
-  BatchMode batch_mode;
-  select (Query.batch_mode) {
-    case time_interval: Interval batch_interval;
-    case leader_selected: Empty;
-  }
-} Query;
-~~~
```
cjpatton, line 928: Moved to where it's first used, similar to BatchSelector and PartialBatchSelector.

draft-ietf-ppm-dap.md
```diff
@@ -2632,52 +2574,169 @@ distinct aggregation parameters. If the batch has been queried with more than
 one distinct aggregation parameter, the Aggregator MUST abort with error of type
 "batchQueriedMultipleTimes".
 
-Finally, the Aggregator checks that the batch does not contain a report that was
-included in any previous batch. If this batch overlap check fails, then the
-Aggregator MUST abort with error of type "batchOverlap". For time-interval
-tasks, it is sufficient (but not necessary) to check that the batch interval
-does not overlap with the batch interval of any previous query. If this batch
-interval check fails, then the Aggregator MAY abort with error of type
-"batchOverlap".
```
cjpatton, line 2641: Implied by new langauge.

draft-ietf-ppm-dap.md
cjpatton: Done, please take another look at the commit message before going through the whole PR.

cjpatton:
@branlwyd please take another look.

draft-ietf-ppm-dap.md
```diff
@@ -2632,52 +2573,171 @@ distinct aggregation parameters. If the batch has been queried with more than
...
+Next, the Aggregator checks if the set of batch bucket identified by by the
```
branlwyd, line 2576:
```suggestion
Next, the Aggregator checks if the set of batch buckets identified by by the
```

(ooh, this is a very nice way to phrase this requirement now that we have the concept of "batch buckets" explicitly specified)

draft-ietf-ppm-dap.md
```diff
@@ -2632,52 +2573,171 @@ distinct aggregation parameters. If the batch has been queried with more than
...
+first element of the pair denotes the start of the batch interval and the second
+denotes the duration.) However, this is not a requirement--the Collector may
+decide to issue queries out-of-order. In addition, the Collector may need to
+vary the duration to adjust to changing report upload rates.
+
+### Query Configuration
+
+The payload of `Query.config` is
+
+struct {
```
branlwyd, line 2626:
```suggestion
~~~ tls-presentation
struct {
```

draft-ietf-ppm-dap.md
```diff
@@ -2632,52 +2573,171 @@ distinct aggregation parameters. If the batch has been queried with more than
...
+
+where `batch_interval` is the batch interval requested by the Collector.
+
+### Partial Batch Selector Configuration
+
+The payload of `PartialBatchSelector.config` is empty.
+
+### Batch Selector Configuration
+
+The payload of `Query.config` is
```
branlwyd, line 2639:
```suggestion
The payload of `BatchSelector.config` is
```

draft-ietf-ppm-dap.md
```diff
@@ -2632,52 +2573,171 @@ distinct aggregation parameters. If the batch has been queried with more than
...
   effectively, the minimum batch duration)
 
 * both `batch_interval.start` and `batch_interval.duration` are divisible by
   `time_precision`
 
-These measures ensure that Aggregators can efficiently "pre-aggregate" output
-shares recovered during the aggregation interaction.
+A batch consists of a sequence of contiguous batch buckets. That is, the set of
+batch buckets identifiers for the batch interval is
+`(batch_interval.start, batch_interval.start + batch_interval.duration)`,
```
branlwyd, line 2672: In this example, I think we want to replace `batch_interval.duration` with `time_precision` throughout. Also, I think the duration for all of the batch bucket identifiers should be `time_precision`.

draft-ietf-ppm-dap.md
```diff
@@ -3922,4 +3982,34 @@ Index value:  No transformation needed.
...
+1. a new batch mode ({{batch-mode}})
+1. a new report extension ({{report-extensions}})
+1. a new report error ({{aggregation-helper-init}})
+1. a new entry in the URN sub-namespace for DAP ({{urn-space-errors}})
+
+Each of these requires registration of a codepoint or other value; see
+{{iana}}. No other considerations are required except in the following cases:
+
+* When a document defines a new batch mode, it MUST include a section titled
+  "DAP Batch Mode Considerations" specifying the following:
```
branlwyd, line 3999:
```suggestion
* When a document defines a new batch mode, it MUST include a section specifying the following:
```

This is opinionated, but I don't think we should be so prescriptive as to include the name of the section in the MUST. :)

draft-ietf-ppm-dap.md
```diff
@@ -3922,4 +3982,34 @@ Index value:  No transformation needed.
...
+
+    * The value of the `config` field of `Query`, `PartialBatchSelector`, and
+      `BatchSelector`
+
+    * Batch buckets ({{batch-buckets}}): how reports are assigned to batch
+      buckets; how each bucket is identified; and how batch buckets are mapped to
+      batches and how batch buckets
```
branlwyd, line 4006: This sentence ends prematurely.

branlwyd (approved):
Looks good! Comments are almost entirely typo/grammar changes.

cjpatton (commented):
Rebased, squashed, and self review

PR #628, opened 2024-10-25
Require rejection of reports outside of the task's validity window

### LINKED ISSUE #625: Rejecting reports outside the task's validity window
[Rejecting reports with timestamps larger than the task expiration date is optional](https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/blob/13c7e905c12fa2b2460f2a18f8b6a39ab53472ac/draft-ietf-ppm-dap.md?plain=1#L1870-L1872):

> Check if the report's timestamp has passed the task's `task_expiration` time.
> If so, the Aggregator MAY mark the input share as invalid with the error
> `task_expired`.

Likewise for #624, which adds a task start time.

Should this MAY be a MUST? It's easy enough for each Aggregator to enforce, since it just has to compare the timestamp reported by the client to the expiration date associated with the task. (The Aggregator doesn't need to check what time it is, for example.)

Rejecting reports that fall outside the task validity window seems like a good idea because it reduces the amount of time the VDAF verification key is in use.

I did some `git blame` splunking and traced this to the following PR from @wangshan: https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/pull/304. We didn't discuss this much, and based on the conversation on the linked issue, it sounds like it wasn't really our highest priority. Our top line concern was making sure the Aggregators know when they can clean up their storage.

cjpatton:
Closes #625.

If a report's timestamp is before `tast_start` or after `task_start + task_duration`, then Aggregators MUST reject it.

### DIFF
--- a/draft-ietf-ppm-dap.md
+++ b/draft-ietf-ppm-dap.md
@@ -1009,14 +1009,13 @@ the following parameters associated with it:
   are grouped into batches and the properties that all batches for this task
   must have. The party MUST NOT configure the task if it does not recognize the
   batch mode.
-* `task_start`: The time from which the Clients will start uploading reports
-  to a task. The Client is not expected to upload its report if the Client's
-  clock is earlier than `task_start`. Aggregators MAY reject reports with
-  timestamps earlier than `task_start`.
+* `task_start`: The time from which the Clients will start uploading reports to
+  a task. Aggregators MUST reject reports with timestamps earlier than
+  `task_start`.
 * `task_duration`: The duration of a task. The task is considered completed
-  after the end time `task_start + task_duration`. Aggregators MAY reject
-  reports that have timestamps later than the end time, and MAY reject the task
-  if `task_duration` is too long.
+  after the end time `task_start + task_duration`. Aggregators MUST reject
+  reports that have timestamps later than the end time, and MAY choose to opt
+  out of the task if `task_duration` is too long.
 * A unique identifier for the VDAF in use for the task, e.g., one of the VDAFs
   defined in {{Section 10 of !VDAF}}.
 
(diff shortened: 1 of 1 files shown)


cjpatton:
@branlwyd let's merge by 2024/10/31 unless we get push back.

tgeoghegan (approved):

draft-ietf-ppm-dap.md
```diff
@@ -1263,11 +1262,11 @@ violation. Note that this is also enforced by the Helper during the aggregation
 interaction. The Leader MAY also abort the upload interaction and alert the
 Client with error `reportRejected`.
 
-The Leader MAY ignore any report whose timestamp is before the task's
+The Leader MUST ignore any report whose timestamp is before the task's
 `task_start`, or is past the end time `task_start + task_duration`. When it does
 so, it SHOULD also abort the upload interaction and alert the Client with error
-`reportRejected`. Client MAY choose to opt out of the task if its own clock is
-earlier than `task_start`, or is past `task_start + task_duration`.
+`reportRejected`. Client MUST NOT upload a report if its own clock is earlier
```
branlwyd, line 1268:
```suggestion
`reportRejected`. Clients MUST NOT upload a report if its own clock is earlier
```

(typographical)

branlwyd, line 1268: We say that client's shouldn't upload a report if their own clock is outside of the appropriate interval, but the timestamp in a report doesn't necessarily correspond to the client's clock.

Maybe something like this?

> Clients MUST NOT upload a report if the timestamp included in the report would be earlier than `task_start` or past `task_start + task_duration`.

branlwyd (approved):

draft-ietf-ppm-dap.md

Issue #619, opened 2024-10-18
Probing for extension support


martinthomson:
Extensions are mandatory to understand, which is a good model for something like this.
A server is required to send invalid_message if it doesn't support an extension.

If an extension (or set thereof) is ever defined that is optional, this model is not particularly friendly to probing for support.
It would be easier for client to recognize the error and try again without the optional extension
if there was an unsupported_extension message.

That sort of thing is potentially unwise if the response could be spoofed, because it would allow a leader to cause a report to be included more than once in an aggregate, but that risk seems manageable to me.

cjpatton:
Historical note: We made the call to make extensions mandatory to implement some time shortly after draft 02. In fact, your feedback was the main decision point.

Two more notable features of report extensions:
1. They are associated with each report share, meaning each aggregator may get a different set of extensions. E.g., in some deployment you might have an extension that is only meaningful to the leader and not to a generic helper.
2. The extensions are encrypted to each aggregator: the leader can't read the helper's extensions and vice versa.

With that in mind, during the upload interaction, the Leader could decrypt, inspect the extensions, and abort if one of them is unrecognized. [Currently the spec requires a pretty opaque error in this case](https://ietf-wg-ppm.github.io/draft-ietf-ppm-dap/draft-ietf-ppm-dap.html#section-4.5.2-24), but we could certainly replace this with something with more information.

That said, there is currently no way to do the same for the Helper's extensions, since the Leader can't decrypt them. Do you have a use case for this?

cjpatton:
2024/10/24 interim: We'll resolve this by adding an abort type that the Leader can use during the upload interaction to let the Client know that some extensions aren't supported. The PR will specify an optional mechanism for conveying programmatically which extensions aren't recognized. We should also add a note in the text that there is no mechanism for the Client to probe the Helper.

PR #630, opened 2024-10-29
Allow Leader to indicate unknown extensions during upload.

branlwyd:
This is done with a new abort type, unsupported_extension; and by specifying a way for the Leader to indicate the unsupported extension code points within the resulting problem document.

### DIFF
--- a/draft-ietf-ppm-dap.md
+++ b/draft-ietf-ppm-dap.md
@@ -791,6 +791,7 @@ standard tokens for use in the "type" field:
 | unauthorizedRequest        | Authentication of an HTTP request failed (see {{request-authentication}}). |
 | stepMismatch               | The Aggregators disagree on the current step of the DAP aggregation protocol. |
 | batchOverlap               | A request's query includes reports that were previously collected in a different batch. |
+| unsupportedExtension       | An upload request's extensions list includes an unknown extension. |
 {: #urn-space-errors = "DAP errors. All are scoped to the errors sub-namespace of the DAP URN, e.g., urn:ietf:params:ppm:dap:error:invalidMessage."}
 
 This list is not exhaustive. The server MAY return errors set to a URI other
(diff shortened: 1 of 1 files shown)


branlwyd:
Closes #619.

draft-ietf-ppm-dap.md
```diff
@@ -1288,10 +1289,21 @@ report for this reason, it SHOULD abort the upload interaction and alert the
...
+If the Leader's input share contains an unrecognized report extension, then the
+Leader MAY abort the upload request with error "unsupportedExtension." If the
+Leader does abort for this reason, it MAY indicate the unsupported extensions
+in the resulting problem document via an extension field
+"unsupported_extensions" on the problem document; this field MUST contain an
+array of numeric values indicating the extension code points which were not
+recognized.
+
+If the Leader's input share contains two extensions with the same type, then
+the Leader MAY abort th eupload request with error "invalidMessage".
```
cjpatton, line 1301:
```suggestion
the Leader MAY abort the upload request with error "invalidMessage".
```

draft-ietf-ppm-dap.md
```diff
@@ -1288,10 +1289,21 @@ report for this reason, it SHOULD abort the upload interaction and alert the
...
-If the Leader's input share contains an unrecognized report extension, or if
-two report extensions have the same ExtensionType, then the Leader MAY abort
-the upload request with error "invalidMessage". Note that this behavior is not
-mandatory because it requires the Leader to decrypt its input share.
+If the Leader's input share contains an unrecognized report extension, then the
+Leader MAY abort the upload request with error "unsupportedExtension." If the
+Leader does abort for this reason, it MAY indicate the unsupported extensions
+in the resulting problem document via an extension field
+"unsupported_extensions" on the problem document; this field MUST contain an
+array of numeric values indicating the extension code points which were not
```
cjpatton, line 1297: I think we need to spell out "array of numeric values" in more detail. Is this just a vector of bytes, where each byte is a different error variant? Should this be base64url encoded?

branlwyd: A problem details document is a JSON object; with that context, I think this meaning is clear -- an example value might be `[23, 45, 52]`.

I added a reference to the relevant section of the problem details RFC 9457 to make this clear.

cjpatton: Great point. In that case, my only suggestion would be to add the example to the text.

cjpatton (approved):

branlwyd:
Rebased & squashed.

Issue #626, opened 2024-10-24
@martinthomson: The Helper MUST support skew recovery


cjpatton:
IIRC the language in the draft says something like: "The Leader MAY retry from the previous step; the Helper SHOULD help it recover". @martinthomson pointed out during the 2024/10/24 interim (in the Zulip) that the Leader's MAY implies a MUST for the Helper. I believe that's correct.

cjpatton:
Specifically, in reference to [slide 7](https://datatracker.ietf.org/meeting/interim-2024-ppm-02/materials/slides-interim-2024-ppm-02-sessa-dap-and-taskprov-open-issues-00), Martin writes:

> Any MAY in a protocol is just a MUST in disguise. The entity that exercises the MAY forces other participants to do something to allow that choice. So the "SHOULD" here really needs to be a "MUST" of some sort.

branlwyd:
I'm not sure I agree in this specific case: if the Leader attempts skew recovery, but the Helper has chosen not to implement it, the Helper would respond with a `stepMismatch` error.

So I think the state of things is: the Leader may optionally attempt to recover from skewed requests; the Helper may optionally implement skew recovery support. If both the Leader & the Helper implement their respective pieces, skew recovery is possible. Otherwise, it is not.

cjpatton:
It sounds like you're saying the MUST is not required because the code path is fully specified in case the Helper doesn't implement skew recovery

If so, I don't think that's MT's point. Rather, the code path is only truly optional if the Helper supports it. In other words, the following statement is invalid:

> The Leader MAY do X. If the Leader does X, then the Helper SHOULD do Y.

Rather:

> The Leader MAY do X. If the Leader does X, then the Helper MUST do Y.

martinthomson:
It might still be the case that the helper doesn't have to successfully handle the rollback and retry, but it can't fall over if the leader tries.

branlwyd:
I surveyed the existing text around replay protection.

The Leader Initialization & Continuation sections don't say anything -- retrying by re-sending the same request again is left implicit. I think this is OK, but I could also see us adding some text if folks think it would be valuable.

The Helper aggregate initialization section says:

[...]

PR #631, opened 2024-10-30
Tweak aggregation skew recovery text.

branlwyd:
* Note that if the Helper does not wish to attempt skew recovery, it MUST fail with a stepMismatch error.
* Add text to the Leader Initialization & Continuation sections suggesting that transient failures SHOULD be retried, and referencing the skew recovery section.


branlwyd:
Closes #626.

branlwyd:
+cc @martinthomson

cjpatton (approved):

Issue #620, opened 2024-10-18
Public report extensions


martinthomson:
I don't know how I missed this earlier, but I think that the design of report extensions is very much suboptimal for the sorts of things that are in [this draft](https://github.com/martinthomson/dap-dp-ext).  Those extensions carry public information, so having that data repeated is not ideal. 

[This draft](https://github.com/martinthomson/dap-bulk) makes a case for submissions with a reduced overhead.  The efficiency gains depend on having extensions to the public report metadata.  I'd like to request that report metadata be made extensible.

I can't conceive of any reason to use private report extensions, as currently defined.  I'm not saying that it is a bad idea to provide that extension point, though a lack of reasons to use them makes their inclusion questionable.  I guess I'm saying that moving upload extensions from private to public might be possible, rather than just adding another extension point.

cjpatton:
See https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/issues/619#issuecomment-2422794129 for historical context. Extensions used to be public, but we decided after draft 02 to make them private. In fact, it was your suggestion to encrypt them: see https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/issues/369#issuecomment-1289958837.

I'm open to doing something here, but given how far along we are, I think we need to be careful to not change the expectations of existing implementations. In particular, I'm hesitant to get rid of the private extensions altogether.

I'd be fine with adding public extensions to the report meta data. Ideally these would have precisely the same semantics as the existing, private extensions, that is, they are mandatory to implement, and there can only be one instance of each codepoint among the private and public extensions.

cjpatton:
2024/10/24 interim: We will put up a PR to add a public extension point. If we decide the protocol complexity is too great, then we'll consider removing the private extension points as well. (There are no publicly known use cases for this, but in the future an "encrypted extension" could be defined that makes this use case possible.)

martinthomson:
One thing that came up was that a public "encrypted" extension might end up being a little cumbersome (not just in definition, but also implementation).  If that is the case, and we've removed the private extensions, DAP version 2 is always possible.  A new version doesn't need to be a whole big deal.  QUIC demonstrated that versioning can be done fairly cheaply with the right motivation.

PR #632, opened 2024-10-31
Add public report extensions.

branlwyd:
The existing (private) report extension mechanism is renamed to "private extensions."

A few notes:
* Public extensions are always the same for both aggregators; the public extensions are included in the input share AAD to ensure that the Leader can't modify these while passing them along to the Helper.
* Public & private extensions share the same "namespace", i.e. they use the same set of extension type codepoints, and duplicating an extension in both the public & private extensions is considered invalid.

### DIFF
--- a/draft-ietf-ppm-dap.md
+++ b/draft-ietf-ppm-dap.md
@@ -1089,6 +1089,7 @@ Clients upload reports by using an HTTP POST to
 struct {
   ReportID report_id;
   Time time;
+  Extension public_extensions<0..2^16-1>;
 } ReportMetadata;
 
 struct {
@@ -1109,6 +1110,9 @@ struct {
       round this value down to the nearest multiple of the task's
       `time_precision` in order to ensure that that the timestamp cannot be used
       to link a report back to the Client that generated it.
+
+    * `public_extensions` is the list of public report extensions; see
+      {{report-extensions}}.
 
 * `public_share` is the public share output by the VDAF sharding algorithm. Note
   that the public share might be empty, depending on the VDAF.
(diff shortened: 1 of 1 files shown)


branlwyd:
Closes #620. +cc @martinthomson

draft-ietf-ppm-dap.md
```diff
@@ -1157,6 +1157,7 @@ struct {
 struct {
   ReportMetadata report_metadata;
   opaque public_share<0..2^32-1>;
+  Extension public_extensions<0..2^32-1>;
```
martinthomson, line 1160: This change would be smaller if you added the extensions to ReportMetadata.  Any reason why you chose not to?

martinthomson (approved):

draft-ietf-ppm-dap.md
martinthomson: I should have added.  4 gigabytes is an awful lot.  Do I have to start sharpening my varint axe?  There's a lot of overhead in this protocol that would be trivially cut with a single slice.

cjpatton: +1, let's move this to the report metadata. This will also help with code movement in daphne.

draft-ietf-ppm-dap.md
```diff
@@ -1245,6 +1250,7 @@ struct {
   TaskID task_id;
   ReportMetadata report_metadata;
   opaque public_share<0..2^32-1>;
+  Extension public_extensions<0..2^32-1>;
```
cjpatton, line 1253: Likewise here: if we include it in report metadata then it will also be covered by AAD.

draft-ietf-ppm-dap.md
```diff
@@ -1535,6 +1545,7 @@ message structured as follows:
 struct {
   ReportMetadata report_metadata;
   opaque public_share<0..2^32-1>;
+  Extension public_extensions<0..2^32-1>;
```
cjpatton, line 1548: LIkewise here

draft-ietf-ppm-dap.md
branlwyd: > I should have added. 4 gigabytes is an awful lot. Do I have to start sharpening my varint axe? There's a lot of overhead in this protocol that would be trivially cut with a single slice.

Using varints for length specifiers is not (AFAICT) supported by TLS syntax we currently use; we'd probably extend our syntax somehow to specify this. I'd need to do a little digging to see what varint formats are currently specified/standardized; a very quick glance does reveal anything. (Of course, if we can't lean on prior work we could certainly specify something ourselves.)

cjpatton: Oh, I should also add: I'm in favor of a 2-byte prefix, for consistency with the private extensions.

cjpatton (approved):

branlwyd:
Rebased & squashed.

PR #636, opened 2024-11-27
Small fixes to async aggregation/collection.

branlwyd:
Specifically:
 * Settle on 202 Accepted for the Helper's response code to aggregation continuation requests. Previously, the Leader Continuation section suggested 202 Accepted while the Helper Continuation section suggested 200 OK. Either code is fine by me, but we need to make the specification consistent.
 * Add maximum-value indicators to the aggregation & collection status messages. This is an editorial change: indicating the maximum value is optional per RFC 8446's presentation language, but doing so makes the specification clearer & is consistent with all other enumerated values defined in the specification.

### DIFF
--- a/draft-ietf-ppm-dap.md
+++ b/draft-ietf-ppm-dap.md
@@ -1795,6 +1795,7 @@ message is structured as follows:
 enum {
   processing(0),
   ready(1),
+  (255)
 } AggregationJobStatus;
 
 struct {
@@ -2125,8 +2126,8 @@ The Helper constructs an `AggregationJobResp` message (see
 {{aggregation-helper-init}}) with each prep step. The order of the prep steps
 MUST match the Leader's `AggregationJobContinueReq`.
 
-The Helper responds to the Leader with HTTP status 200 OK, a body consisting
-of the `AggregationJobResp`, and the media type
+The Helper responds to the Leader with HTTP status 202 Accepted, a body
+consisting of the `AggregationJobResp`, and the media type
 "application/dap-aggregation-job-resp".
 
 Depending on the task parameters, processing an aggregation job may take some
@@ -2321,6 +2322,7 @@ structured as follows:
 enum {
   processing(0),
   ready(1),
+  (255)
 } CollectionJobStatus;
 
 struct {


tgeoghegan (approved):

draft-ietf-ppm-dap.md
```diff
@@ -2125,8 +2126,8 @@ The Helper constructs an `AggregationJobResp` message (see
 {{aggregation-helper-init}}) with each prep step. The order of the prep steps
 MUST match the Leader's `AggregationJobContinueReq`.
 
-The Helper responds to the Leader with HTTP status 200 OK, a body consisting
-of the `AggregationJobResp`, and the media type
+The Helper responds to the Leader with HTTP status 202 Accepted, a body
+consisting of the `AggregationJobResp`, and the media type
```
cjpatton, line 2130: This looks like a bug fix. What do you want to do for interop? Should we cut a new draft so folks can implement this? Or are we OK with implementing 200 OK here?

branlwyd: This is indeed a bugfix (or, well, it resolves an ambiguity in the spec).

I think we should eventually cut a bugfix draft w/ this and any other small fixes we find during implementation. For DAP-13 interop testing we can handshake on the appropriate status; this PR switches to 202 everywhere, and that's what I'm currently implementing in Janus, but I don't have a strong opinion either way on the correct status to use -- I'd be happy enough with just about anything as long as we're consistent.

cjpatton: It would be nice to avoid cutting a draft if we can avoid it. If you think the intent is clear enough, then I think we should treat this as non-breaking. My suggestion would be to merge, update the change log, cut a draft, and inform the mailing list.

cjpatton: We could also wait to see if more changes are needed after we've all wrapped up our implementations. I don't think we've caught any bugs in daphne yet.

branlwyd: IMO, let's wait until we think our implementations are finished. Then we'll be in a better position to decide if we need to cut a new draft or just inform folks. In any case, I'll wait a few days to let anyone object & then merge this; heads-up that this specifies 202 Accepted as the correct response code for a Helper receiving an aggregation continuation request.

I don't think the intent is clear, however; the text for Helper aggregation continuation currently says that the Helper should respond with a 202 Accepted in one place, and a 200 OK in another. I think this is something that should be cleaned up before RFC, one way or another.

PR #637, opened 2025-01-13
Move draft-irtf-cfrg-vdaf-13 to 14

cjpatton:
Note that this is a non-breaking change.


tgeoghegan (approved):
LGTM. I filed https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/issues/643 to update this and another I-D reference to an RFC.

PR #640, opened 2025-01-14
Elaborate on example for `report_dropped` rejection case

cjpatton:
Stacked on #639.

The protocol requires Aggregators to prevent report replays within a task. It also allows the Aggregators to eventually clean up state for anti-replay. If a report arrives late, it may be rejected because the Aggregator can't determine if it has been replayed. Clarify this in the example.

### DIFF
--- a/draft-ietf-ppm-dap.md
+++ b/draft-ietf-ppm-dap.md
@@ -1912,10 +1912,10 @@ following checks:
       {{distributed-systems}} for more information.
 
 1. Finally, if an Aggregator cannot determine if an input share is valid, it
-   MUST mark the input share as invalid with error `report_dropped`. For
-   example, if the Aggregator has evicted the state required to perform the
-   check from long-term storage. (See {{reducing-storage-requirements}} for
-   details.)
+   MUST mark the input share as invalid with error `report_dropped`.
+   For example, the report timestamp may be so far in the past that the state
+   required to perform the check has been evicted from the Aggregator's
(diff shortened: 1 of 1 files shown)


branlwyd (approved):

draft-ietf-ppm-dap.md
```diff
@@ -1916,9 +1916,9 @@ following checks:
 
 1. Finally, if an Aggregator cannot determine if an input share is valid, it
    MUST mark the input share as invalid with error `report_dropped`. For
-   example, if the Aggregator has evicted the state required to perform the
-   check from long-term storage. (See {{reducing-storage-requirements}} for
-   details.)
+   example, the report timestamp may be so far in the past that the Aggregator
+   has evicted the state required to perform the check from long-term storage.
+   (See {{reducing-storage-requirements}} for details.)
```
tgeoghegan, line 1921: We can reword to make it less ambiguous what "from long-term storage" applies to. Also the last sentence doesn't need to be in a parenthetical; let your text roam wild and free!
```suggestion
   example, the report timestamp may be so far in the past that the state required to perform the check has been evicted from the Aggregator's long-term storage. See {{reducing-storage-requirements}} for details.
```

Sorry this isn't line wrapped properly.

tgeoghegan (changes requested):

tgeoghegan (approved):

cjpatton:
Unstacked.

PR #641, opened 2025-01-14
upload: Mention that expired tasks might be treated as unrecognized

cjpatton:
Stacked on #639.

Clients are free to pick any timestamp they wish, even one in the
validity range of a task that has expired. The Aggregators need to
prevent themselves from aggregating such reports indefinitely.

Add some text to the upload flow indicating that indicating an
unrecognized task may suggest the client is either misconfigured or
misbehaving.

### DIFF
--- a/draft-ietf-ppm-dap.md
+++ b/draft-ietf-ppm-dap.md
@@ -978,7 +978,7 @@ the following parameters associated with it:
   batch mode.
 * `task_start`: The time from which the Clients will start uploading reports to
   a task. Aggregators MUST reject reports with timestamps earlier than
-  `task_start`.
+  `task_start` as described in {{input-share-validation}}.
 * `task_duration`: The duration of a task. The task is considered completed
   after the end time `task_start + task_duration`. Aggregators MUST reject
   reports that have timestamps later than the end time, and MAY choose to opt
@@ -1255,7 +1255,8 @@ Clients SHOULD NOT upload the same measurement value in more than one report if
 the Leader responds with HTTP status code 201 Created.
 
 If the Leader does not recognize the task ID, then it MUST abort with error
-`unrecognizedTask`.
+`unrecognizedTask`. Note that indicating an unrecognized task might indicate
+that the Client is misconfigured or the task has expired.
 
 The Leader responds to requests whose Leader encrypted input share uses an
 out-of-date or unknown `HpkeConfig.id` value, indicated by


cjpatton:
Question for reviewers: if we take this change for upload and aggregation, should we also do the same for collection?

branlwyd (approved):
> The Aggregators need to prevent themselves from aggregating such reports indefinitely.

I think the Aggregators will generally evict "old" state (for some implementation-defined definition of "old"), which will effectively avoid the risk of needing to receive such reports indefinitely since reports touching on evicted state MUST be dropped with `report_dropped`. However, evicting old state is not required by DAP, so this change SGTM.

> Question for reviewers: if we take this change for upload and aggregation, should we also do the same for collection?

I don't see why not. We might want to add an additional note reminding implementors that the leeway is even more important in this case, since collection typically _must_ occur after the batch interval, so without appropriate leeway the final batch might be uncollectable.

tgeoghegan:
I'm fine with this in the upload interaction, but not in the aggregation or collection interactions. Task expiry is about limiting which client reports get aggregated into a task, but it seems reasonable to allow aggregation and collection to occur after aggregators stop accepting reports. If we forbid aggregation or collection as aggressively as this change suggests, then we run the risk of orphaning valid, aggregatable reports.

I would think that garbage collection accounts for this on its own: an aggregator will throw away reports if the task they belong to is expired for too long, at which point the aggregation or collection fails as described in {#input-share-validation}. I'd support adding an implementor's note discussing this, or something in operational considerations.

draft-ietf-ppm-dap.md
```diff
@@ -1223,7 +1224,10 @@ Clients SHOULD NOT upload the same measurement value in more than one report if
 the Leader responds with HTTP status code 201 Created.
 
 If the Leader does not recognize the task ID, then it MUST abort with error
-`unrecognizedTask`.
+`unrecognizedTask`. If the task has expired, i.e., `task_start + task_duration`
+is greater than the current time, then it SHOULD abort with error
```
tgeoghegan, line 1228: Aside from my other comment, I don't see why this is a SHOULD whereas aborting with `unrecognizedTask` is a MUST for other cases of unrecognized task ID.

cjpatton:
2025/1/15 call: don't abort during aggregation or collection: (1) in practice there is backlog to process for expired tasks; (2) we already allow dropping reports when the releavant replay state is evicted.

Let's keep this for upload, so the Leader has a tool to deal with malicious Clients, but disallow for aggreagtion/collection.

draft-ietf-ppm-dap.md
```diff
@@ -1223,7 +1224,10 @@ Clients SHOULD NOT upload the same measurement value in more than one report if
...
+`unrecognizedTask`. It may choose to provide leeway for Clients with skewed
+clocks.
```
cjpatton, line 1230: 2025/1/15 call: Remove SHOULD, add task expiry as a case in which this happens.

tgeoghegan (approved):

PR #642, opened 2025-01-15
Clarify language around public / private extension conflicts

jhoyla:
Previous language could be interpreted to mean that public and private extensions are allowed to conflict.

### DIFF
--- a/draft-ietf-ppm-dap.md
+++ b/draft-ietf-ppm-dap.md
@@ -1894,9 +1894,9 @@ following checks:
    invalid with error `invalid_message`.
 
 1. Check if any two public or private report extensions have the same extension
-   type. (For the purposes of this check, a private report extension can
-   conflict with a public report extension.) If so, the Aggregator MUST mark the
-   input share as invalid with error `invalid_message`.
+   type. (For the purposes of this check, a public extension MUST NOT conflict
+   with a private extension.) If so, the Aggregator MUST mark the input share as
+   invalid with error `invalid_message`.
 
 1. If the report pertains to a batch that was previously collected, then the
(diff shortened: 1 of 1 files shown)


draft-ietf-ppm-dap.md
```diff
@@ -1894,9 +1894,9 @@ following checks:
    invalid with error `invalid_message`.
 
 1. Check if any two public or private report extensions have the same extension
-   type. (For the purposes of this check, a private report extension can
-   conflict with a public report extension.) If so, the Aggregator MUST mark the
-   input share as invalid with error `invalid_message`.
+   type. (For the purposes of this check, a public extension MUST NOT conflict
+   with a private extension.) If so, the Aggregator MUST mark the input share as
```
tgeoghegan, line 1898: I don't understand the purpose of this parenthetical statement. As jhoyla notes, the old version was ambiguous. The new version is more correct, but now is redundant and I think we should delete it.
```suggestion
   type. If so, the Aggregator MUST mark the input share as
```

@cjpatton do you remember why that sentence in parentheses was ever there?

tgeoghegan (changes requested):

draft-ietf-ppm-dap.md
cjpatton: I don't recall discussing this paranthetical in particular.

I think we're attempting to define the following algorithm: let $S$ be the multiset of extension points among the public and private extension fields: if any value occurs twice in $S$, then reject.

I would probably clarify this as follows: "if any extension type occurs twice across the public and private extension fields, then reject".

cjpatton:
Closing in favor of #646. (Jonathan is on PTO, so i decided to drive this.)

PR #649, opened 2025-01-27
Enforce agg param validity (*)

cjpatton:
Before aggregating a batch, it is necessary to confirm that the aggregation parameter is valid by decoding it and running `is_valid()`.

While at it, remove the `batchQueriedMultipleTimes` abort type, since the condition it's used in is no longer possible.

Note that this is a breaking change, since previously the aggregation parameter would have been ignored.

### DIFF
--- a/draft-ietf-ppm-dap.md
+++ b/draft-ietf-ppm-dap.md
@@ -806,22 +806,22 @@ error (the 4xx or 5xx classes, respectively, from {{Section 15 of !RFC9110}}).
 To facilitate automatic response to errors, this document defines the following
 standard tokens for use in the "type" field:
 
-| Type                       | Description                                                                                  |
-|:---------------------------|:---------------------------------------------------------------------------------------------|
-| invalidMessage             | A message received by a protocol participant could not be parsed or otherwise was invalid. |
-| unrecognizedTask           | A server received a message with an unknown task ID. |
-| unrecognizedAggregationJob | A server received a message with an unknown aggregation job ID. |
-| outdatedConfig             | The message was generated using an outdated configuration. |
-| reportRejected             | Report could not be processed for an unspecified reason. |
-| reportTooEarly             | Report could not be processed because its timestamp is too far in the future. |
-| batchInvalid               | The batch boundary check for Collector's query failed. |
-| invalidBatchSize           | There are an invalid number of reports in the batch. |
-| batchQueriedMultipleTimes  | A batch was queried with multiple distinct aggregation parameters. |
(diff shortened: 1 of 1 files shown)


draft-ietf-ppm-dap.md
```diff
@@ -2341,6 +2371,10 @@ Upon receipt of a `CollectionJobReq`, the Leader begins by checking that it
 recognizes the task ID in the request path. If not, it MUST abort with error
 `unrecognizedTask`.
 
+Next, the Leader checks that the aggregation parameter is valid as described in
+{{agg-param-validation}}. If the aggregation parameter is invalid, then the
+Leader MUST abort with error `invalidAggregationParameter`.
```
tgeoghegan, line 2376: I think we should discuss this either above in L2350-2352 or here, but not in both places. It's confusing for implementers to have the same MUST laid out twice.

draft-ietf-ppm-dap.md
```diff
@@ -1584,7 +1607,9 @@ struct {
 
 This message consists of:
 
-* `agg_param`: The VDAF aggregation parameter.
+* `agg_param`: The VDAF aggregation parameter chosen by the Collector. Before
+  initializing an aggregation job, the Leader MUST validate the parameter as
+  described in {{agg-param-validation}}.
```
tgeoghegan, line 1612: This is already done when the Leader validates `CollectionJobReq`, isn't it? There's no harm in doing it again when scheduling aggregation jobs, but I don't think it's a MUST to do it in both places.

cjpatton: The MUST below is for the Leader: it tells it what to do if the agg param is invalid. Here we're just reminding the Leader that its agg param needs to be valid.

draft-ietf-ppm-dap.md
cjpatton: Just above, the first MUST is to remind the Collector that the agg param needs to be valid. Here we're telling the Leader what to do if the agg param is invalid.

draft-ietf-ppm-dap.md
tgeoghegan: Reading this again, I think the MUST would arguably be satisfied by having checked agg param validity while handling `CollectionReq`, so this line doesn't force implementations to do more.

draft-ietf-ppm-dap.md
tgeoghegan: I wasn't sure if a MUST on both the sender (Collector) and receiver (Leader) side is necessary, but re-reading RFC 2119, I think it's appropriate. Let's keep what you have here.

tgeoghegan (approved):

PR #651, opened 2025-02-04
Specify body of aggregation job GET requests.

branlwyd:
Previously, this was left implied (by mentioning a "status" field). I think it is better if we explicitly declare the body encoding for this request type.


branlwyd:
Mentioning the content type makes sense.

Squashed; I will merge in the next few days if no one objects (or feel free to merge yourself whenever convenient).

PR #653, opened 2025-02-18
Miscellaneous editorial changes

tgeoghegan:
Various minor fixes for typos, mislabeled tables, or other editorial issues.

PR #654, opened 2025-02-18
move and reorg presentation language exceptions

tgeoghegan:
Restructures the exceptions to RFC 8446 presentation language into individual, titled sections to make it more obvious for readers how many deviations there are. Also moves that entire section to section 3 Message Transport. This is because (1) we're discussing encoding messages on the wire so Message Transport is the appropriate place and (2) we should minimize the text before we get to the Overview section so that readers new to DAP get there as soon as possible.

Stacked on #653

### DIFF
--- a/draft-ietf-ppm-dap.md
+++ b/draft-ietf-ppm-dap.md
@@ -711,17 +711,25 @@ report IDs.
 Communications between DAP participants are carried over HTTP {{!RFC9110}}. Use
 of HTTPS is REQUIRED to provide server authentication and confidentiality.
 
-## Representation Language
+## Presentation Language
 
 With some exceptions, we use the presentation language defined in {{!RFC8446,
 Section 3}} to define messages in the DAP protocol. Encoding and decoding of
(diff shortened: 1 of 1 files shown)


draft-ietf-ppm-dap.md
```diff
@@ -750,8 +758,7 @@ variant {
   /* Field exists regardless of variant */
   uint32 always_present;
   ExampleEnum type = number;
-  /* Only fields included in the `type == number`
-     variant is described */
+  /* Only fields included in the `type == number` variant are described */
```
divergentdave, line 761: This line is just barely too long now

divergentdave (approved):

draft-ietf-ppm-dap.md
tgeoghegan: Oh, whoops, I thought this needed to be <80 rather than <72.

branlwyd (approved):

cjpatton (approved):

PR #655, opened 2025-02-19
Illustrate object lifecycles

### LINKED ISSUE #560: Discuss lifecycle of various objects and how they relate to each other

tgeoghegan:
Adds a section to the overview that includes a diagram illustrating the lifecycle of objects, i.e. how reports become aggregation jobs become aggregate shares etc. Also adds a prose section explaining the arity of each object.

Resolves #560

### DIFF
--- a/draft-ietf-ppm-dap.md
+++ b/draft-ietf-ppm-dap.md
@@ -710,6 +710,170 @@ last two weeks of the current time) and reject reports that fall outside of the
 replay window. This allows implementation to save resources by forgetting old
 report IDs.
 
+## Lifecycle of Protocol Objects
+
+The following diagram illustrates how the various objects in the protocol are
+constructed or transformed into other protocol objects. Nodes enclosed in an
+oval are verbs or actions which process, transform or combine one or more
+objects into one or more other objects. Note that this does not necessarily
+illustrate how participants communicate. In particular, the processing of
+aggregation jobs happens in distinct, non-colluding parties.
+
+~~~ aasvg
+                   measurement
+                        |
+                     .--+--.
+                    | shard |
+                     '--+--'
+                        |
+                        v
+      .-----------------+-------------------------.
+      |                 |                         |
+public share   Leader input share         Helper input share
+      |                 |                         |
+      |        .--------+--------.       .--------+--------.
+      |       | encrypt to Leader |     | encrypt to Helper |
+      |        '--------+--------'       '--------+--------'
+      |                 |                         |
+      |                 v                         v
+      |        Leader report share       Helper report share
+      |                 |                         |
+      '---------------->+<------------------------'
+                        |
+                        v           Client 2 report  ...   Client i report
+                 Client 1 report           |                      |
+                        |              .---+--.               .---+--.
+                    .---+--.          | upload |             | upload |
+                   | upload |          '---+--'               '---+--'
+                    '---+--'               |                      |
+                        | .----------------'                      |
+                        | |     .---------------------------------'
+                        | |     |
+                        v v ... v
+               .--------+-+-----+--.
+              |    assign reports   |
+              | to aggregation jobs |<--- aggregation parameter
+               '--------+----------'      chosen by Collector
+                        |
+                        v
+        .---------------+------------------+---------------------- ... -------.
+        |                                  |                                  |
+aggregation job 1                   aggregation job 2          aggregation job j
+        |                                  |                                  |
+        |                                  '-------------------------.        |
++==+====+==========================================================+ |        |
+║                  report 1         report 2 ... report k          ║=+=+      |
+║                     |                  |        |                ║   ║      |
+║  +==================+================+ |        |   aggregation  ║   ║      |
+║  ║   Leader    Helper report  public ║=+=+      |    parameter   ║   ║      |
+║  ║ report share 1  share 1   share 1 ║   ║      |        |       ║   ║ ... =++
+║  ║      |            |           |   ║   ║      |        |       ║   ║       ║
+║  ║  .---+---.    .---+---.       |   ║   ║      |        |       ║   ║       ║
+║  ║ | decrypt |  | decrypt |      |   ║   ║ ... =++       |       ║   ║       ║
+║  ║  '---+---'    '---+---'       |   ║   ║       ║       |       ║   ║       ║
+║  ║      |            |           |   ║   ║       ║       |       ║   ║       ║
+║  ║      v            v           |   ║   ║       ║       |       ║   ║       ║
+║  ║ Leader input   Helper input   |   ║   ║       ║       |       ║   ║       ║
+║  ║   share 1      share 1     .--'   ║   ║       ║       |       ║   ║       ║
+║  ║      |            |        |      ║   ║       ║       |       ║   ║       ║
+║  ║      v            v        v      ║   ║       ║       |       ║   ║       ║
+║  ║  .---+------------+--------+.     ║   ║       ║       |       ║   ║       ║
+║  ║ |    prepare input shares    |<---║---║-------║-------+       ║   ║       ║
+║  ║  '----+---------------+-----'     ║   ║       ║       |       ║   ║       ║
+║  ║       |               |           ║   ║       ║       |       ║   ║       ║
+║  ║       v               v           ║   ║<------║-------+       ║   ║       ║
+║  ║  Leader output   Helper output    ║   ║       ║       |       ║   ║       ║
+║  ║    share 1         share 1        ║   ║       ║               ║   ║       ║
+║  ║       |               |           ║   ║       ║      ...      ║   ║       ║
+║  +=======+===============+===========+   ║       ║               ║   ║       ║
+║    ║     | Leader output | Helper output ║       ║       |       ║   ║       ║
+║    ║     | share 2       | share 2       ║       ║<------'       ║   ║       ║
+║    ║     | |             | |             ║       ║               ║   ║       ║
+║    +=====+=+=============+=+=============+       ║               ║   ║       ║
+║      ... | |  Leader     | |   Helper            ║               ║   ║       ║
+║          | |  output     | |   output            ║               ║   ║       ║
+║       ║  | |  share k    | |   share k           ║               ║   ║       ║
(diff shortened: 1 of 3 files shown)


tgeoghegan:
I welcome any and all feedback here. I hope the diagram is helpful, but I might be too familiar with DAP to assess that fairly. I think it might help to use some kind of notation to distinguish things (e.g., "Helper input share" or "Leader batch bucket 2") from actions (e.g., "uploaded" or "aggregation jobs run"). Maybe verbs could be marked `*like this*` with an explanatory note?

tgeoghegan:
Hmm, yeah, the arrows are even prettier with the aasvg business. I'll take a look at incorporating that.

cjpatton:
Yeah I wouldn't worry about it for the other images for now, but I think we should make the new figure aasvg now so that we don't end up needing to rework it later.

tgeoghegan:
I adopted aasvg. This looks much prettier in the [editor's copy](https://ietf-wg-ppm.github.io/draft-ietf-ppm-dap/timg-lifecycle-diagram/draft-ietf-ppm-dap.html#name-lifecycle-of-protocol-objec).

draft-ietf-ppm-dap.md
```diff
@@ -706,6 +706,176 @@ last two weeks of the current time) and reject reports that fall outside of the
...
+oval are verbs or actions which process, transform or combine one or more
+objects into one or more other objects. Note that this does not necessarily
+illustrate how participants communicate. In particular, the processing of
+aggregation jobs happens in distinct, non-colluding parties.
+
+~~~ aasvg
+                   measurement
+                        |
+                    .---+---.
+                   | sharded |
```
cjpatton, line 722: s/sharded/shard/g

Likewise below.

draft-ietf-ppm-dap.md
```diff
@@ -706,6 +706,176 @@ last two weeks of the current time) and reject reports that fall outside of the
...
+  |          |            |             | | reports prepared |   |       |
+  |          v            v             |  '--+---------+---'    |       |
+  |    Leader input   Helper input      |     |         |   .----+-------+---.
+  |       share 1      share 1          |     |         |  | aggregation jobs |
+  |          |            |             |     |         |  |       run        |
+  '--------->+----------->|             |     |         |   '----+-------+---'
+             |<-----------+<------------'     |         |        |       |
+             |            |                   |         |        |       |
+         .---+------------+----.              |         |        |       |
+        | input shares prepared |             v         |        |       |
```
cjpatton, line 784: What's the difference between "reports prepared" and "input shares prepared"?

draft-ietf-ppm-dap.md
```diff
@@ -706,6 +706,176 @@ last two weeks of the current time) and reject reports that fall outside of the
...
+~~~
+{: #object-lifecycle title="Lifecycles of protocol objects" }
+
+### Arity of Protocol Objects
+
+DAP reports are 1:1 to with measurements. In this illustration, `i` distinct
+Clients upload a distinct report, but a single Client could upload multiple
+reports to a task (see {{sybil}} for some implications of this). The process of
+sharding measurements, constructing reports and uploading them is discussed in
+detail in {{upload-flow}}.
```
cjpatton, line 842:
```suggestion
sharding measurements, constructing reports and uploading them is specified in {{upload-flow}}.
```

draft-ietf-ppm-dap.md
```diff
@@ -706,6 +706,176 @@ last two weeks of the current time) and reject reports that fall outside of the
...
+               '----+----'
+                    |
+                    v
+             aggregate result
+~~~
+{: #object-lifecycle title="Lifecycles of protocol objects" }
+
+### Arity of Protocol Objects
+
+DAP reports are 1:1 to with measurements. In this illustration, `i` distinct
```
cjpatton, line 838: Likewise below
```suggestion
DAP reports are 1-to-1 to with measurements. In this illustration, `i` distinct
```

draft-ietf-ppm-dap.md
```diff
@@ -706,6 +706,176 @@ last two weeks of the current time) and reject reports that fall outside of the
...
+Clients upload a distinct report, but a single Client could upload multiple
+reports to a task (see {{sybil}} for some implications of this). The process of
+sharding measurements, constructing reports and uploading them is discussed in
+detail in {{upload-flow}}.
+
+Reports are many:1 with aggregation jobs. The Leader assigns each of the `i`
+reports into one of `j` different aggregation jobs, which can be run in parallel
+by the Aggregators. Each aggregation job contains `k` report shares. `k` is
```
cjpatton, line 846: ?
```suggestion
by the Aggregators. Each aggregation job contains `k` reports. `k` is
```

draft-ietf-ppm-dap.md
```diff
@@ -706,6 +706,176 @@ last two weeks of the current time) and reject reports that fall outside of the
...
+shares are decrypted into input shares and then prepared into output shares. The
+scheduling of aggregation jobs and their execution by the Aggregators is
+discussed in detail in {{aggregate-flow}}.
+
+Output shares are accumulated into batch buckets, and so have a many:1
+relationship. In this example, we show one batch bucket for each aggregation
+job, but aggregation jobs and batch buckets are not necessarily 1:1. Multiple
+aggregation jobs could contribute to the same batch bucket, and a single
+aggregation job could contribute to multiple batch buckets. The assignation of
+output shares to batch buckets is an implementation detail of each Aggregator.
```
cjpatton, line 861: I'm wondering to what extent this is true. The batch mode defines an infinite sequence of non-overlapping batch buckets. I suppose the aggregators could sub-divide these further if they wanted to. Is that what you mean?

cjpatton: I'm not following. My understanding of "batch bucket" as we define it in the spec:

1. time_interval: [0, t), [t, 2*t), [2*t, 3*t), ..., where t is short for time_precision
2. leader_selected: batch_id.

At aggregation time, the aggregator always knows which batch bucket a report belongs to.

draft-ietf-ppm-dap.md
cjpatton: I think I was missing, and what's not clear, is that we're "zooming in" on a part of the pipeline. Perhaps we should split out the zoomed-in part to its own figure?

cjpatton (commented):
Two comments

draft-ietf-ppm-dap.md
tgeoghegan: You are correct. I hadn't noticed we'd defined batch buckets that rigorously in [4.8.2.3](https://datatracker.ietf.org/doc/html/draft-ietf-ppm-dap-14#section-4.8.2.3), [5.1.4](https://datatracker.ietf.org/doc/html/draft-ietf-ppm-dap-14#section-5.1.4) and [5.2.4](https://datatracker.ietf.org/doc/html/draft-ietf-ppm-dap-14#section-5.2.4). Thanks for pointing that out. I fixed this text to refer to `{{batch-buckets}}`.

draft-ietf-ppm-dap.md
tgeoghegan: Fixed everywhere that I can see

draft-ietf-ppm-dap.md
tgeoghegan: Actually, I'm going to try one more thing that might make this more clear. It'll take a while, though.

tgeoghegan: OK, I reworked the diagram so that we illustrate the preparation of multiple reports, and then the running of multiple aggregation jobs, as stacks of boxes. Hopefully this makes it clear that all reports are handled the same way while being reasonably compact.

I'm trying to optimize for vertical space so that readers don't have to scroll too much when they look at this, but I'm open to using more room in exchange for greater clarity.

draft-ietf-ppm-dap.md
```diff
@@ -710,6 +710,163 @@ last two weeks of the current time) and reject reports that fall outside of the
...
+        | |  Leader batch               | |  Helper batch                      ║
+    ║   | |    bucket j                 | |    bucket j                        ║
+    +===|=|=====|=======================|=|=====|==============================+
+        v v ... v      query chosen     v v ... v
+     .--+-+-----+--.   by collector  .--+-+-----+--.
+    |  merge leader |       |       | merge helper  |
+    | batch buckets |<------+------>| batch buckets |
+     '------+------'                 '------+------'
+            |                               |
+            '---------------+---------------'
```
cjpatton, line 818: Perhaps we should show aggregate shares as intermediate values here.

cjpatton (approved):
This looks complete. There's a lot going on here ... I still think it would be helpful to split this out into a couple of figures. The second figure might just cover aggregation jobs, which would be an oval box in the first figure.

draft-ietf-ppm-dap.md
tgeoghegan: Oops, you're absolutely right. Those got lost when I reorganized the diagram.

draft-ietf-ppm-dap.md
```diff
@@ -710,6 +710,163 @@ last two weeks of the current time) and reject reports that fall outside of the
...
+║  ║    share 1         share 1        ║   ║       ║      ...      ║   ║       ║
+║  +=======|===============|===========+   ║       ║               ║   ║       ║
+║    ║     | Leader output | Helper output ║       ║       |       ║   ║       ║
+║    ║     | share 2       | share 2       ║       ║<------'       ║   ║       ║
+║    +=====|=|=============|=|=============+       ║               ║   ║       ║
+║      ... | |  Leader     | | Helper              ║               ║   ║       ║
+║          | |  output     | | output              ║               ║   ║       ║
+║       ║  | |  share k    | | share k             ║               ║   ║       ║
+║       +==|=|=====|=======|=|=====|===============+               ║   ║       ║
+║          v v ... v       v v ... v                               ║   ║       ║
```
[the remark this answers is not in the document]
tgeoghegan: I had to use a bit more vertical space to make this work but I think it's worth it.

PR #660, opened 2025-03-05
glossary: tidy protocol roles and add task

tgeoghegan:
Make the glossary's definitions of protocol roles use a consistent form and add VDAF references. Harmonize those definitions with the enumeration of protocol roles in section 2.1.

While we're in the glossary, also add a definition of "Task".


draft-ietf-ppm-dap.md
```diff
@@ -526,6 +528,11 @@ Report share:
   The report share also includes some associated data used for processing the
   report.
 
+Task:
+: A measurement task in which measurements of an understood type will be
```
cjpatton, line 532: Low confidence nit
```suggestion
: An aggregation task in which measurements of an understood type will be
```

cjpatton (approved):

draft-ietf-ppm-dap.md
tgeoghegan: We use the phrase "measurement task" in a couple other places in the doc, so there's precedent there, but now that I look at this again, I don't like using "measurement" twice in this sentence.

PR #665, opened 2025-03-06
edit Overview text

tgeoghegan:
Edits for concision, consistency and occasionally grammar.

### DIFF
--- a/draft-ietf-ppm-dap.md
+++ b/draft-ietf-ppm-dap.md
@@ -537,20 +537,19 @@ Task:
 
 # Overview {#overview}
 
-The protocol is executed by a large set of Clients and a pair of servers
-referred to as "Aggregators". Each Client's input to the protocol is its
-measurement (or set of measurements, e.g., counts of some user behavior). Given
-the input set of measurements `meas_1, ..., meas_N` held by the Clients, and an
-"aggregation parameter" `agg_param` shared by the Aggregators, the goal of DAP
-is to compute `agg_result = F(agg_param, meas_1, ..., meas_N)` for some
-function `F` while revealing nothing else about the measurements. We call `F`
-the "aggregation function" and `agg_result` the "aggregate result".
+The protocol is executed by a large set of Clients and a pair of Aggregators.
+Each Client's input to the protocol is its measurement (or set of measurements,
+e.g., counts of some user behavior). Given a set of measurements
+`meas_1, ..., meas_N` held by the Clients, and an "aggregation parameter"
+`agg_param` shared by the Aggregators, the goal of DAP is to compute
+`agg_result = F(agg_param, meas_1, ..., meas_N)` for some function `F` while
+revealing nothing else about the measurements. We call `F` the "aggregation
+function" and `agg_result` the "aggregate result".
 
 DAP is extensible in that it allows for the addition of new cryptographic
-schemes that compute different aggregation functions. In particular, the
-aggregation function is determined by the Verifiable Distributed Aggregation
-Function, or VDAF {{!VDAF=I-D.draft-irtf-cfrg-vdaf-14}}, used to securely
-compute it.
+schemes that compute different aggregation functions, determined by the
+Verifiable Distributed Aggregation Function, or
+{{!VDAF=I-D.draft-irtf-cfrg-vdaf-14}}, used to compute it.
 
 VDAFs rely on secret sharing to protect the privacy of the measurements. Rather
 than sending its measurement in the clear, each Client shards its measurement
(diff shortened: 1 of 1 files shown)


draft-ietf-ppm-dap.md
```diff
@@ -624,87 +622,77 @@ Helper:
...
-In DAP, input validation is accomplished by an interactive computation between
-the Leader and Helper. At the beginning of this computation, each Aggregator is
-in possession of an input share uploaded by the Client. At the end of the
-computation, each Aggregator is in possession of either an "output share" that
-is ready to be aggregated or an indication that the underlying data is invalid,
-in which case the report is rejected.
+DAP validates inputs using an interactive computation between the Leader and
+Helper. At the beginning of this computation, each Aggregator holds an input
+share uploaded by the Client. At the end of the computation, each Aggregator
+either holds an output share that is ready to be aggregated or rejects the
```
divergentdave, line 654: Having an active and passive verb on either side of this "either ... or" feels unusual to me
```suggestion
either obtains an output share that is ready to be aggregated or rejects the
```

divergentdave (approved):

draft-ietf-ppm-dap.md
```diff
@@ -537,20 +537,19 @@ Task:
...
+function" and `agg_result` the "aggregate result".
 
 DAP is extensible in that it allows for the addition of new cryptographic
-schemes that compute different aggregation functions. In particular, the
-aggregation function is determined by the Verifiable Distributed Aggregation
-Function, or VDAF {{!VDAF=I-D.draft-irtf-cfrg-vdaf-14}}, used to securely
-compute it.
+schemes that compute different aggregation functions, determined by the
+Verifiable Distributed Aggregation Function, or
+{{!VDAF=I-D.draft-irtf-cfrg-vdaf-14}}, used to compute it.
```
jcjones, line 552:
```suggestion
{{!VDAF=I-D.draft-irtf-cfrg-vdaf-14}}, used in the computation.
```
?

draft-ietf-ppm-dap.md
```diff
@@ -624,87 +622,77 @@ Helper:
...
 "report ID". Each Aggregator is responsible for storing the IDs of reports it
-has aggregated for a given task. To check whether a report has been replayed,
-it checks whether the report's ID is in the set of stored IDs.
-
-Note that IDs do not need to be stored indefinitely. The protocol allows
-Aggregators to enforce replay only for a sliding time window (e.g., within the
-last two weeks of the current time) and reject reports that fall outside of the
-replay window. This allows implementation to save resources by forgetting old
-report IDs.
+has aggregated and rejecting replayed reports.
```
jcjones, line 686: Should we say `repeated` instead of `replayed`, to be more encompassing of errors rather than malicious action?

tgeoghegan: I don't think "repeated" would be wrong, but we frequently discuss replay attacks and replayed reports elsewhere in the draft, so I think consistency is important.

draft-ietf-ppm-dap.md
tgeoghegan: I don't disagree with your phrasing but am not sure why it's better. Is the idea to allow for things besides the VDAF to be involved in the computation?

cjpatton: I prefer the current verbiage: a VDAF computes an aggregation function, it's not just one part of the computation.

draft-ietf-ppm-dap.md
```diff
@@ -561,20 +560,19 @@ Aggregators. This scheme has two important properties:
...
 
-Note that some VDAFs allow measurements to be aggregated multiple times, each
-time with a different aggregation parameter; however, DAP only allows each
-measurement to be aggregated once. Similarly, some VDAFs produce aggregate
-values which depend on the order in which the measurements are aggregated;
-however, DAP only supports VDAFs whose aggregation results are independent of
-the order in which measurements are aggregated (see {{Section 4.4.1 of !VDAF}}).
+DAP is not compatible with all VDAFs. Some VDAFs allow measurements to be
+aggregated multiple times with a different aggregation parameter. DAP only
+allows each measurement to be aggregated once. Some VDAFs produce aggregate
```
cjpatton, line 568: At first I read this as "DAP doesn't support Mastic", which I don't think is the intent. It's true that there are certain things we can do with Poplar1/Mastic that we can't do in DAP, but there are still useful applications of Mastic (and Poplar1, kind of) that DAP supports.

Basically I would just avoid conflating multi-collection and order-sensitive VDAFs here.

draft-ietf-ppm-dap.md
```diff
@@ -624,87 +622,77 @@ Helper:
...
 
 In the course of a measurement task ({{task-configuration}}), each Client
-records its own measurement value(s), packages them up into a report, and sends
-them to the Leader. Each share is encrypted to only one of the two Aggregators
-so that even though they pass through the Leader, the Leader is unable to see or
-modify the Helper's share. Depending on the task, the Client may only send one
-report or may send many reports over time.
+records its own measurement, packages it up into a report, and sends it to the
+Leader. Each share is encrypted to only one of the two Aggregators so that even
+though they pass through the Leader, the Leader is unable to see or modify the
```
cjpatton, line 632:
```suggestion
though both pass through the Leader, the Leader is unable to see or modify the
```

draft-ietf-ppm-dap.md
```diff
@@ -624,87 +622,77 @@ Helper:
...
 The Leader distributes the shares to the Helper and orchestrates the process of
-verifying them (see {{validating-inputs}}) and assembling them into an aggregate
-result for the Collector. Depending on the VDAF, it may be possible to
-incrementally process each report as it is uploaded, or it may be necessary to
-wait until the Collector transmits a collection request before processing can
-begin.
+verifying them and assembling them into a aggregate shares for the Collector.
+Depending on the VDAF, it may be possible to process each report as it is
+uploaded, or it may be necessary to wait until the Collector transmits a
+collection request before processing can begin.
```
cjpatton, line 640: WDYT?
```suggestion
uploaded, or it may be necessary to wait until the Collector initializes a collection job before processing can begin.
```

tgeoghegan: Sure, that's more consistent with how we discuss collection jobs elsewhere.

draft-ietf-ppm-dap.md
```diff
@@ -561,20 +560,20 @@ Aggregators. This scheme has two important properties:
...
-time with a different aggregation parameter; however, DAP only allows each
-measurement to be aggregated once. Similarly, some VDAFs produce aggregate
-values which depend on the order in which the measurements are aggregated;
-however, DAP only supports VDAFs whose aggregation results are independent of
-the order in which measurements are aggregated (see {{Section 4.4.1 of !VDAF}}).
+  their shares locally into "aggregate shares", which may later be merged into
+  the aggregate result.
+
+DAP is not compatible with all VDAFs. Some VDAFs produce aggregate
+values which depend on the order in which the measurements are aggregated. DAP
```
cjpatton, line 567: might: This is situation is very hypothetical
values -> results: consistency with terminology
```suggestion
DAP is not compatible with all VDAFs. Some VDAFs might produce aggregate
results which depend on the order in which the measurements are aggregated. DAP
```

cjpatton (approved):

jcjones (approved):

PR #666, opened 2025-03-06
Other problem types may be any URI

divergentdave:
This is a follow-up to #661. While the problem types defined in this document are all URNs, we should still allow for servers to use other URIs that are not URNs as problem types. In particular, our implementation, Janus, uses `about:blank` in one case, where want to add a detail field while keeping the semantics of one of the HTTP status codes. We also use `https:` URLs for a couple of implementation-specific error cases.


tgeoghegan (approved):
Oops, you're right, I forgot that Janus can provide links to docs.divviup.org.

jcjones (approved):

Issue #647, opened 2025-01-23
Support bulk submissions in DAP


akoshelev:
There is currently [interest](https://github.com/patcg/ppa-api/issues/36) from the PPA group to use DAP as the main protocol to aggregate measurement results. However, there is a significant difference in how reports are submitted to the MPC: Rather than coming from end user devices, they are collected by an intermediary and later submitted to the Leader in [bulk](https://patcg.github.io/ppa-api/#overview). The PPA group concluded that network bandwidth savings are worth having a DAP extension that enables bulk uploads.

In https://github.com/patcg/ppa-api/issues/36, @cjpatton expressed support to have this functionality implemented in DAP, so this issue is open to see if there is appetite from this group to support it. 

The formal proposal to change DAP can be found [here](https://datatracker.ietf.org/doc/html/draft-thomson-ppm-dap-bulk-00) (aside from section 4.1 that is I believe is addressed in [dap-13](https://datatracker.ietf.org/doc/html/draft-ietf-ppm-dap-13#name-upload-request))

### Previous work

This issue has been discussed in https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/issues/64 and https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/pull/78. This group concluded that the complexity of supporting batch submissions (particularly handling partial failures) was not worth the network savings obtained from grouping request metadata in one header.

cjpatton:
My main recollection from previous discussions is that if batched uploading could be handled at the HTTP level, and that we would perhaps better serve higher level applications to force them to do just that.

If the reports your batch-uploading have fields in common, then you can in theory save a little bandwidth:
1. the task ID (32 bytes per upload)
2. any public extensions that have identical payloads (savings vary based on the contents of extensions)
3. the [taskprov advertisement](https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap-taskprov), if you're using that extension and using it to provision tasks (savings vary, perhaps a 100 or so bytes per report)

Regardless of how much we might save, I think if sending a very large HTTP request instead of many small HTTP requests is convenient for PPA, then we should support it.

akoshelev:
> How "big" might report extensions get for PPA? 

At this moment we expect at least [3 extensions](https://w3c.github.io/ppa/#prio) (not including the one proposed here) to be used for PPA submissions.

> Would the payloads be the same across all or at least many of the reports?

[...]

PR #667, opened 2025-03-11
Adopt Bulk Submission protocol in DAP

akoshelev:
This addresses #647 by incorporating the functionality to support bulk uploads into DAP spec. 

## References
* [dap-bulk](https://martinthomson.github.io/dap-bulk/draft-thomson-ppm-dap-bulk.html) extension proposal to DAP. This change builds on top of it

## Outstanding issues
* The mechanism to report upload progress is described at the very high-level. It tries to avoid relying on any specific HTTP protocol version. Is it what we want here? 

## Preview

[HTML](https://akoshelev.github.io/draft-ietf-ppm-dap/dap-bulk/draft-ietf-ppm-dap.html#name-bulk-upload)

### DIFF
--- a/draft-ietf-ppm-dap.md
+++ b/draft-ietf-ppm-dap.md
@@ -1382,7 +1382,7 @@ Note: Long cache lifetimes may result in Clients using stale HPKE
 configurations; Aggregators SHOULD continue to accept reports with old keys for
 at least twice the cache lifetime in order to avoid rejecting reports.
 
-### Upload Request
+### Upload Request  {#upload-request}
 
 Clients upload reports by using an HTTP POST to
 `{leader}/tasks/{task-id}/reports`. The payload is a `Report`, with media type
@@ -1551,6 +1551,37 @@ abort the upload request with error "invalidMessage".
 the Leader to decrypt its input share. The Leader also cannot validate the
 Helper's extensions because it cannot decrypt the Helper's input share.
 Mandatory validation of extensions will occur during aggregation.)
+
+### Bulk upload
+
+Clients or intermediates with multiple reports in hand can use bulk upload endpoint
+and send an HTTP POST to `{leader}/tasks/{task-id}/reports/bulk` with media type
+"application/dap-bulk-report". A bulk request is equivalent to multiple separate
+submissions and is strictly a performance optimization.
+
+The "application/dap-bulk-report" format contains a prologue that encodes extensions that are
+common to all reports. The header is followed by any number of reports, from which those
+shared fields are removed.
+
+~~~ tls-presentation
+struct {
+Extension common_extensions<0..2^16-1>;
+Report report[REPORT_COUNT];
+} BulkReport;
+~~~
+{: #syn-bulk-report title="Bulk Report Format"}
+
+The use of `REPORT_COUNT` in {{syn-bulk-report}} is a small abuse of the TLS syntax
+to signify that any number of reports are included. This "value" could be any
+positive integer. Unlike a variable-length field, as denoted with `&lt;` and `>`,
+this format does not require that the size of all included reports
+be known before constructing a request.
+
+The `common_extensions` field contains common extensions that are added to the set of
+public report extensions in each report that follows. Reports that include values for
+any common extension override the value in the common extension.
+
+Each individual report from the bulk upload request will be processed as described in {{upload-request}}.
 
 ### Report Extensions {#report-extensions}
 
--- a/draft-ietf-ppm-dap.md
+++ b/draft-ietf-ppm-dap.md
@@ -1552,7 +1552,7 @@ the Leader to decrypt its input share. The Leader also cannot validate the
 Helper's extensions because it cannot decrypt the Helper's input share.
 Mandatory validation of extensions will occur during aggregation.)
 
-### Bulk upload
+### Bulk upload {#bulk-upload}
 
 Clients or intermediates with multiple reports in hand can use bulk upload endpoint
 and send an HTTP POST to `{leader}/tasks/{task-id}/reports/bulk` with media type
@@ -1582,6 +1582,35 @@ public report extensions in each report that follows. Reports that include value
 any common extension override the value in the common extension.
 
 Each individual report from the bulk upload request will be processed as described in {{upload-request}}.
+
+### Partial failures
+
+Bulk uploads take longer time to complete and have higher chance to fail, leaving clients
+uncertain about the status of their uploads. Replaying the entire request again
+wastes both server and client resources and network bandwidth.
+
+To provide  more robust failure handling, reports in the upload MUST be implicitly
+enumerated by the Server using 0-based indexing. Monotonically increasing report index
+is used to communicate the upload status to the Client.
+
+Any report in the upload may be rejected for reasons listed in ({{upload-request}}).
+
+Leader SHOULD provide a means to Clients to obtain status of the upload.
+
+~~~ tls-presentation
+struct {
+/* The index of the latest report processed so far by the server */
+uint32 latest_index,
+/* The list of indices of reports rejected on the server side */
+uint32 rejected<..>,
+} BulkReportStatus;
+~~~
+
+Server MAY compress large sequences of rejected reports into ranges, to avoid payload blowout.
+
+If bulk upload request fails for any reason, Client MAY choose to submit bulk upload request again.
+Client SHOULD only include reports not confirmed by the Leader. Clients MUST handle
+`reportRejected` errors ({{upload-request}}) during this upload.
 
 ### Report Extensions {#report-extensions}
 
--- a/draft-ietf-ppm-dap.md
+++ b/draft-ietf-ppm-dap.md
@@ -81,7 +81,10 @@ contributor:
        name: Charlie Harrison
        org: Google
        email: csharrison@chromium.org
-
+ -
+       name: Alex Koshelev
+       org: Meta
+       email: koshelev@meta.com
  -
        name: Peter Saint-Andre
        email: stpeter@gmail.com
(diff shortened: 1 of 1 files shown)


draft-ietf-ppm-dap.md
```diff
@@ -106,6 +106,11 @@ contributor:
        org: Apple
        email: shan_wang@apple.com
 
+ -
+       name: Alex Koshelev
+       org: Meta
+       email: koshelev@meta.com
+
```
cjpatton, line 113: Let's keep this list in alphabetical order.

draft-ietf-ppm-dap.md
```diff
@@ -1547,6 +1552,66 @@ the Leader to decrypt its input share. The Leader also cannot validate the
...
+Report report[REPORT_COUNT];
+} BulkReport;
+~~~
+{: #syn-bulk-report title="Bulk Report Format"}
+
+The use of `REPORT_COUNT` in {{syn-bulk-report}} is a small abuse of the TLS syntax
+to signify that any number of reports are included. This "value" could be any
+positive integer. Unlike a variable-length field, as denoted with `&lt;` and `>`,
+this format does not require that the size of all included reports
+be known before constructing a request.
```
cjpatton, line 1578: For this to work, the Leader would need to know what value of `REPORT_COUNT` to expect. This could be an API parameter, for example.

A more conventional alternative would be to add a "length prefix" to the `report` field so that the Leader knows how many total bytes it expects to parse. Something like this:

```
struct {
  Report reports_list<0..2^32-1>;
} BulkReport;
```

The `<0..2^32-1>` signifies that the list of reports can be as long as `2^32-1` bytes. This also implies that that the list is prefixed by a four byte string that encodes the length of the remainder.

FWIW I don't think this is an abuse of TLS syntax :) It's pretty common.

draft-ietf-ppm-dap.md
```diff
@@ -1547,6 +1552,66 @@ the Leader to decrypt its input share. The Leader also cannot validate the
...
+
+The `common_extensions` field contains common extensions that are added to the set of
+public report extensions in each report that follows. Reports that include values for
+any common extension override the value in the common extension.
```
cjpatton, line 1582: This behavior can trigger report rejection in two cases:
1. If we override a public extension, and the payload is different, then HPKE decryption will fail.
2. If we add a public extension, but there is a private extension with the same codepoint, [then the aggregator is suppoesd to reject](https://ietf-wg-ppm.github.io/draft-ietf-ppm-dap/draft-ietf-ppm-dap.html#section-4.7.1.4-3.7.1).

Also, note that each client would need to bind the common extensions to its report via the [`InputShareAad`](https://ietf-wg-ppm.github.io/draft-ietf-ppm-dap/draft-ietf-ppm-dap.html#section-4.6.2-16).

I think to deal with both of these cases, the Leader needs to convey the common extensions to the Helper during aggregation. This presents another questions: are common extensions scoped to a batch, an aggregation job, or  bulk uploads? My suspicion is that the protocol logic would get really hairy unless we scope this to a batch.

draft-ietf-ppm-dap.md
```diff
@@ -1547,6 +1552,66 @@ the Leader to decrypt its input share. The Leader also cannot validate the
...
+uncertain about the status of their uploads. Replaying the entire request again
+wastes both server and client resources and network bandwidth.
+
+To provide  more robust failure handling, reports in the upload MUST be implicitly
+enumerated by the Server using 0-based indexing. Monotonically increasing report index
+is used to communicate the upload status to the Client.
+
+Any report in the upload may be rejected for reasons listed in ({{upload-request}}).
+
+Leader SHOULD provide a means to Clients to obtain status of the upload.
```
cjpatton, line 1598: This PR should specify this mechanism.

draft-ietf-ppm-dap.md
```diff
@@ -1547,6 +1552,66 @@ the Leader to decrypt its input share. The Leader also cannot validate the
...
+~~~ tls-presentation
+struct {
+/* The index of the latest report processed so far by the server */
+uint32 latest_index,
+/* The list of indices of reports rejected on the server side */
+uint32 rejected<..>,
+} BulkReportStatus;
+~~~
+
+Server MAY compress large sequences of rejected reports into ranges, to avoid payload blowout.
```
cjpatton, line 1609: This should be fully specified, since "compress ... into ranges" can have multiple interpretations.

draft-ietf-ppm-dap.md
```diff
@@ -3533,6 +3598,17 @@ report extensions {{report-extensions}} used by the task. Since meaningful
 privacy requires that multiple Clients contribute to a task, they should also
 share a consistent view of the task configuration.
 
+## Bulk uploads
+
+The potential for a single client to generate large amounts of work using
+{{bulk-upload}} for a DAP service is a serious threat to service availability.
```
cjpatton, line 3604: I think "serious" is perhaps too strong of a word :)

draft-ietf-ppm-dap.md
```diff
@@ -1547,6 +1552,66 @@ the Leader to decrypt its input share. The Leader also cannot validate the
...
+
+### Partial failures
+
+Bulk uploads take longer time to complete and have higher chance to fail, leaving clients
+uncertain about the status of their uploads. Replaying the entire request again
+wastes both server and client resources and network bandwidth.
+
+To provide  more robust failure handling, reports in the upload MUST be implicitly
+enumerated by the Server using 0-based indexing. Monotonically increasing report index
+is used to communicate the upload status to the Client.
```
tgeoghegan, line 1594: What do you mean by "implicitly" here? That the Leader MUST assign indices that match the order implied by the order in which reports appear in `BulkReport.reports`?

tgeoghegan (commented):
I left a review with some specific questions and comments on the PR as it stands. But I have some higher level discussion as well.

I support taking a change in DAP because uploading multiple reports at once to deduplicate content and save on other resources is a reasonable feature, and we have some time left before WGLC. So if we can take something that isn't too disruptive to the existing protocol and solves your use case, then great!

Given that context, a design restriction I want to place here is that we should publish DAP with only a single mechanism for uploads, where the upload message contains a vector of reports. The current case of DAP uploads would simply be a vector of length 1.

But I don't think that such an approach would solve your use case, and what you're working toward here is richer than that. What we have here is the beginnings of an interaction wherein a client can stream a large number of reports into the leader, perhaps over a long-lived connection. Parallel to that, the leader streams errors back into the client (though the API metaphor may be different, using a paginated list or a queue) to enable efficient retries. This interaction exists alongside the existing upload interaction.

My concern is that we will struggle to design a solution that satisfies your use case while confining itself to the parts of HTTP DAP currently uses, or without introducing unsatisfactory complexity to the common upload case.

I think it might help if instead of starting by articulating things in POSTs of messages in TLS presentation language syntax, why don't you start by laying out for us your ideal conception of how this works? Don't worry about whether you're being consistent with DAP's established conventions, or if you're relying on transport mechanisms that we have RFCs for. Just tell us how you'd do this if you owned both endpoints and could use whatever primitives you want, and then I think we can try to figure out together if that fits into DAP, or if we should try it in another document.

draft-ietf-ppm-dap.md
```diff
@@ -1547,6 +1552,66 @@ the Leader to decrypt its input share. The Leader also cannot validate the
...
 
+### Bulk upload {#bulk-upload}
+
+Clients or intermediates with multiple reports in hand can use bulk upload endpoint
+and send an HTTP POST to `{leader}/tasks/{task-id}/reports/bulk` with media type
+"application/dap-bulk-report". A bulk request is equivalent to multiple separate
+submissions and is strictly a performance optimization.
+
+The "application/dap-bulk-report" format contains a prologue that encodes extensions that are
+common to all reports. The header is followed by any number of reports, from which those
```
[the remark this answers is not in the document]
akoshelev: I forgot to change this when doing s/header/prologue. Will fix

draft-ietf-ppm-dap.md
akoshelev: I was thinking that we could leverage end of stream messages (or [upload complete](https://httpwg.org/http-extensions/draft-ietf-httpbis-resumable-upload.html#name-completeness)) header to indicate that this is the last report included in the payload.

In my opinion, the latter model where clients communicate end of stream rather than telling the payload size in advance is a bit friendlier. Often, in order to compute the total size, clients need to perform some non-trivial IO. On the other hand, clients almost always know where and when the report stream ends

akoshelev:
I spoke with @cjpatton yesterday and we arrived to a conclusion that the shape of the API proposed here is not a good fit for core DAP for the following reasons

* Streaming API is not something that is used in other parts of DAP. A similar problem for aggregation job was solved by chunking requests.
* There may be implementation challenges for Leaders to support the streaming API. DAP servers backed by serverless compute may struggle to keep active requests for minutes to receive all the data. 

The conclusion we arrived at was to enrich the `{leader}/tasks/{task-id}/reports` endpoint to accept more than one report, but number of reports needs to be known in advance and try to make it part of core DAP spec. If PPA team finds it insufficient, then we can continue pursuing an extension to DAP to support streaming. 

@cjpatton I was thinking that we still create a new endpoint for this `{leader}/tasks/{task-id}/reports/bulk`, but since our last conversation I changed my mind because:

* If we were to implement a streaming API, then there will be 3 endpoints which makes it harder to choose. 
* Given that the size of payload is limited, the risk of having DDoS attacks on this endpoint is not that different from what we have in DAP currently.

The next step for me is to close this PR and open a new one that proposes changing `{leader}/tasks/{task-id}/reports` and make it work for 1..N reports

PR #668, opened 2025-03-12
Move and edit the protocol definition's discussion of batch modes

tgeoghegan:
The existing discussion of batch modes, batches and queries in the protocol definition section:

- discusses tasks, but comes before the {#task-configuration} section that introduces that concept
- contains normative information like batch mode identifiers and generally has more information in it than is strictly necessary to get the reader moving through the text

In this change:

- the overview is re-ordered so that batch modes and queries are introduced as a sub-section of {#task-configuration}
- some material is deleted from the protocol definition and moved into {#batch-modes}, so that there is a single, authoritative section normatively describing batch modes

I structured the PR as two commits for clarity. The first just moves the entire block of text, and the second makes edits.

### DIFF
--- a/draft-ietf-ppm-dap.md
+++ b/draft-ietf-ppm-dap.md
@@ -1166,8 +1166,8 @@ following parameters associated with it:
   in a `Prio3Histogram`).
 * A URL relative to which the Leader's API resources can be found.
 * A URL relative to which the Helper's API resources can be found.
-* The batch mode for this task (see {{batch-mode}}), which determines how reports
-  are grouped into batches.
+* The batch mode for this task (see {{batch-modes-overview}}), which determines
+  how reports are grouped into batches.
 * `task_interval` (`Interval`): Reports whose timestamp is outside of this
   interval will be rejected by the Aggregators.
 * `time_precision` (`Duration`): Used to compute timestamps in DAP. See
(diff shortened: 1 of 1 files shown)


divergentdave (approved):

draft-ietf-ppm-dap.md
```diff
@@ -1239,6 +1198,23 @@ A task's parameters are immutable for the lifetime of that task. The only way to
 change parameters or to rotate secret values like collector HPKE configuration
 or the VDAF verification key is to configure a new task.
 
+### Batch Modes, Batches, and Queries {#batch-modes-overview}
+
+An aggregate result is computed from a set of reports, called a "batch". The
+Collector requests the aggregate result by making a "query" and the Aggregators
+use this query to construct a batch for aggregation.
```
cjpatton, line 1205:
```suggestion
use this query to select a batch for aggregation.
```

draft-ietf-ppm-dap.md
```diff
@@ -1239,6 +1198,23 @@ A task's parameters are immutable for the lifetime of that task. The only way to
 change parameters or to rotate secret values like collector HPKE configuration
 or the VDAF verification key is to configure a new task.
 
+### Batch Modes, Batches, and Queries {#batch-modes-overview}
```
cjpatton, line 1201: Worth mentioning here that batches are always disjoint sets of reports.

cjpatton (approved):

Issue #671, opened 2025-03-20
RFC 9205 alignment


tgeoghegan:
At the most recent PPM interim, I discussed doing a pass on the document to align with RFC 9205/BCP 56. The most substantial change there is #669, but there's also smaller changes like how we discuss HTTP status codes or use of HTTPS, and I want to track and make those changes separately.

tgeoghegan:
This is in #673, #674, #675

tgeoghegan:
All done!

PR #673, opened 2025-03-20
RFC 9205 alignment, part 1

tgeoghegan:
Edits to the sections on message transport, HPKE config request and upload request to align with RFC 9205.
    
- provide examples of requests and responses
- use broadest possible status codes
- refer to RFC 9110 for HTTPS cert validation
    
Part of #671

### DIFF
--- a/draft-ietf-ppm-dap.md
+++ b/draft-ietf-ppm-dap.md
@@ -857,8 +857,16 @@ aggregate results once an hour, or every time 10,000,000 reports are uploaded.
 
 # Message Transport {#message-transport}
 
-Communications between DAP participants are carried over HTTP {{!RFC9110}}. Use
-of HTTPS is REQUIRED to provide server authentication and confidentiality.
+Communications between participants are carried over HTTP {{!RFC9110}}. Use of
+HTTPS is REQUIRED to provide server authentication and confidentiality. TLS
+certificates MUST be checked according to {{!RFC9110, Section 4.3.4}}.
+
+## HTTP Status Codes
+
+HTTP servers participating in DAP MAY use any status code from the applicable
+class when constructing HTTP responses, but HTTP clients MAY treat any status
+code as the most general code of that class. For example, 202 may be handled as
+200, or 499 as 400.
 
 ## Presentation Language
 
@@ -974,15 +982,14 @@ are supported by a participant is outside of this document's scope.
 
 ## Errors
 
-Errors can be reported in DAP both as HTTP status codes and as problem detail
-objects {{!RFC9457}} in the response body. For example, if the HTTP client sends
-a request using a method not allowed in this document, then the server MAY
-return HTTP status 405 Method Not Allowed.
+Errors are reported as HTTP status codes. Any of the standard client or server
+errors (the 4xx or 5xx classes, respectively, from {{Section 15 of !RFC9110}})
+are permitted.
 
 When the server responds with an error status code, it SHOULD provide additional
-information using a problem detail object. If the response body does consist of
-a problem detail object, the HTTP status code MUST indicate a client or server
-error (the 4xx or 5xx classes, respectively, from {{Section 15 of !RFC9110}}).
+information using a problem detail object {{!RFC9457}} in the response body. If
+the response body does consist of a problem detail object, the status code MUST
+indicate a client or server error.
 
 To facilitate automatic response to errors, this document defines the following
 standard tokens for use in the "type" field:
@@ -999,7 +1006,6 @@ standard tokens for use in the "type" field:
 | invalidBatchSize            | There are an invalid number of reports in the batch. |
 | invalidAggregationParameter | The aggregation parameter assigned to a batch is invalid. |
 | batchMismatch               | Aggregators disagree on the report shares that were aggregated in a batch. |
-| unauthorizedRequest         | Authentication of an HTTP request failed (see {{request-authentication}}). |
 | stepMismatch                | The Aggregators disagree on the current step of the DAP aggregation protocol. |
 | batchOverlap                | A request's query includes reports that were previously collected in a different batch. |
 | unsupportedExtension        | An upload request's extensions list includes an unknown extension. |
@@ -1026,7 +1032,8 @@ refer to error types, rather than the full URNs. For example, an "error of type
 
 This document uses the verbs "abort" and "alert with \[some error message\]" to
 describe how protocol participants react to various error conditions. This
-implies HTTP status code 400 Bad Request unless explicitly specified otherwise.
+implies that the response's status code will indicate a client error unless
+specified otherwise.
 
 # Protocol Definition {#protocol}
 
@@ -1038,12 +1045,10 @@ DAP has three major interactions which need to be defined:
 * The Collector collects aggregated results from the Aggregators, specified in
   {{collect-flow}}
 
-Each of these interactions is defined in terms of "resources". In this section
-we define these resources and the messages used to act on them.
+Each of these interactions is defined in terms of HTTP resources. In this
+section we define these resources and the messages used to act on them.
 
 ## Basic Type Definitions {#basic-definitions}
-
-The following are some basic type definitions used in other messages.
 
 A `ReportID` is used to uniquely identify a report in the context of a DAP task.
 
@@ -1221,15 +1226,14 @@ conveyed from the Leader to the Helper when initializing aggregation jobs
 
 ## Resource URLs
 
-DAP is defined in terms of "resources", such as reports ({{upload-flow}}),
+DAP is defined in terms of HTTP resources, such as reports ({{upload-flow}}),
 aggregation jobs ({{aggregate-flow}}), and collection jobs ({{collect-flow}}).
-Each resource has an associated URL. Resource URLs are specified by a sequence
-of string literals and variables. Variables are expanded into strings according
-to the following rules:
+Each resource has a URL. Resource URLs are specified as string literals
+containing variables. Variables are expanded into strings according to the
+following rules:
 
 * Variables `{leader}` and `{helper}` are replaced with the base API URL of the
-  Leader and Helper respectively (the base URLs are defined in
-  {{task-configuration}}).
+  Leader and Helper respectively.
 * Variables `{task-id}`, `{aggregation-job-id}`, and `{collection-job-id}` are
   replaced with the task ID ({{task-configuration}}), aggregation job ID
   ({{agg-init}}), and collection job ID ({{collect-init}}) respectively. The
@@ -1274,14 +1278,14 @@ Before the Client can upload its report to the Leader, it must know the HPKE
 configuration of each Aggregator. See {{compliance}} for information on HPKE
 algorithm choices.
 
-Clients retrieve the HPKE configuration from each Aggregator by sending an HTTP
-GET request to `{aggregator}/hpke_config`.
+Clients retrieve the HPKE configuration from each Aggregator by sending a GET to
+`{aggregator}/hpke_config`.
 
-An Aggregator responds to well-formed requests with HTTP status code 200 OK and
-an `HpkeConfigList` value, with media type "application/dap-hpke-config-list".
-The `HpkeConfigList` structure contains one or more `HpkeConfig` structures in
-decreasing order of preference. This allows an Aggregator to support multiple
-HPKE configurations simultaneously.
+An Aggregator responds with an `HpkeConfigList`, with media type
+"application/dap-hpke-config-list". The `HpkeConfigList` contains one or more
+`HpkeConfig`s in decreasing order of preference. This allows an Aggregator to
+support multiple HPKE configurations and multiple sets of algorithms
+simultaneously.
 
 ~~~ tls-presentation
 HpkeConfig HpkeConfigList<0..2^16-1>;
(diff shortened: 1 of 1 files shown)


draft-ietf-ppm-dap.md
```diff
@@ -853,8 +853,16 @@ aggregate results once an hour, or every time 10,000,000 reports are uploaded.
 
 # Message Transport {#message-transport}
 
-Communications between DAP participants are carried over HTTP {{!RFC9110}}. Use
-of HTTPS is REQUIRED to provide server authentication and confidentiality.
+Communications between participants are carried over HTTP {{!RFC9110}}. Use of
+HTTPS is REQUIRED to provide server authentication and confidentiality. TLS
+certificates MUST be checked according to {{!RFC9110, Section 4.3.4}}.
```
tgeoghegan, line 858: https://datatracker.ietf.org/doc/html/rfc9205#section-4.3-4.6

draft-ietf-ppm-dap.md
```diff
@@ -853,8 +853,16 @@ aggregate results once an hour, or every time 10,000,000 reports are uploaded.
...
+
+## HTTP Status Codes
+
+HTTP servers participating in DAP MAY use any status code from the applicable
+class when constructing HTTP responses, but HTTP clients MAY treat any status
+code as the most general code of that class. For example, 202 may be handled as
+200, or 499 as 400.
```
tgeoghegan, line 865: https://datatracker.ietf.org/doc/html/rfc9205#section-4.6-4

draft-ietf-ppm-dap.md
```diff
@@ -970,15 +978,14 @@ are supported by a participant is outside of this document's scope.
 
 ## Errors
 
-Errors can be reported in DAP both as HTTP status codes and as problem detail
-objects {{!RFC9457}} in the response body. For example, if the HTTP client sends
-a request using a method not allowed in this document, then the server MAY
-return HTTP status 405 Method Not Allowed.
+Errors are reported as HTTP status codes. Any of the standard client or server
+errors (the 4xx or 5xx classes, respectively, from {{Section 15 of !RFC9110}})
+are permitted.
```
tgeoghegan, line 983: https://datatracker.ietf.org/doc/html/rfc9205#section-4.6-4

draft-ietf-ppm-dap.md
```diff
@@ -995,7 +1002,6 @@ standard tokens for use in the "type" field:
 | invalidBatchSize            | There are an invalid number of reports in the batch. |
 | invalidAggregationParameter | The aggregation parameter assigned to a batch is invalid. |
 | batchMismatch               | Aggregators disagree on the report shares that were aggregated in a batch. |
-| unauthorizedRequest         | Authentication of an HTTP request failed (see {{request-authentication}}). |
```
tgeoghegan, line 998: This error type was never used, and besides, doing this at the DAP layer is inappropriate:

- in some deployments, authentication may happen before a request ever gets to a DAP implementation (e.g. in a load balancer or other proxy) so mandating a DAP specific problem document is a burden
- if you're using something like mTLS for auth, auth failures won't be represented as HTTP responses at all, much less a DAP specific problem document

draft-ietf-ppm-dap.md
```diff
@@ -1022,7 +1028,8 @@ refer to error types, rather than the full URNs. For example, an "error of type
 
 This document uses the verbs "abort" and "alert with \[some error message\]" to
 describe how protocol participants react to various error conditions. This
-implies HTTP status code 400 Bad Request unless explicitly specified otherwise.
+implies that the response's status code will indicate a client error unless
+specified otherwise.
```
tgeoghegan, line 1032: https://datatracker.ietf.org/doc/html/rfc9205#section-4.6-4

draft-ietf-ppm-dap.md
```diff
@@ -1034,13 +1041,11 @@ DAP has three major interactions which need to be defined:
 * The Collector collects aggregated results from the Aggregators, specified in
   {{collect-flow}}
 
-Each of these interactions is defined in terms of "resources". In this section
-we define these resources and the messages used to act on them.
+Each of these interactions is defined in terms of HTTP resources. In this
+section we define these resources and the messages used to act on them.
```
tgeoghegan, line 1045: Avoid scare quotes and make explicit reference to HTTP, which has a strong concept of a [resource](https://datatracker.ietf.org/doc/html/rfc9110#section-3.1).

draft-ietf-ppm-dap.md
```diff
@@ -1217,15 +1222,14 @@ conveyed from the Leader to the Helper when initializing aggregation jobs
 
 ## Resource URLs
 
-DAP is defined in terms of "resources", such as reports ({{upload-flow}}),
+DAP is defined in terms of HTTP resources, such as reports ({{upload-flow}}),
```
tgeoghegan, line 1225: Avoid scare quotes and make explicit reference to HTTP, which has a strong concept of a [resource](https://datatracker.ietf.org/doc/html/rfc9110#section-3.1).

draft-ietf-ppm-dap.md
```diff
@@ -1217,15 +1222,14 @@ conveyed from the Leader to the Helper when initializing aggregation jobs
...
-of string literals and variables. Variables are expanded into strings according
-to the following rules:
+Each resource has a URL. Resource URLs are specified as string literals
+containing variables. Variables are expanded into strings according to the
+following rules:
 
 * Variables `{leader}` and `{helper}` are replaced with the base API URL of the
-  Leader and Helper respectively (the base URLs are defined in
-  {{task-configuration}}).
+  Leader and Helper respectively.
```
tgeoghegan, line 1232: This is a bit of cleanup: "sequence of string literals" was misleading. I considered adding text about escaping `{` and `}` if they don't occur in a variable, but decided that we're not spelling out a general mechanism for variable expansion here. Rather we only spell out the expansion of five literals (`{leader}`, `{helper}`, `{task-id}`, `{aggregation-job-id}`, `{collection-job-id}`) and nothing else. If we want anything richer (or if something like httpdir or AD review asks for something better), then we should bite the bullet and do RFC 6570 templates, but I don't want to do something in between.

draft-ietf-ppm-dap.md
```diff
@@ -1270,14 +1274,14 @@ Before the Client can upload its report to the Leader, it must know the HPKE
 configuration of each Aggregator. See {{compliance}} for information on HPKE
 algorithm choices.
 
-Clients retrieve the HPKE configuration from each Aggregator by sending an HTTP
-GET request to `{aggregator}/hpke_config`.
+Clients retrieve the HPKE configuration from each Aggregator by sending a GET to
+`{aggregator}/hpke_config`.
```
tgeoghegan, line 1278: It's already clear we're talking about HTTP

draft-ietf-ppm-dap.md
```diff
@@ -1270,14 +1274,14 @@ Before the Client can upload its report to the Leader, it must know the HPKE
...
-An Aggregator responds to well-formed requests with HTTP status code 200 OK and
-an `HpkeConfigList` value, with media type "application/dap-hpke-config-list".
-The `HpkeConfigList` structure contains one or more `HpkeConfig` structures in
-decreasing order of preference. This allows an Aggregator to support multiple
-HPKE configurations simultaneously.
+An Aggregator responds with an `HpkeConfigList`, with media type
+"application/dap-hpke-config-list". The `HpkeConfigList` contains one or more
+`HpkeConfig`s in decreasing order of preference. This allows an Aggregator to
+support multiple HPKE configurations and multiple sets of algorithms
+simultaneously.
```
tgeoghegan, line 1284: Here I remove the specific prescription to respond with 200 OK. I think this is consistent with [RFC 9205](https://datatracker.ietf.org/doc/html/rfc9205#section-3.1-1). I'd also point to [ACME](https://datatracker.ietf.org/doc/html/rfc8555#section-7.1.1), which describes what resource you get from some request, but only mentions status codes in non-normative examples.

draft-ietf-ppm-dap.md
```diff
@@ -1291,44 +1295,70 @@ struct {
...
+
+Aggregators MUST allocate distinct `id` values for each `HpkeConfig` in an
 `HpkeConfigList`.
 
-The Client MUST abort if any of the following happen for any HPKE config
-request:
+The Client MUST abort if:
 
-* the GET request did not return a valid HPKE config list;
+* the response is not a valid HPKE config list;
```
tgeoghegan, line 1311: no need to repeat that it was a GET

draft-ietf-ppm-dap.md
```diff
@@ -1291,44 +1295,70 @@ struct {
...
 * the HPKE config list is empty; or
-* no HPKE config advertised by the Aggregator specifies a supported a KEM, KDF,
-  or AEAD algorithm triple.
-
-Aggregators SHOULD use HTTP caching to permit client-side caching of this
-resource {{!RFC5861}}. Aggregators SHOULD favor long cache lifetimes to avoid
-frequent cache revalidation, e.g., on the order of days. Aggregators can control
-this cached lifetime with the Cache-Control header, as in this example:
```
tgeoghegan, line 1313: I remixed the caching discussion: I think the appropriate reference is RFC 9111, no need to point to 5861. Further I moved the `Cache-Control` example into the example of an HPKE config request and response pair, below.

draft-ietf-ppm-dap.md
```diff
@@ -1291,44 +1295,70 @@ struct {
...
+    public_key = [0x01, 0x02, 0x03, 0x04, ...],
+  } HpkeConfig,
+  struct {
+    id = 17,
+    kem_id = 0x0020,
+    kdf_id = 0x0001,
+    aead_id = 0x0003,
+    public_key = [0x04, 0x03, 0x02, 0x01, ...],
+  } HpkeConfig,
+])
```
tgeoghegan, line 1354: Add an example per https://datatracker.ietf.org/doc/html/rfc9205#section-4.1-5

I had to come up with a syntax for examples of TLS presentation language structs, with things like an implied `encoded()` function and the `[0x01, 0x02, ...]` syntax for byte arrays. Since this text is non-normative, I think this is good enough. For precedent, ACME doesn't exhaustively spell out what `base64url()` does in [its examples](https://datatracker.ietf.org/doc/html/rfc8555#section-7.3).

In the example, we don't say "200 OK". This is nitty, but 9205 says not to include the reason phrase: https://datatracker.ietf.org/doc/html/rfc9205#section-4.6-7

draft-ietf-ppm-dap.md
```diff
@@ -1291,44 +1295,70 @@ struct {
...
-at least twice the cache lifetime in order to avoid rejecting reports.
-
 ### Upload Request
 
-Clients upload reports by using an HTTP POST to
-`{leader}/tasks/{task-id}/reports`. The payload is a `Report`, with media type
-"application/dap-report", structured as follows:
+Clients upload reports by sending a POST to `{leader}/tasks/{task-id}/reports`.
+The body is a `Report`, with media type "application/dap-report", structured as
+follows:
```
tgeoghegan, line 1361: no need to remind readers that POST is an HTTP request, and "body" is more appropriate than "payload"

draft-ietf-ppm-dap.md
```diff
@@ -1393,8 +1423,8 @@ of !VDAF}}), using the report ID as the nonce:
   nonce.
 
 * `rand` is a random byte string of length specified by the VDAF. Each report's
-  `rand` value MUST be independently sampled from a cryptographically secure
-  random number generator.
+  `rand` MUST be independently sampled from a cryptographically secure random
+  number generator.
```
tgeoghegan, line 1427: I scrubbed a few occurrences of the world "value" that aren't clarifying

draft-ietf-ppm-dap.md
```diff
@@ -1440,21 +1470,17 @@ struct {
 } InputShareAad;
 ~~~
 
-The Leader responds to well-formed requests with HTTP status code 201
```
tgeoghegan, line 1443: This nixes the specific prescription of 201 Created. See comment on the HPKE config section.

draft-ietf-ppm-dap.md
```diff
@@ -1488,13 +1514,43 @@ extensions with code points `23` and `42`, the "unsupported_extensions" member
...
+  } HpkeCiphertext,
+  helper_encrypted_input-share = struct {
+    config_id = 2,
+    enc = [0x0c, 0x0d, 0x0e, 0x0f, ...],
+    payload = [0x08, 0x00, 0x0a, 0x0b, ...],
+  } HpkeCiphertext,
+} Report)
+
+HTTP/1.1 200
+~~~
```
tgeoghegan, line 1553: Add an example per https://datatracker.ietf.org/doc/html/rfc9205#section-4.1-5

draft-ietf-ppm-dap.md
```diff
@@ -1440,21 +1470,17 @@ struct {
 } InputShareAad;
 ~~~
 
-The Leader responds to well-formed requests with HTTP status code 201
-Created. Malformed requests are handled as described in {{errors}}.
```
tgeoghegan, line 1444: We need to be more specific than this in describing how to handle malformed requests

draft-ietf-ppm-dap.md
```diff
@@ -1440,21 +1470,17 @@ struct {
...
-Clients SHOULD NOT upload the same measurement value in more than one report if
-the Leader responds with HTTP status code 201 Created.
```
tgeoghegan, line 1446: This unhelpfully restates the semantics of 201, plus we no longer specifically require that status anyway

draft-ietf-ppm-dap.md
```diff
@@ -1440,21 +1470,17 @@ struct {
...
+If the upload request is malformed, the Leader aborts with error
+`invalidMessage`.
 
-If the Leader does not recognize the task ID, then it MUST abort with error
-`unrecognizedTask`. Note that indicating an unrecognized task might indicate
-that the Client is misconfigured or the task has expired.
```
tgeoghegan, line 1450: I don't think the note is helpful, in the sense that it doesn't tell implementers what their implementation should do.

draft-ietf-ppm-dap.md
```diff
@@ -1488,13 +1514,43 @@ extensions with code points `23` and `42`, the "unsupported_extensions" member
...
 If the same extension type appears more than once among the public extensions
-and the private extensions in the Leader's input share, then the Leader MAY
-abort the upload request with error "invalidMessage".
-
-(Note that validation of extensions is not mandatory here because it requires
-the Leader to decrypt its input share. The Leader also cannot validate the
-Helper's extensions because it cannot decrypt the Helper's input share.
-Mandatory validation of extensions will occur during aggregation.)
+and the private extensions in the Leader's input share, then the Leader MUST
+ignore the report and MAY abort with error `invalidMessage`.
```
tgeoghegan, line 1518: Leader MAY abort, but it MUST ignore the report

draft-ietf-ppm-dap.md
```diff
@@ -1488,13 +1514,43 @@ extensions with code points `23` and `42`, the "unsupported_extensions" member
...
+
+Validation of anti-replay and extensions is not mandatory during the handling of
+upload requests to avoid blocking on storage transactions or decryption of input
+shares. The Leader also cannot validate the Helper's extensions because it
+cannot decrypt the Helper's input share. Validation of report IDs and extensions
+will occur before aggregation.
```
tgeoghegan, line 1524: don't note this, just say it

draft-ietf-ppm-dap.md
```diff
@@ -974,15 +982,14 @@ are supported by a participant is outside of this document's scope.
...
+errors (the 4xx or 5xx classes, respectively, from {{Section 15 of !RFC9110}})
+are permitted.
 
 When the server responds with an error status code, it SHOULD provide additional
-information using a problem detail object. If the response body does consist of
-a problem detail object, the HTTP status code MUST indicate a client or server
-error (the 4xx or 5xx classes, respectively, from {{Section 15 of !RFC9110}}).
+information using a problem detail object {{!RFC9457}} in the response body. If
+the response body does consist of a problem detail object, the status code MUST
+indicate a client or server error.
```
cjpatton, line 992: Unrelated to this PR, but I'm confused by "client or server error". What do we mean by this?

draft-ietf-ppm-dap.md
cjpatton: IIRC this is a relic of when we were doing HMAC in band.

draft-ietf-ppm-dap.md
cjpatton: Maybe "HPKE config list" -> "`HpkeConfigList`"

draft-ietf-ppm-dap.md
```diff
@@ -1295,44 +1299,70 @@ struct {
...
-  Cache-Control: max-age=86400
+* no HPKE config advertised by the Aggregator specifies a supported KEM, KDF and
+  AEAD algorithm triple.
+
+Aggregators SHOULD use caching to permit client-side caching of this resource
+{{!RFC9111}}. Aggregators SHOULD favor long cache lifetimes to avoid frequent
+cache revalidation, e.g., on the order of days. Aggregators can control this
+cached lifetime with the Cache-Control header.
+
+Servers should choose a Cache-Control value appropriate to the lifetime of their
```
cjpatton, line 1325: Less squishy
```suggestion
Servers choose a Cache-Control value appropriate to the lifetime of their
```

draft-ietf-ppm-dap.md
```diff
@@ -1295,44 +1299,70 @@ struct {
...
+keys. Clients SHOULD follow the usual caching semantics {{!RFC9111}} for HPKE
+configurations.
+
+Long cache lifetimes may result in Clients using stale HPKE configurations.
```
cjpatton, line 1329: Unrelated to this PR, but it's not clear to me how a long cache lifetime would result in stale configs being used.

draft-ietf-ppm-dap.md
```diff
@@ -1444,21 +1474,17 @@ struct {
...
-The Leader responds to requests whose Leader encrypted input share uses an
-out-of-date or unknown `HpkeConfig.id` value, indicated by
-`HpkeCiphertext.config_id`, with error of type 'outdatedConfig'. When the Client
-receives an 'outdatedConfig' error, it SHOULD invalidate any cached
-HpkeConfigList and retry with a freshly generated Report. If this retried upload
-does not succeed, the Client SHOULD abort and discontinue retrying.
+If the Leader does not recognize the `config_id` in the encrypted input share,
+it aborts with an error of type `outdatedConfig`. When the Client receives an
+`outdatedConfig` error, it SHOULD invalidate any cached `HpkeConfigList` and
+retry with a freshly generated Report. If this retried upload does not succeed,
```
cjpatton, line 1486:
```suggestion
retry with a freshly generated `Report`. If this retried upload does not succeed,
```

cjpatton (commented):
Looks good overall, just some minor comments/clarifications.

draft-ietf-ppm-dap.md
tgeoghegan: HTTP 4xx or 5xx. This is explained in the paragraph above with a reference to RFC 9110.

draft-ietf-ppm-dap.md
tgeoghegan: Yeah, since this is already in the context of the SHOULD in the paragraph above, I agree!

draft-ietf-ppm-dap.md
tgeoghegan: The case that comes to mind is if an aggregator rotates its HPKE config(s) ahead of the lifetime implied by `Cache-Control`. But that sounds like dubious practice on the part of an Aggregator: either the rotation period should be very close to the cache lifetime, or the key was rotated early because of a known compromise, in which case the server shouldn't honor messages encrypted to it anymore.

I'd be OK with deleting this sentence. What really matters is the following one about how long to accept reports with old keys.

cjpatton: Agreed. Let's delete the sentence.

cjpatton (approved):
One unresolved comment, but we agree on how to resolve it, so here's a stamp in advance!

PR #675, opened 2025-03-21
Align collection jobs, aggregate shares w RFC 9205

tgeoghegan:
- Allow either resource to be handled sync or async
- use empty response body to indicate deferred handling
- add aggregate share ID and make the request a PUT
- define DELETE semantics
- add examples of HTTP request/responses

### DIFF
--- a/draft-ietf-ppm-dap.md
+++ b/draft-ietf-ppm-dap.md
@@ -1234,11 +1234,12 @@ following rules:
 
 * Variables `{leader}` and `{helper}` are replaced with the base API URL of the
   Leader and Helper respectively.
-* Variables `{task-id}`, `{aggregation-job-id}`, and `{collection-job-id}` are
-  replaced with the task ID ({{task-configuration}}), aggregation job ID
-  ({{agg-init}}), and collection job ID ({{collect-init}}) respectively. The
-  value MUST be encoded in its URL-safe, unpadded Base 64 representation as
-  specified in {{Sections 5 and 3.2 of !RFC4648}}.
+* Variables `{task-id}`, `{aggregation-job-id}`, `{aggregate-share-id}`, and
+  `{collection-job-id}` are replaced with the task ID ({{task-configuration}}),
+  aggregation job ID ({{agg-init}}), aggregate share ID ({{collect-aggregate}})
+  and collection job ID ({{collect-init}}) respectively. The value MUST be
+  encoded in its URL-safe, unpadded Base 64 representation as specified in
+  {{Sections 5 and 3.2 of !RFC4648}}.
 
 For example, given a helper URL "https://example.com/api/dap", task ID "f0 16 34
 47 36 4c cf 1b c0 e3 af fc ca 68 73 c9 c3 81 f6 4a cd f9 02 06 62 f8 3f 46 c0 72
@@ -1787,8 +1788,6 @@ Once all the report shares have been initialized, the Leader creates an
 `AggregationJobInitReq` message.
 
 ~~~ tls-presentation
-opaque BatchID[32];
-
 struct {
   BatchMode batch_mode;
   opaque config<0..2^16-1>;
@@ -2048,22 +2047,22 @@ response body and sets a Location header field set to the relative reference
 `/tasks/{task-id}/aggregation_jobs/{aggregation-job-id}?step=0`. The response
 SHOULD include a Retry-After header field to suggest a polling interval to the
 Leader. The Leader then polls the state of the job by sending GET requests to the
-resolved URL. The Helper responds the same way until the job is ready, at which
-point it responds with the encoded `AggregationJobResp`.
+resolved URL. The Helper responds the same way until the job is ready, from
+which point it responds with the encoded `AggregationJobResp`.
 
 The Helper MAY also wait to respond to the Leader's PUT until the aggregation
 job initialization is ready. In this case, it responds with the encoded
 `AggregationJobResp`.
 
-Changing an aggregation job's parameters is illegal, so further PUT requests to
-the aggregation job with a different `AggregationJobInitReq` in the body MUST
-fail with a client error status code. For further requests with the same
-`AggregationJobInitReq` in the body, the Helper SHOULD respond as it did for the
-original `AggregationJobInitReq`.
+Changing an aggregation job's parameters is illegal. If the Helper receives
+further PUT requests to the aggregation job with a different
+`AggregationJobInitReq` in the body, it MUST abort with a client error. For
+further requests with the same `AggregationJobInitReq` in the body, the Helper
+SHOULD respond as it did for the original `AggregationJobInitReq`.
 
-It is illegal to rewind or reset the state of an aggregation job. Once an
-aggregation job has been continued at least once (see {{agg-continue-flow}}),
-further requests to initialize that aggregation job MUST fail.
+It is illegal to rewind or reset the state of an aggregation job. If the Helper
+receives requests to initialize an aggregation job once it has been continued at
+least once (see {{agg-continue-flow}}), it MUST abort with a client error.
 
 #### Input Share Decryption {#input-share-decryption}
 
@@ -2167,7 +2166,7 @@ encoded(struct {
     config = encoded(struct {
       batch_id = [0x1f, 0x1e, ..., 0x00],
     } LeaderSelectedPartialBatchSelectorConfig),
-  },
+  } PartialBatchSelector,
   prepare_inits,
 } AggregationJobInitReq)
 
@@ -2427,8 +2426,8 @@ response body and sets the Location header field to the relative reference
 `step` is set to `AggregationJobContinueReq.step`. The response SHOULD include a
 Retry-After header field to suggest a polling interval to the Leader. The Leader
 then polls the state of the job by sending GET requests to the resolved URL. The
-Helper responds the same way until the job is ready, at which point it responds
-with the encoded `AggregationJobResp`.
+Helper responds the same way until the job is ready, from which point it
+responds with the encoded `AggregationJobResp`.
 
 The Helper MAY also wait to respond to the Leader's POST until the aggregation
 job continuation is ready. In this case, it responds with the encoded
@@ -2578,13 +2577,11 @@ HTTP/1.1 200
 
 ## Collecting Results {#collect-flow}
 
-In this phase, the Collector requests aggregate shares from each Aggregator and
-then locally combines them to yield a single aggregate result. In particular,
-the Collector issues a query to the Leader ({{batch-modes}}), which the
-Aggregators use to select a batch of reports to aggregate. Each Aggregator emits
-an aggregate share encrypted to the Collector so that it can decrypt and combine
-them to yield the aggregate result. This process is composed of two
-interactions:
+The Collector initiates this phase with a query to the Leader ({{batch-modes}}),
+which the Aggregators use to select a batch of reports to aggregate. Each
+Aggregator emits an aggregate share encrypted to the Collector so that it can
+decrypt and combine them to yield the aggregate result. This process is referred
+to as a "collection job" and is composed of two interactions:
 
 1. Collect request and response between the Collector and Leader, specified in
    {{collect-init}}
@@ -2594,22 +2591,23 @@ interactions:
 Once complete, the Collector computes the final aggregate result as specified in
 {{collect-finalization}}.
 
-This overall process is referred to as a "collection job".
-
-### Collection Job Initialization {#collect-init}
-
-First, the Collector chooses a collection job ID:
+Collection jobs are identified by a 16-byte job ID:
 
 ~~~ tls-presentation
 opaque CollectionJobID[16];
 ~~~
 
-This ID value MUST be unique within the scope of the corresponding DAP task.
+A collection job is an HTTP resource served by the Leader at the URL
+`{leader}/tasks/{task-id}/collection_jobs/{collection-job-id}`.
 
-To initiate the collection job, the collector issues a PUT request to
-`{leader}/tasks/{task-id}/collection_jobs/{collection-job-id}`. The body of the
-request has media type "application/dap-collection-job-req", and it is structured
-as follows:
+### Collection Job Initialization {#collect-init}
+
+First, the Collector chooses a collection job ID, which MUST be unique within
+the scope of the corresponding DAP task.
+
+To initiate the collection job, the Collector issues a PUT request to the
+collection job with media type "application/dap-collection-job-req", and a body
+structured as follows:
 
 ~~~ tls-presentation
 struct {
@@ -2623,24 +2621,11 @@ struct {
 } CollectionJobReq;
 ~~~
 
-The named parameters are:
-
 * `query`, the Collector's query. The content of this field depends on the
-  indicated batch mode: for time-interval mode, the `config` field contains the
-  requested time interval ({{time-interval-batch-mode}}); for leader-selected
-  mode, the `config` field is empty, and the Collector gets the aggregate
-  result for the next batch selected by the Leader
-  ({{leader-selected-batch-mode}}).
-
-  Batch modes defined by future documents MUST specify the content of this field;
-  see {{extending-this-doc}} for details.
-
-  The indicated batch mode MUST match the task's batch mode. Otherwise, the
-  Leader MUST abort with error "invalidMessage".
+  indicated batch mode ({{batch-modes}}).
 
 * `agg_param`, an aggregation parameter for the VDAF being executed. This is
-  the same value as in `AggregationJobInitReq` (see {{leader-init}}). The
-  aggregation parameter MUST be valid as defined in {{agg-param-validation}}.
+  the same value as in `AggregationJobInitReq` (see {{leader-init}}).
 
 Collectors MUST authenticate their requests to Leaders using a scheme that meets
 the requirements in {{request-authentication}}.
@@ -2653,58 +2638,75 @@ running aggregation jobs ({{aggregate-flow}}) until the Collector initiates a
 collection job. This is because, in general, the aggregation parameter is not
 known until this point. In certain situations it is possible to predict the
 aggregation parameter in advance. For example, for Prio3 the only valid
-aggregation parameter is the empty string. For these reasons, the collection job
-is handled asynchronously. If aggregation is performed eagerly, the Leader MUST
-validate that the aggregation parameter received in the `CollectionJobReq`
-matches the aggregation parameter used in aggregations.
+aggregation parameter is the empty string.
 
-Upon receipt of a `CollectionJobReq`, the Leader begins by checking that it
-recognizes the task ID in the request path. If not, it MUST abort with error
+The Leader MAY defer handling the collection request. In this case, it indicates
+that the collection job is not yet ready by immediately sending an empty
+response body. The Leader SHOULD include a Retry-After header field to suggest a
+polling interval to the Collector. The Collector then polls the state of the job
+by sending GET requests to the collection job. The Collector SHOULD use each
+response's Retry-After header field to decide when to try again. The Leader
+responds the same way until either the job is ready, from which point it
+responds with a `CollectionJobResp` (defined below), or the job fails, from
+which point the Leader MUST abort with the error that caused the failure.
+
+The Leader MAY instead handle the request immediately. It waits to respond to
+the Collector's PUT until the collection job is ready, in which case it responds
+with the `CollectionJobResp`, or the job fails, in which case the Leader MUST
+abort with the error that caused the failure.
+
+The Leader begins handling a `CollectionJobReq` by checking that it recognizes
+the task ID. If not, it MUST fail the collection job with error
 `unrecognizedTask`.
 
+Next, the Leader checks that the indicated batch mode matches the task's batch
+mode. If not, it MUST fail the job with error `invalidMessage`.
+
 Next, the Leader checks that the aggregation parameter is valid as described in
-{{agg-param-validation}}. If the aggregation parameter is invalid, then the
-Leader MUST abort with error `invalidAggregationParameter`.
+{{agg-param-validation}}. If not, it MUST fail the job with error
+`invalidAggregationParameter`.
 
-The Leader MAY further validate the request according to the requirements in
-{{batch-validation}} and abort with the indicated error, though some conditions
-such as the number of valid reports may not be verifiable while handling the
-`CollectionJobReq` message, and the batch will have to be re-validated later on
-regardless.
+If aggregation is performed eagerly, then the Leader checks that the aggregation
+parameter received in the `CollectionJobReq` matches the aggregation parameter
+used in each aggregation job pertaining to the batch. If not, the Leader MUST
+fail the job with error `invalidMessage`.
 
-Changing a collection job's parameters is illegal, so further requests to
-`PUT /tasks/{task-id}/collection_jobs/{collection-job-id}` for the same
-`collection-job-id` but with a different `CollectionJobReq` in the body MUST
-fail with an HTTP client error status code.
+Having validated the `CollectionJobReq`, the Leader begins working with the
+Helper to aggregate the reports satisfying the query (or continues this process,
+depending on whether the Leader is aggregating eagerly) as described in
+{{aggregate-flow}}.
 
-The Leader responds to `CollectionJobReq`s with a `CollectionJobResp`, which is
-structured as follows:
+If the Leader has a pending aggregation job that overlaps with the batch for the
+collection job, the Leader MUST first complete the aggregation job before
+proceeding and requesting an aggregate share from the Helper. This avoids a race
+condition between aggregation and collection jobs that can yield batch mismatch
+errors.
+
+The Leader then checks whether it can construct a valid batch for the collection
+job by applying the requirements in {{batch-validation}}. If so, then the Leader
+obtains the Helper's aggregate share following the aggregate-share request flow
+described in {{collect-aggregate}}. If not, it either fails the collection job
+with error `invalidBatch` or tries again later, depending on which requirement
+in {{batch-validation}} was not met.
+
+If obtaining the aggregate share fails, then the Leader MUST fail the collection
+job with the error that caused the failure.
+
+Once the Leader has the Helper's aggregate share and has computed its own, the
+collection job is ready. Its results are represented by a `CollectionJobResp`,
+which is structured as follows:
 
 ~~~ tls-presentation
-enum {
-  processing(0),
-  ready(1),
-  (255)
-} CollectionJobStatus;
-
 struct {
   PartialBatchSelector part_batch_selector;
   uint64 report_count;
   Interval interval;
   HpkeCiphertext leader_encrypted_agg_share;
   HpkeCiphertext helper_encrypted_agg_share;
-} Collection;
-
-struct {
-  CollectionJobStatus status;
-  select (CollectionJob.status) {
-    case processing: Empty;
-    case ready:      Collection;
-  }
 } CollectionJobResp;
 ~~~
 
-The body's media type is "application/dap-collection-job-resp". The `Collection`
+A `CollectionJobResp`'s media type is "application/dap-collection-job-resp". The
 structure includes the following:
 
 * `part_batch_selector`: Information used to bind the aggregate result to the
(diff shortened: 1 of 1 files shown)


draft-ietf-ppm-dap.md
```diff
@@ -2789,16 +2860,26 @@ of this task:
...
+~~~ tls-presentation
+opaque AggregateShareID[16];
+~~~
+
+The Helper's aggregate share is an HTTP resource served by the Helper at the URL
+`{helper}/tasks/{task-id}/aggregate_shares/{aggregate-share-id}`. To obtain it,
+the Leader first chooses an aggregate share ID, which MUST be unique within the
+scope of the corresponding DAP task.
+
+Then the Leader sends a PUT request to the aggregate share with the body:
```
tgeoghegan, line 2882: The aggregate share ID is new, and we need it so that the leader can poll on something for asynchronous agg share handling. The alternative would be to not specify an aggregate share ID and let the helper choose a unique URL for the aggregate share, but we chose long ago to have HTTP clients choose resource identifiers (i.e. leader chooses agg job and agg share IDs, collector chooses collection job IDs).

draft-ietf-ppm-dap.md
```diff
@@ -2814,26 +2895,13 @@ struct {
...
-The media type of the request is "application/dap-aggregate-share-req". The
-structure contains the following parameters:
+The media type of `AggregateShareReq` is "application/dap-aggregate-share-req".
+The structure contains the following parameters:
 
 * `batch_selector`: The "batch selector", the contents of which depends on the
-  indicated batch mode: for time-interval mode, the `config` field contains the
-  time interval selected by the Collector ({{time-interval-batch-mode}}); in
-  leader-selected mode, the field contains the batch ID selected by the Leader
-  ({{leader-selected-batch-mode}}).
```
tgeoghegan, line 2824: Defer discussion of batch modes to {{batch-modes}}

draft-ietf-ppm-dap.md
```diff
@@ -2814,26 +2895,13 @@ struct {
...
-
-  Future documents that new batch modes MUST specify the contents of the
-  `config` field; see {{extending-this-doc}} for details.
-
-  The indicated batch mode MUST match the task's batch mode. Otherwise, the
-  Helper MUST abort with "invalidMessage".
```
tgeoghegan, line 2830: Put all the validation MUSTs in one place, below

draft-ietf-ppm-dap.md
```diff
@@ -2843,28 +2911,30 @@ structure contains the following parameters:
...
+
+The indicated batch mode MUST match the task's batch mode. If not, the Helper
+MUST abort with `invalidMessage`.
+
+The Helper then verifies that the request meets the requirements in
+{{batch-validation}}. If not, it aborts with the indicated error.
+
+The aggregation parameter MUST match the aggregation parameter used in
+aggregation jobs pertaining to this batch. If not, the Helper aborts with
+`invalidMessage`.
```
tgeoghegan, line 2925: Break each validation MUST into its own paragraph for clarity.

draft-ietf-ppm-dap.md
```diff
@@ -2843,28 +2911,30 @@ structure contains the following parameters:
...
+the values provided in the `AggregateShareReq`, it MUST abort with an error of
+type `batchMismatch`.
 
 The Helper then encrypts `agg_share` under the Collector's HPKE public key as
 described in {{aggregate-share-encrypt}}, yielding `encrypted_agg_share`.
-Encryption prevents the Leader from learning the actual result, as it only has
-its own aggregate share and cannot compute the Helper's.
 
-The Helper responds to the Leader with HTTP status code 200 OK and a body
-consisting of an `AggregateShare`, with media type
```
tgeoghegan, line 2867: Don't prescribe HTTP status

draft-ietf-ppm-dap.md
```diff
@@ -2843,28 +2911,30 @@ structure contains the following parameters:
...
+request using the same process used by the Leader (described at the beginning of
+this section), arriving at its aggregate share, report count, and checksum
+values. If the Helper's computed report count and checksum values do not match
+the values provided in the `AggregateShareReq`, it MUST abort with an error of
+type `batchMismatch`.
 
 The Helper then encrypts `agg_share` under the Collector's HPKE public key as
 described in {{aggregate-share-encrypt}}, yielding `encrypted_agg_share`.
-Encryption prevents the Leader from learning the actual result, as it only has
-its own aggregate share and cannot compute the Helper's.
```
tgeoghegan, line 2864: I deleted this for brevity but could be convinced to put it back.

draft-ietf-ppm-dap.md
```diff
@@ -2878,33 +2948,140 @@ struct {
 computed above and `encrypted_aggregate_share.ciphertext` is the ciphertext
 `encrypted_agg_share` computed above.
 
-The Helper's handling of this request MUST be idempotent. That is, if multiple
-identical, valid `AggregateShareReq`s are received, they should all yield the
-same response.
```
tgeoghegan, line 2883: The request is now a PUT, which implies idempotence in HTTP, so we don't need this remark

draft-ietf-ppm-dap.md
```diff
@@ -2878,33 +2948,140 @@ struct {
...
+`AggregateShare`. The Leader SHOULD use each response's Retry-After header field
+to decide when to try again.
+
+The Helper MAY also block responding to the Leader's PUT until the aggregate
+share is ready. In this case, it responds with the `AggregateShare`.
 
-After receiving the Helper's response, the Leader uses the HpkeCiphertext to
-finalize a collection job (see {{collect-finalization}}).
+After receiving the Helper's response, the Leader includes the HpkeCiphertext in
+its response to the Collector (see {{collect-finalization}}).
```
tgeoghegan, line 2964: I found it awkward to imply that the Leader uses the Helper's agg share for anything, as well as the implication that the leader finalizes the collection job. Seems like that happens when the collector unshards the agg shares.

draft-ietf-ppm-dap.md
```diff
@@ -2925,11 +3102,13 @@ done as follows:
...
-where `pk` is the HPKE public key encoded by the Collector's HPKE key,
-`server_role` is the Role of the encrypting server (`0x02` for the Leader and
-`0x03` for a Helper), `0x00` represents the Role of the recipient (always the
-Collector), and `agg_share_aad` is a value of type `AggregateShareAad`. The
-`SealBase()` function is as specified in {{!HPKE, Section 6.1}} for the
+* `pk` is the Collector's HPKE public key
+* `server_role` is the Role of the encrypting server (`0x02` for the Leader and
+  `0x03` for a Helper)
+* `0x00` represents the Role of the recipient (always the Collector)
+* `agg_share_aad` is an `AggregateShareAad` (defined below).
```
tgeoghegan, line 3109: explicitly enumerate the variables for clarity.

draft-ietf-ppm-dap.md
```diff
@@ -2979,25 +3160,22 @@ ciphersuite indicated by the HPKE configuration.
...
-Next, the Aggregator checks that batch contains a valid number of reports. The
-Aggregator checks that `len(X) >= min_batch_size`, where `X` is the set of
-reports successfully aggregated into the batch and `min_batch_size` is the
-minimum batch size for the task. If this check fails, then Helpers MUST abort
-with an error of type "invalidBatchSize". Leaders SHOULD wait for more reports
-to be validated and try the collection job again later.
+Next, the Aggregator checks that the number of reports in the batch is equal to
+or greater than the task's minimum batch size. If not, then the Helper MUST
+abort with error `invalidBatchSize`. The Leader SHOULD wait for more reports to
+be validated and try the collection job again later.
```
tgeoghegan, line 3175: I think that explaining the comparison to minimum batch size is perfectly clear and more concise than the IMO awkward definition of `X`.

draft-ietf-ppm-dap.md
```diff
@@ -2979,25 +3160,22 @@ ciphersuite indicated by the HPKE configuration.
...
 
-Next, the Aggregator checks if the set of batch buckets identified by the
-request overlaps with the batch buckets that have already been collected. If
-so, it MUST abort with "batchOverlap".
+Next, the Aggregator checks if any of the batch buckets identified by the
+request have already been collected. If so, it MUST abort with "batchOverlap".
```
tgeoghegan, line 3178: *Batches* (which consist of one or more batch buckets) can overlap but I'm pretty sure *batch buckets* can't.

draft-ietf-ppm-dap.md
```diff
@@ -3030,20 +3208,36 @@ Each batch mode specifies the following:
...
+  struct {
+    start = 1659546000,
+    duration = 1000,
+  } Interval,
+  struct {
+    start = 1659547000,
+    duration = 1000,
+  } Interval
+]
+~~~
```
tgeoghegan, line 3236: This is longer than the existing example of batch intervals, but uses a syntax we already introduced. I can go either way with it.

draft-ietf-ppm-dap.md
```diff
@@ -3077,60 +3271,67 @@ where `batch_interval` is the batch interval requested by the Collector.
 
 ### Batch Buckets {#time-interval-batch-buckets}
 
-Each batch bucket is identified by an interval of time whose beginning is a
-multiple of the task's `time_precision` and whose duration is equal to the
+Each batch bucket is identified by an `Interval` whose duration is equal to the
```
tgeoghegan, line 3274: Use the `Interval` type so that we implicitly get the definition in terms of `time_precision` as well as the comparison and inclusion semantics.

draft-ietf-ppm-dap.md
```diff
@@ -3077,60 +3271,67 @@ where `batch_interval` is the batch interval requested by the Collector.
...
+available batch. The Leader selects a recent batch to aggregate which MUST not
+yet have been associated with a collection job.
 
 The Aggregators can output batches of any size that is larger than or equal to
-`min_batch_size`. The target batch size, if any, is implementation-specific, and
-may be equal to or greater than the minimum batch size. Deciding how soon
-batches should be output is also implementation-specific. Exactly sizing batches
-may be challenging for Leader deployments in which multiple, independent nodes
-running the aggregate interaction (see {{aggregate-flow}}) need to be
-coordinated.
```
tgeoghegan, line 3129: Seems unhelpful to me to say that this part of the protocol is hard without saying what to do about it.

draft-ietf-ppm-dap.md
```diff
@@ -3077,60 +3271,67 @@ where `batch_interval` is the batch interval requested by the Collector.
...
+    start = batch_interval.start + time_precision,
+    duration = time_precision,
+  } Interval,
+  ...
+  struct {
+    start = batch_interval.start + batch_interval.duration - time_precision,
+    duration = time_precision,
+  } Interval
+]
+~~~
```
tgeoghegan, line 3310: As above, we use an existing syntax to represent these intervals, at the cost of more vertical space.

draft-ietf-ppm-dap.md
```diff
@@ -2623,24 +2621,11 @@ struct {
...
-  mode, the `config` field is empty, and the Collector gets the aggregate
-  result for the next batch selected by the Leader
-  ({{leader-selected-batch-mode}}).
-
-  Batch modes defined by future documents MUST specify the content of this field;
-  see {{extending-this-doc}} for details.
-
-  The indicated batch mode MUST match the task's batch mode. Otherwise, the
-  Leader MUST abort with error "invalidMessage".
+  indicated batch mode ({{batch-modes}}).
```
cjpatton, line 2625: Please double check that {{batch-modes}} has a reference to {{extending-this-doc}}.

draft-ietf-ppm-dap.md
```diff
@@ -2653,59 +2638,57 @@ running aggregation jobs ({{aggregate-flow}}) until the Collector initiates a
...
 
 Next, the Leader checks that the aggregation parameter is valid as described in
-{{agg-param-validation}}. If the aggregation parameter is invalid, then the
-Leader MUST abort with error `invalidAggregationParameter`.
+{{agg-param-validation}}. If not, then the Leader MUST abort with error
+`invalidAggregationParameter`.
+
+If aggregation is performed eagerly, the Leader MUST validate that the
+aggregation parameter received in the `CollectionJobReq` matches the aggregation
+parameter used in aggregations.
```
cjpatton, line 2655:
```suggestion
If aggregation is performed eagerly, the Leader MUST validate that the
aggregation parameter received in the `CollectionJobReq` matches the aggregation
parameter used in each aggregation job pertaining to the batch.
```

Also, what causes this check to fail? Would this be a internal error on the part of the Leader?

draft-ietf-ppm-dap.md
```diff
@@ -2653,59 +2638,57 @@ running aggregation jobs ({{aggregate-flow}}) until the Collector initiates a
...
+
+Changing a collection job's parameters is illegal, so further PUT requests to
+the collection job with a different `CollectionJobReq` MUST fail.
```
cjpatton, line 2658: Let's replace "MUST fail" with a specific action, like an abort.

draft-ietf-ppm-dap.md
```diff
@@ -2653,59 +2638,57 @@ running aggregation jobs ({{aggregate-flow}}) until the Collector initiates a
...
+
+After receiving the `CollectionJobReq`, the Leader begins working with the
+Helper to aggregate the reports satisfying the query (or continues this process,
+depending on the VDAF) as described in {{aggregate-flow}}.
```
cjpatton, line 2662:
```suggestion
depending on whether the Leader is aggregating eagerly) as described in {{aggregate-flow}}.
```

draft-ietf-ppm-dap.md
```diff
@@ -2653,59 +2638,57 @@ running aggregation jobs ({{aggregate-flow}}) until the Collector initiates a
...
-The Leader MAY further validate the request according to the requirements in
-{{batch-validation}} and abort with the indicated error, though some conditions
-such as the number of valid reports may not be verifiable while handling the
-`CollectionJobReq` message, and the batch will have to be re-validated later on
-regardless.
+The Leader first checks whether it can construct a valid batch for the
+collection job by applying the requirements in {{batch-validation}}. If so, then
+the Leader obtains the Helper's aggregate share following the aggregate-share
+request flow described in {{collect-aggregate}}. If not, it either aborts the
+collection job or tries again later, depending on which requirement in
```
cjpatton, line 2668: By "abort the collection job" do you mean abort the PUT request from the Collector?

draft-ietf-ppm-dap.md
```diff
@@ -2653,59 +2638,57 @@ running aggregation jobs ({{aggregate-flow}}) until the Collector initiates a
...
+{{batch-validation}} was not met.
 
-Changing a collection job's parameters is illegal, so further requests to
-`PUT /tasks/{task-id}/collection_jobs/{collection-job-id}` for the same
-`collection-job-id` but with a different `CollectionJobReq` in the body MUST
-fail with an HTTP client error status code.
+If the Leader has a pending aggregation job that overlaps with the batch and
+aggregation parameter for the collection job, the Leader MUST first complete the
```
cjpatton, line 2672: Multi-collection is no longer possible, so I actually think it's unnecessary to say this.
```suggestion
If the Leader has a pending aggregation job that overlaps with the batch for the collection job, the Leader MUST first complete the
```

draft-ietf-ppm-dap.md
```diff
@@ -2653,59 +2638,57 @@ running aggregation jobs ({{aggregate-flow}}) until the Collector initiates a
...
+aggregation job before proceeding and requesting an aggregate share from the
+Helper. This avoids a race condition between aggregation and collection jobs
+that can yield batch mismatch errors.
 
-The Leader responds to `CollectionJobReq`s with a `CollectionJobResp`, which is
+A collection job's results are represented by a `Collection`, which is
```
cjpatton, line 2677: I think I prefer `CollectionJobResp`, for symmetry with aggregation jobs.

draft-ietf-ppm-dap.md
```diff
@@ -2653,59 +2638,57 @@ running aggregation jobs ({{aggregate-flow}}) until the Collector initiates a
...
-  select (CollectionJob.status) {
-    case processing: Empty;
-    case ready:      Collection;
-  }
-} CollectionJobResp;
 ~~~
 
-The body's media type is "application/dap-collection-job-resp". The `Collection`
-structure includes the following:
+A `Collection`'s media type is "application/dap-collection". The structure
```
cjpatton, line 2690: Likewise here: dap-collection-job-resp might be better than dap-collection

draft-ietf-ppm-dap.md
```diff
@@ -2725,56 +2708,144 @@ structure includes the following:
 * `helper_encrypted_agg_share`: The Helper's aggregate share, encrypted to the
   Collector (see {{aggregate-share-encrypt}}).
```
cjpatton, line 2710: Self-note: got here

draft-ietf-ppm-dap.md
tgeoghegan: It does, in its very first paragraph.

draft-ietf-ppm-dap.md
tgeoghegan: It's hard to say. If aggregation was performed eagerly, then either:

- the aggregation parameter for the VDAF is the unit type (`()` in Rust), like for Prio3
- the VDAF does have a non-unit aggregation parameter, but for some reason (special knowledge at the application layer above DAP) its value is known to the Leader such that it could perform eager aggregation.

In the first case, it's not possible for the aggregation parameter values to disagree because there's no aggregation parameter values.

In the second case, I guess it would make sense for the Leader to check that the aggregation parameter it gets from the Collector matches what it guessed when it ran the aggregation jobs.

That's me reaching for a reason to have this check, though.

draft-ietf-ppm-dap.md
tgeoghegan: Sure. I made a similar change to the equivalent text in the aggregation job init section.

draft-ietf-ppm-dap.md
tgeoghegan: The former. The Leader's validation of  doesn't necessarily happen synchronously during handling of the PUT from the Collector. So I think we should avoid the verb "abort" here. I'll be more explicit here and reuse the language from later where we describe what to do if getting the aggregate share fails.

draft-ietf-ppm-dap.md
cjpatton: I think validation makes sense, but let's say waht happens when the check fails. I.e., instead of "MUST validate" write "if validation fails, then MUST abort with error BLAH".

draft-ietf-ppm-dap.md
```diff
@@ -2653,58 +2638,60 @@ running aggregation jobs ({{aggregate-flow}}) until the Collector initiates a
...
+{{agg-param-validation}}. If not, then the Leader MUST abort with error
+`invalidAggregationParameter`.
+
+If aggregation is performed eagerly, the Leader MUST validate that the
+aggregation parameter received in the `CollectionJobReq` matches the aggregation
+parameter used in aggregations.
+
+Changing a collection job's parameters is illegal, so if there are further PUT
+requests to the collection job with a different `CollectionJobReq`, the Leader
+MUST abort with a client error.
```
cjpatton, line 2659: What is a "client error"?

tgeoghegan: HTTP status in the 400 range. I think it's clear enough from the context that we're discussing HTTP requests, and we lay out what it means to "abort with..." in [the section on errors](https://ietf-wg-ppm.github.io/draft-ietf-ppm-dap/draft-ietf-ppm-dap.html#section-3.4).

cjpatton: I guess what I'm asking for is to be prescriptive about what error to use from that table.

tgeoghegan: OK, I think that's fair. I also need to do a pass here to make sure we distinguish clearly between those conditions that get checked during the handling of the PUT request regardless of whether the job will be run asynchronously (so stuff like "is the message well-formed" that's cheap) vs. checks that you'd want to defer (so stuff like "is the right agg param used" that requires fetching a bunch of data from storage).

draft-ietf-ppm-dap.md
```diff
@@ -3025,25 +3216,41 @@ Each batch mode specifies the following:
...
-`(1659546000, 1000)`, `(1659547000, 1000)` satisfies these conditions. (The
-first element of the pair denotes the start of the batch interval and the second
-denotes the duration.) However, this is not a requirement--the Collector may
-decide to issue queries out-of-order. In addition, the Collector may need to
-vary the duration to adjust to changing report upload rates.
+reports are grouped by an interval of time. The Collector specifies a "batch
+interval" into which report timestamps must fall.
+
+The Collector can issue queries whose batch intervals are continuous,
+monotonically increasing, and have the same duration. For example:
```
cjpatton, line 3228:
```suggestion
monotonically increasing, and have the same duration. For example, the following sequence of batch intervals satisfies these conditions
```

draft-ietf-ppm-dap.md
```diff
@@ -3025,25 +3216,41 @@ Each batch mode specifies the following:
...
+    duration = 1000,
+  } Interval,
+  struct {
+    start = 1659546000,
+    duration = 1000,
+  } Interval,
+  struct {
+    start = 1659547000,
+    duration = 1000,
+  } Interval
```
cjpatton, line 3247: unless this would be inconsistent
```suggestion
  } Interval,
```

draft-ietf-ppm-dap.md
```diff
@@ -3077,60 +3284,67 @@ where `batch_interval` is the batch interval requested by the Collector.
...
+  start = 1729629000,
+  duration = 1000,
+} Interval
+~~~
 
-* both `batch_interval.start` and `batch_interval.duration` are divisible by
-  `time_precision`
+The `Query` received by the Leader or `BatchSelector` received by the Helper
+determines a valid set of batch bucket identifiers if
+`batch_interval.duration >= time_precision`.
```
cjpatton, line 3302: Is `batch_interval` defined?

draft-ietf-ppm-dap.md
```diff
@@ -3077,60 +3284,67 @@ where `batch_interval` is the batch interval requested by the Collector.
...
+  } Interval,
+  struct {
+    start = batch_interval.start + time_precision,
+    duration = time_precision,
+  } Interval,
+  ...
+  struct {
+    start = batch_interval.start + batch_interval.duration - time_precision,
+    duration = time_precision,
+  } Interval
```
cjpatton, line 3321:
```suggestion
  } Interval,
```

draft-ietf-ppm-dap.md
```diff
@@ -3077,60 +3284,67 @@ where `batch_interval` is the batch interval requested by the Collector.
...
+opaque BatchID[32];
+~~~
 
 The Collector will not know the set of batch IDs available for collection. To
-get the aggregate of a batch, the Collector issues a query, which does not
-include any information specifying a particular batch (see {{batch-modes}}). The
-Leader selects a recent batch to aggregate. The Leader MUST select a batch that
-has not yet been associated with a collection job.
+get the aggregate of a batch, the Collector issues a query for the next
+available batch. The Leader selects a recent batch to aggregate which MUST not
```
cjpatton, line 3337:
```suggestion
available batch. The Leader selects a recent batch to aggregate which MUST NOT
```

draft-ietf-ppm-dap.md
cjpatton: we're saying that coordination in a massively distributed leader may make it impossible to hit the target batch size exactly.

draft-ietf-ppm-dap.md
```diff
@@ -2623,24 +2621,11 @@ struct {
...
-  see {{extending-this-doc}} for details.
-
-  The indicated batch mode MUST match the task's batch mode. Otherwise, the
-  Leader MUST abort with error "invalidMessage".
+  indicated batch mode ({{batch-modes}}).
 
 * `agg_param`, an aggregation parameter for the VDAF being executed. This is
-  the same value as in `AggregationJobInitReq` (see {{leader-init}}). The
-  aggregation parameter MUST be valid as defined in {{agg-param-validation}}.
+  the same value as in `AggregationJobInitReq` (see {{leader-init}}).
```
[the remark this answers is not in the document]
tgeoghegan: Do you mean that the line "This is the same value as in `AggregationJobInitReq` (see {{leader-init}})." isn't relevant anymore or that my statement from last week isn't relevant anymore?

draft-ietf-ppm-dap.md
```diff
@@ -2653,58 +2638,74 @@ running aggregation jobs ({{aggregate-flow}}) until the Collector initiates a
...
+by sending GET requests to the collection job. The Collector SHOULD use each
+response's Retry-After header field to decide when to try again. The Leader
+responds the same way until either the job is ready, from which point it
+responds with a `CollectionJobResp` (defined below), or the job fails, from
+which point the Leader MUST abort with the error that caused the failure.
+
+The Leader MAY also wait to respond to the Collector's PUT until the collection
+job is ready, in which case it responds with the `CollectionJobResp`, or the job
+fails, in which case the Leader MUST abort with the error that caused the
+failure.
```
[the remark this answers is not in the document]
tgeoghegan: 0. Sure, your phrasing is more explicit. I made similar changes to corresponding text in aggregate share handling.
1. Not explicitly, but I added a line of text later indicating when the job should be considered ready. And made a corresponding change to the aggregate share text.
2. In the paragraph above, we introduce it as "`CollectionJobResp` (defined below)". I can hoist up the definition of that structure but I think it's clear enough as is.

draft-ietf-ppm-dap.md
```diff
@@ -2653,58 +2638,74 @@ running aggregation jobs ({{aggregate-flow}}) until the Collector initiates a
...
+Next, the Leader checks that the indicated batch mode matches the task's batch
+mode. If not, it MUST fail the job with error `invalidMessage`.
+
 Next, the Leader checks that the aggregation parameter is valid as described in
-{{agg-param-validation}}. If the aggregation parameter is invalid, then the
-Leader MUST abort with error `invalidAggregationParameter`.
+{{agg-param-validation}}. If not, it MUST fail the job with error
+`invalidAggregationParameter`.
+
+If aggregation is performed eagerly, then the Leader checks that the aggregation
```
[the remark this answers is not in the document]
tgeoghegan: It's discussed in [{#aggregate-flow}](https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/blob/6d3be110d1422385944664d21ed5a186e5b4ff28/draft-ietf-ppm-dap.md?plain=1#L1680), several sections before this text. If we want to make the discussion of eager aggregation more robust, I think we should do that in a distinct PR.

draft-ietf-ppm-dap.md
tgeoghegan: We discussed this at the last PPM interim. What I proposed was to let the leader choose it and nobody argued against that ([1]). My rationale was that we should be consistent with all the other interactions where HTTP clients always choose identifiers (clients choose report IDs, leaders choose aggregation job IDs, collectors choose collection job IDs).

[1]: https://datatracker.ietf.org/meeting/interim-2025-ppm-01/materials/slides-interim-2025-ppm-01-sessa-tim-geoghegan-aligning-dap-with-rfc-9205-00

draft-ietf-ppm-dap.md
```diff
@@ -2843,28 +2918,44 @@ structure contains the following parameters:
...
-requirements for batch parameters following the procedure in
-{{batch-validation}}. The Helper MUST validate that the aggregation parameter
-matches the aggregation parameter used during aggregation of this batch;
-otherwise, it aborts with "invalidMessage".
+The Helper MAY defer handling the aggregate share request. In this case, it
+indicates that the aggregate share is not yet ready by immediately sending an
+empty response body. The Helper SHOULD include a Retry-After header field to
+suggest a polling interval to the Leader. The Leader then polls the state of the
+job by sending GET requests to the aggregate share. The Leader SHOULD use each
+response's Retry-After header field to decide when to try again. The Helper
```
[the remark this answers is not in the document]
tgeoghegan: Seems reasonable, but there's probably other standard HTTP headers we refer to that want the same treatment. How about if we punt to a later PR? I promise I'll do it, it's just easier to extract that from a tricky change like this.

https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/issues/681

draft-ietf-ppm-dap.md
tgeoghegan: That's a fair point. I put this back.

draft-ietf-ppm-dap.md
tgeoghegan: No. I rewrote this in prose without `batch_interval` and I think it's clear.

draft-ietf-ppm-dap.md
tgeoghegan: Fair enough. I put the original paragraph back.

tgeoghegan (commented):
We're getting there, thanks @cjpatton!

draft-ietf-ppm-dap.md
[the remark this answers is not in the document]
cjpatton: What about just sticking in a backwards reference to that section? Though I can see an argument for making this a section, say in operational considerations? (And yeah if we do this, let's do it in a separate PR.)

cjpatton:
I see one more decision here (https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/pull/675#discussion_r2045446641), then I think this is ready to merge.

cjpatton (commented):
I see one more decision here (https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/pull/675#discussion_r2045446641) then I think this is ready to merge.

draft-ietf-ppm-dap.md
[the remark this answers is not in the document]
tgeoghegan: We'll come back to this in #682.

cjpatton (approved):

PR #678, opened 2025-04-15
Fold batch validation into collection job, aggregate share sections

### LINKED ISSUE #662: Actor ambiguity in asynchronous Batch Validation
In **Batch Validation**, it says Leaders should try again later when the helpers haven't validated enough reports yet:

https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/blob/f1ff40d75d372786727ff1d56ac88c51fc12e2ae/draft-ietf-ppm-dap.md?plain=1#L2923-L2928

Since this is asynchronous, that seems fine, but I think it's also natural to imagine this to be a typo as this whole section starts with the Collector as the initial actor, and it could also be reasonable that it's the _Collector_'s job to try again later. 

Maybe I'm overthinking this, but it might be good to add a few words here like: "Leaders SHOULD wait for more reports to be validated and try the collection job again later, and only after a suitable timeout report the failure to the Collector", if the idea is that the Leader might be jumping the gun here and waiting a little bit will yield success.

Additionally, perhaps there should be some discussion of what happens if the Collector does come to the Leader before the Leader has enough validated reports, and whether that should be an error or a retry-later.

I'm obviously lacking deep understanding of how this is intended to work.

tgeoghegan:
Issue #662 highlights ambiguities in the specification of Leader handling of collection jobs or Helper handling of aggregate shares. In my opinion, this ambiguity was caused in part by trying to write a single batch validation section fro both, which forces readers to flip back and forth between sections and carefully parse nuanced protocol text to figure out exactly what requirements apply to either role. This PR addresses that problem by removing the {{batch-validation}} section and "inlining" its contents into the respective Leader and Helper sections. We also clarify who is responsible for retrying collection jobs.

See individual commits for more discussion.

Resolves #662

### DIFF
--- a/draft-ietf-ppm-dap.md
+++ b/draft-ietf-ppm-dap.md
@@ -2138,10 +2138,10 @@ for each input share in the job:
       has completed a collection job for a CollectionJobReq message from the
       Collector; the Helper considers a batch to be collected once it has
       responded to an `AggregateShareReq` message from the Leader. A batch is
-      determined by query ({{batch-modes}}) conveyed in these messages. Queries
-      must satisfy the criteria covered in {{batch-validation}}. These criteria
-      are meant to restrict queries in a way that makes it easy to determine
-      whether a report pertains to a batch that was collected. See
+      determined by query conveyed in these messages. Queries must satisfy the
+      criteria defined by their batch mode ({{batch-modes}}). These criteria are
+      meant to restrict queries in a way that makes it easy to determine whether
+      a report pertains to a batch that was collected. See
       {{distributed-systems}} for more information.
 
 1. If an Aggregator cannot determine if an input share is valid, it MUST mark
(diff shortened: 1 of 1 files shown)


draft-ietf-ppm-dap.md
```diff
@@ -2139,9 +2139,9 @@ for each input share in the job:
       Collector; the Helper considers a batch to be collected once it has
       responded to an `AggregateShareReq` message from the Leader. A batch is
       determined by query ({{batch-modes}}) conveyed in these messages. Queries
-      must satisfy the criteria covered in {{batch-validation}}. These criteria
-      are meant to restrict queries in a way that makes it easy to determine
-      whether a report pertains to a batch that was collected. See
+      must satisfy the criteria defined by their batch mode. These criteria are
```
cjpatton, line 2142: I think a reference to the batch mode section would be appropriate here.

draft-ietf-ppm-dap.md
```diff
@@ -2662,6 +2662,10 @@ the Collector's PUT until the collection job is ready, in which case it responds
 with the `CollectionJobResp`, or the job fails, in which case the Leader MUST
 abort with the error that caused the failure.
 
+If the job fails with `invalidBatchSize`, then the Collector MAY retry it later,
+once it believes enough new reports have been uploaded and aggregated to allow
+the collection job to succeed.
```
cjpatton, line 2667: At first I thought this failure was impossible, but I now see that it depends on new behavior in this PR. Currently (before this PR) I believe the Leader accepts a collection job regardless of how many reports are in the batch so far and only completes the job once a sufficient number of reports have been aggregated. Now (as of this PR) the leader can quit if it doesn't thin it's going to get more reports. Is my understanding correct?

cjpatton (commented):
This looks good modulo new behavior that I want to discuss.

draft-ietf-ppm-dap.md
tgeoghegan: Good idea. I moved the reference from the previous sentence into this one.

draft-ietf-ppm-dap.md
tgeoghegan: Yeah, well spotted, that is a change worth discussing and highlighting. The previous text about this was:

```
Next, the Aggregator checks that the number of reports in the batch is equal to
or greater than the task's minimum batch size. If not, then the Helper MUST
abort with error `invalidBatchSize`. The Leader SHOULD wait for more reports to
be validated and try the collection job again later.
```

That was a SHOULD, which means the Leader was also permitted to give up on the job. We just didn't specify what error it should give up with (or indeed that it should give up at all). So I think this is something that could happen before anyway, but with this change, we're more clear about what the Collector should be ready to handle.

tgeoghegan: Oh, and for context, the new text describing the Leader behavior:

```
If the number of validated reports in the batch is not equal to or greater than
the task's minimum batch size, then the Leader SHOULD wait for more reports to
be uploaded and aggregated and try the collection job again later. The Leader
MAY give up on the collection job (for example, if it decides that no new
reports satisfying the query are likely to ever arrive), in which case it MUST
fail the job with error `invalidBatchSize`.
```

So it's the same SHOULD, but now we offer another option.

cjpatton: Oh yes, I see! This text is strictly clearer. I'm happy.

cjpatton (approved):

PR #679, opened 2025-04-15
Security consideration for predictable task IDs

### LINKED ISSUE #664: Security consideration for non-random task IDs
DAP mandates that task IDs be 32 bytes long and globally unique, but not necessarily random. I'm concerned that this could open the door for task enumeration attacks, which I discuss a bit [over in taskprov](https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap-taskprov/issues/102). I don't think DAP should start mandating random task IDs, but we should have a security considerations explaining the risks of enumeration attacks if task IDs aren't random.

tgeoghegan:
Resolves #664

### DIFF
--- a/draft-ietf-ppm-dap.md
+++ b/draft-ietf-ppm-dap.md
@@ -3784,6 +3784,23 @@ involving crafted DAP task parameters can be mitigated by having the Aggregators
 refuse shared parameters that are trivially insecure (e.g., a minimum batch size
 of 1 report).
 
+### Predictable or Enumerable Task IDs
+
+This specification imposes no requirements on task IDs except that they be
+globally unique. One way to achieve this is to use random task IDs, but
+deployments can also use schemes like {{?I-D.draft-ietf-ppm-dap-taskprov-01}}
+where task IDs are deterministically generated from some set of task parameters.
+
+In such settings, deployments should consider whether an Aggregator
+acknowledging the existence of a task (by accepting report uploads or
+aggregation jobs, for example) could unintentionally leak information such as a
+label describing the task, the identities of participating Aggregators or the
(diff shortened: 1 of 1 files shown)


draft-ietf-ppm-dap.md
```diff
@@ -3784,6 +3784,23 @@ involving crafted DAP task parameters can be mitigated by having the Aggregators
 refuse shared parameters that are trivially insecure (e.g., a minimum batch size
 of 1 report).
 
+### Predictable or Enumerable Task IDs
+
+This specification imposes no requirements on task IDs except that they be
+globally unique. One way to achieve this is to use random task IDs, but
+deployments can also use schemes like {{?I-D.draft-ietf-ppm-dap-taskprov-01}}
```
cjpatton, line 3791: I'm a little worried about this dependency, but I suppose it's likely we'll do Last Call for both DAP and taskprov around the same time.

draft-ietf-ppm-dap.md
```diff
@@ -3784,6 +3784,23 @@ involving crafted DAP task parameters can be mitigated by having the Aggregators
...
+where task IDs are deterministically generated from some set of task parameters.
+
+In such settings, deployments should consider whether the an Aggregator
```
cjpatton, line 3794:
```suggestion
In such settings, deployments should consider whether the Aggregator
```

cjpatton (approved):

draft-ietf-ppm-dap.md
tgeoghegan: We could omit the reference, but then I think it becomes unhelpfully vague. Let's see what happens as we make our way through LC and the rest of the review process.

PR #680, opened 2025-04-15
Tidy up anti-replay and late report checks

### LINKED ISSUE #670: Tidy up checks for anti-replay and collected batches
@cjpatton and I had [a discussion in IETF Slack](https://ietf.slack.com/archives/C02P9620XT6/p1741988726043879) about a couple of the more nuanced bits of validation aggregators must do on input shares, which I'll reproduce here:

There's a touchy inconsistency in input share validation that I'd like input on. First, let's recall a few things. Sorry, this is going to have a lot of links to specific paragraphs, because at this stage of the document, I'm being very fussy about making sure we say exactly what we mean to say and refer to exactly the right things.

- anti-replay consists of checking two things:
  a. has the batch this report pertains to already been collected?
  b. has this report been aggregated in the task before?
- both aggregators must enforce these conditions at some point before the end of an aggregation job.
- {{input-share-validation}} [includes condition (a)](https://datatracker.ietf.org/doc/html/draft-ietf-ppm-dap-14#section-4.8.1.4-3.8.1), but not (b).
- leader aggregation job initialization (4.8.1.1) requires checking both anti-replay conditions.
  - [4.8.1.1-2](https://datatracker.ietf.org/doc/html/draft-ietf-ppm-dap-14#section-4.8.1.1-2) discusses condition (b)
  - [4.8.1.1-3](https://datatracker.ietf.org/doc/html/draft-ietf-ppm-dap-14#section-4.8.1.1-3) references {{input-share-validation}} and so covers condition (a).
  - I think this is fine and consistent with our intention.
- helper aggregation job initialization (4.8.1.2) is designed to allow the helper to defer certain checks to the end of an aggregation job, when output shares are recovered.
  - [4.8.1.2-7](https://datatracker.ietf.org/doc/html/draft-ietf-ppm-dap-14#section-4.8.1.2-7) discusses condition (b) as a MAY
  - [4.8.1.2-10](https://datatracker.ietf.org/doc/html/draft-ietf-ppm-dap-14#section-4.8.1.2-10) references {{input-share-validation}} and so covers condition (a) as a MUST
  - [4.8.1.2-21](https://datatracker.ietf.org/doc/html/draft-ietf-ppm-dap-14#section-4.8.1.2-21) discusses condition (b) as a MUST, in the case where we have recovered an output share

The rationale was that some of the anti-replay checks can be exceptionally expensive, depending on the implementation and its storage architecture. We want helpers to be easier/cheaper to run than leaders, so the intent is to give helper implementations some flexibility. I'm OK with all that. What I find surprising is that we allow the helper to defer checks on condition (b) but not condition (a). Was that our intention?

tgeoghegan:
Among many other things about reports, Aggregators are required to check that reports only get aggregated once and that no report be added to a collected  batch. These two checks differ from the numerous other per-report validations in that they can require expensive storage queries and/or synchronization across many jobs. And while it's not OK to include a report that fails either check into a batch bucket or batch, there's no harm in preparing/refining/validating a report that will eventually be rejected for either reason. Therefore, we don't want to require that these checks be done as early as possible during aggregation job initialization, but want to allow them to be run at any time before they get added to a batch bucket.

The existing text for this was spread out across
{{input-share-validation}}, Leader agg job init, Leader agg job continue, Helper agg job init and Helper agg job continue. In particular, the Helper's anti-replay requirement appeared as a MAY in one place but a MUST in another, which is confusing for implementors.

To address this, I beef up the notion of "committing an output share" to mean the irreversible action of including an output share into a batch bucket. This allows us to more crisply discuss it in the different init and continue sections. We also move the anti-replay and late report checks into {{batch-buckets}} (which was already pointed to by all the relevant sections).

Reviewers might ask why #678 inlines some common validation logic into multiple sections while this PR does the opposite. Subjectively, I feel that {#batch-buckets} applies more neatly to the four places that we refer to it from than did the {#batch-validation} section removed in PR #678.

Closes #670

### DIFF
--- a/draft-ietf-ppm-dap.md
+++ b/draft-ietf-ppm-dap.md
@@ -680,7 +680,7 @@ with any measurement system and DAP does not attempt to address it. DAP merely
 ensures that the data is within the chosen limits, so the Client could not
 report `10^6`  or `-20` seconds.
 
-## Replay Protection {#replay-protection}
+## Replay Protection and Double Collection {#replay-protection}
 
 Another goal of DAP is to mitigate replay attacks in which a report is
 aggregated in multiple batches or multiple times in a single batch. This would
@@ -690,6 +690,14 @@ than it would otherwise.
 When a Client generates a report, it also generates a random nonce, called the
 "report ID". Each Aggregator is responsible for storing the IDs of reports it
 has aggregated and rejecting replayed reports.
+
+DAP must also ensure that any batch is only collected once, even if new reports
+arrive that would fall into that batch. Otherwise, comparing the new aggregate
+result to the previous aggregate result can violate the privacy of the added
+reports.
+
+Aggregators are responsible for refusing new reports if the batch they fall into
+has been collected ({{collect-flow}}).
 
 ## Lifecycle of Protocol Objects
 
@@ -1487,9 +1495,8 @@ If a report's ID matches that of a previously uploaded report, the Leader MUST
 ignore it. In addition, it MAY alert the Client with error `reportRejected`.
 
 The Leader MUST ignore any report pertaining to a batch that has already been
-collected (see {{input-share-validation}} for details). Otherwise, comparing
-the aggregate result to the previous aggregate result may result in a privacy
-violation. The Leader MAY also abort with error `reportRejected`.
+collected (see {{replay-protection}} for details). The Leader MAY also abort
+with error `reportRejected`.
 
 The Leader MUST ignore any report whose timestamp is outside of the task's
 `time_interval`. When it does so, it SHOULD also abort with error
@@ -1692,16 +1699,20 @@ An aggregation job has three phases:
   for each report.
 - Continuation: Exchange prep shares and messages until preparation completes or
   an error occurs.
-- Completion: Yield an output share for each report share in the aggregation job.
+- Completion: Yield an output share for each report share in the aggregation
+  job.
 
-After an aggregation job is completed, each Aggregator updates running-total
-aggregate shares and other values for each "batch bucket" associated with a
-prepared output share, as described in {{batch-buckets}}. These values are
-stored until the batch is collected as described in {{collect-flow}}.
+After an aggregation job is completed, each Aggregator commits to the output
+shares by updating running-total aggregate shares and other values for each
+batch bucket associated with a prepared output share, as described in
+{{batch-buckets}}. These values are stored until a batch that includes the batch
+bucket is collected as described in {{collect-flow}}.
 
-The aggregation interaction provides replay protection ({{replay-protection}}).
-Before committing to an output share, the Aggregators check whether its report
-ID has already been aggregated.
+The aggregation interaction provides protection against including reports in
+more than one batch and against adding reports to already collected batches,
+both of which can violate privacy ({{replay-protection}}). Before committing to
+an output share, the Aggregators check whether its report ID has already been
+aggregated and whether the batch bucket being updated has been collected.
 
 ### Aggregate Initialization {#agg-init}
 
@@ -1717,10 +1728,6 @@ The Leader and Helper initialization behavior is detailed below.
 
 The Leader begins an aggregation job by choosing a set of candidate reports that
 belong to the same task and a job ID which MUST be unique within the task.
-
-For each report in the candidate set, it checks if the report ID has already
-been aggregated for this task. If so, it rejects the report and removes it from
-the candidate set.
 
 Next, the Leader decrypts each of its report shares as described in
 {{input-share-decryption}}, then checks input share validity as described in
@@ -1859,9 +1866,9 @@ The Leader proceeds as follows with each report:
    the preparation process is complete: either `state == Rejected()`, in which
    case the Leader rejects the report and removes it from the candidate set; or
    `state == Finished(out_share)`, in which case preparation is complete and the
-   Leader updates the relevant batch bucket with `out_share` as described in
-   {{batch-buckets}}. The Leader stores the report ID corresponding to the
-   `out_share` for replay protection.
+   Leader commits `out_share` as described in {{batch-buckets}}. If commitment
+   fails, then the Leader rejects the report and removes it from the candidate
+   set.
 
 1. Else if the type is "reject", then the Leader rejects the report and removes
    it from the candidate set. The Leader MUST NOT include the report in a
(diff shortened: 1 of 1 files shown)


draft-ietf-ppm-dap.md
```diff
@@ -2130,19 +2112,6 @@ for each input share in the job:
    private extension fields. If so, the Aggregator MUST mark the input share as
    invalid with error `invalid_message`.
 
-1. If the report pertains to a batch that was previously collected, then the
-   input share MUST be marked as invalid with error `batch_collected`.
-
-    * Implementation note: The Leader considers a batch to be collected once it
```
tgeoghegan, line 2136: I should put this back, but I'm not sure if it should go inline in `{#batch-buckets}` or in Operational Considerations.

tgeoghegan: I put the implementation note unchanged at the bottom of {{batch-buckets}}.

draft-ietf-ppm-dap.md
```diff
@@ -2029,21 +2014,18 @@ variant {
...
+described in {{batch-buckets}}. If commitment fails with error `commit_error`,
+then instead of the `continue` response above, the Helper responds with
 
 ~~~ tls-presentation
 variant {
   ReportID report_id;
-  PrepareRespState prepare_resp_state = reject;
-  Reporterror report_error = report_replayed;
+  PrepareRespstate prepare_resp_state = reject;
+  ReportError = commit_error;
```
cjpatton, line 2025: Syntax looks inconsistent with what we had before. Should it be `ReportError report_error = commit_error`?

Likewise below.

draft-ietf-ppm-dap.md
```diff
@@ -2751,7 +2737,7 @@ structure includes the following:
 
 Once the `Leader` has constructed a `CollectionJobResp` for the Collector, the
 Leader considers the batch to be collected, and further aggregation jobs MAY NOT
```
cjpatton, line 2739: I'm pretty sure MAY NOT should be MUST NOT

draft-ietf-ppm-dap.md
```diff
@@ -3009,11 +2995,11 @@ computed above and `encrypted_aggregate_share.ciphertext` is the ciphertext
...
 its response to the Collector (see {{collect-finalization}}).
 
-Once an `AggregateShareReq` has been issued for the batch determined by a given
-query, the Helper considers the batch to be collected. It is an error for the
-Leader to issue any more aggregation jobs for additional reports that satisfy
-the query. These reports will be rejected by the Helper as described in
-{{input-share-validation}}.
+Once an `AggregateShareReq` has been constructed for the batch determined by a
+given query, the Helper considers the batch to be collected. The Helper MUST NOT
+commit any more output shares to the batch and it is an error for the Leader to
```
cjpatton, line 3000:
```suggestion
commit any more output shares to the batch. It is an error for the Leader to
```

draft-ietf-ppm-dap.md
```diff
@@ -2029,21 +2014,18 @@ variant {
 If `state == Continued(prep_state)`, then the Helper stores `state` to
 prepare for the next continuation step ({{aggregation-helper-continuation}}).
 
-If `state == Finished(out_share)`, the Helper MUST resolve replay of the
-report. It does so by checking if the report ID was previously aggregated for
-this task. If so, it responds with
+If `state == Finished(out_share)`, then the Helper commits to `out_share` as
+described in {{batch-buckets}}. If commitment fails with error `commit_error`,
```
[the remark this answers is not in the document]
tgeoghegan: What I mean is that it could fail with some `ReportError` value that should be used in the `PrepareResp`. I made this more clear (hopefully).

draft-ietf-ppm-dap.md
tgeoghegan: Good catch!

draft-ietf-ppm-dap.md
```diff
@@ -1692,16 +1692,20 @@ An aggregation job has three phases:
...
+"batch bucket" associated with a prepared output share, as described in
+{{batch-buckets}}. These values are stored until the batch is collected as
+described in {{collect-flow}}.
 
-The aggregation interaction provides replay protection ({{replay-protection}}).
-Before committing to an output share, the Aggregators check whether its report
-ID has already been aggregated.
+The aggregation interaction provides protection against including reports in
+more than one batch ({{replay-protection}}) and against adding reports to
+already collected batches, both of which can violate privacy. Before committing
```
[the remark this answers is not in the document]
tgeoghegan: Yeah, good call. To address this, I expanded {{replay-protection}} to also discuss late reports and batches being collected multiple times. Doing this made me notice that the language in the upload section had an outdated reference to {{input-share-validation}} which I updated.

draft-ietf-ppm-dap.md
```diff
@@ -1487,9 +1495,8 @@ If a report's ID matches that of a previously uploaded report, the Leader MUST
 ignore it. In addition, it MAY alert the Client with error `reportRejected`.
 
 The Leader MUST ignore any report pertaining to a batch that has already been
-collected (see {{input-share-validation}} for details). Otherwise, comparing
-the aggregate result to the previous aggregate result may result in a privacy
```
cjpatton, line 1491: Why remove "Otherwise, ..."? This seems like a useful clarification.

draft-ietf-ppm-dap.md
```diff
@@ -2029,21 +2021,19 @@ variant {
 If `state == Continued(prep_state)`, then the Helper stores `state` to
 prepare for the next continuation step ({{aggregation-helper-continuation}}).
 
-If `state == Finished(out_share)`, the Helper MUST resolve replay of the
-report. It does so by checking if the report ID was previously aggregated for
-this task. If so, it responds with
+If `state == Finished(out_share)`, then the Helper commits to `out_share` as
+described in {{batch-buckets}}. If commitment fails with some `ReportError`
+value `commit_error`, then instead of the `continue` response above, the Helper
```
cjpatton, line 2026: I now see what you're doing. But "`commit_error`" reads as if `commit_error` is a variant of `ReportError`. I would just do this:
```suggestion
, then instead of the `continue` response above, the Helper
```

cjpatton: I would also clarify how this might happen in a parantehtical:

> (i.e., the report was replayed or its batch bucket was collected)

draft-ietf-ppm-dap.md
```diff
@@ -2400,23 +2374,20 @@ variant {
 If `state == Continued(prep_state)`, then the Helper stores `state` to
 prepare for the next continuation step ({{aggregation-helper-continuation}}).
 
-If `state == Finished(out_share)`, the Helper MUST resolve replay of the
-report. It does so by checking if the report ID was previously stored for this
-task. If so, it responds with
+If `state == Finished(out_share)`, the Helper commits `out_share` as described
+in {{batch-buckets}}. If commitment fails with some some `ReportError` value
```
cjpatton, line 2378:
```suggestion
in {{batch-buckets}}. If commitment fails with some `ReportError` value
```

draft-ietf-ppm-dap.md
tgeoghegan: Because I think that's now well explained in {#replay-protection}, so no need to repeat it here.

cjpatton (approved):

PR #684, opened 2025-04-16
Expand on "eager aggregation"

### LINKED ISSUE #682: Add heading for eager aggregation and refer to it from protocol text
There's a couple spots where we include extra prescriptions in the case of eager aggregation (e.g. checking that a collector's agg param matches the agg param eagerly used by the aggregators). These should refer to a section which explicitly defines eager aggregation.

We'll take the [existing paragraph on this](https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/blob/6d3be110d1422385944664d21ed5a186e5b4ff28/draft-ietf-ppm-dap.md?plain=1#L1680), break it out into `### Eager aggregation`, and add a sentence or two explaining that if there is an agg param mismatch, the relevant reports can never be collected.

tgeoghegan:
closes #682

### DIFF
--- a/draft-ietf-ppm-dap.md
+++ b/draft-ietf-ppm-dap.md
@@ -1685,13 +1685,9 @@ model that best fits its architecture and use case. For instance replay checks
 across vast numbers of reports and preparation of large histograms, may be
 better suited for the asynchronous model.
 
-In general, aggregation cannot begin until the Collector specifies a query and
-an aggregation parameter. However, depending on the VDAF and batch mode in use,
-it is often possible to begin aggregation as soon as reports arrive. For
-example, Prio3 has just one valid aggregation parameter (the empty string), and
-thus allows for eager aggregation. Both the time-interval and leader-selected
-batch modes defined in this document ({{batch-modes}}) allow for eager
-aggregation.
+Aggregation cannot begin until the Collector specifies a query and an
+aggregation parameter, except where eager aggregation ({{eager-aggregation}}) is
+possible.
 
 An aggregation job has three phases:
 
@@ -1713,6 +1709,29 @@ more than one batch and against adding reports to already collected batches,
 both of which can violate privacy ({{replay-protection}}). Before committing to
 an output share, the Aggregators check whether its report ID has already been
 aggregated and whether the batch bucket being updated has been collected.
+
+### Eager Aggregation {#eager-aggregation}
+
+In general, aggregation cannot begin until the Collector specifies a query and
+an aggregation parameter. However, depending on the VDAF and batch mode in use,
+it is often possible to begin aggregation as soon as reports arrive.
+
+For example, Prio3 has just one valid aggregation parameter (the empty string),
+and so allows for eager aggregation. Both the time-interval and leader-selected
+batch modes defined in this document ({{batch-modes}}) allow for eager
+aggregation, but future batch modes might preclude it.
+
+Even when the VDAF uses a non-empty aggregation parameter, there still might be
+some applications in which the Aggregators can anticipate the parameter the
+Collector will choose and begin aggregation.
+
+For example, when using Poplar1 ({{Section 8 of !VDAF}}), the Collector and
+Aggregators might agree ahead of time on the set of candidate prefixes to use.
+In such cases, it is important that Aggregators ensure that the parameter
+eventually chosen by the Collector matches what they used. Depending on the
+VDAF, aggregating reports with multiple aggregation parameters may impact
+privacy. Aggregators must therefore ensure they only ever use the aggregation
+parameter chosen by the Collector.
 
 ### Aggregate Initialization {#agg-init}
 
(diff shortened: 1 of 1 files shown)


draft-ietf-ppm-dap.md
```diff
@@ -1714,6 +1706,22 @@ both of which can violate privacy ({{replay-protection}}). Before committing to
...
+example, Prio3 has just one valid aggregation parameter (the empty string), and
+thus allows for eager aggregation. Both the time-interval and leader-selected
+batch modes defined in this document ({{batch-modes}}) allow for eager
+aggregation. Or if the VDAF in use does have a non-trivial aggregation
+parameter, there might be some way the Aggregators can anticipate the parameter
+the Collector will choose and begin aggregation. In such cases, it's important
+that Aggregators ensure that the parameter eventually chosen by the Collector
+matches what they used. Re-preparing the reports with a new aggregation
+parameter can violate privacy, so Aggregators must take care. If they guess the
+wrong parameter, the reports can never be aggregated.
```
cjpatton, line 1723:
```suggestion
matches what they used. Depending on the VDAF, aggregating reports with multiple aggregation parameters new aggregation may impact privacy. Aggregators must therefore ensure they only use the aggregation parameter chosen by the Collector.
```

draft-ietf-ppm-dap.md
```diff
@@ -1714,6 +1706,22 @@ both of which can violate privacy ({{replay-protection}}). Before committing to
...
+
```
cjpatton, line 1724: We should also mention in this section how batch modes might or might not affect eager aggregation. I imagine this matters only if the batch mode requires so much flexibility that we can't define batch buckets. Imagine for example you wanted SQL-like queries.

draft-ietf-ppm-dap.md
```diff
@@ -1714,6 +1706,22 @@ both of which can violate privacy ({{replay-protection}}). Before committing to
...
+In general, aggregation cannot begin until the Collector specifies a query and
+an aggregation parameter. However, depending on the VDAF and batch mode in use,
+it is often possible to begin aggregation as soon as reports arrive. For
+example, Prio3 has just one valid aggregation parameter (the empty string), and
+thus allows for eager aggregation. Both the time-interval and leader-selected
+batch modes defined in this document ({{batch-modes}}) allow for eager
+aggregation. Or if the VDAF in use does have a non-trivial aggregation
+parameter, there might be some way the Aggregators can anticipate the parameter
+the Collector will choose and begin aggregation. In such cases, it's important
```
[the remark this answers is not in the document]
tgeoghegan: Are attribute based metrics possible in Poplar1? I thought you needed Mastic for that. I put together a Poplar1 based example but I agree that it's contrived.

draft-ietf-ppm-dap.md
```diff
@@ -1714,6 +1710,29 @@ both of which can violate privacy ({{replay-protection}}). Before committing to
...
+In general, aggregation cannot begin until the Collector specifies a query and
+an aggregation parameter. However, depending on the VDAF and batch mode in use,
+it is often possible to begin aggregation as soon as reports arrive.
+
+For example, Prio3 has just one valid aggregation parameter (the empty string),
+and so allows for eager aggregation. Both the time-interval and leader-selected
+batch modes defined in this document ({{batch-modes}}) allow for eager
+aggregation, but future batch modes might preclude it.
+
+If the VDAF in use does have a non-trivial aggregation parameter, there might be
```
cjpatton, line 1724:
```suggestion
Even when the VDAF uses a non-empty aggregation parameter, there still might be
```

draft-ietf-ppm-dap.md
```diff
@@ -1714,6 +1710,29 @@ both of which can violate privacy ({{replay-protection}}). Before committing to
...
+some way the Aggregators can anticipate the parameter the Collector will choose
+and begin aggregation. For example, when using Poplar1 ({{Section 8 of !VDAF}}),
+the Collector and Aggregators might agree ahead of time on the set of candidate
+prefixes to use.
```
cjpatton, line 1728:
```suggestion
prefixes to use. However this depends on the application.
```

draft-ietf-ppm-dap.md
```diff
@@ -1714,6 +1710,29 @@ both of which can violate privacy ({{replay-protection}}). Before committing to
...
+### Eager Aggregation {#eager-aggregation}
+
+In general, aggregation cannot begin until the Collector specifies a query and
+an aggregation parameter. However, depending on the VDAF and batch mode in use,
+it is often possible to begin aggregation as soon as reports arrive.
+
+For example, Prio3 has just one valid aggregation parameter (the empty string),
+and so allows for eager aggregation. Both the time-interval and leader-selected
+batch modes defined in this document ({{batch-modes}}) allow for eager
+aggregation, but future batch modes might preclude it.
```
cjpatton, line 1722: This sentence ("Both the time-interval, ...") doesn't really belong here. Perhaps move to a new paragraph below as part of https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/pull/684/files#r2054422312.

draft-ietf-ppm-dap.md
cjpatton: Bump

tgeoghegan: It's at the end of the paragraph above:

> Both the time-interval and leader-selected batch modes defined in this document ({{batch-modes}}) allow for eager aggregation, but future batch modes might preclude it.

draft-ietf-ppm-dap.md
tgeoghegan: This paragraph discusses how VDAF and batch mode affect the possibiity of eager aggregation, following directly from the previous paragraph. It makes sense to me.

draft-ietf-ppm-dap.md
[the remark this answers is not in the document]
cjpatton: It is indeed possible, but you get pretty limited functionality. Basically you can only do Prio3Count.

draft-ietf-ppm-dap.md
tgeoghegan: Sure, I added the word "application" earlier in the paragraph.

draft-ietf-ppm-dap.md
```diff
@@ -1714,6 +1710,29 @@ both of which can violate privacy ({{replay-protection}}). Before committing to
...
+
+Even when the VDAF uses a non-empty aggregation parameter, there still might be
+some applications inw which the Aggregators can anticipate the parameter the
```
cjpatton, line 1725:
```suggestion
some applications in which the Aggregators can anticipate the parameter the
```

draft-ietf-ppm-dap.md
```diff
@@ -1714,6 +1710,29 @@ both of which can violate privacy ({{replay-protection}}). Before committing to
...
+Collector will choose and begin aggregation. For example, when using Poplar1
+({{Section 8 of !VDAF}}), the Collector and Aggregators might agree ahead of
+time on the set of candidate prefixes to use.
+
+In such cases, it's important that Aggregators ensure that the parameter
```
cjpatton, line 1730: Eh
```suggestion
Collector will choose and begin aggregation.

For example, when using Poplar1
({{Section 8 of !VDAF}}), the Collector and Aggregators might agree ahead of
time on the set of candidate prefixes to use.
In such cases, it's important that Aggregators ensure that the parameter
```

cjpatton (approved):

PR #685, opened 2025-04-29
draft-ietf-ppm-dap-15 chores

tgeoghegan:
- flesh out the changelog
- update domain separation tags for -15


draft-ietf-ppm-dap.md
```diff
@@ -160,10 +160,23 @@ aggregator.
 
 15:
 
-- Reorganize text for clarity and flow.
+- Specify body of responses to aggregation job GET requests. (#651)
```
cjpatton, line 163: Is this a breaking change?

draft-ietf-ppm-dap.md
```diff
@@ -160,10 +160,23 @@ aggregator.
...
+- Add diagram illustrating object lifecycles and relationships. (#655)
+
+- Use aasvg for prettier diagrams. (#657)
+
+- More precise description of time and intervals. (#658)
+
+- Reorganize text for clarity and flow. (#659, #660, #661, #663, #665, #666,
+  #668, #672, #678, #680, #684, #653, #654)
+
+- Align with RFC 9205 recommendations. (#673, #683)
```
cjpatton, line 174: Is this a breaking change?

cjpatton (commented):
* nit: Each of the bullets should follow conventions for git messages, e.g., "Add a more precise description of..." rather than "More precise description of..."
* nit: Previously we haven't had links to issues/PRs. I'd define "#XXX" in this section so that readers unfamiliar with this convention know what we mean. You could also just leave them out for consistency.

draft-ietf-ppm-dap.md
tgeoghegan: No, AFAICT it only clarified implicit behavior.

draft-ietf-ppm-dap.md
tgeoghegan: Not as far as I can tell.

cjpatton: In my view it is. Previously we were prescribing specific HTTP status codes. Now we're not, but a DAP-14 aggregator might expect us to.

cjpatton (approved):

PR #686, opened 2025-04-30
Support more than one request in report upload API

akoshelev:
As discussed in #667, this PR proposes changing the upload request to support more than one report in one upload. It does so by employing a technique similar to [aggregation jobs](https://www.ietf.org/archive/id/draft-ietf-ppm-dap-14.html#section-4.8.1.2-6). Clients submit a sequence of 1..N reports and the Leader responds with a sequence of statuses for each upload. 

It suggests an alternative approach to the one taken in #667 to support #647

~~This is a draft as this PR hasn't been polished yet. I am seeking early feedback on the direction where this PR takes DAP.~~

## Preview
[HTML](https://akoshelev.github.io/draft-ietf-ppm-dap/dap-bulk-multipart/draft-ietf-ppm-dap.html)

## Improvements
* ~~It does seem to me that `ReportError` now needs to live in a dedicated section as it's been used across two unrelated functionalities~~ Done
* ~~Maybe there is a way to express `None` variant better than `ReportError::None`. A Rust analog would be `Option<ReportError>`, but it does not seem that [Section 3 of RFC8446](https://www.rfc-editor.org/rfc/rfc8446#section-3) allows for it~~ no longer required as we skip successful reports in the response

### DIFF
--- a/draft-ietf-ppm-dap.md
+++ b/draft-ietf-ppm-dap.md
@@ -83,6 +83,11 @@ contributor:
        email: csharrison@chromium.org
 
  -
+       name: Alex Koshelev
+       org: Meta
+       email: koshelev@meta.com
+
+ -
        name: Peter Saint-Andre
        email: stpeter@gmail.com
 
@@ -958,9 +963,6 @@ standard tokens for use in the "type" field:
 | invalidMessage              | A message received by a protocol participant could not be parsed or otherwise was invalid. |
 | unrecognizedTask            | A server received a message with an unknown task ID. |
 | unrecognizedAggregationJob  | A server received a message with an unknown aggregation job ID. |
-| outdatedConfig              | The message was generated using an outdated configuration. |
-| reportRejected              | Report could not be processed for an unspecified reason. |
-| reportTooEarly              | Report could not be processed because its timestamp is too far in the future. |
 | batchInvalid                | The batch boundary check for Collector's query failed. |
 | invalidBatchSize            | There are an invalid number of reports in the batch. |
 | invalidAggregationParameter | The aggregation parameter assigned to a batch is invalid. |
@@ -1050,6 +1052,27 @@ in different situations.
 
 ~~~ tls-presentation
 struct {} Empty;
+~~~
+
+Errors that occurred while handling individual reports in the upload or
+aggregation interactions are represented by the following enum:
+
+~~~ tls-presentation
+enum {
+  reserved(0),
+  batch_collected(1),
+  report_replayed(2),
+  report_dropped(3),
+  hpke_unknown_config_id(4),
+  hpke_decrypt_error(5),
+  vdaf_prep_error(6),
+  task_expired(7),
+  invalid_message(8),
+  report_too_early(9),
+  task_not_started(10),
+  outdated_config(11),
+  (255)
+} ReportError;
 ~~~
 
 ### Times, Durations and Intervals {#timestamps}
@@ -1347,8 +1370,8 @@ encoded([
 ### Upload Request
 
 Clients upload reports by sending a POST to `{leader}/tasks/{task-id}/reports`.
-The body is a `Report`, with media type "application/dap-report", structured as
-follows:
+The body is an `UploadRequest`, with media type "application/dap-upload-req",
+structured as follows:
 
 ~~~ tls-presentation
 struct {
@@ -1363,7 +1386,16 @@ struct {
   HpkeCiphertext leader_encrypted_input_share;
   HpkeCiphertext helper_encrypted_input_share;
 } Report;
+
+struct {
+  Report reports[message_length];
+} UploadRequest;
 ~~~
+
+Here `message_length` denotes the length of the HTTP message content ({{!RFC9110,
+Section 6.4}}).
+
+Each upload request contains a sequence of `Report` with the following fields:
 
 * `report_metadata` is public metadata describing the report.
 
@@ -1466,30 +1498,60 @@ If the upload request is malformed, the Leader aborts with error
 If the Leader does not recognize the task ID, then it aborts with error
 `unrecognizedTask`.
 
+Otherwise, the Leader responds with a body consisting of an `UploadResponse`
+with the media type `application/dap-upload-resp`. The structure of the response
+is as follows:
+
+~~~ tls-presentation
+struct {
+  ReportID id;
+  ReportError error;
+} ReportUploadStatus;
+
+struct {
+  ReportUploadStatus status[message_length];
+} UploadResponse;
+~~~
+
+Here `message_length` denotes the length in bytes of the concatenated
+`ReportUploadStatus` objects.
+
+The Leader only includes reports that failed processing in the response. Reports
+that are accepted do not have a response.
+
+Reports in the response MUST appear in the same order as in the request.
+
+For each report that failed to upload, the Leader creates a `ReportUploadStatus`
+and includes the `ReportId` from the input and a `ReportError`
+({{basic-definitions}}) that describes the failure. The length of this sequence
+is always less than or equal to the length of the upload sequence.
+
 If the Leader does not recognize the `config_id` in the encrypted input share,
-it aborts with an error of type `outdatedConfig`. When the Client receives an
-`outdatedConfig` error, it SHOULD invalidate any cached `HpkeConfigList` and
-retry with a freshly generated `Report`. If this retried upload does not
-succeed, the Client SHOULD abort and discontinue retrying.
+it sets the corresponding error field to `outdated_config`. When the Client
+receives an `outdated_config` error, it SHOULD invalidate any cached
+`HpkeConfigList` and retry with a freshly generated `Report`. If this retried
+upload does not succeed, the Client SHOULD abort and discontinue retrying.
 
 If a report's ID matches that of a previously uploaded report, the Leader MUST
-discard it. In addition, it MAY alert the Client with error `reportRejected`.
+discard it. In addition, it MAY set the corresponding error field to
+`report_replayed`.
 
 The Leader MUST discard any report pertaining to a batch that has already been
-collected (see {{replay-protection}} for details). The Leader MAY also abort
-with error `reportRejected`.
+collected (see {{replay-protection}} for details). The Leader MAY also set the
+corresponding error field to `report_replayed`.
 
 The Leader MUST discard any report whose timestamp is outside of the task's
-`time_interval`. When it does so, it SHOULD also abort with error
-`reportRejected`.
+`time_interval`. When it does so, it SHOULD set the corresponding error field to
+`report_dropped`.
 
 The Leader may need to buffer reports while waiting to aggregate them (e.g.,
 while waiting for an aggregation parameter from the Collector; see
 {{collect-flow}}). The Leader SHOULD NOT accept reports whose timestamps are too
 far in the future. Implementors MAY provide for some small leeway, usually no
 more than a few minutes, to account for clock skew. If the Leader rejects a
-report for this reason, it SHOULD abort with error `reportTooEarly`. In this
-situation, the Client MAY re-upload the report later on.
+report for this reason, it SHOULD set the corresponding error field to
+`report_too_early`. In this situation, the Client MAY re-upload the report later
+on.
 
 If the report contains an unrecognized public report extension, or if the
 Leader's input share contains an unrecognized private report extension, then the
@@ -1514,31 +1576,101 @@ will occur before aggregation.
 
 #### Example
 
+Successful upload
+
 ~~~ http
 POST /leader/tasks/8BY0RzZMzxvA46_8ymhzycOB9krN-QIGYvg_RsByGec/reports
 Host: example.com
-Content-Type: application/dap-report
+Content-Type: application/dap-upload-req
+Content-Length: 100
 
 encoded(struct {
-  report_metadata = struct {
-    report_id = [0x0a, 0x0b, 0x0c, 0x0d, ...],
-    time = 1741986088,
-    public_extensions = [0x00, 0x00],
-  } ReportMetadata,
-  public_share = [0x0a, 0x0b, ...],
-  leader_encrypted_input-share = struct {
-    config_id = 1,
-    enc = [0x0f, 0x0e, 0x0d, 0x0c, ...],
-    payload = [0x0b, 0x0a, 0x09, 0x08, ...],
-  } HpkeCiphertext,
-  helper_encrypted_input-share = struct {
-    config_id = 2,
-    enc = [0x0c, 0x0d, 0x0e, 0x0f, ...],
-    payload = [0x08, 0x00, 0x0a, 0x0b, ...],
-  } HpkeCiphertext,
-} Report)
+  reports = [
+    struct {
+      report_metadata = struct {
+        report_id = [0x0a, 0x0b, 0x0c, 0x0d, ...],
+        time = 1741986088,
+        public_extensions = [0x00, 0x00],
+      } ReportMetadata,
+      public_share = [0x0a, 0x0b, ...],
+      leader_encrypted_input-share = struct {
+        config_id = 1,
+        enc = [0x0f, 0x0e, 0x0d, 0x0c, ...],
+        payload = [0x0b, 0x0a, 0x09, 0x08, ...],
+      } HpkeCiphertext,
+      helper_encrypted_input-share = struct {
+        config_id = 2,
+        enc = [0x0c, 0x0d, 0x0e, 0x0f, ...],
+        payload = [0x08, 0x00, 0x0a, 0x0b, ...],
+      } HpkeCiphertext,
+    } Report,
+  ],
+} UploadRequest)
 
 HTTP/1.1 200
+~~~
+
+Failed upload of 1/2 reports submitted in one bulk upload
+
+~~~ http
+POST /leader/tasks/8BY0RzZMzxvA46_8ymhzycOB9krN-QIGYvg_RsByGec/reports
+Host: example.com
+Content-Type: application/dap-upload-req
+Content-Length: 200
+
+encoded(struct {
+  reports = [
+    struct {
+      report_metadata = struct {
+        report_id = [0x0a, 0x0b, 0x0c, 0x0d, ...],
+        time = 2000000000,
+        public_extensions = [0x00, 0x01],
+      } ReportMetadata,
+      public_share = [0x0a, 0x0b, ...],
+      leader_encrypted_input-share = struct {
+        config_id = 1,
+        enc = [0x0f, 0x0e, 0x0d, 0x0c, ...],
+        payload = [0x0b, 0x0a, 0x09, 0x08, ...],
+      } HpkeCiphertext,
+      helper_encrypted_input-share = struct {
+        config_id = 2,
+        enc = [0x0c, 0x0d, 0x0e, 0x0f, ...],
+        payload = [0x08, 0x00, 0x0a, 0x0b, ...],
+      } HpkeCiphertext,
+    } Report,
+    struct {
+      report_metadata = struct {
+        report_id = [0x0z, 0x0y, 0x0x, 0x0w, ...],
+        time = 2000000001,
+        public_extensions = [0x00, 0x01],
+      } ReportMetadata,
+      public_share = [0x0a, 0x0b, ...],
+      leader_encrypted_input-share = struct {
+        config_id = 1,
+        enc = [0x0f, 0x0e, 0x0d, 0x0c, ...],
+        payload = [0x0b, 0x0a, 0x09, 0x08, ...],
+      } HpkeCiphertext,
+      helper_encrypted_input-share = struct {
+        config_id = 2,
+        enc = [0x0c, 0x0d, 0x0e, 0x0f, ...],
+        payload = [0x08, 0x00, 0x0a, 0x0b, ...],
+      } HpkeCiphertext,
+    } Report,
+  ],
+} UploadRequest)
+
+HTTP/1.1 200
+Content-Type: application/dap-upload-resp
+Content-Length: 20
+
+encoded(struct {
+  reports = [
+    struct {
+      id = [0x0z, 0x0y, 0x0x, 0x0w, ...],
+      error = report_replayed,
+    },
+  ],
+} UploadResponse)
 ~~~
 
 ### Report Extensions {#report-extensions}
@@ -1961,21 +2093,6 @@ enum {
   reject(2),
   (255)
 } PrepareRespType;
-
-enum {
-  reserved(0),
-  batch_collected(1),
-  report_replayed(2),
-  report_dropped(3),
-  hpke_unknown_config_id(4),
-  hpke_decrypt_error(5),
-  vdaf_prep_error(6),
-  task_expired(7),
-  invalid_message(8),
-  report_too_early(9),
-  task_not_started(10),
-  (255)
-} ReportError;
 
 struct {
   ReportID report_id;
@@ -3540,7 +3657,7 @@ ensure correctness during concurrent operation. This section describes the
 relevant concerns and makes suggestions as to potential implementation
 tradeoffs.
 
-* The upload interaction requires the Leader to ignore uploaded reports with a
+* The upload interaction requires the Leader to discard uploaded reports with a
   duplicated ID, including concurrently-uploaded reports. This might be
   implemented by synchronization or via an eventually-consistent process. If the
   Leader wishes to alert the Client with a `reportRejected` error,
@@ -3876,7 +3993,8 @@ This specification defines the following protocol messages, along with their
 corresponding media types:
 
 - HpkeConfigList {{hpke-config}}: "application/dap-hpke-config-list"
-- Report {{upload-request}}: "application/dap-report"
+- UploadRequest {{upload-request}}: "application/dap-upload-req"
+- UploadResponse {{upload-request}}: "application/dap-upload-resp"
 - AggregationJobInitReq {{leader-init}}: "application/dap-aggregation-job-init-req"
 - AggregationJobResp {{aggregation-helper-init}}: "application/dap-aggregation-job-resp"
 - AggregationJobContinueReq {{aggregation-leader-continuation}}: "application/dap-aggregation-job-continue-req"
@@ -3899,7 +4017,7 @@ component is using. This MAY be used as a hint by the receiver of the request
 to do compatibility checks between client and server.
 For example, A report submission to leader from a client that supports
 draft-ietf-ppm-dap-09 could have the header
-`Content-Type: application/dap-report;version=09`.
+`Content-Type: application/dap-upload-req;version=09`.
 
 The "Media Types" registry at https://www.iana.org/assignments/media-types will
 be (RFC EDITOR: replace "will be" with "has been") updated to include each of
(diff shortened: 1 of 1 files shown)


draft-ietf-ppm-dap.md
```diff
@@ -1479,36 +1487,58 @@ struct {
 } InputShareAad;
 ~~~
 
+~~~ tls-presentation
+struct {
+ReportID id;
```
bemasc, line 1492: Would it be helpful in terms of efficiency to omit the ID here?  It can be inferred from the order of responses in the vector.

draft-ietf-ppm-dap.md
```diff
@@ -1382,8 +1384,14 @@ struct {
   HpkeCiphertext leader_encrypted_input_share;
   HpkeCiphertext helper_encrypted_input_share;
 } Report;
+
+struct {
+  Report reports<0..2^32-1>;
```
tgeoghegan, line 1389: My understanding was that one of the motivations here is to deduplicate public extensions across reports, so I would think we'd have a field for that here.

draft-ietf-ppm-dap.md
```diff
@@ -1479,36 +1487,58 @@ struct {
...
+~~~
+
 If the upload request is malformed, the Leader aborts with error
 `invalidMessage`.
 
 If the Leader does not recognize the task ID, then it aborts with error
 `unrecognizedTask`.
 
+Otherwise, the Leader responds with HTTP status 200, a body consisting
+of `UploadResponse` and the media type `application/dap-report-resp`. If
```
tgeoghegan, line 1508: Per RFC 9205, we don't prescribe specific HTTP statuses, but leave it up to implementations to choose which success status to use.
```suggestion
Otherwise, the Leader responds with a body consisting
of an `UploadResponse` and the media type `application/dap-report-resp`. If
```

draft-ietf-ppm-dap.md
```diff
@@ -3906,7 +4005,8 @@ This specification defines the following protocol messages, along with their
 corresponding media types:
 
 - HpkeConfigList {{hpke-config}}: "application/dap-hpke-config-list"
-- Report {{upload-request}}: "application/dap-report"
+- UploadRequest {{upload-request}}: "application/dap-report"
+- UploadResponse {{upload-request}}: "application/dap-report-resp"
```
tgeoghegan, line 4009: I think we should use `application/dap-upload-req` and `dap-upload-resp`.

draft-ietf-ppm-dap.md
tgeoghegan: Based on the "Failed upload of 1/2 reports submitted in one bulk upload" example, it seems like the intent might be that the response can include `ReportUploadStatus` values only for the reports that weren't successfully uploaded, and at that point you do need to report ID to be able to tell which it was. I kinda like that, because as we expect the vast majority of reports to be uploaded without issue, it massively cuts down on the size of responses. 

You might be able to save even more room on the wire by including an index into the vector in the `UploadRequest` instead of the report ID, but that would require clients (or that entity which uploads, which may not be a client in proxy arrangements) to maintain an index of upload requests to report IDs, which might not be tractable.

tgeoghegan (commented):
Thanks for this contribution, Alex!

> It does seem to me that ReportError now needs to live in a dedicated section as it's been used across two unrelated functionalities

Yeah, I think we would hoist it out into {#errors} or {#basic-definitions}. Slightly unfortunate to add another table before we get to the meat of the protocol, but I think it's right.


> Maybe there is a way to express None variant better than ReportError::None. A Rust analog would be Option<ReportError>, but it does not seem that [Section 3 of RFC8446](https://www.rfc-editor.org/rfc/rfc8446#section-3) allows for it

MLS defines [`struct optional<T>`](https://datatracker.ietf.org/doc/html/rfc9420#section-2.1.1) which we could use, but I don't think it'd be better than defining `ReportError::none`. Either way, in the success case, you're encoding a 0, either to indicate `ReportError::none` or `optional::none`.

In any case, I think I prefer the (perhaps unintentional?) semantic implied by your second example: the `UploadResponse` should only contain `ReportUploadStatus` for those reports that weren't accepted by the leader, to make the response smaller. In the success case, a `ReportUploadStatus` would be 17 bytes (16 byte report ID + 1 byte ReportError) to convey no information, which can add up fast if we assume that batches will be large and uploads will succeed in the vast majority of cases.

The obvious point of comparison is `AggregationJobResp`, which unconditionally includes a `PrepareResp` for each report. The difference there is that in either the success or failure case, the Helper needs to convey information to the Leader (either the next prepare step or that a failure occurred). In the success case of an upload, the Client doesn't need to know anything except that it succeeded.

draft-ietf-ppm-dap.md
cjpatton, line 1492: Need a tab here (4 spaces). Likewise below.
```suggestion
    ReportID id;
```

draft-ietf-ppm-dap.md
```diff
@@ -1479,36 +1487,58 @@ struct {
...
 If the upload request is malformed, the Leader aborts with error
 `invalidMessage`.
 
 If the Leader does not recognize the task ID, then it aborts with error
 `unrecognizedTask`.
 
+Otherwise, the Leader responds with HTTP status 200, a body consisting
+of `UploadResponse` and the media type `application/dap-report-resp`. If
+the entire upload has been successfully completed, the Leader MAY respond with
+an empty body.
```
cjpatton, line 1510: I'm not in favor of this MAY. It's better I think if there's only one way to handle the response: expect an `UploadResponse`.

draft-ietf-ppm-dap.md
```diff
@@ -1479,36 +1487,58 @@ struct {
...
+
+For each report in the upload, the Leader creates `ReportUploadStatus` struct
+and adds the `ReportId` to it. If there was no error during the processing,
+the Leader sets the `error` to `none`.
```
cjpatton, line 1514: Thanks for implementing this suggestion here @akoshelev. However, after thinking this more, I think this is may not a good idea. The problem I see is that during the aggregation flow, a `ReportError error = none` is not a valid response. If there was really no report error, then the aggregator would expect a different response than a `ReportError`.

That said, I think you and @tgeoghegan have already suggested a much better idea (https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/pull/686/files#r2070551993): if a report was uploaded successfully, then simply don't include it in the `UploadResponse`.

draft-ietf-ppm-dap.md
akoshelev: it seems that we use [2 spaces](https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/blob/main/draft-ietf-ppm-dap.md?plain=1#L1100C1-L1100C3) everywhere? 

still need to fix this though

draft-ietf-ppm-dap.md
akoshelev: I was a bit worried that it may violate the assumption of upload API to support (1..N) reports. Requests to upload more than one report will have extra `public_extensions` field. Or, if we decide to always include them, requests to upload 1 report will need to omit `public_extensions` or duplicate them across two fields.

draft-ietf-ppm-dap.md
tgeoghegan: I don't agree. We use empty response bodies in the aggregation and collection interactions to indicate that aggregation, aggregate share and collection jobs are being handled asynchronously, so using an empty body here seems fine, and saves a few bytes on the wire. The encoding of an empty upload response is bigger than nothing.

draft-ietf-ppm-dap.md
tgeoghegan: I think we should hoist `public_extensions` out of `struct ReportMetadata` and up into `struct UploadRequest` (though of course we have to pay careful attention to how per-report AAD is computed to make sure all extensions are still mixed in). Suppose reports with no extensions get uploaded. Right now we have:

```
struct {
 ReportID report_id; // 16 bytes
 Time time; // 8 bytes
 Extension public_extensions<0..2^16-1>; // 2 bytes for empty
} ReportMetadata;

struct {
 ReportMetadata report_metadata; // 26 bytes
 HpkeCiphertext leader_encrypted_input_share;
 HpkeCiphertext helper_encrypted_input_share;
} Report;

struct {
  Report reports<0..2^32-1>; // 26 bytes + public share and encrypted shares per report
} UploadRequest;
```

If we hoist the public extensions out, we get (`struct Report` doesn't change):

```
struct {
    ReportID report_id; // 16 bytes
    Time time; // 8 bytes
} ReportMetadata;

struct {
    Extension public_extensions<0..2^16-1>; // 2 for empty
    Report reports<0..2^32-1>; // 24 bytes + public share and encrypted shares per report
}
```

So it seems to me we save 2 bytes per report past the first one by hoisting out public extensions.

IMO de-duplicating the repeated public extensions is the principal reason to make this change. If we don't do that, what's the motivation here?

draft-ietf-ppm-dap.md
```diff
@@ -1546,33 +1596,100 @@ will occur before aggregation.
...
+    leader_encrypted_input-share = struct {
+      config_id = 1,
+      enc = [0x0f, 0x0e, 0x0d, 0x0c, ...],
+      payload = [0x0b, 0x0a, 0x09, 0x08, ...],
+    } HpkeCiphertext,
+    helper_encrypted_input-share = struct {
+      config_id = 2,
+      enc = [0x0c, 0x0d, 0x0e, 0x0f, ...],
+      payload = [0x08, 0x00, 0x0a, 0x0b, ...],
+    } HpkeCiphertext,
```
cjpatton, line 1623: This may need to be indented one more level

draft-ietf-ppm-dap.md
```diff
@@ -4660,17 +4835,18 @@ The initial contents of this registry are listed below in {{report-error-id}}.
...
+| `0x02` | `report_replayed`        | {{basic-definitions}} of RFX XXXX |
+| `0x03` | `report_dropped`         | {{basic-definitions}} of RFX XXXX |
+| `0x04` | `hpke_unknown_config_id` | {{basic-definitions}} of RFX XXXX |
+| `0x05` | `hpke_decrypt_error`     | {{basic-definitions}} of RFX XXXX |
+| `0x06` | `vdaf_prep_error`        | {{basic-definitions}} of RFX XXXX |
+| `0x07` | `task_expired`           | {{basic-definitions}} of RFX XXXX |
+| `0x08` | `invalid_message`        | {{basic-definitions}} of RFX XXXX |
+| `0x09` | `report_too_early`       | {{basic-definitions}} of RFX XXXX |
+| `0xA`  | `task_not_started`       | {{basic-definitions}} of RFX XXXX |
+| `0xB`  | `outdated_config`        | {{basic-definitions}} of RFX XXXX |
```
cjpatton, line 4849:
```suggestion
| `0x0A`  | `task_not_started`       | {{basic-definitions}} of RFX XXXX |
| `0x0B`  | `outdated_config`        | {{basic-definitions}} of RFX XXXX |
```

draft-ietf-ppm-dap.md
cjpatton: Well, there's an important difference I think.

During aggregation, an empty response indicates async handling, which means the Leader needs to poll later. But there is no async processing of the upload request. Should there be?

tgeoghegan: Sure, it's not exactly the same, but the point is that we already assign semantic meaning to the absence of a body. Further, the real issue here is that there's no point in constructing and parsing a message that conveys no information (the message would just be 4 0 values bytes indicating an empty vector of reports) when literally nothing would serve just as well.

cjpatton: What I think is odd about this is that we have to special case the empty body. That is the code is something like:

```rust
fn parse_request_body(bytes: &[u8]) -> Vec<ReportStatus> {
   if bytes.empty() {
      return Vec::new();
   }
   parse_report_status_list(bytes)
}
```

But tbh I don't feel very strongly about this. I'm ok with whatever we we do here.

draft-ietf-ppm-dap.md
jcjones: If the public extensions really are repeated, HTTP compression should handle them fine, as long as the report structures are all aligned the same way anyway. Seems like that should be testable, and easier than this?

tgeoghegan: Request body compression is a really good point I hadn't considered. I think it makes more sense to rely on that than to allow for extensions in yet another place.

martinthomson: I like the basic idea @cjpatton described for compression here  (I would call it `common_extensions` rather than `public_extensions`), but do the reports really need to have a length prefix?  That could be a compromise between different approaches.  It's not like we can't rely on the outer message framing (like Content-Length) as an indication of the end of the message.  It saves 4 bytes.  And it opens up the possibility of indeterminate length requests, if an implementation is able to handle that.

draft-ietf-ppm-dap.md
```diff
@@ -1492,36 +1522,56 @@ struct {
...
 ~~~
 
+~~~ tls-presentation
+struct {
+  ReportID id;
+  ReportError error;
+} ReportUploadStatus;
+
+struct {
+  ReportUploadStatus status<0..2^32-1>;
```
martinthomson, line 1532: Same comment as above.  There's no need for another length on this.

martinthomson (commented):
This seems to be moving in a good direction.

draft-ietf-ppm-dap.md
cjpatton: Removing the length prefix sounds good to me, but I don't know how to spell this in TLS' presentation language.@martinthomson do you have a suggestion?

We could just say in prose: "the payload of the HTTP request is a sequence of `Report` structures concatenated together". This is a little ugly but I suppose it would work.

martinthomson: From memory:

```
~~~ tls-syntax
struct Whatever {
  HeaderType header;
  ValueType values[value_count];
};
~~~
```

That is, you use the fixed-size vector syntax, but use a variable instead of a number.

> The size of the vector may be specified at documentation time or left unspecified until runtime.  

-- https://datatracker.ietf.org/doc/html/rfc8446#section-3.4

Of course, this isn't perfectly clear there, because it doesn't say how to do that.

cjpatton: That works for me!

tgeoghegan (commented):
I'm not crazy about requiring outside context (the `Content-Length`) to parse `UploadRequest` and `UploadResponse` messages, but it does have the merits argued by Martin, and it seems like that's where the rough consensus is going.

I have a some changes to the text here in https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/commit/4867378c805dfb961aacd4be0c975d26946c1922 that address outstanding review comments and clean up some other problems.

I've (mostly) implemented this in Janus [here](https://github.com/divviup/janus/pull/3831). It wasn't too bad to implement, with the bulk of the work being updating test cases to look for appropriate errors inside an `UploadResponse` instead of a problem document.

My only gripe is that there is some awkwardness introduced by having to convey many statuses in a single HTTP request/response. For example, what should a leader do if every single report in the `UploadRequest` is somehow invalid? I think the right thing is to return HTTP 200 OK and have the body include `ReportUploadStatus` describing each failure, because the exchange of HTTP messages was successful. We should probably make that clear in the text.

draft-ietf-ppm-dap.md
akoshelev: I've changed <0...2^32-1> to [2^32-1] as it seems the most appropriate syntax for this

> Here, T' occupies n bytes in the data stream, where n is a multiple
   of the size of T.

akoshelev:
One remaining unresolved issue is the common extensions for report uploads. One way to save on the wire is to rely on HTTP compression as suggested by @jcjones, another way is to be explicit about re-use of these fields at the DAP protocol layer and specify them in the upload request message.

I may have a slight preference towards the HTTP compression as it seems a bit more straightforward to use and helps prevent implementation errors including extensions into AAD, but I am open to other suggestions as well.

draft-ietf-ppm-dap.md
martinthomson: > I've changed <0...2^32-1> to [2^32-1]

The brackets indicate a fixed size, so that says that you will always have 4GiB of data. That's probably not what you intended.  You need a variable, like: `Whatever thing[VARIABLE_NAME_HERE];`

akoshelev: changed this to [report_count]/[status_count]

draft-ietf-ppm-dap.md
```diff
@@ -1446,6 +1446,9 @@ Each upload request contains a sequence of `Report` messages constructed as foll
 
 * `helper_encrypted_input_share` is the Helper's encrypted input share.
 
+The length of the sequence is specified at runtime, but it must not exceed
+2^32 - 1 bytes.
```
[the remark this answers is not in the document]
cjpatton: I don't think the 2GB limit we impose elsewhere in the draft applies here. The limit exists only because we need to specify the length of the length prefix of fields in the wire format of messages. since we're not encoding the length prefix, we don't need to specify an upper bound ... at least not for the same reason.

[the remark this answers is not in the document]
cjpatton: I think we should just remove this paragraph.

draft-ietf-ppm-dap.md
```diff
@@ -1395,8 +1419,14 @@ struct {
   HpkeCiphertext leader_encrypted_input_share;
   HpkeCiphertext helper_encrypted_input_share;
 } Report;
+
+struct {
+  Report reports[report_count];
```
tgeoghegan, line 1424: `report_count` is misleading. Per RFC 8446 vector syntax ([1](https://datatracker.ietf.org/doc/html/rfc8446#section-3.4)), the value in square brackets is the length *in bytes*, not the count of elements in the vector. Let's use `content_length` as in https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/commit/4867378c805dfb961aacd4be0c975d26946c1922.

draft-ietf-ppm-dap.md
```diff
@@ -1498,30 +1531,57 @@ If the upload request is malformed, the Leader aborts with error
...
+is shown below:
+
+~~~ tls-presentation
+struct {
+  ReportID id;
+  ReportError error;
+} ReportUploadStatus;
+
+struct {
+  ReportUploadStatus status[status_count];
```
tgeoghegan, line 1545: As above, `status_count` is misleading because the value here is a number of bytes. Let's use `content_length`. And we should make it it clear that the response's `Content-Length` header MUST match the value used here.

draft-ietf-ppm-dap.md
```diff
@@ -1498,30 +1531,57 @@ If the upload request is malformed, the Leader aborts with error
...
+} UploadResponse;
+~~~
+
+The Leader only includes reports that failed processing in the response.
+Reports that are accepted do not have a response.
+
+Reports in the response follow the same order as specified in the request.
```
tgeoghegan, line 1552:
```suggestion
Reports in the response MUST appear in the same order as in the request.
```

draft-ietf-ppm-dap.md
```diff
@@ -1498,30 +1531,57 @@ If the upload request is malformed, the Leader aborts with error
...
+
+For each report that failed to upload, the Leader creates a
+`ReportUploadStatus` struct and includes the `ReportId` from the input and a
+`ReportError` {{basic-definitions}} that describes the failure. The length of
```
tgeoghegan, line 1556:
```suggestion
`ReportError` ({{basic-definitions}}) that describes the failure. The length of
```

draft-ietf-ppm-dap.md
```diff
@@ -1498,30 +1531,57 @@ If the upload request is malformed, the Leader aborts with error
...
+this sequence is always less than or equal to the length of the upload sequence.
+
 If the Leader does not recognize the `config_id` in the encrypted input share,
-it aborts with an error of type `outdatedConfig`. When the Client receives an
-`outdatedConfig` error, it SHOULD invalidate any cached `HpkeConfigList` and
+it sets the error field to `outdated_config`
+(see {{aggregation-helper-init}}). When the Client receives an
+`outdated_config` error, it SHOULD invalidate any cached `HpkeConfigList` and
```
tgeoghegan, line 1562:
```suggestion
it sets the error field to `outdated_config`. When the Client receives an
`outdated_config` error, it SHOULD invalidate any cached `HpkeConfigList` and
```
Is this reference correct? The `ReportError` enum was moved up in the document. Actually, since the preceding paragraph already points to {{basic-definitions}}, let's delete this altogether.

draft-ietf-ppm-dap.md
```diff
@@ -1498,30 +1531,57 @@ If the upload request is malformed, the Leader aborts with error
...
 
 The Leader may need to buffer reports while waiting to aggregate them (e.g.,
 while waiting for an aggregation parameter from the Collector; see
 {{collect-flow}}). The Leader SHOULD NOT accept reports whose timestamps are too
 far in the future. Implementors MAY provide for some small leeway, usually no
 more than a few minutes, to account for clock skew. If the Leader rejects a
-report for this reason, it SHOULD abort with error `reportTooEarly`. In this
-situation, the Client MAY re-upload the report later on.
+report for this reason, it SHOULD indicate it with `report_too_early` error.
+In this situation, the Client MAY re-upload the report later on.
```
tgeoghegan, line 1584: Let's use consistent language for error handling to avoid implying that the server should do different things in different cases. "set the error field to `XXX` for the corresponding report in the response" should work

draft-ietf-ppm-dap.md
```diff
@@ -1546,33 +1606,100 @@ will occur before aggregation.
 
 #### Example
 
+Successful upload
+
 ~~~ http
 POST /leader/tasks/8BY0RzZMzxvA46_8ymhzycOB9krN-QIGYvg_RsByGec/reports
 Host: example.com
-Content-Type: application/dap-report
+Content-Type: application/dap-upload-req
```
tgeoghegan, line 1614: The examples should include `Content-Length` since we're going to have a MUST about the requests and responses including them.

draft-ietf-ppm-dap.md
```diff
@@ -1114,6 +1116,28 @@ in different situations.
...
+  vdaf_prep_error(6),
+  task_expired(7),
+  invalid_message(8),
+  report_too_early(9),
+  task_not_started(10),
+  outdated_config(11),
+  (255)
+} ReportError;
+~~~
+
```
tgeoghegan, line 1139:
```suggestion
```

draft-ietf-ppm-dap.md
```diff
@@ -1379,7 +1403,7 @@ encoded([
 ### Upload Request
 
 Clients upload reports by sending a POST to `{leader}/tasks/{task-id}/reports`.
-The body is a `Report`, with media type "application/dap-report", structured as
+The body is a `UploadRequest`, with media type "application/dap-upload-req", structured as
```
tgeoghegan, line 1406:
```suggestion
The body is an `UploadRequest`, with media type "application/dap-upload-req", structured as
```

tgeoghegan:
> One remaining unresolved issue is the common extensions for report uploads. One way to save on the wire is to rely on HTTP compression as suggested by @jcjones, another way is to be explicit about re-use of these fields at the DAP protocol layer and specify them in the upload request message.
> 
> I may have a slight preference towards the HTTP compression as it seems a bit more straightforward to use and helps prevent implementation errors including extensions into AAD, but I am open to other suggestions as well.

I think we should leave compression to the HTTP layer because we won't do a better job at the DAP layer.

draft-ietf-ppm-dap.md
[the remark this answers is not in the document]
tgeoghegan: No, I think this is a misinterpretation of RFC 8446 vectors. https://datatracker.ietf.org/doc/html/rfc8446#section-3.4:

```
T T'[n];

   Here, T' occupies n bytes in the data stream, where n is a multiple
   of the size of T.
```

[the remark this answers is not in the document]
martinthomson: I'm not happy with mandating the use of Content-Length.  We discourage it in HTTP/2 and HTTP/3 and it would discourage the use of chunked transfer encoding in HTTP/1.1, which is a perfectly good thing to use.

Yes, it deprives servers of certainty.  They need to learn to get over that.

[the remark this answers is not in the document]
tgeoghegan: @martinthomson OK, I'm willing to take your word on this being a problem in HTTP/2 or 3. If there's no Content-Length and no length in the protocol-level message, how then are servers supposed to know when the message ends?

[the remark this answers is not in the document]
tgeoghegan: Oh, I think I get what you're saying: a server _could_ use `Content-Length` if it happens to be using HTTP 1.1 without chunked transfer, but we don't want to force all possible configurations to do so. So I guess we'd have to write some text here saying roughly "the length of the vector has to be determined from the transport layer, which could be Content-Length in some HTTP/1.1 settings, END_STREAM in HTTP/2 or something else"?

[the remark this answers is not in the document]
martinthomson: Both HTTP/2 and HTTP/3 have a clear end of stream indication, yes.  The [latest text](https://datatracker.ietf.org/doc/html/rfc9113#section-8.1.1-2) in RFC 9113 doesn't exactly forbid the use of `Content-Length`, but it does imply strongly that you can and should avoid it.  At a minimum, it's redundant.  At worst, it disagrees with the other parts of the protocol, causing errors.

> a server could use Content-Length if it happens to be using HTTP 1.1 without chunked transfer, but we don't want to force all possible configurations to do so

Precisely.  What you said is where I'd look to go.

akoshelev:
> I think this is nearly there. I have changes that address Martin's point about allowing various HTTP message length indications, address rebase conflicts and clean up a few other nits. @akoshelev if you're OK with it, I can (force) push those changes onto this branch so we can all review and then move ahead with this change.

I was planning to do the rebase this weekend, but if you have those changes ready - that's even better! Feel free to update this PR

jcjones (commented):
While I'm not an official reviewer, I find the changes here straightforward and clearly specified.

draft-ietf-ppm-dap.md
cjpatton: @tgeoghegan is the answer "no" here given the discussion on https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/pull/686#discussion_r2108267711?

cjpatton (commented):
This is in good shape. Once rebased I'll go ahead and approve.

I have two follow up questions, which might address in a future PR:

1. In the operational considerations, should we should suggest using HTTP compression when doing bulk upload?
2. Should this interaction be allowed to be asynchronous, just as every other interaction is?

cjpatton (approved):
Whoops, I just noticed you already rebased! Approved modulo nits.

draft-ietf-ppm-dap.md
tgeoghegan: Is the answer "no" to what?

cjpatton: sorry, I mean this example shouldn't include the Content-Length header.

tgeoghegan:
Merged! Thanks to @akoshelev for this valuable contribution and to everyone who helped refine and improve it. n.b. it's not too late to make further changes in case there are still typos lurking.

PR #699, opened 2025-05-05
Use minimum length to denote non-empty vectors

### LINKED ISSUE #687: Minimum vector lengths
S 3.2.1 says:

>  {{Section 3.4 of !RFC8446}} describes a syntax for variable-length vectors where
  a range of permitted lengths is provided. This document always specifies the
  lower length limit of variable-length vectors to be `0`.

And yet there are settings where the lower limit is not 0. For instance, S 4.5.1 sayus

>    An Aggregator responds with an HpkeConfigList, with media type
   "application/dap-hpke-config-list".  The HpkeConfigList contains one
   or more HpkeConfigs in decreasing order of preference.  This allows
   an Aggregator to support multiple HPKE configurations and multiple
   sets of algorithms simultaneously.

And yet:
```
   HpkeConfig HpkeConfigList<0..2^16-1>;
```

So, clearly it's not actually legal to have this list be empty. 

I know that it's irritating to have to compute the actual lower bounds when it's some struct, so perhaps you are trying to say "we always use 0 in the syntax" but you have to enforce limits anyway? However, there are other places where it's possibly legal to actually have an empty vector, for instance `ReportMetadata.public_extensions`.

cjpatton:
Closes #687.

In the definition of protocol messages, if a vector is expected to be non-empty, then set its minimum length to be 1.

### DIFF
--- a/draft-ietf-ppm-dap.md
+++ b/draft-ietf-ppm-dap.md
@@ -902,12 +902,6 @@ Section 3}} to define messages in the protocol. Encoding and decoding of these
 messages as byte strings also follows {{!RFC8446}}. We enumerate the exceptions
 below.
 
-### Variable-Length Vectors
-
-{{Section 3.4 of !RFC8446}} describes a syntax for variable-length vectors where
-a range of permitted lengths is provided. This document always specifies the
-lower length limit of variable-length vectors to be `0`.
-
 ### Structure Variants
 
 {{Section 3.7 of !RFC8446}} defines a syntax for structure fields whose values
@@ -1104,8 +1098,8 @@ uint8 HpkeConfigId;
 
 struct {
   HpkeConfigId config_id;
-  opaque enc<0..2^16-1>;
-  opaque payload<0..2^32-1>;
+  opaque enc<1..2^16-1>;
+  opaque payload<1..2^32-1>;
 } HpkeCiphertext;
 ~~~
 
@@ -1316,7 +1310,7 @@ support multiple HPKE configurations and multiple sets of algorithms
 simultaneously.
 
 ~~~ tls-presentation
-HpkeConfig HpkeConfigList<0..2^16-1>;
+HpkeConfig HpkeConfigList<1..2^16-1>;
 
 struct {
   HpkeConfigId id;
@@ -1326,7 +1320,7 @@ struct {
   HpkePublicKey public_key;
 } HpkeConfig;
 
-opaque HpkePublicKey<0..2^16-1>;
+opaque HpkePublicKey<1..2^16-1>;
 uint16 HpkeAeadId;
 uint16 HpkeKemId;
 uint16 HpkeKdfId;
@@ -1462,7 +1456,7 @@ The Client then wraps each input share in the following structure:
 ~~~ tls-presentation
 struct {
   Extension private_extensions<0..2^16-1>;
-  opaque payload<0..2^32-1>;
+  opaque payload<1..2^32-1>;
 } PlaintextInputShare;
 ~~~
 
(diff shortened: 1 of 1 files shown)


draft-ietf-ppm-dap.md
```diff
@@ -1310,7 +1304,7 @@ support multiple HPKE configurations and multiple sets of algorithms
 simultaneously.
 
 ~~~ tls-presentation
-HpkeConfig HpkeConfigList<0..2^16-1>;
+HpkeConfig HpkeConfigList<1..2^16-1>;
```
cjpatton, line 1307: We can compute a tighter bound here.

draft-ietf-ppm-dap.md
```diff
@@ -1835,7 +1829,7 @@ struct {
 struct {
   opaque agg_param<0..2^32-1>;
   PartialBatchSelector part_batch_selector;
-  PrepareInit prepare_inits<0..2^32-1>;
+  PrepareInit prepare_inits<1..2^32-1>;
```
cjpatton, line 1832: We can compute a tighter bound here.

draft-ietf-ppm-dap.md
```diff
@@ -2074,7 +2068,7 @@ which is structured as follows:
 
 ~~~ tls-presentation
 struct {
-  PrepareResp prepare_resps<0..2^32-1>;
+  PrepareResp prepare_resps<1..2^32-1>;
```
cjpatton, line 2071: We can compute a tighter bound here.

draft-ietf-ppm-dap.md
```diff
@@ -2265,7 +2259,7 @@ Next, the Leader sends a POST to the aggregation job with media type
 ~~~ tls-presentation
 struct {
   uint16 step;
-  PrepareContinue prepare_continues<0..2^32-1>;
+  PrepareContinue prepare_continues<1..2^32-1>;
```
cjpatton, line 2262: We can compute a tighter bound here.

cjpatton:
Rebased.

draft-ietf-ppm-dap.md
```diff
@@ -1843,7 +1837,7 @@ struct {
 struct {
   opaque agg_param<0..2^32-1>;
   PartialBatchSelector part_batch_selector;
-  PrepareInit prepare_inits<0..2^32-1>;
+  PrepareInit prepare_inits<41..2^32-1>;
```
tgeoghegan, line 1840:
```suggestion
  PrepareInit prepare_inits<46..2^32-1>;
```

draft-ietf-ppm-dap.md
```diff
@@ -1843,7 +1837,7 @@ struct {
 struct {
   opaque agg_param<0..2^32-1>;
   PartialBatchSelector part_batch_selector;
-  PrepareInit prepare_inits<0..2^32-1>;
+  PrepareInit prepare_inits<46..2^32-1>;
```
cjpatton, line 1840: I got 44:
```
﻿﻿PrreapreInit {
  ReportShare {
    ReportMetadata {
      ReportID 16
      Time 8
      Extensions 2 + 0
    }
    opaque 4 + 0
    HpkeCiphertext {
      HpkeConfigID 1
      opaque 2 + 1
      opaque 4 + 1
    }
  }
  opaque 4 + 1
}
```

tgeoghegan: You are correct, I misinterpreted the public extensions and public share fields as being at least 1 byte long, but they're both allowed to be empty.

cjpatton: Will update, then merge.

cjpatton: You got there first!

Issue #689, opened 2025-05-05
Can you really store a new state of Rejected() or Finished()


ekr:
>        If outbound != None, then the Leader stores state and outbound
       and proceeds to Section 4.6.3.1.  If outbound == None, then the
       preparation process is complete: either state == Rejected(), in
       which case the Leader rejects the report and removes it from the
       candidate set; or state == Finished(out_share), in which case
       preparation is complete and the Leader commits out_share as
       described in Section 4.6.3.3.  If commitment fails, then the
       Leader rejects the report and removes it from the candidate set.

This suggests I might store `(Rejected, [stuff])` which seems unlikely.

See also: https://github.com/cfrg/draft-irtf-cfrg-vdaf/issues/542

tgeoghegan:
Hopefully this text can be made simpler once https://github.com/cfrg/draft-irtf-cfrg-vdaf/issues/542 is resolved. I think we should make a PR on draft-ietf-ppm-dap that adopts the changes in https://github.com/cfrg/draft-irtf-cfrg-vdaf/pull/543 to make sure this ends up where we want.

cjpatton:
I've got a PR up for this: https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/pull/705

I think this is a good change overall. The PR includes some orthogonal improvements that we can split out into separate PRs if we want to.

PR #702, opened 2025-05-05
Minor text improvements from my review.

ekr:

PR #704, opened 2025-05-07
Editorial and a bit of renaming

cjpatton:
Just a few things i noticed while working on a PR for #689. The PR is split into minimal commits.

### DIFF
--- a/draft-ietf-ppm-dap.md
+++ b/draft-ietf-ppm-dap.md
@@ -1920,8 +1920,8 @@ The Leader proceeds as follows with each report:
    job.
 
 Since VDAF preparation completes in a constant number of rounds, it will never
-be the case that some reports in an aggregation job are completed and others are
-not.
+be the case that preparation is complete for some of the reports in an
+aggregation job but not others.
 
 If the Leader fails to process the response from the Helper, for example because
 of a transient failure such as a network connection failure or process crash,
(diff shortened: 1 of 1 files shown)


draft-ietf-ppm-dap.md
```diff
@@ -1845,9 +1846,9 @@ This message consists of:
   initializing an aggregation job, the Leader MUST validate the parameter as
   described in {{agg-param-validation}}.
 
-* `part_batch_selector`: The "partial batch selector" used by the Aggregators to
-  determine how to aggregate each report. Its contents depends on the indicated
-  batch mode. This field is called the "partial" batch selector because
+* `part_batch_selector`: The "partial batch selector" used by the Aggregators
+  to determine how to aggregate each report. Its contents depends on the batch
+  mode of the task. This field is called the "partial" batch selector because
```
tgeoghegan, line 1851: We put the batch mode in `PartialBatchSelector` precisely so that implementations wouldn't need to go load a task from storage in order to parse `AggregationJobInitReq`, so _actually_ (🤓) I think it's more correct to point the reader to `part_batch_selector.batch_mode` than to the task. Perhaps there's a better way to do that.

"Its contents depend on the batch mode indicated by `part_batch_selector.batch_mode`."?

note also we should remove the "s" at the end of "depends" in this sentence.

draft-ietf-ppm-dap.md
```diff
@@ -1912,7 +1913,7 @@ The Leader proceeds as follows with each report:
    job.
 
 Since VDAF preparation completes in a constant number of rounds, it will never
-be the case that some reports are completed and others are not.
+be the case that preparation is complete for some reports but not others.
```
tgeoghegan, line 1916: I think this will conflict with #702. Should we resolve and merge that PR before this one?

draft-ietf-ppm-dap.md
cjpatton: I'll just drop it.

tgeoghegan (approved):

tgeoghegan (approved):

PR #705, opened 2025-05-07
Align ping pong API with upstream changes

### LINKED ISSUE #689: Can you really store a new state of Rejected() or Finished()

cjpatton:
Stacked on #704.
Closes #689.

As of https://github.com/cfrg/draft-irtf-cfrg-vdaf/pull/543, the ping pong API for VDAFs has been modified by merging the outbound message into the report state. Adjust our calls to this API accordingly.

While working this change I notice a small problem. We say the Leader must reject a report if commitment fails. However in every case the Leader would reject, the Helper has already committed. Thus such a failure always leads to a batch mismatch.

Resolve this by ensuring commitment doesn't fail. It's sufficient to require the Leader to check for `report_replayed` and `batch_collected` prior to selecting the initial candidate set.

Finally, a couple more minor improvements:

1. Clarify that the Leader can retry if it encounters `report_too_early` during aggregation continuation.

1. Add a note about rejection agreement. There is a case where the Leader might reject after the Helper commits, but this shouldn't happen if (1) the VDAF is robust and (2) the Helper is not misbehaving.

### DIFF
--- a/draft-ietf-ppm-dap.md
+++ b/draft-ietf-ppm-dap.md
@@ -1689,6 +1689,14 @@ The number of steps, and the type of the responses, depends on the VDAF. The
 message structures and processing rules are specified in the following
 subsections.
 
+Each Aggregator maintains some state for each report. A state transition is
+triggered by receiving a message from the Aggregator's peer. Eventually this
+process results in a terminal state, either rejecting the report or recovering
+an output share. Once a report has reached a terminal state, no more messages
+will be processed for it. There are four possible states (see {{Section 5.7 of
+!VDAF}}: `Continued`, `FinishedWithOutbound`, `Finished` and `Rejected`. The
+first two states include an outbound message to be processed by the peer.
+
 The Helper can either process each step synchronously, meaning it computes each
 preparation step before producing a response to the Leader's HTTP request, or
 asynchronously, meaning it responds immediately and defers processing to a
@@ -1758,8 +1766,24 @@ The Leader and Helper initialization behavior is detailed below.
 
 #### Leader Initialization {#leader-init}
 
+~~~ aasvg
+                     PrepareResp
+Report --> Continued -----+----> Continued
+                          |
+                          +----> FinishedWithOutbound
+                          |
+                          +----> Finished (*commit)
+                          |
+                          +----> Rejected (*reject)
+~~~
+{: #leader-init-state title="Leader state transition triggered by aggregation initialization. (\*) indicates a terminal state." }
+
 The Leader begins an aggregation job by choosing a set of candidate reports that
 belong to the same task and a job ID which MUST be unique within the task.
+
+First, the Leader MUST ensure each report in the candidate set can be committed
+per the criteria detailed in {{batch-buckets}}. If a report cannot be committed,
+then the Leader rejects it and removes it from the candidate set.
 
 Next, the Leader decrypts each of its report shares as described in
 {{input-share-decryption}}, then checks input share validity as described in
@@ -1769,7 +1793,7 @@ and removes it from the candidate set.
 For each report the Leader executes the following procedure:
 
 ~~~ pseudocode
-(state, outbound) = Vdaf.ping_pong_leader_init(
+state = Vdaf.ping_pong_leader_init(
     vdaf_verify_key,
     "dap-15" || task_id,
     agg_param,
@@ -1790,13 +1814,14 @@ where:
 * `plaintext_input_share` is the Leader's `PlaintextInputShare`
 
 `ping_pong_leader_init` is defined in {{Section 5.7.1 of !VDAF}}. This process
-determines the initial per-report `state`, as well as the initial `outbound`
-message to the Helper. If `state` is of type `Rejected`, then the report is
-rejected and removed from the candidate set, and no message is sent to the
-Helper for this report.
+determines the initial per-report `state`. If `state` is of type `Rejected`
+({{Section 5.7 of !VDAF}}, then the report is rejected and removed from the
+candidate set, and no message is sent to the Helper for this report.
 
-If `state` is of type `Continued`, then the Leader constructs a `PrepareInit`
-structure for that report.
+Otherwise, if `state` is of type `Continued` (no other state is reachable at
+this point), then the state includes an outbound message denoted
+`state.outbound`. The Leader uses it to construct a `PrepareInit` structure for
+that report.
 
 ~~~ tls-presentation
 struct {
@@ -1819,10 +1844,7 @@ This message consists of:
 
   * `report_share.encrypted_input_share`: The Helper's encrypted input share.
 
-  * `payload`: The `outbound` message computed by the previous step.
-
-It is not possible for `state` to be of type `Finished` during Leader
-initialization.
+  * `payload`: The outbound message, set to `state.outbound`.
 
 Once all the report shares have been initialized, the Leader creates an
 `AggregationJobInitReq` message containing the `PrepareInit` structures
@@ -1879,10 +1901,10 @@ The Leader proceeds as follows with each report:
 1. If the inbound prep response has type "continue", then the Leader computes
 
    ~~~ pseudocode
-   (state, outbound) = Vdaf.ping_pong_leader_continued(
+   state = Vdaf.ping_pong_leader_continued(
        "dap-15" || task_id,
        agg_param,
-       prev_state,
+       state,
        inbound,
    )
    ~~~
@@ -1892,26 +1914,35 @@ The Leader proceeds as follows with each report:
    * `task_id` is the task ID
    * `agg_param` is the VDAF aggregation parameter provided by the Collector
      (see {{collect-flow}})
-   * `prev_state` is the state computed earlier by calling
-     `Vdaf.ping_pong_leader_init` or `Vdaf.ping_pong_leader_continued`
-   * `inbound` is the message in the `PrepareResp`
+   * `state` is the report's initial prep state
+   * `inbound` is the payload of the `PrepareResp`
 
-   If `outbound != None`, then the Leader stores `state` and `outbound` and
-   proceeds to {{aggregation-leader-continuation}}. If `outbound == None`, then
-   the preparation process is complete: either `state == Rejected()`, in which
-   case the Leader rejects the report and removes it from the candidate set; or
-   `state == Finished(out_share)`, in which case preparation is complete and the
-   Leader commits `out_share` as described in {{batch-buckets}}. If commitment
-   fails, then the Leader rejects the report and removes it from the candidate
-   set.
+   If the new `state` has type `Continued` or `FinishedWithOutbound`,
+   then there is at least one more outbound message to send before preparation
+   is complete. The Leader stores `state` and proceeds as in
+   {{aggregation-leader-continuation}}.
 
-1. Else if the type is "reject", then the Leader rejects the report and removes
-   it from the candidate set. The Leader MUST NOT include the report in a
-   subsequent aggregation job, unless the error is `report_too_early`, in which
-   case the Leader MAY include the report in a subsequent aggregation job.
+   Else if the new `state` has type `Finished`, then preparation is complete and
+   the state includes an output share, denoted `state.out_share`. The Leader
+   commits to `state.out_share` as described in {{batch-buckets}}.
 
-1. Else the type is invalid, in which case the Leader MUST abort the aggregation
-   job.
+   Else if `state` has type `Rejected`, then the Leader rejects the
+   report and removes it from the candidate set.
+
+   Note on rejection agreement: rejecting at this point would result in a batch
+   mismatch if the Helper had already committed to its output share. This is
+   impossible due to the robustness of the VDAF: if the underlying measurement
+   were invalid, then the Helper would have indicated rejection in its
+   response.
+
+1. Else if the `PrepareResp` has type "reject", then the Leader rejects the
+   report and removes it from the candidate set. The Leader MUST NOT include
+   the report in a subsequent aggregation job, unless the report error is
+   `report_too_early`, in which case the Leader MAY include the report in a
+   subsequent aggregation job.
+
+1. Otherwise the inbound message type is invalid for the Leader's current
+   state, in which case the Leader MUST abort the aggregation job.
 
 Since VDAF preparation completes in a constant number of rounds, it will never
 be the case that preparation is complete for some of the reports in an
@@ -1923,6 +1954,15 @@ the Leader SHOULD re-send the original request unmodified in order to attempt
 recovery (see {{aggregation-step-skew-recovery}}).
 
 #### Helper Initialization {#aggregation-helper-init}
+
+~~~ aasvg
+PrepareInit --+--> Continued
+              |
+              +--> FinishedWithOutbound (*commit)
+              |
+              +--> Rejected (*reject)
+~~~
+{: #helper-init-state title="Helper state transition triggered by aggregation initialization. (\*) indicates a terminal state." }
 
 The Helper begins an aggregation job when it receives an `AggregationJobInitReq`
 message from the Leader. For each `PrepareInit` in this message, the Helper
@@ -1999,8 +2039,8 @@ struct {
 } PrepareResp;
 ~~~
 
-The Helper processes each of the remaining report shares in turn:
-It decrypts each of its remaining report shares as described in
+The Helper processes each of the remaining report shares in turn.
+First, the Helper decrypts each report share as described in
 {{input-share-decryption}}, then checks input share validity as described in
 {{input-share-validation}}. If either decryption of validation fails, the Helper
 sets the corresponding outbound preparation response to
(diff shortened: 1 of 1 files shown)


draft-ietf-ppm-dap.md
```diff
@@ -1758,9 +1767,28 @@ The Leader and Helper initialization behavior is detailed below.
...
+
 The Leader begins an aggregation job by choosing a set of candidate reports that
 belong to the same task and a job ID which MUST be unique within the task.
 
+First, the Leader MUST ensure each report in the candidate set can be
+committed. That is: the Leader MUST ensure it has not committed a report with
+the same ID and task; and the Leader MUST ensure the batch bucket to which the
+report belongs has not been collected. See {{batch-buckets}} for details. If a
+report is not cannot be committed, then the Leader rejects it and removes it
+from the candidate set.
```
cjpatton, line 1790: Leader MUST ensure commitment won't fail. This amounts to resolving race conditions across aggregation jobs and collection jobs, both of which are already operational requirements.

draft-ietf-ppm-dap.md
```diff
@@ -1819,10 +1847,7 @@ This message consists of:
 
   * `report_share.encrypted_input_share`: The Helper's encrypted input share.
 
-  * `payload`: The `outbound` message computed by the previous step.
-
-It is not possible for `state` to be of type `Finished` during Leader
-initialization.
```
cjpatton, line 1825: Already implied by text added above

draft-ietf-ppm-dap.md
```diff
@@ -1891,26 +1916,34 @@ The Leader proceeds as follows with each report:
    * `task_id` is the task ID
    * `agg_param` is the VDAF aggregation parameter provided by the Collector
      (see {{collect-flow}})
-   * `prev_state` is the state computed earlier by calling
-     `Vdaf.ping_pong_leader_init` or `Vdaf.ping_pong_leader_continued`
```
cjpatton, line 1895: Consistency: Elsewhere we write `state = Vdaf.ping_pong(state)` rather than `state = Vdaf.ping_pong(prev_state)`

draft-ietf-ppm-dap.md
```diff
@@ -1891,26 +1916,34 @@ The Leader proceeds as follows with each report:
...
-   * `inbound` is the message in the `PrepareResp`
-
-   If `outbound != None`, then the Leader stores `state` and `outbound` and
-   proceeds to {{aggregation-leader-continuation}}. If `outbound == None`, then
-   the preparation process is complete: either `state == Rejected()`, in which
-   case the Leader rejects the report and removes it from the candidate set; or
-   `state == Finished(out_share)`, in which case preparation is complete and the
-   Leader commits `out_share` as described in {{batch-buckets}}. If commitment
-   fails, then the Leader rejects the report and removes it from the candidate
-   set.
```
cjpatton, line 1905: Rejection does nothing because the Helper has already committed. This will always result in a batch mismatch, unless the Leader has already ensured that the report was not replayed and the bucket was not collected.

draft-ietf-ppm-dap.md
```diff
@@ -1997,7 +2039,7 @@ struct {
 } PrepareResp;
 ~~~
 
-Next the Helper decrypts each of its remaining report shares as described in
```
cjpatton, line 2000: This is the first step of computing `AggregationJobResp`, not an intermediate step.

draft-ietf-ppm-dap.md
```diff
@@ -2046,23 +2087,27 @@ variant {
...
 
-If `state == Finished(out_share)`, then the Helper commits to `out_share` as
-described in {{batch-buckets}}. If commitment fails with some `ReportError`
-(e.g., the report was replayed or its batch bucket was collected), then instead
-of the `continue` response above, the Helper responds with
+Else if `state` has type `FinishedWithOutbound`, then the Helper commits to
+`state.out_share` as described in {{batch-buckets}}. If commitment fails with
+some `ReportError` (e.g., the report was replayed or its batch bucket was
+collected), then instead of the `continue` response above, the Helper responds
+with
```
cjpatton, line 2110: @tgeoghegan in a later PR it would be good to avoid

> then instead of the [...] response above, the Helper responds with

Ideally the text matches chronologically what we would do in the code.

draft-ietf-ppm-dap.md
```diff
@@ -2297,36 +2362,50 @@ MUST abort the aggregation job.
...
    ~~~
 
-   where `task_id` is the task ID and `inbound` is the message payload. If
-   `outbound != None`, then the Leader stores `state` and `outbound` and
-   proceeds to another continuation step. If `outbound == None`, then the
-   preparation process is complete: either `state == Rejected()`, in which case
-   the Leader rejects the report and removes it from the candidate set or
-   `state == Finished(out_share)`, in which case the Leader commits to
-   `out_share` as described in {{batch-buckets}}. If commitment fails, then the
-   Leader rejects the report and removes it from the candidate set.
```
cjpatton, line 2319: Here again, I believe rejection has no affect and that a batch mismatch is unavoidable ... unless the Leader prevents it by being careful about how it chooses candidate sets.

draft-ietf-ppm-dap.md
```diff
@@ -2297,36 +2362,50 @@ MUST abort the aggregation job.
...
+   where `task_id` is the task ID and `inbound` is the payload of the inbound `PrepareResp`.
+   It then processes the next state transition:
+
+   * If the type of the new `state` is `Continued` or
+     `FinishedWithOutbound`, then the Leader stores `state` and proceeds
+     to the next continuation step.
 
-1. Else if the type is "finish" and `state == Finished(out_share)`, then
-   preparation is complete and the Leader stores the output share for use in
-   the collection interaction ({{collect-flow}}).
```
cjpatton, line 2323: @tgeoghegan we missed a "stores the output share" -> "commits the output share" here.

cjpatton:
Rebased.

draft-ietf-ppm-dap.md
```diff
@@ -1695,6 +1695,15 @@ The number of steps, and the type of the responses, depends on the VDAF. The
 message structures and processing rules are specified in the following
 subsections.
 
+Each Aggregator maintains some state for each report. A state transition is
+triggered by receiving a message from the Aggregatro's peer. Eventually this
```
tgeoghegan, line 1693:
```suggestion
triggered by receiving a message from the Aggregator's peer. Eventually this
```

draft-ietf-ppm-dap.md
```diff
@@ -1764,9 +1773,28 @@ The Leader and Helper initialization behavior is detailed below.
 
 #### Leader Initialization {#leader-init}
 
+~~~ aasvg
+                     PrepareResp
+Report --> Continued -----+----> Continued
+                          |
+                          +----> FinishedWithOutbound
```
tgeoghegan, line 1774:
```suggestion
                          +----> FinishedWithOutbound (*commit)
```


Isn't `FinishedWithOutbound` terminal? Sure, you send a message to the peer, but the aggregator that reaches this state doesn't expect to receive any further message.

draft-ietf-ppm-dap.md
```diff
@@ -1764,9 +1773,28 @@ The Leader and Helper initialization behavior is detailed below.
...
+~~~
+{: #leader-init-state title="Leader state transition triggered by aggregation initialization. (\*) indicates a terminal state." }
+
 The Leader begins an aggregation job by choosing a set of candidate reports that
 belong to the same task and a job ID which MUST be unique within the task.
 
+First, the Leader MUST ensure each report in the candidate set can be
+committed. That is: the Leader MUST ensure it has not committed a report with
+the same ID and task; and the Leader MUST ensure the batch bucket to which the
+report belongs has not been collected. See {{batch-buckets}} for details. If a
```
tgeoghegan, line 1788:
```suggestion
First, the Leader MUST ensure each report in the candidate set can be
committed per the criteria detailed in {{batch-buckets}}. If a
```

I buy that we need to do these checks at job scheduling time rather than at output share commitment. However I'd prefer not to re-state the criteria here as we describe them carefully in {#batch-buckets} already. Also, I think we should consider tweaking the text there given that the leader now wants to use it slightly differently. Currently, in {#batch-buckets}, we have:

```
Aggregators MUST check the following conditions before committing an output
share:

* If the batch bucket has been collected, the Aggregator MUST mark the report
  invalid with error `batch_collected`.
* If the report ID associated with `out_share` has been aggregated in the task,
  the Aggregator MUST mark the report invalid with error `report_replayed`.

Aggregators may perform these checks at any time before commitment (i.e., during
either aggregation initialization or continuation).
```

I think maybe we should take these lines and break them out into their own section so that this text right here can point at that, without the description of how to commit an output share to a batch bucket. Maybe a sub-section within {#batch-buckets} like `##### Commitment eligibility`? Or do you think it's clear enough if we point to all of {#batch-buckets} that these are the only fallible checks in there?

Further, the last sentence about doing the checks at anytime is no longer true for the Leader, right? It now has to do it when constructing the candidate set. So it could say:

```
Helpers may perform these checks at any time before commitment (i.e., during
either aggregation initialization or continuation). Leaders MUST perform these
checks before adding a report to an aggregation job.
```

draft-ietf-ppm-dap.md
tgeoghegan: I don't mind renaming it from `prev_state`, but we should still describe the `state` value input to `Vdaf.ping_pong_leader_continued`

draft-ietf-ppm-dap.md
```diff
@@ -2049,23 +2091,27 @@ variant {
 } PrepareResp;
 ~~~
 
-Otherwise the Helper responds with
+Otherwise `state` has type `Continued` or `FinishedWithOutbound` and
+there is at least one more outbound message to process. (State `Finished` is
+not reachable at this point.) The Helper responds with
```
tgeoghegan, line 2096:
```suggestion
there is at least one more outbound message to process. State `Finished` is
not reachable at this point. The Helper responds with
```

draft-ietf-ppm-dap.md
```diff
@@ -2049,23 +2091,27 @@ variant {
...
-  opaque payload<1..2^32-1> = outbound;
+  opaque payload<1..2^32-1> = state.outbound;
 } PrepareResp;
 ~~~
 
-If `state == Continued(prep_state)`, then the Helper stores `state` to
-prepare for the next continuation step ({{aggregation-helper-continuation}}).
+If `state` has type `Continued`, then the Helper stores `state`
+for use in the first continuation step in
+({{aggregation-helper-continuation}}).
```
tgeoghegan, line 2108:
```suggestion
{{aggregation-helper-continuation}}.
```

draft-ietf-ppm-dap.md
tgeoghegan: Glad you spotted it

draft-ietf-ppm-dap.md
cjpatton: Not for the Leader! It's still possible for the Helper to reject at this point. The Helper always commits first.

draft-ietf-ppm-dap.md
cjpatton: > I buy that we need to do these checks at job scheduling time rather than at output share commitment. However I'd prefer not to re-state the criteria here as we describe them carefully in {#batch-buckets} already.

Good idea, done. Regarding changes to {{batch-buckets}}:

> I think maybe we should take these lines and break them out into their own section so that this text right here can point at that, without the description of how to commit an output share to a batch bucket. Maybe a sub-section within {#batch-buckets} like `##### Commitment eligibility`? Or do you think it's clear enough if we point to all of {#batch-buckets} that these are the only fallible checks in there?

I wouldn't want to add another sub-sub-sub-sub-sub-section and have to define "commitment eligiibility" :)

I word-smithed the section a bit to make it more clear, let me know what you think

>
> Further, the last sentence about doing the checks at anytime is no longer true for the Leader, right? It now has to do it when constructing the candidate set. So it could say:
> 
> ```
> Helpers may perform these checks at any time before commitment (i.e., during
> either aggregation initialization or continuation). Leaders MUST perform these
> checks before adding a report to an aggregation job.
> ```

Good catch, added your text verbatim.

draft-ietf-ppm-dap.md
cjpatton: Done.

draft-ietf-ppm-dap.md
```diff
@@ -1892,26 +1915,35 @@ The Leader proceeds as follows with each report:
...
-
-1. Else if the type is "reject", then the Leader rejects the report and removes
-   it from the candidate set. The Leader MUST NOT include the report in a
-   subsequent aggregation job, unless the error is `report_too_early`, in which
-   case the Leader MAY include the report in a subsequent aggregation job.
-
-1. Else the type is invalid, in which case the Leader MUST abort the aggregation
-   job.
+   * `inbound` is the payload of the `PrepareResp`
+   * `state` is the report's initial prep state
```
tgeoghegan, line 1919: nit: match the order of the function arguments
```suggestion
   * `state` is the report's initial prep state
   * `inbound` is the payload of the `PrepareResp`
```

tgeoghegan (approved):
One nit but I think this is sound. Before we merge it, I'd like a chance to implement this in Janus as a further check that it's sound.

cjpatton:
Squashed.

draft-ietf-ppm-dap.md
```diff
@@ -1689,6 +1689,15 @@ The number of steps, and the type of the responses, depends on the VDAF. The
...
 subsections.
 
+Each Aggregator maintains some state for each report. A state transition is
+triggered by receiving a message from the Aggregator's peer. Eventually this
+process results in a terminal state, either rejecting the report or recovering
+an output share. Once a report has reached a terminal state, no more messages
+will be processed for it. Each state is said to have one of four possible
+"types" (see {{Section 5.7 of !VDAF}}: `Continued`, `FinishedWithOutbound`,
+`Finished` and `Rejected`. The first two types include an outbound message to
+be processed by the peer.
```
tgeoghegan, line 1699:
```suggestion
will be processed for it. There are four possible states (see {{Section 5.7 of
!VDAF}}: `Continued`, `FinishedWithOutbound`, `Finished` and `Rejected`. The
first two states include an outbound message to be processed by the peer.
```

draft-ietf-ppm-dap.md
```diff
@@ -1758,9 +1767,25 @@ The Leader and Helper initialization behavior is detailed below.
...
+                          +----> Rejected (*reject)
+~~~
+{: #leader-init-state title="Leader state transition triggered by aggregation initialization. (\*) indicates a terminal state." }
+
 The Leader begins an aggregation job by choosing a set of candidate reports that
 belong to the same task and a job ID which MUST be unique within the task.
 
+First, the Leader MUST ensure each report in the candidate set can be committed
+per the criteria detailed in {{batch-buckets}}. If a If a report is not cannot
+be committed, then the Leader rejects it and removes it from the candidate set.
```
tgeoghegan, line 1787:
```suggestion
per the criteria detailed in {{batch-buckets}}. If a report cannot be committed,
then the Leader rejects it and removes it from the candidate set.
```

draft-ietf-ppm-dap.md
```diff
@@ -1892,26 +1915,35 @@ The Leader proceeds as follows with each report:
...
+   * `inbound` is the payload of the `PrepareResp`
+
+   If the new `state` has type `Continued` or `FinishedWithOutbound`,
+   then there is at least one more outbound message to send before preparation
+   is complete. The Leader stores `state` and proceeds as in
+   {{aggregation-leader-continuation}}.
+
+   Else if the new  `state` has type `Finished`, then preparation is complete
+   and the state includes an output share, denoted `state.out_share`. The
+   Leader commits to `state.out_share` as described in {{batch-buckets}}.
```
tgeoghegan, line 1928:
```suggestion
   Else if the new `state` has type `Finished`, then preparation is complete and
   the state includes an output share, denoted `state.out_share`. The Leader
   commits to `state.out_share` as described in {{batch-buckets}}.
```

draft-ietf-ppm-dap.md
```diff
@@ -2481,16 +2573,17 @@ A few different pieces of information are associated with each batch bucket:
...
   invalid with error `batch_collected`.
 * If the report ID associated with `out_share` has been aggregated in the task,
   the Aggregator MUST mark the report invalid with error `report_replayed`.
 
-Aggregators may perform these checks at any time before commitment (i.e., during
-either aggregation initialization or continuation).
+Aggregators MUST check the following conditions before committing an output
+share. Helpers may perform these checks at any time before commitment (i.e.,
+during either aggregation initialization or continuation), but Leaders MUST
+perform these checks before adding a report to an aggregation job.
```
tgeoghegan, line 2586:
```suggestion
Aggregators MUST check these conditions before committing an output share.
Helpers may perform these checks at any time before commitment (i.e., during
either aggregation initialization or continuation), but Leaders MUST perform
these checks before adding a report to an aggregation job.
```

tgeoghegan (approved):
I implemented this in https://github.com/divviup/janus/pull/3826 and I think it looks OK. I have a few more nits but this change is sound.

PR #710, opened 2025-05-27
HTTP resource deletion vs. replay protection

### LINKED ISSUE #700: Share deletion
S 4.7.3 says

> The Leader can send a DELETE request to the aggregate share, which indicates to
the Helper that it can abandon the aggregate share and discard all state related
to it.

This text doesn't seem quite right because the replay state is "related" to it, and we don't want to let you do that after commitment.

tgeoghegan:
Clarify that DELETEing aggregation jobs, aggregate shares or collection jobs MUST NOT delete any information needed by aggregators for {{replay-protection}}.

Resolves #700

### DIFF
--- a/draft-ietf-ppm-dap.md
+++ b/draft-ietf-ppm-dap.md
@@ -2701,7 +2701,8 @@ encoded(struct { prepare_resps } AggregationJobResp)
 
 If for whatever reason, the Leader must abandon an aggregation job, it SHOULD
 let the Helper know it can clean up its state by sending a DELETE request to the
-job.
+job. Deletion of an aggregation job MUST NOT delete information needed for
+replay or double collection checks ({{replay-protection}}).
 
 #### Example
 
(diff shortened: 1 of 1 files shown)


draft-ietf-ppm-dap.md
```diff
@@ -2701,7 +2701,8 @@ encoded(struct { prepare_resps } AggregationJobResp)
 
 If for whatever reason, the Leader must abandon an aggregation job, it SHOULD
 let the Helper know it can clean up its state by sending a DELETE request to the
-job.
+job. Deletion of an aggregation job MUST NOT delete information needed for
+replay or double collection checks ({{replay-protection}}).
```
cjpatton, line 2705: I wonder if {{batch-buckets}} would be a better reference?

tgeoghegan: It's counter intuitive but you're right, {{batch-buckets}} has more of the fine details of how to enforce replay and double collection.

cjpatton (approved):

PR #715, opened 2025-06-11
Security Considerations: Describe the trust model more accurately

cjpatton:
Say explicitly which parties are controlled by the attacker. These differ between privacy and robustness. Also, we don't allow the attacker full control over the network, as this would allow for Sybil attacks.

### DIFF
--- a/draft-ietf-ppm-dap.md
+++ b/draft-ietf-ppm-dap.md
@@ -3722,14 +3722,19 @@ suite:
 # Security Considerations {#sec-considerations}
 
 DAP aims to achieve the privacy and robustness security goals defined in
-{{Section 9 of !VDAF}}, even in the presence of an active attacker. It is
-assumed that the attacker may control the network and have the ability to
-control a subset of of Clients, one of the Aggregators, and the Collector for a
-given task.
+{{Section 9 of !VDAF}}. That is, an active attacker that controls a subset of
+the Clients, one of the Aggregators, and the Collector learns nothing about the
+honest Clients' measurements beyond their aggregate result. At the same time,
+an attacker that controls a subset of Clients cannot force the Collector to
+compute anything but the aggregate result over the honest Clients'
+measurements.
 
-In the presence of this adversary, there are some threats DAP does not defend
-against and which are considered outside of DAP's threat model. These are
-enumerated below, along with potential mitigations.
+Since DAP requires HTTPS ({{message-transport}}), the attacker cannot tamper
+with messages delivered by honest parties or forge messages from honest,
+authenticated parties; but it can drop messages or forge messages from
+unauthenticated parties. Thus there are some threats that DAP does not defend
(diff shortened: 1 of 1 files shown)


tgeoghegan:
Do we need to relax the attacker capability from "controls the network" to "can eavesdrop"? Since we use TLS, an attacker that controls the network can't tamper with messages but can still drop them, meaning it can DoS the protocol. I think it's reasonable to concede that capability to the attacker, or more to the point, acknowledge that DAP doesn't mitigate it.

cjpatton:
> Do we need to relax the attacker capability from "controls the network" to "can eavesdrop"? Since we use TLS, an attacker that controls the network can't tamper with messages but can still drop them, meaning it can DoS the protocol. I think it's reasonable to concede that capability to the attacker, or more to the point, acknowledge that DAP doesn't mitigate it.

Agreed, but I'm uncertain how to wordsmith this. We do get into some of the consequences of dropping or tampering messages in the text below, e.g., Sybil attacks.

tgeoghegan:
How about:

```
DAP aims to achieve the privacy and robustness security goals defined in
{{Section 9 of !VDAF}}, even in the presence of an active attacker. It is
assumed that the attacker may control the network and have the ability to
control a subset of of Clients, one of the Aggregators, and the Collector for a
given task.

Since DAP requires HTTPS ({{message-transport}}), the attacker cannot tamper with messages, but can still prevent their delivery, blocking the execution of the protocol. DAP does not mitigate such denial of service attacks. However, the attacker learns nothing about the honest Clients' measurements beyond their aggregate result. At the same time, an attacker that controls a subset of Clients and controls the network cannot force the Collector to compute anything but the aggregate result over the honest Clients' measurements.

There are some threats DAP does not defend against and which are considered
outside of DAP's threat model. These are enumerated below, along with potential
mitigations.
```

cjpatton:
> How about:
> 
> ```
> DAP aims to achieve the privacy and robustness security goals defined in
> {{Section 9 of !VDAF}}, even in the presence of an active attacker. It is
> assumed that the attacker may control the network and have the ability to
> control a subset of of Clients, one of the Aggregators, and the Collector for a
> given task.
> 
> Since DAP requires HTTPS ({{message-transport}}), the attacker cannot tamper with messages, but can still prevent their delivery, blocking the execution of the protocol. DAP does not mitigate such denial of service attacks. However, the attacker learns nothing about the honest Clients' measurements beyond their aggregate result. At the same time, an attacker that controls a subset of Clients and controls the network cannot force the Collector to compute anything but the aggregate result over the honest Clients' measurements.
> 
> There are some threats DAP does not defend against and which are considered
> outside of DAP's threat model. These are enumerated below, along with potential
> mitigations.
> ```

I've tacked on a commit with this massaged in. PTAL.

tgeoghegan (approved):

PR #716, opened 2025-06-11
Add guidance for relaxing report processing rules

### LINKED ISSUE #703: Replay and privacy
This version of the document has an absolute ban on replay but then when it comes to batch size is like ¯\_(ツ)_/¯

>    An important parameter of a DAP deployment is the minimum batch size.
   If a batch includes too few reports, then the aggregate result can
   reveal information about individual measurements.  Aggregators
   enforce the agreed-upon minimum batch size during collection, but
   implementations SHOULD also opt out of participating in a DAP task if
   the minimum batch size is too small.  This document does not specify
   how to choose an appropriate minimum batch size, but an appropriate
   value may be determined from the differential privacy (Section 8.5)
   parameters in use, if any.

The issue here is that it's actually quite useful to be able to do replay in some cases, for instance if you want to do crosstabs or the like. I recognize that it's difficult to specify what's safe there, but it's *also* difficult to specify safe batch sizes, which are arguably even more unsafe and we don't provide guidance there. 

I don't think it's out of the question that someone could come up with a safe rule that was less extreme, though I agree that we don't have something good now. What I'd like to suggest is that this is the default rule and that we explicitly allow for application profiles to relax it and add a section about the potential challenges (e.g., tracking what has happened).

cjpatton:
Closes #703. Alternative to #713.

cc @ekr

### DIFF
--- a/draft-ietf-ppm-dap.md
+++ b/draft-ietf-ppm-dap.md
@@ -3953,6 +3953,32 @@ document does not specify how to choose an appropriate minimum batch size, but
 an appropriate value may be determined from the differential privacy ({{dp}})
 parameters in use, if any.
 
+### Relaxing Report Processing Rules
+
+DAP Aggregators enforce several rules for report processing related to the
+privacy of individual measurements:
+
+1. Each report may be aggregated at most once ({{batch-buckets}})
+
+1. A batch bucket may be collected at most once (reports pertaining to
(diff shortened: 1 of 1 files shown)


tgeoghegan (approved):
This text is good (I especially appreciate the reminder to check the VDAF's requirements) but I can't be certain it addresses the concern raised in #703.

PR #717, opened 2025-06-16
Use HTTP message content length where possible

### LINKED ISSUE #714: Consider removing length prefixes of messages in the aggregation flow
While wrangling the bulk upload PR, we opted to not have a length prefix for the sequence of reports in the request and the sequence of upload statuses in the response. The same logic applies to the messages in the aggregation flow. See [this thread](https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/pull/686#discussion_r2108267711).

tgeoghegan:
In #686, we use the length of the enclosing HTTP message to figure out how long protocol messages are. For consistency, we now do the same thing for:

- AggregationJobInitReq
- AggregationJobResp
- AggregationJobContinueReq
- CollectionJobReq
- AggregateShareReq

In the last case, the message fields had to be re-ordered so that `agg_param` comes last.

Additionally, the word "denotes" was misused in several places. For example, we had:

```
struct {
  Report reports[message_length];
} UploadRequest;

Here `message_length` denotes the length of the HTTP message content...
```

Per Merriam-Webster, to denote means "to serve as an indication of" ([1]). Implementations are not meant to use the length of `reports` as an indication of how long the HTTP message content is. It's the other way around. Let's just say "is" and avoid this trouble.

[1]: https://www.merriam-webster.com/dictionary/denote

Closes #714

### DIFF
--- a/draft-ietf-ppm-dap.md
+++ b/draft-ietf-ppm-dap.md
@@ -1392,7 +1392,7 @@ struct {
 } UploadRequest;
 ~~~
 
-Here `message_length` denotes the length of the HTTP message content ({{!RFC9110,
+Here `message_length` is the length of the HTTP message content ({{!RFC9110,
 Section 6.4}}).
 
 Each upload request contains a sequence of `Report` with the following fields:
@@ -1960,7 +1960,7 @@ struct {
 struct {
   opaque agg_param<0..2^32-1>;
   PartialBatchSelector part_batch_selector;
-  PrepareInit prepare_inits<44..2^32-1>;
+  PrepareInit prepare_inits[prepare_inits_length];
 } AggregationJobInitReq;
 ~~~
 
@@ -1977,7 +1977,10 @@ This message consists of:
   {{batch-modes}}.
 
 * `prepare_inits`: the sequence of `PrepareInit` messages constructed in the
-  previous step.
+  previous step. Here `prepare_inits_length` is the length of the HTTP message
+  content ({{!RFC9110, Section 6.4}}), minus the lengths in octets of
+  `agg_param` and `part_batch_selector.` That is, the remainder of the HTTP
+  message consists of `prepare_inits`.
 
 The Leader sends the `AggregationJobInitReq` in the body of a PUT request to the
 aggregation job with a media type of "application/dap-aggregation-job-init-req".
@@ -2163,12 +2166,15 @@ which is structured as follows:
 
 ~~~ tls-presentation
 struct {
-  PrepareResp prepare_resps<17..2^32-1>;
+  PrepareResp prepare_resps[message_length];
 } AggregationJobResp;
 ~~~
 
-where `prepare_resps` are the outbound `PrepareResp` messages for each report
-computed in the previous step. The order MUST match
+Here `message_length` is the length of the HTTP message content ({{!RFC9110,
+Section 6.4}}).
+
+`prepare_resps` is the outbound `PrepareResp` messages for each report computed
+in the previous step. The order MUST match
 `AggregationJobInitReq.prepare_inits`. The media type for `AggregationJobResp`
 is "application/dap-aggregation-job-resp".
 
@@ -2259,6 +2265,7 @@ The Helper handles the aggregation job initialization synchronously:
 PUT /helper/tasks/8BY0RzZMzxvA46_8ymhzycOB9krN-QIGYvg_RsByGec/aggregation_jobs/lc7aUeGpdSNosNlh-UZhKA
 Host: example.com
 Content-Type: application/dap-aggregation-job-init-req
+Content-Length: 100
 Authorization: Bearer auth-token
 
 encoded(struct {
@@ -2274,6 +2281,7 @@ encoded(struct {
 
 HTTP/1.1 200
 Content-Type: application/dap-aggregation-job-resp
+Content-Length: 100
 
 encoded(struct { prepare_resps } AggregationJobResp)
 ~~~
@@ -2284,6 +2292,7 @@ Or asynchronously:
 PUT /helper/tasks/8BY0RzZMzxvA46_8ymhzycOB9krN-QIGYvg_RsByGec/aggregation_jobs/lc7aUeGpdSNosNlh-UZhKA
 Host: example.com
 Content-Type: application/dap-aggregation-job-init-req
+Content-Length: 100
 Authorization: Bearer auth-token
 
 encoded(struct {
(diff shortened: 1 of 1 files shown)


cjpatton (commented):
Thanks for the PR. I don't think this change makes sense in cases where you have length prefixes for opaque strings, e.g., `AggregateShareReq.agg_param`. Imagine for example if the `AggregteShareReq` had two length prefixed fields in it: you'd have to somehow convey the length of each, and we can't rely on HTTP for that.

Likewise for `AggregationJobInitReq` and `AggregationJobContinueReq`.

I think the only place this makes sense is `AggregationJobResp`, which like `UploadReq` is merely a sequence of identical messages.

Now that I'm looking at a PR, I'm less enthusiastic about this change. At the very least, we should deal with the possible ambiguities that would arise for messages we think might possibly have multiple length prefixed fields.

tgeoghegan:
I agree with you as far as `CollectionJobReq` and `AggregateShareReq` go:  we don't expect anyone to need to stream those requests into aggregators. So it makes sense to only do this for the `AggregationJob` messages, to allow implementations to start sending the sub-messages as they work through aggregation jobs. However, I don't think it makes sense to do this just for `AggregationJobResp`: we'd want both the requests and responses to be streamable for this to be useful.

> Imagine for example if the AggregteShareReq had two length prefixed fields in it: you'd have to somehow convey the length of each, and we can't rely on HTTP for that.

I don't follow your argument. Since we agreed to leave `AggregateShareReq` alone, let's look at the version of `AggregationJobInitReq` I want to introduce here:

```
struct {
  opaque agg_param<0..2^32-1>;
  PrepareInit prepare_inits[prepare_inits_length];
} AggregationJobInitReq;
```

I agree we'd be in a bind if we added another vector field like so:

```
struct {
  opaque agg_param<0..2^32-1>;
  PrepareInit prepare_inits[prepare_inits_length];
  SubMessage sub_messages[sub_messages_length];
} AggregationJobInitReq;
```

Even with the total HTTP content length, there would be no way to tell how much of the message is `prepare_inits_length` vs. `sub_messages_length`. But if we had another length-prefixed, variable length vector field:

```
struct {
  opaque agg_param<0..2^32-1>;
  SubMessage sub_messages<0..2^16-1>;
  PrepareInit prepare_inits[prepare_inits_length];
} AggregationJobInitReq;
```

Then so long as the vector comes last, I don't think there's any ambiguity here, is there? I may not know the total length of the `AggregationJobInitReq` when I start handling the message, but because `agg_param` and `sub_messages` are both length prefixed, I can tell how long they are. `PartialBatchSelector` also contains enough information to determine its length and parse it. If I know the total HTTP message content length (whether via `Content-Length` header or because I received an indication the message is finished and kept track of how many total bytes I received), I can determine what `prepare_inits_length` is by substraction. Am I missing something here?

cjpatton:
> Then so long as the vector comes last, I don't think there's any ambiguity here, is there? I may not know the total length of the `AggregationJobInitReq` when I start handling the message, but because `agg_param` and `sub_messages` are both length prefixed, I can tell how long they are. `PartialBatchSelector` also contains enough information to determine its length and parse it. If I know the total HTTP message content length (whether via `Content-Length` header or because I received an indication the message is finished and kept track of how many total bytes I received), I can determine what `prepare_inits_length` is by substraction. Am I missing something here?

I think you're right, as long as our convention is that only the part of the body that is meant to be "streamed" is not length prefixed, then we're good. By the way, have we articulated this in operational considerations? If not, I think we should do so in this PR.

draft-ietf-ppm-dap.md
```diff
@@ -1960,7 +1960,7 @@ struct {
 struct {
   opaque agg_param<0..2^32-1>;
   PartialBatchSelector part_batch_selector;
-  PrepareInit prepare_inits<44..2^32-1>;
+  PrepareInit prepare_inits[prepare_inits_length];
```
cjpatton, line 1963: It might be nice to use the same variable name for all messages. What about "suffix_length" to denote that this is supposed to be the length of the rest of the body? It also fits nicer with "message_length".

draft-ietf-ppm-dap.md
```diff
@@ -2259,6 +2265,7 @@ The Helper handles the aggregation job initialization synchronously:
 PUT /helper/tasks/8BY0RzZMzxvA46_8ymhzycOB9krN-QIGYvg_RsByGec/aggregation_jobs/lc7aUeGpdSNosNlh-UZhKA
 Host: example.com
 Content-Type: application/dap-aggregation-job-init-req
+Content-Length: 100
```
cjpatton, line 2268: I find it odd that the content length doesn't match the length of the example payload.

cjpatton (commented):
I'm happy with this, just a couple of suggestions.

Also, how much time should we give this before merging? Maybe worth bringing it up at the interim?

draft-ietf-ppm-dap.md
tgeoghegan: Well, we can't compute a real content length when some of the fields are `agg_param = [0x01, 0x02, ...]`. I think this is good enough for an example. The point is to illustrate that there would be a `Content-Length`.

draft-ietf-ppm-dap.md
tgeoghegan: If we used `suffix_length`, then we'd have to explain what we mean by "suffix" which unhelpfully complicates the explanatory text that follows. I think it's perfectly clear that `prepare_inits_length` is the length of `prepare_inits`, and then we explain in prose how to compute `prepare_inits_length`.

tgeoghegan:
> Also, how much time should we give this before merging? Maybe worth bringing it up at the interim?

Yeah, the interim seems like a good place to talk about this.

tgeoghegan:
From the 6/30 PPM interim: we'll keep this open until EOD on Thursday July 3 so that the WG can offer ideas and then merge.

draft-ietf-ppm-dap.md
```diff
@@ -1977,7 +1977,10 @@ This message consists of:
   {{batch-modes}}.
 
 * `prepare_inits`: the sequence of `PrepareInit` messages constructed in the
-  previous step.
+  previous step. Here `prepare_inits_length` is the length of the HTTP message
+  content ({{!RFC9110, Section 6.4}}), minus the lengths in octets of
+  `agg_param` and `part_batch_selector.` That is, the remainder of the HTTP
```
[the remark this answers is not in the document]
tgeoghegan: Maybe it would be more clear to say "the lengths in octets of the encoded `agg_param` and `part_batch_selector` fields"?

Issue #706, opened 2025-05-15
Make error type unambiguous for time_precision misalignments


jcjones:
It's not completely clear which error type should be returned for reports rejected on the leader side due to their timestamps not aligning to the task's `time_precision`, nor for the various jobs with intervals unaligned.

What we have on the leader side is only for [violating time bounds](https://ietf-wg-ppm.github.io/draft-ietf-ppm-dap/draft-ietf-ppm-dap.html#section-4.5.2-23), not for alignment. One could argue the [generic "malformed" category](https://ietf-wg-ppm.github.io/draft-ietf-ppm-dap/draft-ietf-ppm-dap.html#section-4.5.2-18) would then govern, which uses `invalid_message`.

On the helper side we have explicit guidance for [Input Share Validation](https://ietf-wg-ppm.github.io/draft-ietf-ppm-dap/draft-ietf-ppm-dap.html#section-4.6.2.4-2.2.1) which says the correct answer is `invalid_message`.

That seems to be two votes for `invalid_message`, although I easily could see an argument could be made for `reportRejected`, particularly for _reports_, but not for other jobs.

Anyway, we should make it unambiguous for all time precision misalignments.

cjpatton:
Generally I think the abort message should have parity with the corresponding report error during aggregation. Therefore I'd vote vote for "invalidMessage". An implementation could choose to provide detail in the JSON response.

cjpatton:
I've reviewed the PR. It seems like it solves the problem, and I feel comfortable merging. However, the language about "Time or Duration being malformed implies a message containing it is malformed" feels very fussy. It makes me wish that there weren't a way to represent an invalid Time or Duration on the wire.

In fact, we might be able to do this. Our convention could be that `Time` is an integer representing a number of `time_precision` intervals since the UNIX epoch, rather than seconds. Likewise for `Duration`.

This would be a breaking change of course, and would be kind of a pain to maintain alongside previous versions of DAP. But it would simplify the text and reduce some bug surface.

jcjones:
We discussed this a bit as well, with the same outcomes. 

Given the convergence, maybe we should try and make it so - get a conversation going with at the WG for one last major breaking change before WGLC?

PR #720, opened 2025-07-11
Redefine Time and Duration as multiples of Task time_precision

jcjones:
Represent the `Time` and `Duration` types as the number of `time_precision` intervals since the UNIX epoch, rather than seconds. This eliminates the error condition where `Timestamps`, `Durations`, and `Intervals` could be malformed for a given task.

This is a follow-up to #719 and #706.

### DIFF
--- a/draft-ietf-ppm-dap.md
+++ b/draft-ietf-ppm-dap.md
@@ -83,6 +83,11 @@ contributor:
        email: csharrison@chromium.org
 
  -
+       name: J.C. Jones
+       org: ISRG
+       email: ietf@insufficient.coffee
+
+ -
        name: Alex Koshelev
        org: Meta
        email: koshelev@meta.com
@@ -167,6 +172,9 @@ aggregator.
 
 - Bump draft-irtf-cfrg-vdaf-13 to 14 {{!VDAF}} and adopt changes to the
   ping-pong API (\#705, \#718).
+
+- Represent the Time and Duration types as a number of time_precision
+  intervals, rather than seconds (\*) (#720).
 
 15:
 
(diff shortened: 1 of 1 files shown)


draft-ietf-ppm-dap.md
```diff
@@ -168,6 +168,9 @@ aggregator.
 - Bump draft-irtf-cfrg-vdaf-13 to 14 {{!VDAF}} and adopt changes to the
   ping-pong API (\#705, \#718).
 
+- Represent the Time and Duration types as the number of time_precision
+  intervals since the UNIX epoch, rather than seconds (\*) (TKTK).
```
jcjones, line 172:
```suggestion
  intervals since the UNIX epoch, rather than seconds (\*) (#720).
```

draft-ietf-ppm-dap.md
```diff
@@ -1594,7 +1587,7 @@ encoded(struct {
     struct {
       report_metadata = struct {
         report_id = [0x0a, 0x0b, 0x0c, 0x0d, ...],
-        time = 1741986088,
+        time = 17419860,
```
jcjones, line 1590: I've shaved a few significant digits off all the times and durations fairly arbitrarily. I'm open to any adjustments.

draft-ietf-ppm-dap.md
```diff
@@ -3452,16 +3445,16 @@ batch bucket identifiers for the batch interval is
 [
   struct {
     start = batch_interval.start,
-    duration = time_precision,
+    duration = 1,
```
jcjones, line 3448: Check this set over carefully - I think the meaning is still here, but I'm not an expert!

draft-ietf-ppm-dap.md
```diff
@@ -168,6 +173,9 @@ aggregator.
 - Bump draft-irtf-cfrg-vdaf-13 to 14 {{!VDAF}} and adopt changes to the
   ping-pong API (\#705, \#718).
 
+- Represent the Time and Duration types as the number of time_precision
+  intervals since the UNIX epoch, rather than seconds (\*) (#720).
```
tgeoghegan, line 177: This isn't quite right: a Duration isn't measured relative to any epoch or any starting point at all. A Duration relative to a starting point is an Interval.

tgeoghegan (commented):
- How does this relate to #719? IMO we should take this change and then I'll rebase #719. I think we can remove the language it adds about malformed messages (because it's now impossible to malform them, yay), but I think the stuff I added about rejecting otherwise malformed messages is still useful, if pedantic.

- Do we need to revisit the language on how to determine if a Time falls into an Interval? I think it might be clear enough as is:

> Intervals are half-open; that is, `start` is included and `(start + duration)` is excluded. A time that is before the `start` of an `Interval` is said to be before that interval. A time that is equal to or after `Interval.start + Interval.duration` is said to be after the interval.

So if `(start + duration)` is excluded, then `(start + duration - 1)` is included and anything between `(start + duration - 1)` and `(start + duration)` is impossible to represent, which is our intent. @jcjones did any other interesting edge cases come up when you implemented time validation that you think would be worth surfacing here?

- I think we want to remove the mention of "truncated" time on [L1411](https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/blob/1f44b4b7756b6a3bda665cc252ac0a1a4073fc09/draft-ietf-ppm-dap.md?plain=1#L1411) as it's no longer possible for a time to not be truncated.

draft-ietf-ppm-dap.md
```diff
@@ -3436,8 +3434,8 @@ the task's `time_precision` is 1000 seconds and the report's timestamp is
```
tgeoghegan, line 3434: The example timestamp on L3433 should be updated to match the interval start, I believe.

jcjones: Should it? I think the `report's timestamp` is different than the `Time` representation -- but maybe I could rephrase that to add `raw` or similar?

```diff
For example, if
- the task's `time_precision` is 1000 seconds and the report's timestamp is
+ the task's `time_precision` is 1000 seconds and the report's raw timestamp is
1729629081, the relevant batch bucket identifier is
```

tgeoghegan: It's verbose, but how about "...is 1000 seconds and the report was generated at 1729629081 seconds after the start of the UNIX epoch, the relevant..."

tgeoghegan (approved):

tgeoghegan:
@jcjones Could you please squash into a single commit so the history on main looks nicer when I merge? Thanks!

jcjones:
@tgeoghegan Squashed.

PR #721, opened 2025-07-16
update changelog with work since draft 15

tgeoghegan:


draft-ietf-ppm-dap.md
```diff
@@ -165,8 +165,17 @@ aggregator.
 
 16:
 
-- Bump draft-irtf-cfrg-vdaf-13 to 14 {{!VDAF}} and adopt changes to the
-  ping-pong API (\#705, \#718).
+- Bump draft-irtf-cfrg-vdaf-13 to 15 {{!VDAF}} and adopt changes to the
+  ping-pong API. (\*) (#705, #718)
```
cjpatton, line 169: "#" gets interpreted by markdown ... I believe you want to escape it. Here and below.
```suggestion
  ping-pong API. (\*) (\#705, \#718)
```

cjpatton, line 169: "*" means wire breaking change. This isn't a wire breaking change, is it?

draft-ietf-ppm-dap.md
```diff
@@ -165,8 +165,17 @@ aggregator.
...
+
+- Numerous editorial changes for clarity and concision.
```
cjpatton, line 171: I don't think we need to list this.

draft-ietf-ppm-dap.md
```diff
@@ -165,8 +165,17 @@ aggregator.
...
+
+- Allow many reports to be uploaded at once. (\*) (#686)
+
+- Remove TLS presentation language syntax extensions. (#707)
+
+- Use HTTP message content length to determine length of vectors in messages
+  where possible. (\*) (#717)
```
cjpatton, line 178: I would list the messages for affected by this change.

draft-ietf-ppm-dap.md
tgeoghegan: It doesn't appear to confuse GitHub's markdown renderer:

Image

I think the escapes are unnecessary.

tgeoghegan: You're right, I forgot that all the context strings or labels that incorporate draft versions are at the DAP layer.

cjpatton (approved):

PR #722, opened 2025-07-16
Incorporate VDAF configuration into InputShareAad

tgeoghegan:
In recent discussion on the ppm list ([1]), it was observed that since only task ID gets incorporated into `InputShareAad`, the VDAF in use and any parameters for it (e.g., number of elements in `Prio3SumVec`) are only commited to implicitly, by assuming that all honest protocol participants have agreed on the VDAF configuration for the task ID. This is a problem if the proposed late task binding extension is in use ([2]), and generally it seems safer to explicitly commit to VDAF configuration.

This commit achieves this by:

- adding a `vdaf_configuration` field to `InputShareAad`
- state that VDAFs that want to be compatible with DAP must provide an encoding of their configuration
- providing encodings for the Prio3 family of VDAFs (lifted from [3]) in an appendix

[1]: https://mailarchive.ietf.org/arch/msg/ppm/wtlTZV1Fy6RUltWCCYLl2deTJrc/
[2]: https://datatracker.ietf.org/doc/draft-thomson-ppm-dap-dp-ext/02/
[3]: https://datatracker.ietf.org/doc/html/draft-ietf-ppm-dap-taskprov-02#section-3.2

### DIFF
--- a/draft-ietf-ppm-dap.md
+++ b/draft-ietf-ppm-dap.md
@@ -168,6 +168,9 @@ aggregator.
 - Bump draft-irtf-cfrg-vdaf-13 to 14 {{!VDAF}} and adopt changes to the
   ping-pong API (\#705, \#718).
 
+- Include VDAF configuration in InputShareAad to mitigate cross-VDAF confusion
+  attacks. (\*) (#TODO(timg): fill in PR number)
+
 15:
 
 - Specify body of responses to aggregation job GET requests. (#651)
@@ -1141,6 +1144,13 @@ VDAF types:
 * PrepShare
 * PrepMessage
 
+Additionally, to prevent cross-VDAF confusion attacks, in which an input
+intended for a particular VDAF configuration is evaluated under another VDAF
+configuration, Clients commit to the configuration in use by incorporating it
+into {{!HPKE}} authenticated data (see {{upload-request}}). This requires VDAFs
+compatible with DAP to specify an encoding of their configuration. Encodings for
+the Prio3 family of VDAFs are provided in {{vdaf-configuration-encodings}}.
+
 ## Task Configuration {#task-configuration}
 
 A task represents a single measurement process, though potentially aggregating
@@ -1483,8 +1493,7 @@ enc, payload = SealBase(pk,
 * `server_role` is the `Role` of the recipient (`0x02` for the Leader and `0x03`
    for the Helper).
 * `plaintext_input_share` is the Aggregator's `PlaintextInputShare`.
-* `input_share_aad` is an encoded `InputShareAad`, constructed from the
-  corresponding fields in the report per the definition below.
+* `input_share_aad` is an encoded `InputShareAad`, defined below.
 
 The `SealBase()` function is as specified in {{!HPKE, Section 6.1}} for the
 ciphersuite indicated by the Aggregator's HPKE configuration.
@@ -1494,8 +1503,17 @@ struct {
   TaskID task_id;
   ReportMetadata report_metadata;
   opaque public_share<0..2^32-1>;
+  opaque vdaf_configuraton<0..2^16-1>;
 } InputShareAad;
 ~~~
+
+* `task_id`, `report_metadata` and `public_share` are the corresponding fields
+  from the `Report` structure.
+
+* `vdaf_configuration` is the encoding of the VDAF's configuration. Encodings
+  for the Prio3 family of VDAFs are specified in
+  {{vdaf-configuration-encodings}}. Other VDAFs will provide encodings of their
+  configuration (see {{extending-this-document}}).
 
 If the upload request is malformed, the Leader aborts with error
 `invalidMessage`.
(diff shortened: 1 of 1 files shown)


draft-ietf-ppm-dap.md
```diff
@@ -1483,20 +1493,34 @@ enc, payload = SealBase(pk,
...
 ciphersuite indicated by the Aggregator's HPKE configuration.
 
 ~~~ tls-presentation
+uint32 VdafIdentifier;
+
 struct {
   TaskID task_id;
   ReportMetadata report_metadata;
   opaque public_share<0..2^32-1>;
+  VdafIdentifier vdaf_identifier;
```
tgeoghegan, line 1508: I went back and forth on whether we need `vdaf_identifier`. `InputShareAad` never gets transmitted or decoded, it's only constructed by clients to encrypt messages and then constructed by aggregators to decrypt them. So you don't need an enum discriminant to tell you how to decode `vdaf_configuration`. However, it's possible that the configurations for two different VDAF types could have the same encoding, so I believe we need the VDAF codepoint to reliably distinguish them.

draft-ietf-ppm-dap.md
```diff
@@ -4936,4 +4961,56 @@ Each of these requires registration of a codepoint or other value; see
...
+
+## Prio3MultihotCountVec
+
+~~~ tls-presentation
+struct {
+    uint32 length;       /* length of the vector */
+    uint32 chunk_length; /* size of each proof chunk */
+    uint32 max_weight;   /* largest vector weight /
+} Prio3MultihotCountVecConfig;
+~~~
```
tgeoghegan, line 5016: I didn't include Poplar1 here because DAP isn't capable of running Poplar1 or any VDAF that collects batches multiple times.

cjpatton: Poplar1 can be used in "attributed-based metrics" mode, providing a limited version of Mastic's capability. I think it should be included.

draft-ietf-ppm-dap.md
```diff
@@ -1141,6 +1144,13 @@ VDAF types:
 * PrepShare
 * PrepMessage
 
+Additionally, to prevent cross-VDAF confusion attacks, in which an input
+intended for a particular VDAF configuration is evaluated under another VDAF
```
cjpatton, line 1148:
```suggestion
Additionally, to prevent cross-VDAF confusion attacks, in which a message
intended for a particular VDAF configuration is evaluated under another VDAF
```

cjpatton (commented):
As I mentioned on the list, I don't think we should modify DAP try to address VDAF-confusion attacks at this stage. I agree late binding makes these attacks more likely, but this is something I think @martinthomson's draft could address on its own. That said, if most people are in favor of this change, then I won't stand in the way. 

My main concern with the current PR is the new requirement for VDAFs to define an encoding of their configurations. We'll have to do something about the fact that draft-ietf-ppm-dap-taskprov already defines the same structures. We should pick one place to put them. The most sensible place would be draft-irtf-cfrg-vdaf, but  this would be a pretty expensive change to that document.

My second concern is that binding only to the VDAF config and not the entire task config (i.e., `TaskConfig` in draft-ietf-ppm-dap-taskprov) feels like a half measure. Many of the arguments to defeating VDAF-confusion attacks equally apply to confusing other parameters. For example, it's easy to imagine how confusing two batch modes or swapping one Helper URL for another could lead to attacks. Again, avoiding these vectors is a key motivation for taskprov. Perhaps we should ask ourselves whether we should be merging taskprov into the core DAP spec (except for the task ID derivation or the task provisioning bits).

I think I'd be comfortable with a less invasive change. To mitigate the specific attacks described in the thread, including for late binding, it would be sufficient to bind just to just the VDAF identifier (the 32-bit codepoint). This is probably the most important thing to bind to: in the case of prio3, it identifies the circuit in use. (Running the wrong circuit on a measurement is the primary concern.) It's less obvious to me that binding to the rest of the parameters is particularly useful. At least in the case of Prio3, the parameters impact the size of the encoded input, not the circuit in use. If you run preparation on the wrong input length, then it'll fail right away.

The advantage of this approach is that we don't have to make any changes to taskprov or VDAF drafts. What do folks think of this compromise? cc @martinthomson


> should we include VDAF configuration in AggregateShareAad? 

We should do so, for the same reason we include the task ID: it ensures that both the data producer (the Aggregators) and data consumer (the Collector) agree on the parameters we want them to agree on.

draft-ietf-ppm-dap.md
```diff
@@ -4936,4 +4961,56 @@ Each of these requires registration of a codepoint or other value; see
...
+  encoding of that VDAF's configuration like the ones in
+  {{vdaf-configuration-encodings}}.
+
 --- back
+
+# VDAF Configuration Encodings
+
+This section provides encodings of configuration of the Prio3 family of VDAFs.
+These are included in `InputShareAad` structures ({{upload-request}}) to prevent
+cross-VDAF confusion attacks.
```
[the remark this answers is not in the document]
tgeoghegan: Yeah, my thinking is that if we do this in DAP, we'd delete these from taskprov and have it refer to the DAP appendix.

PR #723, opened 2025-08-14
Fix typo in HpkeConfigList cardinality presentation

jcjones:
Text says 'The `HpkeConfigList` contains one or more `HpkeConfig`s' but we went from `0` to `10`, when we meant to go to `1`.

### DIFF
--- a/draft-ietf-ppm-dap.md
+++ b/draft-ietf-ppm-dap.md
@@ -1312,7 +1312,7 @@ support multiple HPKE configurations and multiple sets of algorithms
 simultaneously.
 
 ~~~ tls-presentation
-HpkeConfig HpkeConfigList<10..2^16-1>;
+HpkeConfig HpkeConfigList<1..2^16-1>;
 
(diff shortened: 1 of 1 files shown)


divergentdave (approved):
LGTM. If we wanted tighter bounds, I think it would be `<8..2^16-1>`, but it looks like we standardized on using ranges that start at zero or one.

cjpatton:
Actually @ekr  asked us to tighten the lower bounds

> LGTM. If we wanted tighter bounds, I think it would be `<8..2^16-1>`, but it looks like we standardized on using ranges that start at zero or one.

Actually, @ekr asked us to tighten the lower bounds as much as possible: https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/issues/687

I thought that 10 was the smallest possible HpkeConfig, but did we count incorrectly?

divergentdave:
No, 10 is right, I missed an inner length prefix before the public key.

PR #741, opened 2025-09-23
break up object lifecycle diagram

### LINKED ISSUE #727: Figure 2 is two and a half pages
It's a quite a thing, but it's also very large.  It seems like it could be broken up.

Also, this isn't particularly readable:

Image

tgeoghegan:
Breaks the object lifecycle diagram into 5 sub-diagrams, each illustrating a particular transformation. The text from the existing "Arity" section is redistributed among the diagrams.

The diagrams also now fit within 72 columns, which was an idnits problem.

Resolves #727

### DIFF
--- a/draft-ietf-ppm-dap.md
+++ b/draft-ietf-ppm-dap.md
@@ -752,15 +752,25 @@ has been collected ({{collect-flow}}).
 
 ## Lifecycle of Protocol Objects
 
-The following diagram illustrates how the various objects in the protocol are
+The following diagrams illustrate how the various objects in the protocol are
 constructed or transformed into other protocol objects. Oval nodes are verbs or
 actions which process, transform or combine one or more objects into one or more
-other objects. This diagram does not necessarily illustrate how participants
+other objects.
+
+The diagrams in this section do not necessarily illustrate how participants
 communicate. In particular, the processing of aggregation jobs happens in
 distinct, non-colluding parties.
 
+### The Upload Interaction
+
+Reports are 1 to 1 with measurements. In this illustration, `i` distinct Clients
+upload `i` distinct reports, but a single Client could upload multiple reports
+to a task (see {{sybil}} for some implications of this). The process of sharding
+measurements, constructing reports and uploading them is specified in
+{{upload-flow}}.
+
 ~~~ aasvg
-                   measurement
+                  measurement 1
                         |
                      .--+--.
                     | shard |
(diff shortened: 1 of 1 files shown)


draft-ietf-ppm-dap.md
```diff
@@ -778,127 +786,157 @@ public share   Leader input share         Helper input share
...
+                         |
+      .------------------+-----------------------.
+      |                  |                       |
+      v                  v                       v
+Client report 1    Client report 2   ...   Client report i
+      |                  |                       |
+      '----------------. |     .-----------------'
+                       | | ... |
+               .-------+-+-----+---.
+              |    assign reports   +---- aggregation parameter
```
martinthomson, line 839: does the aggregation parameter need an arrowhead?

```suggestion
              |    assign reports   |<--- aggregation parameter
```

draft-ietf-ppm-dap.md
```diff
@@ -914,6 +952,45 @@ may occur multiple times for each task, with the Collector obtaining multiple
...
+     .--+-+-----+--.        |        .--+-+-----+--.
+    |  merge Leader |       v       | merge Helper  |
+    | batch buckets |<------+------>| batch buckets |
+     '------+------'                 '------+------'
+            |                               |
+            v                               v
+  Leader aggregate share          Helper aggregate share
+            |                               |
+            '---------------+---------------'
+                            |
```
martinthomson, line 981: Should this indicate how the shares are encrypted, sent to the collector (via the leader, I believe), decrypted, then unsharded.  That's probably another diagram's worth.

martinthomson (approved):
This is much better, interleaving description and pictures, just a few nits, I think.

draft-ietf-ppm-dap.md
martinthomson: > This diagram deliberately does not illustrate how this plays out over a network. 

I didn't pick up on this.  It might be worth making that explicit.

tgeoghegan: It says as much in the first paragraph under "Lifecycle of Protocol Objects":

> The following diagrams illustrate how the various objects in the protocol are
> constructed or transformed into other protocol objects. Oval nodes are verbs or
> actions which process, transform or combine one or more objects into one or more
> other objects. **This diagram does not necessarily illustrate how participants
> communicate. In particular, the processing of aggregation jobs happens in
> distinct, non-colluding parties.**

I'm open to suggestions for making that more clear.

tgeoghegan:
I'm going to merge this at the end of the week if there's no other remarks about it.

PR #743, opened 2025-09-25
remove DAP-Auth-Token citation

### LINKED ISSUE #728: DAP-Auth-Token citation can probably be removed
The citation itself is a little odd (The inclusion of "draft" in "[I-D.draft-etc]" is redundant), but the main point I'd make is that this is a bearer token of the sort that is commonly included in a standard `Authorization` field: https://www.rfc-editor.org/rfc/rfc6750#section-2.1

tgeoghegan:
Resolves #728


cjpatton (approved):
We might note that this header was in use in some deployments that predated the published standard.

Issue #742, opened 2025-09-23
Deal with idnits


tgeoghegan:
`make idnits` yields the following results:

```
❌ Review the 6 errors, 4 warnings, 2 comments listed below.

███████▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀▀
 ERROR  6 nits of ⛔ error severity
███████▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄▄

 1  INVALID_INTRODUCTION_SECTION_CHILD
 └- Desc - The introduction section must consist of <artwork>, <aside>, <blockquote>, <dl>, <figure>, <iref>, <ol>, <sourcecode>, <t>, <table>, <texttable>, <ul> elements only.
 └- Ref  - https://www.rfc-editor.org/rfc/rfc7991.html#section-2.46
 └- Path - rfc.middle.section[name=Introduction]

 2  INVALID_SECURITY_CONSIDERATIONS_SECTION_CHILD
 └- Desc - The security considerations section must consist of <artwork>, <aside>, <blockquote>, <dl>, <figure>, <iref>, <ol>, <sourcecode>, <t>, <table>, <texttable>, <ul> elements only.
 └- Ref  - https://www.rfc-editor.org/rfc/rfc7991.html#section-2.46

 3  INVALID_REFERENCES_NAME
 └- Desc - The references section <name> element should be either Normative or Informative.
 └- Ref  - https://authors.ietf.org/required-content#references
 └- Path - rfc.back.references[0].name

 4  INVALID_IANA_CONSIDERATIONS_SECTION_CHILD
 └- Desc - The security considerations section must consist of <artwork>, <aside>, <blockquote>, <dl>, <figure>, <iref>, <ol>, <sourcecode>, <t>, <table>, <texttable>, <ul> elements only.
 └- Ref  - https://www.rfc-editor.org/rfc/rfc7991.html#section-2.46
 └- Path - rfc.middle.section[8].section

 5  OBSOLETE_DOCUMENT
[...]

PR #744, opened 2025-09-25
address some idnits

tgeoghegan:
Part of #742

With this change, we still trip a few idnits,  but I'm not sure if they're legitimate.

- Missing BCP 78/79 boilerplate: is this something that gets added by the RFC editor?
- Obsolete reference to RFC 7807: this occurs in the changelog which will be deleted from the final text
- INVALID_IANA_CONSIDERATIONS_SECTION_CHILD, INVALID_INTRODUCTION_SECTION_CHILD, INVALID_SECURITY_CONSIDERATIONS_SECTION_CHILD : I'm not sure what element it's complaining about.
- INVALID_REFERENCES_NAME: AFAICT all the references are properly tagged either normative or informative.
- We do indeed have a downref to HPKE, which is technically an informative RFC, but until HPKE gets republished in the HPKE WG, we don't really have a choice.

### DIFF
--- a/draft-ietf-ppm-dap.md
+++ b/draft-ietf-ppm-dap.md
@@ -600,7 +600,7 @@ function" and `agg_result` the "aggregate result".
 DAP is extensible in that it allows for the addition of new cryptographic
 schemes that compute different aggregation functions, determined by the
 Verifiable Distributed Aggregation Function, or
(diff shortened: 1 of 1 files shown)


cjpatton (approved):
Oh it's awesome there's a tool for this.

PR #746, opened 2025-09-29
upload: introduce Client and Leader sub-sections

### LINKED ISSUE #730: Section 4.5.2 is extraordinarily long
There's an awful lot of stuff here.  It would probably benefit from some breaking up into subsections.

tgeoghegan:
Breaks 4.5.2 into a couple new sections, addressing the interaction from the point of view of either the Client or the Leader. This should make it easier for implementers to zero in on the normative text they need.

Resolves #730

### DIFF
--- a/draft-ietf-ppm-dap.md
+++ b/draft-ietf-ppm-dap.md
@@ -1383,9 +1383,10 @@ encoded([
 
 ### Upload Request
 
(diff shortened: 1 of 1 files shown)


tgeoghegan:
I used your names. Somehow just having more than one word feels better there.

PR #748, opened 2025-10-01
use a single `application/dap` media type

### LINKED ISSUE #738: Use a single `application/dap` media-type
> The document currently defines ten different media types. My assumption is that
> most implementations will need to support all of these media types.  I would
> strongly consider defining a single media type and a parameter such as
> `message` to distinguish the different message types.
> 
> e.g. `application/dap; message=hpke-config-list`
> 
> I do not see value in making distinct media type registrations as it does not
> seem like they could be used independently. This will simplify IANA
> registration and if there are future major versions of the protocol it will be
> easier to content negotation for the version.

tgeoghegan:
I also cleaned up a couple of leftover mentions of `UploadResp` and a couple places where we used `collection` instead of `collection-job-req`.

Resolves #738

### DIFF
--- a/draft-ietf-ppm-dap.md
+++ b/draft-ietf-ppm-dap.md
@@ -1404,10 +1404,10 @@ Clients retrieve the HPKE configuration from each Aggregator by sending a GET to
 `{aggregator}/hpke_config`.
 
 An Aggregator responds with an `HpkeConfigList`, with media type
-"application/dap-hpke-config-list". The `HpkeConfigList` contains one or more
-`HpkeConfig`s in decreasing order of preference. This allows an Aggregator to
-support multiple HPKE configurations and multiple sets of algorithms
-simultaneously.
+"application/ppm-dap;message=hpke-config-list". The `HpkeConfigList` contains
+one or more `HpkeConfig`s in decreasing order of preference. This allows an
+Aggregator to support multiple HPKE configurations and multiple sets of
+algorithms simultaneously.
 
 ~~~ tls-presentation
 HpkeConfig HpkeConfigList<10..2^16-1>;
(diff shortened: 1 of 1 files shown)


draft-ietf-ppm-dap.md
```diff
@@ -1404,9 +1404,9 @@ Clients retrieve the HPKE configuration from each Aggregator by sending a GET to
 `{aggregator}/hpke_config`.
 
 An Aggregator responds with an `HpkeConfigList`, with media type
-"application/dap-hpke-config-list". The `HpkeConfigList` contains one or more
-`HpkeConfig`s in decreasing order of preference. This allows an Aggregator to
-support multiple HPKE configurations and multiple sets of algorithms
+"application/dap;message=hpke-config-list". The `HpkeConfigList` contains one or
```
cjpatton, line 1407: Should there be a space here? Here and below.
```suggestion
"application/dap; message=hpke-config-list". The `HpkeConfigList` contains one or
```

draft-ietf-ppm-dap.md
```diff
@@ -4215,665 +4226,23 @@ component is using. This MAY be used as a hint by the receiver of the request
...
+`application/dap` media type.
 
 Type name:
 
 : application
 
 Subtype name:
 
-: dap-aggregation-job-init-req
+: dap
```
cjpatton, line 4241: "dap" seems likely to have name collisions. Perhaps "ppm-dap" would be a better name.

cjpatton (commented):
Nice to get rid of all this heft!

draft-ietf-ppm-dap.md
tgeoghegan: Either is valid according to RFC 9110
https://www.rfc-editor.org/rfc/rfc9110.html#section-8.3.1

cjpatton (approved):

PR #751, opened 2025-10-06
Extract HTTP text into section 3

### LINKED ISSUE #731: Reorganize text from section 4.3 into 3 and rename to "HTTP usage"
From httpdir review:

> I wonder if the section would be more appropriately titled "HTTP Usage". It may be worth considering extracting some of the text from 4.3 that discusses resource URL construction and moving it to this section. ... The section on handling asynchronous requests seems more appropriate in section 3 as it is not specific to any particular resource.

### LINKED ISSUE #736: Is it OK to GET an aggregation or collection job handled synchronously?
> The examples demonstrate use of GET on the resource URL returned in the
> Location header when the response is asynchronous. However it is not clear if
> GET can be called on this resource when the resource is created synchronously.
> Is it required to provide the correct step query parameter? What happens if the
> step is not included?

Same applies to collection jobs, though without the step query param. When addressing this, consider garbage collection and data retention: how long after completing an aggregation or collection job should a helper or leader be expected to successfully serve GET requests?

tgeoghegan:
Per httpdir review, the discussion of resources (their URLs, how to handle them async or sync) is moved to section 3, which is renamed "HTTP Usage". Additionally we clarify:

- whether the `step` query param is necessary when GETting an aggregation job (yes it is)
- whether a server that handes jobs sync has to support GET on those resources (not required but you can)

Resolves #731
Resolves #736

### DIFF
--- a/draft-ietf-ppm-dap.md
+++ b/draft-ietf-ppm-dap.md
@@ -1006,11 +1006,74 @@ Leader batch bucket 1            Helper batch bucket 1
   title="Lifecycles of protocol objects in the collection interaction"
 }
 
-# Message Transport {#message-transport}
+# HTTP Usage {#http-usage}
 
-Communications between participants are carried over HTTP {{!RFC9110}}. Use of
-HTTPS is REQUIRED to provide server authentication and confidentiality. TLS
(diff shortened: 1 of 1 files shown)


cjpatton (approved):
🚢 !

PR #752, opened 2025-10-07
Rename VDAF preparation to verification

### LINKED ISSUE #729: use "verify" instead of "prepare"
In [Section 4.5.2](https://www.ietf.org/archive/id/draft-ietf-ppm-dap-16.html#section-4.5.2), there are two fields: `leader_encrypted_input_share` and `helper_encrypted_input_share`.  Earlier, in the description of preparation (not a great name: "validation" was right there for the taking; we can talk about the fiction that is offline computation in MPC some time), these were called "report shares".  Should these be renamed?

cjpatton:
Closes #729.


draft-ietf-ppm-dap.md
```diff
@@ -926,16 +926,16 @@ report share       share        share        |
 input share     input share       |          |
      |               |            |          |
  .---+---------------+------------+.         |
-|        prepare input share        +--------+
+|        Verify input share         +--------+
```
cjpatton, line 929: @tgeoghegan what do you think? I'm worried "verify input share" is confusing.
```suggestion
|        Verification         +--------+
```

PR #757, opened 2025-12-18
Wrap long lines in examples per RFC 8792

### LINKED ISSUE #755: RFC 8792 for examples
You have a few that are overflowing the 72 columns.  RFC 8792 is a bit gross, but you probably need it given how some of those examples are constructed (I can't see a way to make them narrower).

tgeoghegan:
We use the single slash strategy for message examples enclosed in `~~~ http` blocks. While we're at it, clean up a couple other incorrectly wrapped lines in the Markdown.

Closes #755

### DIFF
--- a/draft-ietf-ppm-dap.md
+++ b/draft-ietf-ppm-dap.md
@@ -493,6 +493,9 @@ aggregator.
 
 {::boilerplate bcp14-tagged}
 
+All examples in this document wrap long lines using the Single Backslash
+Strategy of {{!RFC8792, Section 7}}.
+
 ### Glossary of Terms
 
 Aggregate result:
@@ -2144,8 +2147,8 @@ The Leader proceeds as follows with each report:
    is complete. The Leader stores `state` and proceeds as in
    {{aggregation-leader-continuation}}.
 
-   Else if the new `state` has type `Finished`, then verification is complete and
-   the state includes an output share, denoted `state.out_share`. The Leader
+   Else if the new `state` has type `Finished`, then verification is complete
+   and the state includes an output share, denoted `state.out_share`. The Leader
    commits to `state.out_share` as described in {{batch-buckets}}.
 
    Else if `state` has type `Rejected`, then the Leader rejects the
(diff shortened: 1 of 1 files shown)


draft-ietf-ppm-dap.md
```diff
@@ -2391,7 +2391,10 @@ If all of the above checks succeed, the input share is valid.
 The Helper handles the aggregation job initialization synchronously:
 
 ~~~ http
-PUT /helper/tasks/8BY0RzZMzxvA46_8ymhzycOB9krN-QIGYvg_RsByGec/aggregation_jobs/lc7aUeGpdSNosNlh-UZhKA
+========== NOTE: '\' line wrapping per RFC 8792 ===========
```
martinthomson: I'm not a fan of the inclusion of this in every example, but it's what the RFC recommends.

The key text is the "NOTE: '\' line wrapping per RFC 8792" part.  How you decorate that to make it clear it's not part of the actual example is up to you.  So this is fine.

Personally, I would probably add something to your "[Conventions and Definitions](https://datatracker.ietf.org/doc/html/draft-ietf-ppm-dap#name-conventions-and-definitions)" section instead.  Just say that all examples use RFC 8792 "\\" wrapping (one backslash).  That works as long as you weren't planning on having a trailing backslash in any figure (I couldn't find one).

draft-ietf-ppm-dap.md
```diff
@@ -2391,7 +2391,10 @@ If all of the above checks succeed, the input share is valid.
...
+
+PUT /helper/tasks/8BY0RzZMzxvA46_8ymhzycOB9krN-QIGYvg_RsByGec/\
+aggregation_jobs/lc7aUeGpdSNosNlh-UZhKA
```
martinthomson: When wrapping, you can add a small indent to the continuation line (which the 8792 processing skips).  For human readers, that's handy.

martinthomson (approved):
This is a "good, but" review :)

tgeoghegan:
Thanks @martinthomson, those are very helpful suggestions.

PR #759, opened 2025-12-18
Clarify definitions for time types

### LINKED ISSUE #756: Time, truncation, and time_precision
> time is the truncated (see [Section 4.1.1](https://datatracker.ietf.org/doc/html/draft-ietf-ppm-dap#timestamps)) time at which the report was generated.

-- https://datatracker.ietf.org/doc/html/draft-ietf-ppm-dap#section-4.5.2-5.1.2.2.1

Section 4.1.1 doesn't really say how that truncation works.  Given that `Time` is a 64-bit value that is *probably* multiplied out by a value greater than a second, I can't see why you would need to mention truncation at all.

However, I say "probably" there advisedly.  Chasing the definition of `time_precision` from Section 4.1.1 to [Section 4.2](https://datatracker.ietf.org/doc/html/draft-ietf-ppm-dap#section-4.2-5.6.1), I find that `time_precision` is a `Duration`.  That's a circular definition.  I would have to guess that `time_precision` is actually a `Duration` with a corresponding `time_precision` in seconds.  But it could equally be milliseconds.

tgeoghegan:
Issue #756 points out some problems with time types, and also discussion in https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap-taskprov/issues/120

- Fix circular definitions of `Duration` and `time_precision`
- Remove confusing mention of "truncation" in the definition of `Report.report_metadata.time` included a confusing
- Add simple examples of `Time`, `Duration` and `Interval` types

Resolves #756

### DIFF
--- a/draft-ietf-ppm-dap.md
+++ b/draft-ietf-ppm-dap.md
@@ -1257,22 +1257,33 @@ enum {
 ### Times, Durations and Intervals {#timestamps}
 
 ~~~ tls-presentation
+uint64 TimePrecision;
+
 uint64 Time;
 ~~~
 
-Times are represented by POSIX timestamps, defined to be integers representing
-a number of `time_precision`s since the Epoch ({{task-configuration}}), defined
-in section 4.16 of {{POSIX}}. That is, the number of seconds after 1970-01-01
-00:00:00 UTC, excluding leap seconds, divided by the task's `time_precision`.
+A `TimePrecision` is an integer number of seconds, used to compute times and
+durations in DAP. The time precision is a parameter of a task
+({{task-configuration}}).
+
+Times are integers, representing a number of `TimePrecision`s since the Epoch,
+defined in section 4.16 of {{POSIX}}. That is, the number of seconds after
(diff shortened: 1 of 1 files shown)


draft-ietf-ppm-dap.md
```diff
@@ -1260,10 +1260,11 @@ enum {
 uint64 Time;
 ~~~
 
-Times are represented by POSIX timestamps, defined to be integers representing
-a number of `time_precision`s since the Epoch ({{task-configuration}}), defined
-in section 4.16 of {{POSIX}}. That is, the number of seconds after 1970-01-01
-00:00:00 UTC, excluding leap seconds, divided by the task's `time_precision`.
+Times are integers, representing a number of `time_precision`s
+({{task-configuration}}) since the Epoch, defined in section 4.16 of {{POSIX}}.
```
martinthomson: My suggestion here would be to move the definition of `time_precision` to this section.  This is where I would look for that and you avoid having to jump forward just to understand a very basic concept.

PR #766, opened 2026-01-21
Add verification key id to the aggregation job

martinthomson:
The goal here is not to describe how verification keys might be updated, but to build the protocol so that they could be, provided that the analysis supports the protocol being safe.

Chris and I discussed this for a while.  A few things came out of that, none of which seemed insurmountable.  So I think we convinced ourselves that a different method for managing verification keys is *possible*, even though we agree that there is no need to go through the complete exercise of defining an alternative method right now.

Firstly, privacy needs to be maintained.  The analysis in the paper (https://ia.cr/2023/130) suggests that you need to commit to the verification key prior to generating a report in order to prevent the key being selected in a way that might allow privacy to be compromised. My sense of the current set of VDAFs is that there isn't an easy path to privacy compromise even if an attacker could select the verification key, but this is not something that is generally true of any VDAF in DAP.

In practice, using a key agreement protocol with contributory behavior is likely to avoid this class of problem.  Think Diffie-Hellman or one of the modern KEMs that fold entropy from both parties into the result (or even commit and reveal public coin protocols). But we'd need a fresh analysis to demonstrate that this is truly safe.

Second, there is the question of correctness.  With knowledge of the verification key, either Aggregator can generate a report that will poison the result.  This is somewhat academic, because the VDAFs we have are vulnerable to an Aggregator tweaking its own share of the output, but it's worth noting as the other security reason.  It also doesn't bear much on this situation, as this is already a vulnerability in the protocol.

(And no, you can't fix this with an online key agreement.  A Leader that wants to poison the aggregate can just insert an extra record after the key agreement finishes, putting it in the past as necessary.)

So, I think we convinced ourselves that this change makes it possible to do things in the future with the verification key without protocol changes, which is a nice feature.  That opens the path to making extra-long-lived tasks.  That will come after some restructuring.

As a practical matter, I folded the value into the initialization of the aggregation job, but not in any continuations.  Happy to add that, which would make that parameter consistent throughout, but it isn't really necessary. I just wanted to flag that as something to think about.

### DIFF
--- a/draft-ietf-ppm-dap.md
+++ b/draft-ietf-ppm-dap.md
@@ -137,6 +137,17 @@ informative:
     date: 2016
     target: "https://privacytools.seas.harvard.edu/files/privacytools/files/complexityprivacy_1.pdf"
 
+  DPRS23:
+    title: "Verifiable Distributed Aggregation Functions"
+    author:
+      - name: Hannah Davis
+      - name: Christopher Patton
+      - name: Mike Rosulek
+      - name: Phillipp Schoppmann
+    date: 2024-09-25
+    refcontent: "Privacy Enhancing Technologies Symposium (PETS)"
+    target: https://ia.cr/2023/130/20240925:184929
+
 --- abstract
 
 There are many situations in which it is desirable to take measurements of data
@@ -166,7 +177,11 @@ aggregator.
 
 (RFC EDITOR: Remove this section.)
 
-(\*) Indicates a change that breaks wire compatibility with the previous draft.
+
+18:
+
+- Add verification key ID to aggregation jobs to enable but not require
+  verification key management. (\*) (#766)
 
 17:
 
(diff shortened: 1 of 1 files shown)


draft-ietf-ppm-dap.md
```diff
@@ -2136,6 +2149,12 @@ struct {
 
 This message consists of:
 
+* `verification_key_id`: An identifier for the shared secret
+  that will be used for this aggregation job.
```
cjpatton:
```suggestion
* `verification_key_id`: An identifier for the VDAF verification key ({{Section 5.2 of !VDAF}}), the shared secret used by the Aggregators to verify the report's validity.
```

cjpatton (commented):
Nice! Just one minor change needed.

cjpatton (approved):

draft-ietf-ppm-dap.md
```diff
@@ -2152,6 +2172,14 @@ This message consists of:
   encoded `agg_param` and `part_batch_selector` fields. That is, the remainder
   of the HTTP message consists of `verify_inits`.
 
+{:aside}
```
tgeoghegan: Do `{:aside}`s appear in the final document?

tgeoghegan (approved):

PR #769, opened 2026-01-21
Collection job extensions

martinthomson:
This defines extensions for the collection job.

There are two main commits.  The first adds the fields so that the Collector can add extensions.  The second makes sure that information gets to the Helper and then is bound to the final shares, so that we can be confident that all three entities agree (Collector, Leader, and Helper).  A third commit really just covers a gap in the others, which is to describe what the Leader and Helper are expected to do in processing extensions, which is just whatever the extension definition says.  No doubt I'll add to the pile in response to comments as well.

Commit messages contain more detail, but I wanted to highlight one thing in particular:

This removes batch ID from the AAD of the aggregate shares.  I think that this is justified, but I want to check that this is OK.  More analysis in the second commit comment on that.  If this is agreed, then I suggest we follow up and remove the PartialBatchSelector concept from the protocol entirely, as a simplification.

I've also defined an `invalidExtension` error code, which covers a few different error conditions (bad encoding of an extension, bad sort order).  In TLS we've found that distinguishing unsupported extensions is worth a code on its own; my theory is that this is even more important here, where an unsupported extension always means failure.

### DIFF
--- a/draft-ietf-ppm-dap.md
+++ b/draft-ietf-ppm-dap.md
@@ -183,6 +183,8 @@ aggregator.
 - Add verification key ID to aggregation jobs to enable but not require
   verification key management. (\*) (#766)
 
+- Define colleciton job extensions. (\*) (#769)
+
 17:
 
 - Bump version tag from "dap-16" to "dap-17". (\*)
@@ -1893,7 +1895,7 @@ enum {
 Field `extension_type` indicates the type of extension, and `extension_data`
 contains the opaque encoding of the extension.
 
-Extensions are mandatory to implement. Unrecognized extensions are handled as
+Extensions are mandatory to support. Unrecognized extensions are handled as
 specified in {{input-share-validation}}.
 
 ## Verifying and Aggregating Reports {#aggregate-flow}
@@ -3084,6 +3086,9 @@ conditions:
   If not, the Leader MUST fail the collection job
   with the error indicated by the processing model for the extension.
 
+The Leader then performs any configuration changes or processing
+defined in the included extensions; see {{collect-ext}}.
+
 Having validated the `CollectionJobReq`, the Leader begins working with the
 Helper to aggregate the reports satisfying the query (or continues this process,
 depending on whether the Leader is aggregating eagerly; {{eager-aggregation}})
@@ -3170,7 +3175,7 @@ encoded(struct {
     query = encoded(Empty),
   } Query,
   agg_param = [0x00, 0x01, ...],
-  extensions = encoded(Empty)
+  extensions = [],
 } CollectionJobReq)
 
 HTTP/1.1 200
@@ -3213,7 +3218,7 @@ encoded(struct {
     } TimeIntervalQueryConfig),
   },
   agg_param = encoded(Empty),
-  extensions = encoded(Empty)
+  extensions = [],
 } CollectionJobReq)
 
 HTTP/1.1 200
@@ -3279,13 +3284,25 @@ a type identifier (`extension_type`)
 and arbitrary data (`extension_data`).
 The data is structured according to the definition of the extension.
 
-Extensions are mandatory to understand and support by both Aggregators.
+Extensions are mandatory to support by both Aggregators.
 A collection job that contains an unrecognized or unsupported extension
-MUST be failed with an `unsupportedExtension` error; see {{errors}}.
+MUST be aborted with an `unsupportedExtension` error; see {{errors}}.
 
 Extensions MUST be encoded in strictly increasing order.
-If any `extension_type` value is equal to or less than the extension that precedes it,
+If any `extension_type` value is equal to or less than that of the extension that precedes it,
 the job MUST be failed with an `invalidExtension` error; see {{errors}}.
+
+Each collection job extension MUST define how Aggregators
+alter their handling of collection jobs
+when the extension is present,
+including any validation of the included `extension_data`.
+
+An extension could limit its compatibility
+to specific VDAFs, batch modes, or those with certain properties.
+For instance, an extension might only apply
+if the VDAF supports eager aggregation; see {{eager-aggregation}}.
+In that case, jobs can be failed with an `invalidExtension` error
+if the extension is incompatible.
 
 Collection job extensions are distinct from report and task extensions.
 Collection job extensions are only agreed between Collector, Leader, and Helper;
@@ -3359,8 +3376,8 @@ struct {
 } BatchSelector;
 
 struct {
+  CollectionJobReq collection_job_req;
   BatchSelector batch_selector;
-  opaque agg_param<0..2^32-1>;
   uint64 report_count;
   opaque checksum[32];
 } AggregateShareReq;
@@ -3370,10 +3387,11 @@ The media type of `AggregateShareReq` is
 "application/ppm-dap;message=aggregate-share-req". The structure contains the
 following parameters:
 
-* `batch_selector`: The "batch selector", the contents of which depends on the
-  indicated batch mode (see {{batch-modes}}.
+* `collection_job_req`: The details of the collection job,
+  as provided by the Collector when initiating the job; see {{collect-init}}.
 
-* `agg_param`: The encoded aggregation parameter for the VDAF being executed.
+* `batch_selector`: The "batch selector", the contents of which depends on the
+  indicated batch mode; see {{batch-modes}}.
 
 * `report_count`: The number number of reports included in the batch, as
   computed above.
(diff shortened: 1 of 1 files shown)


draft-ietf-ppm-dap.md
```diff
@@ -3199,6 +3217,56 @@ encoded(struct {
...
+{: #f-collect-ext title="Collection Job Extensions"}
+
+These extensions (shown in {{f-collect-ext}}) contain
+a type identifier (`extension_type`)
+and arbitrary data (`extension_data`).
+The data is structured according to the definition of the extension.
+
+Extensions are mandatory to understand and support by both Aggregators.
+A collection job that contains an unrecognized or unsupported extension
+MUST be failed with an `unsupportedExtension` error; see {{errors}}.
```
cjpatton: I think this would be more consistent with the language we've been using?
```suggestion
MUST be aborted with an `unsupportedExtension` error; see {{errors}}.
```

draft-ietf-ppm-dap.md
```diff
@@ -3299,8 +3368,10 @@ MUST fail the job with error `invalidMessage`.
 If the `AggregateShareReq` is malformed, the Helper MUST fail the job with error
 `invalidMessage`.
 
-The Helper then verifies that the `BatchSelector` in the Leader's request
-determines a batch that can be collected. If the selector does not identify a
+The Helper then verifies that the `batch_selector` in the Leader's request
+is consistent with the `collection_job_req.query`.
```
cjpatton: This is a bit tricky because the definition "consistency" depends on the batch mode. I expect it'll be usually be pretty obvious in practice. For time interval, the specified interval should match between the query and batch selector; for leader-selected, the query is the empty string, so there's nothing to check. What would we do for the new batch mode for attribution?

cjpatton (commented):
Looks reasonable to me overall. One thing is that we'll need to provide some guidance for "Query" and "BatchSeletor" consistency. In practice, I think batch modes would need to define this for themselves.

> This removes batch ID from the AAD of the aggregate shares. I think that this is 
justified, but I want to check that this is OK. More analysis in the second commit comment on that. If this is agreed, then I suggest we follow up and remove the PartialBatchSelector concept from the protocol entirely, as a simplification.

The main purpose of the PartialBatchSelector is to encode info that the Helper needs to map a report to a bucket. This is needed if the reprots themselves don't encode this info, as is is the case for leader-selected.

draft-ietf-ppm-dap.md
```diff
@@ -3199,6 +3217,56 @@ encoded(struct {
...
+The data is structured according to the definition of the extension.
+
+Extensions are mandatory to understand and support by both Aggregators.
+A collection job that contains an unrecognized or unsupported extension
+MUST be failed with an `unsupportedExtension` error; see {{errors}}.
+
+Extensions MUST be encoded in strictly increasing order.
+If any `extension_type` value is equal to or less than the extension that precedes it,
+the job MUST be failed with an `invalidExtension` error; see {{errors}}.
+
```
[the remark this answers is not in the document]
martinthomson: Report extensions don't have the same requirement because they don't need it.  It wouldn't hurt to have it, because you remove an axis of variability - and a "covert" channel - by doing so.

draft-ietf-ppm-dap.md
martinthomson: It's probably worth saying that explicitly then.

martinthomson: I've taken a swing at this, which ends up levying a requirement on the definition of batch modes.  Take a look and see what you think of the last one.

draft-ietf-ppm-dap.md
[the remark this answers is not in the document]
cjpatton: https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/issues/772

cjpatton (approved):

draft-ietf-ppm-dap.md
```diff
@@ -3312,10 +3384,20 @@ If the number of validated reports in the batch is not equal to or greater than
...
 `invalidBatchSize`.
 
-The aggregation parameter MUST match the aggregation parameter used in
+The aggregation parameter in `collection_job_req.agg_param`
+MUST match the aggregation parameter used in
 aggregation jobs pertaining to this batch. If not, the Helper MUST fail the job
 with error `invalidMessage`.
 
+The Helper then validates any extensions in `collection_job_req.extensions`,
+following the same logic as the Leader; see {{collect-init}}.
```
tgeoghegan:
```suggestion
following the same logic as the Leader; see {{collect-ext}}.
```

Seems like a more specific and useful reference.

draft-ietf-ppm-dap.md
```diff
@@ -3627,6 +3724,9 @@ query does not specify a set of valid batch buckets.
 
 The payload of `PartialBatchSelector.config` is empty.
 
+It is impossible for a `PartialBatchSelector.config` to be inconsistent
+with `Query.config` for this batch mode.
```
tgeoghegan: {{collect-aggregate}} discusses checking consistency between (full, non-partial) `BatchSelector`s and `Query`s:

> The Helper then verifies that the `batch_selector` in the Leader's request is consistent with the `collection_job_req.query`.
>
> Each batch mode needs to define any consistency checks necessary.

But this passage and the corresponding one under {{leader-selected-batch-mode}} discuss checking consistency between `PartialBatchSelector` and `Query`. 

In time interval there could be disagreement between:

```
struct {
  Interval batch_interval;
} TimeIntervalQueryConfig;
```
and
```
struct {
  Interval batch_interval;
} TimeIntervalBatchSelectorConfig;
```

...so this text should direct implementations to compare the two `Interval`s.

draft-ietf-ppm-dap.md
```diff
@@ -3199,6 +3221,56 @@ encoded(struct {
...
+} CollectionJobExtensionType;
+~~~
+{: #f-collect-ext title="Collection Job Extensions"}
+
+These extensions (shown in {{f-collect-ext}}) contain
+a type identifier (`extension_type`)
+and arbitrary data (`extension_data`).
+The data is structured according to the definition of the extension.
+
+Extensions are mandatory to understand and support by both Aggregators.
```
tgeoghegan: Our existing language for report extensions says "Extensions are mandatory to implement." We should be consistent across extension definitions in this doc. Which do we like better?

draft-ietf-ppm-dap.md
```diff
@@ -3199,6 +3221,56 @@ encoded(struct {
...
+A collection job that contains an unrecognized or unsupported extension
+MUST be aborted with an `unsupportedExtension` error; see {{errors}}.
+
+Extensions MUST be encoded in strictly increasing order.
+If any `extension_type` value is equal to or less than the extension that precedes it,
```
tgeoghegan:
```suggestion
If any `extension_type` value is equal to or less than that of the extension that precedes it,
```

draft-ietf-ppm-dap.md
```diff
@@ -3199,6 +3217,56 @@ encoded(struct {
...
+including any validation of the included `extension_data`.
+
+An extension could limit its compatibility
+to specific VDAFs, batch modes, or those with certain properties.
+For instance, an extension might only apply
+if the VDAF supports eager aggregation; see {{eager-aggregation}}.
+In that case, jobs can be failed with an `invalidExtension` error
+if the extension is incompatible.
+
+Collection job extensions are distinct from report and task extensions.
```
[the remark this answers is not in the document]
tgeoghegan: I think I'm OK with a dangling reference here, especially since [the change to define task configuration extensions](https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/pull/774)is in flight and will (🤞🏻) land before terribly long.

draft-ietf-ppm-dap.md
```diff
@@ -3720,6 +3835,9 @@ struct {
 
 where `batch_id` is the batch ID selected by the Leader.
 
+A `PartialBatchSelector.config` is consistent
+with `Query.config` if the `batch_id` values are equal.
```
tgeoghegan: I believe the `Query.config` for leader selected batch mode is empty. The collector (who constructs the query) doesn't know the batch ID; it gets chosen by the leader.

tgeoghegan (changes requested):

draft-ietf-ppm-dap.md
```diff
@@ -168,6 +168,10 @@ aggregator.
 
 (\*) Indicates a change that breaks wire compatibility with the previous draft.
 
+18:
+
```
tgeoghegan: I just merged #766, so this will now conflict with the changelog we started there. Sorry for the runaround here.
```suggestion
```

draft-ietf-ppm-dap.md
```diff
@@ -3199,6 +3221,56 @@ encoded(struct {
 } CollectionJobResp)
 ~~~
 
+### Collection Job Extensions {#collect-ext}
+
+A Collector can add extensions to the creation of a collection job
+to pass additional information to Aggregators (primarily the Leader)
+about how to handle the collection job.
```
tgeoghegan: Did we consider whether it's OK for two different collection jobs in the same task to use different extensions? Mind you that's also something we should consider w.r.t. report extensions, so maybe we should discuss adding such a requirement separately from this change.

draft-ietf-ppm-dap.md
martinthomson: I'll look after this properly.

draft-ietf-ppm-dap.md
martinthomson: Of course.  The original text would have been correct in that case.

draft-ietf-ppm-dap.md
martinthomson: OK, I've gone through and used the word "support".  I think that's OK: it's the word other specs tend to use and - more importantly - the spec is already good about what to do if an unsupported extension appears.

martinthomson:
I've gone over the comments (thanks Tim!) and I think that I resolved them.  Apologies for misunderstanding a few on the way.

tgeoghegan (approved):

PR #777, opened 2026-03-17
Widen types used in VDAF config structs

tgeoghegan:
The VDAF config structs inherited from draft-ietf-ppm-dap-taskprov used `uint32` everywhere. That's probably fine for things like the number of buckets in `Prio3Histogram`: the performance is unlikely to scale to vectors of 2^32-1 or more entries. However it's unfortunate to forbid particularly large measurements in `Prio3Sum` or `Prio3SumVec`. Those VDAFs use `Field64` and `Field128`, respectively, so they're capable of representing measurements exceeding `2^32-1` without a performance hit.

This commit widens all the VDAF config structure fields.

`Prio3Sum::max_measurement` is now `uint64`.
`Prio3SumVec::max_measurement` is now `uint128` (we add a simple definition of that type in {{basic-definitions}}`.

I widened all the remaining types (`length`s, `chunk_length`s, `max_weight`) to `uint64`. We _could_ leave these as `uint32` and it's unlikely it would restrict the utility of VDAFs. But on the other hand, these quantities describe lengths of vectors or slices, and since the majority of deployment targets are going to be LP64, using a 64 bit type for this seems more natural to me.

See also some discussion in the #ppm channel of ietf.slack.com ([1]).

[1]: https://ietf.slack.com/archives/C02P9620XT6/p1773341361058649

### DIFF
--- a/draft-ietf-ppm-dap.md
+++ b/draft-ietf-ppm-dap.md
@@ -4976,43 +4976,64 @@ used.
 
 ~~~ tls-presentation
 struct {
-    uint32 max_measurement; /* largest summand */
+    uint64 max_measurement;
 } Prio3SumConfig;
 ~~~
+
+`max_measurement` is the largest summand permitted in this instantiation of the
+VDAF. The value MUST be less than the modulus of the field associated with the
+circuit; see {{!VDAF, Section 6.1.4}}.
 
 ## Prio3SumVec
 
 ~~~ tls-presentation
 struct {
-    uint32 length;          /* length of the vector */
-    uint32 max_measurement; /* maximum value of each summand */
-    uint32 chunk_length;    /* size of each proof chunk */
+    uint32 length;
+    uint64 max_measurement;
+    uint32 chunk_length;
 } Prio3SumVecConfig;
 ~~~
+
+- `length` is the length of the vectors being summed.
+- `max_measurement` is the largest summand permitted in this instantiation of
+  the VDAF.
+- `chunk_length` is the size of each proof chunk.
 
 ## Prio3Histogram
 
 ~~~ tls-presentation
 struct {
-    uint32 length;       /* number of buckets */
-    uint32 chunk_length; /* size of each proof chunk */
+    uint32 length;
+    uint32 chunk_length;
 } Prio3HistogramConfig;
 ~~~
+
+- `length` is the number of buckets in the histogram.
+- `chunk_length` is the size of each proof chunk.
 
 ## Prio3MultihotCountVec
 
 ~~~ tls-presentation
 struct {
-    uint32 length;       /* length of the vector */
-    uint32 chunk_length; /* size of each proof chunk */
-    uint32 max_weight;   /* largest vector weight */
+    uint32 length;
+    uint32 chunk_length;
+    uint64 max_weight;
 } Prio3MultihotCountVecConfig;
 ~~~
+
+- `length` is the length of the vectors being summed.
+- `chunk_length` is the size of each proof chunk.
+- `max_weight` is the largest weight allowed for measurements; the maximum
+  number of true entries in the bit vector. The value MUST be less than the
+  modulus of the field associated with the circuit; see {{!VDAF, Section
+  6.1.4}}.
 
 ## Poplar1
 
 ~~~ tls-presentation
 struct {
-    uint16 bits; /* bit length of the input string */
+    uint16 bits;
 } Poplar1Config;
 ~~~
+
+`bits` is the number of bits in input strings.


tgeoghegan:
I'm open to pushback on various aspects of this. I think that at a strict minimum, `{Prio3Sum, Prio3SumVec}::max_measurement` has to be raised to `uint64`.

But given that `Prio3SumVec` uses `Field128`, I don't see why we'd restrict it. If we really strongly feel that nobody will ever measure things larger than 2^64-1 with `Prio3SumVec`, then we should instantiate it with `Field64`, no?

As for the length, chunk size, etc. parameters, I can live with those being `uint32` if the WG consensus goes against me.

We'll need a corresponding change on draft-ietf-ppm-l1-bound-sum. I'll draft that PR once we agree on a direction for this one.

divergentdave:
> If we really strongly feel that nobody will ever measure things larger than 2^64-1 with `Prio3SumVec`, then we should instantiate it with `Field64`, no?

The field was originally selected for Fiat-Shamir soundness, since Prio3SumVec uses joint randomness. If we were to instantiate it with Field64, that could work, but we'd have to increase the number of proofs to compensate.

One other nitpick: the constraint that wraparound imposes on field size hinges not on individual measurements, but on the sum of a batch worth of measurements.

draft-ietf-ppm-dap.md
```diff
@@ -4976,43 +4982,63 @@ used.
...
+
 ## Prio3SumVec
 
 ~~~ tls-presentation
 struct {
-    uint32 length;          /* length of the vector */
-    uint32 max_measurement; /* maximum value of each summand */
-    uint32 chunk_length;    /* size of each proof chunk */
+    uint64 length;
+    uint128 max_measurement;
```
martinthomson: 64 should be enough for anyone.

draft-ietf-ppm-dap.md
```diff
@@ -4976,43 +4982,63 @@ used.
...
+6.1.4}}.
+
 ## Prio3SumVec
 
 ~~~ tls-presentation
 struct {
-    uint32 length;          /* length of the vector */
-    uint32 max_measurement; /* maximum value of each summand */
-    uint32 chunk_length;    /* size of each proof chunk */
+    uint64 length;
```
martinthomson: Really?  2^32 is a crazy length.  I can see why you might not want 2^16.

draft-ietf-ppm-dap.md
```diff
@@ -4976,43 +4982,63 @@ used.
...
+    uint128 max_measurement;
+    uint64 chunk_length;
```
martinthomson: This should track `length`.

martinthomson (commented):
I'm not thrilled with the notion of a 128 maximum, which you have only done for Prio3SumVec.  Some other values could be shorter...

The lengths have no business being 64-bit values.

tgeoghegan:
Thanks for the feedback, Martin. I'm comfortable with the position you've laid out here. I want to keep this PR open until the end of the week (Friday March 27)  to see if anyone else shares an opinion.

draft-ietf-ppm-dap.md
```diff
@@ -4976,43 +4982,63 @@ used.
...
 ~~~ tls-presentation
 struct {
-    uint32 max_measurement; /* largest summand */
+    uint64 max_measurement;
 } Prio3SumConfig;
 ~~~
 
+`max_measurement` is the largest summand permitted in this instantiation of the
+VDAF. The value MUST not exceed the modulus of `Field64`; see {{!VDAF, Section
+6.1.4}}.
```
cjpatton: To @divergentdave's earlier point, one could instantiate SumVec with a different field, in which case this MUST wouldn't be applicable.

Stepping back a bit, I'm a little uncomfortable using the size of the _default_ field of the circuit to inform the data type. I would maybe go with `uint128` for ranges (max_measurement and max_weight) but `uint64` for lengths. (Or even `uint32` for length, as @martinthomson suggests.)

cjpatton (commented):
I don't think we should use the _default_ field to inform the type. It ma

cjpatton (commented):
Conclusion from 2026/3/26 sync:

1. values get 64 bits
2. lengths get 32 bits

draft-ietf-ppm-dap.md
tgeoghegan: We agreed to use `uint64` as `max_measurement` irrespective of fields. But one thing I want to point out is that if someone did instantiate e.g. `Prio3Sum` with another field, they'd have to define that VDAF, choose a codepoint and specify a VDAF config encoding. I don't think it's a goal of these VDAF config encodings to support instantiations with fields besides the ones in draft-irtf-cfrg-vdaf.

draft-ietf-ppm-dap.md
```diff
@@ -4976,43 +4976,62 @@ used.
...
-    uint32 chunk_length;    /* size of each proof chunk */
+    uint32 length;
+    uint64 max_measurement;
+    uint32 chunk_length;
 } Prio3SumVecConfig;
 ~~~
 
+- `length` is the length of the vectors being summed.
+- `max_measurement` is the largest summand permitted in this instantiation of
+  the VDAF.
```
tgeoghegan, line 4999: I want to draw attention to this line: `Prio3Sum.max_measurement` is limited to the Field64 modulus, but we don't need any such restriction here since the field is `Field128`.

tgeoghegan (commented):
Updated the PR to reflect WG consensus from the 3/26/26 call.

draft-ietf-ppm-dap.md
```diff
@@ -4976,43 +4976,62 @@ used.
...
 ## Prio3MultihotCountVec
 
 ~~~ tls-presentation
 struct {
-    uint32 length;       /* length of the vector */
-    uint32 chunk_length; /* size of each proof chunk */
-    uint32 max_weight;   /* largest vector weight */
+    uint32 length;
+    uint32 chunk_length;
+    uint32 max_weight;
```
cjpatton: nit: this is the maximum of the sum of a bunch of bits in the vector. I would make this uint64 to match max_measurement semantically.
```suggestion
    uint64 max_weight;
```

draft-ietf-ppm-dap.md
cjpatton: I hear you, but I think the intent of the text is still unclear. I still think we should drop it -- it's more a consideration for the VDAF itself --- but if we want to say something, I would make the statement more precise.
```suggestion
`max_measurement` is the largest summand permitted in this instantiation of the
VDAF. The value MUST be less than the modulus of the field associated with the the circuit; see {{!VDAF, Section
6.1.4}}.
```

draft-ietf-ppm-dap.md
```diff
@@ -4976,43 +4976,62 @@ used.
...
 } Prio3MultihotCountVecConfig;
 ~~~
 
+- `length` is the length of the vectors being summed.
+- `chunk_length` is the size of each proof chunk.
+- `max_weight` is the largest weight allowed for measurments; the maximum number
+  of true entries in the bit vector.
```
cjpatton: For consistency, we might want the same MUST related to the field modulus.

cjpatton (approved):
Beautiful

PR #778, opened 2026-03-26
Remove partial batch selector from collection job response

### LINKED ISSUE #767: Potential bug in batch mode encodings for output shares
`CollectionJobResp` includes the partial batch selector, presumably because you don't want to include all the gruesome details in the public API.  But the `AggregateShareAad` contains the full batch selector, as passed from Leader to Helper.  If these values are different, the Collector will be unable to decrypt the shares.

Also, it's not clear that it is always possible for the Helper to determine what this partial selector is; from the current batch modes, sure, but not necessarily always.

I think that what you want is to have the Leader pass two things to the helper: the Query that was provided by the Collector and the detailed selector that it used to select batches for aggregation.  The Query would be used in encryption and decryption.  If there are details that the Leader needs or wants to provide the Collector regarding the collection job, then the partial selector can be used to communicate that.

cjpatton:
Closes #767.

### DIFF
--- a/draft-ietf-ppm-dap.md
+++ b/draft-ietf-ppm-dap.md
@@ -192,6 +192,9 @@ aggregator.
 
 - Require report extensions to be sorted by type. (\*) (#775)
 
+- Remove the partial batch selector from the collection job response. (\*)
+  (#778)
+
 17:
 
 - Bump version tag from "dap-16" to "dap-17". (\*)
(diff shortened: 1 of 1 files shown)


tgeoghegan (approved):
Looks right to me. For posterity, let's add an entry to the changelog highlighting this wire-breaking change.

martinthomson (approved):

Issue #762, opened 2026-01-08
Batch bucket design constrains implementation of anti-replay


martinthomson:
... but maybe not enough.

The current design for batching operates at two levels.  The most important level is collection jobs.  Collection jobs are what the batch mode primarily governs.  That determines which reports contribute to an aggregate.

Buckets are used internally between the Leader and Helper to split the work they do into smaller slices.  On the face of it, these don't need anything more than the Leader to sweep a bunch of reports into a bucket.  But for efficiency reasons, especially for anti-replay, it helps if the Helper understands what the strategy is.  If the Leader were to do arbitrary selection, the Helper doesn't get to take any shortcuts.

Consider the full leader-selected mode: there the Leader could pick any report it has for inclusion in any batch.  A Helper in that situation that wants to guarantee that it can do anti-replay has no choice but to remember something about every report it processes, enough to ensure that it doesn't accidentally reject non-replayed reports, but that it can guarantee that replay doesn't occur.  In practice, that probably means saving every report ID it has ever seen in the task.  That's pretty bad, but that seems to be the only viable approach if the Leader is in control of how reports are batched[^1].

[^1]: The leader isn't similarly constrained.  It knows what selection criteria it used and can use a far more efficient approach.

The time interval batch buckets use a different descriptor.  The idea there is that each bucket contains all the reports that will ever be aggregated for a slice of time (by the way, the presentation of this concept in the draft is quite badly scattered; it's not very easy to follow).  As a result, Helpers can switch to remembering time spans, rather than individual reports.  The idea is that this is perhaps more efficient a mode[^2].

[^2]: Again, the leader is better positioned to manage this.  For instance, a Leader can choose to aggregate in time order always.  The Leader will know that after aggregating a bucket $(t_1,t_2)$ it never has to accept a report with a time before $t_2$.  A Helper can't assume this and so has to remember every span it ever receives; the only way it can reduce outstanding state is to coalesce adjacent time spans.

The definition of each batch bucket is supposed to be the unit at which the protocol operates its anti-replay.

However, by coupling the concept of how reports are batched with how anti-replay is managed, the protocol simultaneously over-constrains and under-constrains.

A different approach might be both simpler and more powerful.  Let's say that the batch bucket only dealt with the mechanics involved with handling a group of reports.  The protocol would then only have to assign an ID to that batch, with the Leader choosing which reports to include.  The protocol is simplified, because you only have one method for handling buckets.

[...]

PR #779, opened 2026-03-27
Replace partial batch selector with aggregation job extensions

cjpatton:
Based on #778 (merge that first).
Partially addresses #762.

The partial batch selector was used to convey additional information required to map reports to batch buckets as specified 
by the batch mode. The idea now is that the same information would be conveyed by an extension. For example, the leader-selected batch mode now defines an extension to convey the batch ID.

The intent is to allow for extending aggregation in other ways in the future. This also slightly simplifies the description of the protocol, since the reader now only needs to understand the relationship between Query and BatchSelector.

This is a wire-breaking change: the format of AggregationJobInitReq changes from including a PartialBatchSelector field to an extensions list.

### DIFF
--- a/draft-ietf-ppm-dap.md
+++ b/draft-ietf-ppm-dap.md
@@ -192,8 +192,10 @@ aggregator.
 
 - Require report extensions to be sorted by type. (\*) (#775)
 
-- Remove the partial batch selector from the collection job response. (\*)
-  (#778)
+- Replace the partial batch selector with aggregation job extensions. If a
+  batch mode needs to convey information to the Helper during aggregation, it
+  defines an aggregation job extension. For example, the leader-selected batch
+  mode defines an extension that conveys the batch ID. (\*) (#762)
 
 17:
 
@@ -2141,6 +2143,25 @@ both of which can violate privacy ({{replay-protection}}). Before committing to
 an output share, the Aggregators check whether its report ID has already been
 aggregated and whether the batch bucket being updated has been collected.
 
+### Aggregation Job Extensions {#agg-job-extensions}
+
+The Leader may include extensions in the `AggregationJobInitReq` to convey
+additional parameters for the aggregation job to the Helper. Extensions use
+the following types:
+
+~~~ tls-presentation
+enum {
+  reserved(0),
+  (65535)
+} AggregationJobExtensionType;
+~~~
+
+Each extension is identified by an `AggregationJobExtensionType` codepoint.
+Extension type values are defined in {{agg-job-extensions-registry}}.
+
+Extensions are mandatory to support. If the Helper does not recognize an
+extension type, it MUST abort with error `unsupportedExtension`.
+
 ### Eager Aggregation {#eager-aggregation}
 
 In general, aggregation cannot begin until the Collector specifies a query and
@@ -2263,14 +2284,14 @@ for the relevant reports.
 
 ~~~ tls-presentation
 struct {
-  BatchMode batch_mode;
-  opaque config<0..2^16-1>;
-} PartialBatchSelector;
+  AggregationJobExtensionType extension_type;
+  opaque extension_data<0..2^16-1>;
+} AggregationJobExtension;
 
 struct {
   uint8 verification_key_id;
   opaque agg_param<0..2^32-1>;
-  PartialBatchSelector part_batch_selector;
+  AggregationJobExtension extensions<0..2^16-1>;
   VerifyInit verify_inits[verify_inits_length];
 } AggregationJobInitReq;
 ~~~
@@ -2288,16 +2309,14 @@ This message consists of:
   initializing an aggregation job, the Leader MUST validate the parameter as
   described in {{agg-param-validation}}.
 
-* `part_batch_selector`: The "partial batch selector" used by the Aggregators to
-  determine how to aggregate each report. Its contents depends on the indicated
-  batch mode. This field is called the "partial" batch selector because
-  depending on the batch mode, it may only partially determine a batch. See
-  {{batch-modes}}.
+* `extensions`: A list of aggregation job extensions ({{agg-job-extensions}})
+  providing additional parameters for the aggregation job. Extensions are
+  mandatory to support.
 
 * `verify_inits`: the sequence of `VerifyInit` messages constructed in the
   previous step. Here `verify_inits_length` is the length of the HTTP message
   content ({{!RFC9110, Section 6.4}}), minus the lengths in octets of the
-  encoded `agg_param` and `part_batch_selector` fields. That is, the remainder
+  encoded `agg_param` and `extensions` fields. That is, the remainder
   of the HTTP message consists of `verify_inits`.
 
 {:aside}
@@ -2407,9 +2426,11 @@ conditions:
 * Whether the `AggregationJobInitReq` is malformed. If so, the the Helper MUST
   fail the job with error `invalidMessage`.
 
-* Whether the batch mode indicated by `part_batch_selector.batch_mode` matches
-  the task's batch mode. If not, then the Helper MUST fail the job with error
-  `invalidMessage`.
+* Whether the extensions in `AggregationJobInitReq.extensions` are valid:
+  - If any extension type is unrecognized, the Helper MUST fail the job with
+    error `unsupportedExtension`.
+  - If the extensions are not encoded in strictly increasing order of
+    `extension_type`, the Helper MUST fail the job with error `invalidMessage`.
 
 * Whether the aggregation parameter is valid as described in
   {{agg-param-validation}}. If the aggregation parameter is invalid, then the
@@ -2610,12 +2631,12 @@ Authorization: Bearer auth-token
 
 encoded(struct {
   agg_param = [0x00, 0x01, 0x02, 0x04, ...],
-  part_batch_selector = struct {
-    batch_mode = BatchMode.leader_selected,
-    config = encoded(struct {
-      batch_id = [0x1f, 0x1e, ..., 0x00],
-    } LeaderSelectedPartialBatchSelectorConfig),
-  } PartialBatchSelector,
+  extensions = [
+    struct {
+      extension_type = AggregationJobExtensionType.leader_selected_batch_id,
+      extension_data = encoded([0x1f, 0x1e, ..., 0x00] BatchID),
+    } AggregationJobExtension,
+  ],
   verify_inits,
 } AggregationJobInitReq)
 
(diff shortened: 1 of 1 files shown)


draft-ietf-ppm-dap.md
```diff
@@ -192,8 +192,10 @@ aggregator.
 
 - Require report extensions to be sorted by type. (\*) (#775)
 
-- Remove the partial batch selector from the collection job response. (\*)
-  (#778)
```
tgeoghegan, line 196: This deletion isn't right, but I suppose merge conflicts will need to be cleaned up eventually.

draft-ietf-ppm-dap.md
```diff
@@ -4832,8 +4883,11 @@ Each of these requires registration of a codepoint or other value; see
 * When a document defines a new batch mode, it MUST include a section titled
   "DAP Batch Mode Considerations" specifying the following:
 
-    * The value of the `config` field of `Query`, `PartialBatchSelector`, and
-      `BatchSelector`
+    * The value of the `config` field of `Query` and `BatchSelector`
+
+    * Any aggregation job extension defined for the batch mode
```
tgeoghegan: fussy nit:
```suggestion
    * The aggregation job extension, if any, defined for the batch mode
```

draft-ietf-ppm-dap.md
```diff
@@ -3967,20 +3989,22 @@ interaction (see {{aggregate-flow}}) need to be coordinated.
...
+being aggregated to the Helper. An aggregation job extension is defined for
+this purpose:
 
 ~~~ tls-presentation
-struct {
-  BatchID batch_id;
-} LeaderSelectedPartialBatchSelectorConfig;
+enum {
+  leader_selected_batch_id(1),
+  (65535)
```
tgeoghegan: For consistency with the other extension type enums:
```suggestion
  reserved(0),
  leader_selected_batch_id(1),
  (2^16-1)
```

We can also change the other definitions to say 65535. I don't care which so long as we're consistent.

draft-ietf-ppm-dap.md
```diff
@@ -3822,6 +3841,13 @@ Each batch mode specifies the following:
    can be incorporated into `TaskConfiguration` ({{task-configuration}})
    structures.
 
+It may be necessary for the Leader to convey additional information to the
+Helper during aggregation in order to map reports to batch buckets. This may be
```
tgeoghegan: super fussy: elsewhere we say "assign" instead of "map". But I think both are clear enough.

draft-ietf-ppm-dap.md
tgeoghegan: On re-reading, this duplicates the definition in `{{agg-job-extensions}}`.

draft-ietf-ppm-dap.md
```diff
@@ -2141,6 +2143,25 @@ both of which can violate privacy ({{replay-protection}}). Before committing to
...
+  reserved(0),
+  (65535)
+} AggregationJobExtensionType;
+~~~
+
+Each extension is identified by an `AggregationJobExtensionType` codepoint.
+Extension type values are defined in {{agg-job-extensions-registry}}.
+
+Extensions are mandatory to support. If the Helper does not recognize an
+extension type, it MUST abort with error `unsupportedExtension`.
```
tgeoghegan, line 2171: It'd be nice to harmonize this definition with what we do for task and collection job extensions. Which is to say: define `struct AggregationJobExtension` here and also assert that extensions must be in ascending order. But one way or the other we have all the normative text we need.

Orthogonally to that observation, this specific MUST also appears on L2430 and I think we should only say it once.

draft-ietf-ppm-dap.md
```diff
@@ -2141,6 +2143,25 @@ both of which can violate privacy ({{replay-protection}}). Before committing to
...
+
+~~~ tls-presentation
+enum {
+  reserved(0),
+  (65535)
+} AggregationJobExtensionType;
+~~~
+
+Each extension is identified by an `AggregationJobExtensionType` codepoint.
+Extension type values are defined in {{agg-job-extensions-registry}}.
```
martinthomson, line 2168: There's a requirement to sort extensions elsewhere.  Do you want to add that here as well?

Note that while task extensions currently depend on that sorting to ensure that there is a canonical encoding, the decision was to make report extensions sort as well, which have no such requirement.  So this should probably do that.  (Tim mentions the same.)

draft-ietf-ppm-dap.md
```diff
@@ -2288,16 +2309,14 @@ This message consists of:
...
   described in {{agg-param-validation}}.
 
-* `part_batch_selector`: The "partial batch selector" used by the Aggregators to
-  determine how to aggregate each report. Its contents depends on the indicated
-  batch mode. This field is called the "partial" batch selector because
-  depending on the batch mode, it may only partially determine a batch. See
-  {{batch-modes}}.
+* `extensions`: A list of aggregation job extensions ({{agg-job-extensions}})
+  providing additional parameters for the aggregation job. Extensions are
+  mandatory to support.
```
martinthomson: Is this the right ordering?

draft-ietf-ppm-dap.md
```diff
@@ -3967,20 +3989,22 @@ interaction (see {{aggregate-flow}}) need to be coordinated.
...
+During aggregation, the Leader needs to convey the batch ID of the reports
+being aggregated to the Helper. An aggregation job extension is defined for
+this purpose:
 
 ~~~ tls-presentation
-struct {
-  BatchID batch_id;
-} LeaderSelectedPartialBatchSelectorConfig;
+enum {
+  leader_selected_batch_id(1),
```
martinthomson, line 4008: You could add this to the definition of the aggregation job extensions.  Save the effort of reproducing the entire structure here.  Then just say "the leader_selected_batch_id aggregation job extension (codepoint: 1)" when introducing it.

draft-ietf-ppm-dap.md
```diff
@@ -4708,6 +4733,31 @@ The initial contents of this registry are listed in {{task-extension-id}}.
...
+: Where the extension is defined
+
+The initial contents of this registry are listed in {{agg-job-extension-id}}.
+
+| Value    | Name                        | Reference                                             |
+|:---------|:----------------------------|:------------------------------------------------------|
+| `0x0000` | `reserved`                  | {{agg-job-extensions-registry}} of RFC XXXX           |
+| `0x0001` | `leader_selected_batch_id`  | {{leader-selected-batch-id-extension}} of RFC XXXX    |
+{: #agg-job-extension-id title="Initial contents of the DAP Aggregation Job Extension
+Identifiers registry."}
```
```suggestion
Identifiers registry"}
```

draft-ietf-ppm-dap.md
martinthomson: I think that you want:


```suggestion
    * Aggregation job extensions, if any, needed for batch mode operation
```

martinthomson (approved):
Tim caught most of the problems.

draft-ietf-ppm-dap.md
cjpatton: Good catch, I copied the text from the task extensions section.

draft-ietf-ppm-dap.md
cjpatton: Copied the text on extension ordering.

Repeated MUST: We intend the various extension types (collection jobs, aggregation jobs, reports, and tasks) to have the same syntax and core semantics, but for now we're just replicating the relevant text for each. We can refactor to unify this in one section, but I think let's do that in a separate PR.

draft-ietf-ppm-dap.md
cjpatton: Not sure what you're asking. It's the same order as the fields in the struct. Do you want to reorder the fields in the struct?

draft-ietf-ppm-dap.md
```diff
@@ -2263,14 +2284,14 @@ for the relevant reports.
 
 ~~~ tls-presentation
 struct {
-  BatchMode batch_mode;
-  opaque config<0..2^16-1>;
-} PartialBatchSelector;
+  AggregationJobExtensionType extension_type;
+  opaque extension_data<0..2^16-1>;
+} AggregationJobExtension;
```
cjpatton: Noticing this should be moved to {{agg-job-extensions}} for consistency with other extensions.

draft-ietf-ppm-dap.md
cjpatton: I intended to remove all redundant (as in included in {{agg-job-extensions}}) but missed this sentence

```suggestion
  providing additional parameters for the aggregation job.
```

draft-ietf-ppm-dap.md
cjpatton: I weakly prefer defining it here so that the section is self-contained. If you prefer to move it I'm happy to oblige.

draft-ietf-ppm-dap.md
cjpatton: We should still change 65535 to 2^16-1.

draft-ietf-ppm-dap.md

draft-ietf-ppm-dap.md

draft-ietf-ppm-dap.md

draft-ietf-ppm-dap.md
cjpatton: I'm not sure what you mean. At least for TLS, it's common for an extension to repeat the "enum" struct with the new variant. This seems to be always how it's done.

draft-ietf-ppm-dap.md
tgeoghegan: This deletion is still wrong. I think maybe you need a rebase?

draft-ietf-ppm-dap.md
tgeoghegan: Ah, OK, I think I got confused between the section of the text above that defines agg job extensions, and this section which defines the specific batch ID extension. This is fine, except that I think it should be `(2^16-1)` instead of `(65535)` for consistency with `enum CollectionJobExtensionType` in `{#collect-ext}` and `enum TaskExtensionType` in `{#task-extensions}`.

draft-ietf-ppm-dap.md
cjpatton: Restored.

draft-ietf-ppm-dap.md
cjpatton: Got it. Done, here and for report extensions for consistency.

tgeoghegan (approved):

PR #781, opened 2026-04-22
Move Resource ID generation to server-side

### LINKED ISSUE #760: Collection job HTTP interactions
I know this is late, but it's the first I've really looked at this.

Collection jobs are created in response to a PUT request to a URL that follows a particular template.  That's sort of OK, but I think that it creates some weird dynamics.  I'm going through the process of working out how to define a new batch mode for attribution and the current structure is not great.  Let's walk through the issues.

What I'm trying to do is have a collection job created that uses a *Collector*-defined set of reports.  The existing batch modes either use properties of reports to assign them into batches or the Leader chooses, which allows the submission of reports to be effectively an undifferentiated stream arriving at the Leader.  For a Collector-selected batch, when that selection might use properties that DAP doesn't convey to choose reports, there needs to be something extra in the submission process to match reports to batches.

The simplest method I could come up with was to have the Collector create a collection job and to use the collection job ID (or similar) to identify the collection job for reports that are submitted.  The net effect of that is that this is always going to be asynchronous, but that's probably already true unless you are doing a running tally (which, to be fair, is very doable in a lot of cases).

The Collector need to pick the collection job ID.  This is counter to typical URL control arrangements, where the server is preferentially given control over its URL space.  DAP seems not to follow this pattern, which is defensible generally, but I don't think it is here.  Where task IDs generally need to be outside of any actor's direct control (taskprov, etc...), collection job IDs are only used in exactly one place.

Here's the natural request I'd expect:

```http
POST {leader}/task/{task_id}/collection <-- "collection_jobs" is just 5 extra characters, but whatever
Authorization: ...

CollectionJobReq details
```

That would typically result in either an answer:

```http
200 
Content-Location:  {leader}/task/{task_id}/collection/{collection_job_id} <-- optional way to re-fetch the result

CollectionJobResp with all the goodies
```

Or a deferral for asynchronous processing:

```
201
Location:  {leader}/task/{task_id}/collection/{collection_job_id}  <-- server chooses
```

All the other operations for async handling would apply.

The reason I mention this is that it would be idea if the asynchronous deferral could also identify a new URL where reports could be uploaded for that batch.  That's a new type of message content (a status update message), but one that could be seamlessly integrated into the current flow, which currently has an empty message for incomplete collection jobs.  They don't need to be empty.  It might mean that my new batch mode would need to define a media type for that.

The request for DAP would be to move to POST rather than PUT and have the server do the resource creation.  None of that is necessary, but I think it would make it better overall.

jcjones:
Intends to resolve #760 by changing DAP's model such that servers now choose resource IDs and provide them to clients via the `Location` header:

- Replace `PUT {resource-url}` with `POST {creation-url}` for creating aggregation jobs, collection jobs, and aggregate shares. The server assigns the ID and returns the resource location in `Location`.
- Add a new `Resource Creation` covering the shared pattern, including an explicit idempotency requirement
    - I'm considering a truncated HMAC of the request body keyed with a server secret -- hence "content-derived resource IDs" -- for Janus, but that's outside the spec's scope
- Define per-interaction identity criteria for idempotency (`AggregationJobInitReq` byte-equality; `CollectionJobReq` by `query` / `agg_param` / `extensions`; `AggregateShareReq` by `batch_selector`).
- Update  HTTP examples and add a "creation URL" alongside the "resource URL".

### DIFF
--- a/draft-ietf-ppm-dap.md
+++ b/draft-ietf-ppm-dap.md
@@ -200,6 +200,9 @@ aggregator.
   defines an aggregation job extension. For example, the leader-selected batch
   mode defines an extension that conveys the batch ID. (\*) (#762)
 
+- Move resource creation from PUT to POST with server-selected identifiers for
+  aggregation jobs, collection jobs, and aggregate shares. (\*) (#781)
+
 17:
 
 - Bump version tag from "dap-16" to "dap-17". (\*)
@@ -1112,8 +1115,8 @@ by immediately sending an empty response body with a successful status code
 The HTTP client then polls the state of the resource by sending GET requests to
 the resource URL. In some interactions, the resource's location will be
 indicated by a Location header in the HTTP server's response ({{!RFC9110,
-Section 10.2.2}}). Otherwise the resource URL is the URL to which the HTTP
-client initially sent its request.
+Section 10.2.2}}); see {{resource-creation}}. Otherwise the resource URL is the
+URL to which the HTTP client initially sent its request.
 
 The HTTP client SHOULD use each response's Retry-After header field to decide
 when to fetch the resource. The HTTP server responds the same way as it did to
@@ -1131,6 +1134,30 @@ Implementations are not required to support GET on resources if they are served
 synchronously, but they could do so, as a way for other protocol participants to
 retrieve the results of some transaction later on. The retention period for
 job results is an implementation detail.
+
+## Resource Creation {#resource-creation}
+
+Several DAP interactions involve creating new resources on an HTTP server:
+aggregation jobs ({{agg-init}}), collection jobs ({{collect-init}}), and
+aggregate shares ({{collect-aggregate}}). In each case, the HTTP client sends a
+POST request to a creation URL and the HTTP server assigns the new resource a
+unique identifier.
+
+The server responds with a successful status code and a Location header field
+({{!RFC9110, Section 10.2.2}}) indicating the location of the newly created
+resource. If the server handles the request synchronously, the response also
+includes the resource's representation in the body. If the server defers
+handling, the response body is empty and the HTTP client polls the resource at
+the location indicated by the Location header, as described in {{http-usage}}.
+
+Resource creation MUST be idempotent: if the server receives a POST request
+identical to one that created an existing resource (as defined by the interaction-
+specific criteria below), it MUST indicate the existing resource's location
+in the Location header rather than creating a duplicate. This allows HTTP clients
+to safely retry requests without risking side effects.
+
+One way to achieve this is to derive the resource identifier deterministically
+from the request content.
 
 ## HTTP Status Codes
 
@@ -2049,14 +2076,13 @@ refinement process. We instead think of these as properties of the output shares
 themselves: if verification succeeds, then the resulting output shares are
 guaranteed to combine into a valid, refined measurement.
 
-Aggregation jobs are identified by 16-byte job ID, chosen by the Leader:
+Aggregation jobs are identified by a server-selected identifier, unique within the
+scope of the task, assigned during resource creation ({{resource-creation}}).
 
-~~~ tls-presentation
-opaque AggregationJobID[16];
-~~~
-
-An aggregation job is an HTTP resource served by the Helper at the
-URL `{helper}/tasks/{task-id}/aggregation_jobs/{aggregation-job-id}`. VDAF
+Aggregation jobs are created by POSTing to the creation URL served by the Helper at
+`{helper}/tasks/{task-id}/aggregation_jobs`. An existing aggregation job is an
+HTTP resource served by the Helper at the URL
+`{helper}/tasks/{task-id}/aggregation_jobs/{aggregation-job-id}`. VDAF
 verification is mapped onto an aggregation job as illustrated in {{agg-flow}}.
 The first request from the Leader to the Helper includes the aggregation
 parameter, the Helper's report share for each report in the job, and for each
@@ -2223,7 +2249,7 @@ Report --> Continued -----+----> Continued
 initialization. (*) indicates a terminal state." }
 
 The Leader begins an aggregation job by choosing a set of candidate reports that
-belong to the same task and a job ID which MUST be unique within the task.
+belong to the same task.
 
 First, the Leader MUST ensure each report in the candidate set can be committed
 per the criteria detailed in {{batch-buckets}}. If a report cannot be committed,
@@ -2333,10 +2359,17 @@ This message consists of:
 > a different analysis would be required
 > to enable selecting a verification key after a task has started.
 
-The Leader sends the `AggregationJobInitReq` in the body of a PUT request to the
-aggregation job with a media type of
-"application/ppm-dap;message=aggregation-job-init-req". The Leader handles the
-response(s) as described in {{http-usage}} to obtain an `AggregationJobResp`.
+The Leader sends the `AggregationJobInitReq` in the body of a POST request to
+`{helper}/tasks/{task-id}/aggregation_jobs` with a media type of
+"application/ppm-dap;message=aggregation-job-init-req". The Helper creates the
+aggregation job resource and indicates its location in the Location header, as
+described in {{resource-creation}}. The Leader handles the response(s) as
+described in {{http-usage}} to obtain an `AggregationJobResp`.
+
+Two `AggregationJobInitReq` messages are considered identical if they are
+byte-for-byte identical when serialized. The Helper uses this to provide the
+idempotency guarantee described in {{resource-creation}}. A Leader that retries a
+request MUST replay the identical serialized body.
 
 The `AggregationJobResp.verify_resps` field must include exactly the same
 report IDs in the same order as the Leader's `AggregationJobInitReq`. Otherwise,
@@ -2414,12 +2447,12 @@ the Leader does. If successful, it includes the result in its response for the
 Leader to use to continue verifying the report.
 
 The initialization request can be handled either asynchronously or synchronously
-as described in {{http-usage}}. When indicating that the job is not yet
-ready, the response MUST include a Location header field ({{!RFC9110, Section
-10.2.2}}) set to the relative reference
-`/tasks/{task-id}/aggregation_jobs/{aggregation-job-id}?step=0`. Subsequent GET
-requests to the aggregation job MUST include the `step` query parameter so that
-the Helper can figure out which step of preparation the Leader is on (see
+as described in {{http-usage}}. The response MUST include a Location header
+field ({{!RFC9110, Section 10.2.2}}) indicating the location of the aggregation
+job resource. When indicating that the job is not yet ready, the Location header
+MUST include the query parameter `step=0`. Subsequent GET requests to the
+aggregation job MUST include the `step` query parameter so that the Helper can
+figure out which step of preparation the Leader is on (see
 {{aggregation-step-skew-recovery}}). When the job is ready, the Helper responds
 with the `AggregationJobResp` (defined below).
 
@@ -2540,13 +2573,14 @@ in the previous step. The order MUST match
 `AggregationJobInitReq.verify_inits`. The media type for `AggregationJobResp`
 is "application/ppm-dap;message=aggregation-job-resp".
 
-The Helper may receive multiple copies of a given initialization request. The
-Helper MUST verify that subsequent requests have the same
-`AggregationJobInitReq` value and abort with a client error if they do not. It
-is illegal to rewind or reset the state of an aggregation job. If the Helper
-receives requests to initialize an aggregation job once it has been continued at
+If the Helper receives a POST whose content matches an existing aggregation
+job (per the identity criteria in {{leader-init}}), it MUST return the existing
+job's location and current state as described in {{resource-creation}}. It is
+illegal to rewind or reset the state of an aggregation job. If the Helper
+receives a POST that matches an aggregation job which has been continued at
 least once, confirming that the Leader received the Helper's response (see
-{{agg-continue-flow}}), it MUST abort with a client error.
+{{agg-continue-flow}}), it MUST return the existing job's location and current
+state rather than reinitializing the job.
 
 #### Input Share Decryption {#input-share-decryption}
 
@@ -2628,8 +2662,8 @@ If all of the above checks succeed, the input share is valid.
 The Helper handles the aggregation job initialization synchronously:
 
 ~~~ http
-PUT /helper/tasks/8BY0RzZMzxvA46_8ymhzycOB9krN-QIGYvg_RsByGec/\
-  aggregation_jobs/lc7aUeGpdSNosNlh-UZhKA
+POST /helper/tasks/8BY0RzZMzxvA46_8ymhzycOB9krN-QIGYvg_RsByGec/\
+  aggregation_jobs
 Host: example.com
 Content-Type: application/ppm-dap;message=aggregation-job-init-req
 Content-Length: 100
@@ -2647,6 +2681,8 @@ encoded(struct {
 } AggregationJobInitReq)
 
 HTTP/1.1 200
+Location: /helper/tasks/8BY0RzZMzxvA46_8ymhzycOB9krN-QIGYvg_RsByGec/\
+  aggregation_jobs/lc7aUeGpdSNosNlh-UZhKA
 Content-Type: application/ppm-dap;message=aggregation-job-resp
 Content-Length: 100
 
@@ -2656,8 +2692,8 @@ encoded(struct { verify_resps } AggregationJobResp)
 Or asynchronously:
 
 ~~~ http
-PUT /helper/tasks/8BY0RzZMzxvA46_8ymhzycOB9krN-QIGYvg_RsByGec/\
-  aggregation_jobs/lc7aUeGpdSNosNlh-UZhKA
+POST /helper/tasks/8BY0RzZMzxvA46_8ymhzycOB9krN-QIGYvg_RsByGec/\
+  aggregation_jobs
 Host: example.com
 Content-Type: application/ppm-dap;message=aggregation-job-init-req
 Content-Length: 100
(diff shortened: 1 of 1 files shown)


draft-ietf-ppm-dap.md
```diff
@@ -1124,6 +1124,33 @@ synchronously, but they could do so, as a way for other protocol participants to
...
 
+## Resource Creation {#resource-creation}
+
+Several DAP interactions involve creating new resources on an HTTP server:
+aggregation jobs ({{agg-init}}), collection jobs ({{collect-init}}), and
+aggregate shares ({{collect-aggregate}}). In each case, the HTTP client sends a
+POST request to a creation URL and the HTTP server assigns the new resource a
+unique identifier.
+
+The server responds with a successful status and a Location header field
```
```suggestion
The server responds with a successful status code and a Location header field
```

draft-ietf-ppm-dap.md
```diff
@@ -1124,6 +1124,33 @@ synchronously, but they could do so, as a way for other protocol participants to
...
+({{!RFC9110, Section 10.2.2}}) indicating the location of the newly created
+resource. If the server handles the request synchronously, the response also
+includes the resource's representation in the body. If the server defers
+handling, the response body is empty and the HTTP client polls the resource at
+the location indicated by the Location header, as described in {{http-usage}}.
+
+Resource creation MUST be idempotent: if the server receives a POST request
+whose content matches that of an existing resource (as defined by the specific
```
martinthomson: when you say "whose content matches that of an existing resource", I think that you need to be a little more precise.  Resources don't have content in HTTP (unless they do).

Instead, you probably want to say that if the content of a POST is identical (or functionally equivalent to...?  maybe we don't need to recognize that given the way the encodings here tend to be canonical always) the content that was used to create a resource, the same resource should be returned.

Edit: I see what you did below.  Use the word "identical" if that is what you mean.  It looks like identical is possible; I'd rather force things to be identical, rather than have some more fuzzy notion.

draft-ietf-ppm-dap.md
```diff
@@ -1124,6 +1124,33 @@ synchronously, but they could do so, as a way for other protocol participants to
...
+interaction), it MUST indicate the existing resource's location in the Location
+header rather than creating a duplicate. This allows HTTP clients to safely
+retry requests without risking side effects.
+
+One way to achieve this is to derive the resource identifier deterministically
+from the request content. For example, by computing a truncated HMAC of the
+request body keyed with a server secret. This ensures that identical requests
```
martinthomson: I want to briefly consider whether the server secret is a necessary part of this.  Are there any cases where it would be outright harmful if the client were able to predict the identity of a resource?  I can't think of any.  In that case, you probably want to drop mention of keyed HMAC as a deterministic function.  Otherwise, you are talking about a PRF!

draft-ietf-ppm-dap.md
```diff
@@ -2041,14 +2068,17 @@ refinement process. We instead think of these as properties of the output shares
 themselves: if verification succeeds, then the resulting output shares are
 guaranteed to combine into a valid, refined measurement.
 
-Aggregation jobs are identified by 16-byte job ID, chosen by the Leader:
+Aggregation jobs are identified by a 16-byte job ID, chosen by the Helper
```
martinthomson: I get that you are making this an incremental change, so feel free to defer this one...

In this case, I don't think that we are using the aggregation job ID anywhere in the protocol.  So this doesn't need to take a specific form and can instead be "unique, server-selected identifier ... `opaque AggregationJobID[???]`".

draft-ietf-ppm-dap.md
```diff
@@ -3596,9 +3642,11 @@ commit any more output shares to the batch. It is an error for the Leader to
 issue any more aggregation jobs for additional reports that satisfy the query.
 These reports MUST be rejected by the Helper as described in {{batch-buckets}}.
 
-Changing an aggregate share's parameters is illegal, so if there are further PUT
-requests to the aggregate share with a different `AggregateShareReq`, the Helper
-MUST abort with error `invalidMessage`.
+Two `AggregateShareReq` messages are considered identical if they have the same
+`batch_selector`. (Within a task, a given batch can only be collected once, so
```
```suggestion
`batch_selector`. Note that, within a task, a given batch can only be collected once, so
```

draft-ietf-ppm-dap.md
```diff
@@ -3596,9 +3642,11 @@ commit any more output shares to the batch. It is an error for the Leader to
...
+the batch selector is a natural key.) If the Helper receives a POST whose
```
```suggestion
the batch selector is a natural key. If the Helper receives a POST whose
```

martinthomson (commented):
LGTM

draft-ietf-ppm-dap.md
jcjones: Fair. Seems like if we change this, we should do the same to `AggregateShareID` and  `CollectionJobID` -- so I'm treating them equally in 0ce3558

draft-ietf-ppm-dap.md
jcjones: OK, I mean identical, I say identical: 6eb6bda

draft-ietf-ppm-dap.md
jcjones: I don't have a reason to suggest any thing in particular here. I'll just end this after the first sentence. 03e522c

draft-ietf-ppm-dap.md
```diff
@@ -2041,14 +2065,13 @@ refinement process. We instead think of these as properties of the output shares
 themselves: if verification succeeds, then the resulting output shares are
 guaranteed to combine into a valid, refined measurement.
 
-Aggregation jobs are identified by 16-byte job ID, chosen by the Leader:
+Aggregation jobs are identified by a unique, server-selected identifier
```
tgeoghegan: "Unique" is stronger than necessary, and would force implementations to use extra wide job IDs. "unique within the scope of the task" should do.

draft-ietf-ppm-dap.md
```diff
@@ -2305,10 +2328,17 @@ This message consists of:
...
+`{helper}/tasks/{task-id}/aggregation_jobs` with a media type of
+"application/ppm-dap;message=aggregation-job-init-req". The Helper creates the
+aggregation job resource and indicates its location in the Location header, as
+described in {{resource-creation}}. The Leader handles the response(s) as
+described in {{http-usage}} to obtain an `AggregationJobResp`.
+
+Two `AggregationJobInitReq` messages are considered identical if they are
+byte-for-byte identical when serialized. The Helper uses this to provide the
+idempotency guarantee described in {{resource-creation}}. A Leader that retries a
+request after a crash or timeout MUST replay the identical serialized body.
```
tgeoghegan: The Leader should replay the body if it retries for any reason, but this MUST suggests it only applies to crashes and timeouts. Sorry to get snotty but MUST sentences are to be read very literally, I think.

draft-ietf-ppm-dap.md
```diff
@@ -3110,23 +3143,20 @@ to as a "collection job" and is composed of two interactions:
 Once complete, the Collector computes the final aggregate result as specified in
 {{collect-finalization}}.
 
-Collection jobs are identified by a 16-byte job ID, chosen by the Collector:
-
-~~~ tls-presentation
-opaque CollectionJobID[16];
-~~~
+Collection jobs are identified by a unique, server-selected identifier
```
tgeoghegan: As above, let's say "unique in the scope of the task"

draft-ietf-ppm-dap.md
```diff
@@ -3475,18 +3511,13 @@ The Leader then combines the values inside the batch bucket as follows:
 * Report counts are combined via summing.
 * Checksums are combined via bitwise XOR.
 
-A Helper aggregate share is identified by a 16-byte ID:
-
-~~~ tls-presentation
-opaque AggregateShareID[16];
-~~~
+A Helper aggregate share is identified by a unique, server-selected identifier
```
tgeoghegan: As above, "unique within the scope of the task"

draft-ietf-ppm-dap.md
jcjones: Absolutely. So, `"A Leader that retries a request MUST replay the identical serialized body."` maybe?

jcjones: Updated alongside the `unique` updates in dff73b5

tgeoghegan (approved):

draft-ietf-ppm-dap.md
tgeoghegan: The examples were arguably helpful, if we can find a phrasing that doesn't run afoul of the MUST. But the current phrasing is good too.

draft-ietf-ppm-dap.md
```diff
@@ -3273,17 +3304,18 @@ Once the `Leader` has constructed a `CollectionJobResp` for the Collector, the
 Leader considers the batch to be collected, and further aggregation jobs MUST
 NOT commit more reports to the batch (see {{batch-buckets}}).
 
-Changing a collection job's parameters is illegal, so if there are further PUT
-requests to the collection job with a different `CollectionJobReq`, the Leader
-MUST abort with error `invalidMessage`.
+Two `CollectionJobReq` messages are considered identical if they have the same
+`query`, `agg_param`, and `extensions`. If the Leader receives a POST whose
+content matches an existing collection job, it MUST return the existing job's
+location and current state as described in {{resource-creation}}.
```
cjpatton: Consistency with aggregation jobs
```suggestion
Two `CollectionJobReq` messages are considered identical if their encodings are byte-for-byte identical. If the Leader receives a POST whose
content matches an existing collection job, it MUST return the existing job's
location and current state as described in {{resource-creation}}.
```

draft-ietf-ppm-dap.md
```diff
@@ -3596,9 +3627,11 @@ commit any more output shares to the batch. It is an error for the Leader to
 issue any more aggregation jobs for additional reports that satisfy the query.
 These reports MUST be rejected by the Helper as described in {{batch-buckets}}.
 
-Changing an aggregate share's parameters is illegal, so if there are further PUT
-requests to the aggregate share with a different `AggregateShareReq`, the Helper
-MUST abort with error `invalidMessage`.
+Two `AggregateShareReq` messages are considered identical if they have the same
+`batch_selector`. Note that, within a task, a given batch can only be collected once, so
```
cjpatton: Likewise here: the bytes of the DAP message should be identical correct?

cjpatton (approved):

draft-ietf-ppm-dap.md
jcjones: Yes... and that is easier to specify. The batch_selector natural-key argument is useful context. These encodings are canonical, it's functionally equivalent, I think. But I like that `Note that` which Martin contributed. See what you think of this update in 248e078.

tgeoghegan:
@jcjones I think this is good to go once @cjpatton signs off, but it should also add an entry to the changelog at the top of the document since this is a breaking change.

cjpatton (approved):
Changes look good to me. Would you mind squashing the stack of commits into one before merging? I'd prefer to not have merge commits on main.

draft-ietf-ppm-dap.md
```diff
@@ -2006,7 +2028,7 @@ struct {
 
 enum {
   reserved(0),
-  (2^16-1)
+  (65535)
```
jcjones: uh, did I do this? This looks like a merge problem.

draft-ietf-ppm-dap.md
```diff
@@ -2295,10 +2288,15 @@ Once all the report shares have been initialized, the Leader creates an
 for the relevant reports.
 
 ~~~ tls-presentation
+struct {
+  BatchMode batch_mode;
+  opaque config<0..2^16-1>;
+} PartialBatchSelector;
+
```
jcjones: ok, give me a little bit. I mudged the `git rebase --strategy` and got distracted

jcjones:
The rebase went sideways, but here's the diff between what we approved (248e078) and the final version here in (d776008), which seems to match up with https://github.com/ietf-wg-ppm/draft-ietf-ppm-dap/pull/779:

<details><summary>diff</summary>
<p>

```diff
diff draft-ietf-ppm-dap.md draft-ietf-ppm-dap.md-248e078
198,205d197
< - Replace the partial batch selector with aggregation job extensions. If a
<   batch mode needs to convey information to the Helper during aggregation, it
<   defines an aggregation job extension. For example, the leader-selected batch
<   mode defines an extension that conveys the batch ID. (\*) (#762)
< 
< - Move resource creation from PUT to POST with server-selected identifiers for
<   aggregation jobs, collection jobs, and aggregate shares. (\*) (#781)
< 
2036c2028
<   (2^16-1)
---
>   (65535)
2175,2202d2166
< ### Aggregation Job Extensions {#agg-job-extensions}
< 
< The Leader may include extensions in the `AggregationJobInitReq` to convey
< additional parameters for the aggregation job to the Helper. Extensions use
< the following types:
< 
< ~~~ tls-presentation
< struct {
<   AggregationJobExtensionType extension_type;
<   opaque extension_data<0..2^16-1>;
< } AggregationJobExtension;
< 
< enum {
<   reserved(0),
<   (2^16-1)
< } AggregationJobExtensionType;
< ~~~
< 
< Each extension is identified by an `AggregationJobExtensionType` codepoint.
< Extension type values are defined in {{agg-job-extensions-registry}}.
< 
< Extensions are mandatory to support. If the Helper does not recognize an
< extension type, it MUST abort with error `unsupportedExtension`.
< 
< Extensions MUST be encoded in strictly increasing order of `extension_type`.
< That is, each `extension_type` value must be greater than that of the extension
< that precedes it.
< 
2324a2289,2293
>   BatchMode batch_mode;
>   opaque config<0..2^16-1>;
> } PartialBatchSelector;
> 
> struct {
2327c2296
<   AggregationJobExtension extensions<0..2^16-1>;
---
>   PartialBatchSelector part_batch_selector;
2345,2346c2314,2318
< * `extensions`: A list of aggregation job extensions ({{agg-job-extensions}})
<   providing additional parameters for the aggregation job.
---
> * `part_batch_selector`: The "partial batch selector" used by the Aggregators to
>   determine how to aggregate each report. Its contents depends on the indicated
>   batch mode. This field is called the "partial" batch selector because
>   depending on the batch mode, it may only partially determine a batch. See
>   {{batch-modes}}.
2351c2323
<   encoded `agg_param` and `extensions` fields. That is, the remainder
---
>   encoded `agg_param` and `part_batch_selector` fields. That is, the remainder
2468,2472c2440,2442
< * Whether the extensions in `AggregationJobInitReq.extensions` are valid:
<   - If any extension type is unrecognized, the Helper MUST fail the job with
<     error `unsupportedExtension`.
<   - If the extensions are not encoded in strictly increasing order of
<     `extension_type`, the Helper MUST fail the job with error `invalidMessage`.
---
> * Whether the batch mode indicated by `part_batch_selector.batch_mode` matches
>   the task's batch mode. If not, then the Helper MUST fail the job with error
>   `invalidMessage`.
2674,2679c2644,2649
<   extensions = [
<     struct {
<       extension_type = AggregationJobExtensionType.leader_selected_batch_id,
<       extension_data = encoded([0x1f, 0x1e, ..., 0x00] BatchID),
<     } AggregationJobExtension,
<   ],
---
>   part_batch_selector = struct {
>     batch_mode = BatchMode.leader_selected,
>     config = encoded(struct {
>       batch_id = [0x1f, 0x1e, ..., 0x00],
>     } LeaderSelectedPartialBatchSelectorConfig),
>   } PartialBatchSelector,
2704c2674,2677
<   extensions = [],
---
>   part_batch_selector = struct {
>     batch_mode = BatchMode.time_interval,
>     config = encoded(Empty),
>   },
2983c2956
< Batch buckets are indexed by a "batch bucket identifier" as specified by the
---
> Batch buckets are indexed by a "batch bucket identifier" as as specified by the
2986c2959
< * For the time-interval batch mode ({{time-interval-batch-mode}}), the batch
---
> * For the time-interval batch mode ({{time-interval-batch-mode}}, the batch
2990,2992c2963
<   batch bucket identifier is the batch ID carried in the
<   `leader_selected_batch_id` aggregation job extension
<   ({{leader-selected-batch-id-extension}}).
---
>   batch bucket identifier is the batch ID and indicated in the aggregation job.
3881c3852,3853
< 1. The value of the `config` field of `Query` and `BatchSelector`
---
> 1. The value of the `config` field of `Query`, `PartialBatchSelector`, and
>    `BatchSelector`
3892,3898d3863
< It may be necessary for the Leader to convey additional information to the
< Helper during aggregation in order to assign reports to batch buckets. This may
< be done with the aggregation job extension mechanism (see
< {{agg-job-extensions}}). A batch mode that defines such an extension should
< specify how it is used to assign reports to batch buckets. See the
< leader-selected mode ({{leader-selected-batch-mode}}) for an example.
< 
3947a3913,3916
> ### Partial Batch Selector Configuration
> 
> The payload of `PartialBatchSelector.config` is empty.
> 
4040c4009
< ### Aggregation Job Extension {#leader-selected-batch-id-extension}
---
> ### Partial Batch Selector Configuration
4042,4044c4011
< During aggregation, the Leader needs to convey the batch ID of the reports
< being aggregated to the Helper. An aggregation job extension is defined for
< this purpose:
---
> The payload of `PartialBatchSelector.config` is:
4047,4050c4014,4016
< enum {
<   leader_selected_batch_id(1),
<   (2^16-1)
< } AggregationJobExtensionType;
---
> struct {
>   BatchID batch_id;
> } LeaderSelectedPartialBatchSelectorConfig;
4053,4055c4019,4022
< The payload of this extension is a `BatchID` selected by the Leader. The
< Helper MUST abort with error `invalidMessage` if this extension is absent or
< the payload is not a valid batch ID.
---
> where `batch_id` is the batch ID selected by the Leader.
> 
> It is impossible for a `BatchSelector.config` to be inconsistent
> with `Query.config` for this batch mode.
4072,4073c4039
< report is assigned to the batch indicated by the `leader_selected_batch_id`
< aggregation job extension ({{leader-selected-batch-id-extension}}) during
---
> report is assigned to the batch indicated by the `PartialBatchSelector` during
4784,4808d4749
< ### Aggregation Job Extensions Registry {#agg-job-extensions-registry}
< 
< A new registry will be (RFC EDITOR: change "will be" to "has been") created
< called "DAP Aggregation Job Extension Identifiers" for extensions included in
< aggregation job initialization requests ({{agg-job-extensions}}). This registry
< should contain the following columns:
< 
< Value:
< : The two-byte identifier for the aggregation job extension
< 
< Name:
< : The name of the aggregation job extension
< 
< Reference:
< : Where the extension is defined
< 
< The initial contents of this registry are listed in {{agg-job-extension-id}}.
< 
< | Value    | Name                        | Reference                                             |
< |:---------|:----------------------------|:------------------------------------------------------|
< | `0x0000` | `reserved`                  | {{agg-job-extensions-registry}} of RFC XXXX           |
< | `0x0001` | `leader_selected_batch_id`  | {{leader-selected-batch-id-extension}} of RFC XXXX    |
< {: #agg-job-extension-id title="Initial contents of the DAP Aggregation Job Extension
< Identifiers registry"}
< 
4923d4863
< 1. a new aggregation job extension ({{agg-job-extensions}})
4934,4938c4874,4875
<     * The value of the `config` field of `Query` and `BatchSelector`
< 
<     * Aggregation job extensions, if any, needed for batch mode operation
<       ({{agg-job-extensions}}), including how the extension is used to assign
<       reports to batch buckets.
---
>     * The value of the `config` field of `Query`, `PartialBatchSelector`, and
>       `BatchSelector`
```

</p>
</details> 

The real important bit is the PR looks right again.

tgeoghegan:
I rechecked the diff and it looks cromulent, so I'm merging!

PR #784, opened 2026-05-15
Clarify aggregation verification key ID handling

StanFromIreland:
I found it quite confusing and contradictory, so I took a stab at trying to correct a few things :-)
Additionally, I think that the Helper should reject an unknown `verification_key_id`, I've included that here, but if controversial I can revert it.

### DIFF
--- a/draft-ietf-ppm-dap.md
+++ b/draft-ietf-ppm-dap.md
@@ -112,6 +112,11 @@ contributor:
        email: mt@mozilla.com
 
  -
+       name: Stan Ulbrych
+       org: Independent
+       email: stanulbrych@gmail.com
+
+ -
        name: Shan Wang
        org: Apple
        email: shan_wang@apple.com
@@ -181,7 +186,7 @@ aggregator.
 18:
 
 - Add verification key ID to aggregation jobs to enable but not require
-  verification key management. (\*) (#766)
+  verification key management. (\*) (#766, #784)
 
 - Define collection job extensions. (\*) (#769)
 
@@ -1339,6 +1344,7 @@ enum {
   report_too_early(9),
   task_not_started(10),
   outdated_config(11),
+  unknown_verification_key_id(12),
   (255)
 } ReportError;
 ~~~
@@ -2279,7 +2285,8 @@ state = Vdaf.ping_pong_leader_init(
 
 where:
 
-* `vdaf_verify_key` is the VDAF verification key for the task
+* `vdaf_verify_key` is the VDAF verification key identified by
+  `verification_key_id`
 * `task_id` is the task ID
 * `agg_param` is the VDAF aggregation parameter provided by the Collector (see
   {{collect-flow}})
@@ -2340,7 +2347,8 @@ This message consists of:
   see {{Section 5.2 of VDAF}} and {{verification-key}}.
   This allows a Leader to nominate a verification key
   from a set of prearranged keys.
-  This might also allow for verification keys to be updated by Aggregators.
+  The Helper MUST use the verification key identified by this field when
+  initializing VDAF verification.
 
 * `agg_param`: The VDAF aggregation parameter chosen by the Collector. Before
   initializing an aggregation job, the Leader MUST validate the parameter as
@@ -2352,8 +2360,8 @@ This message consists of:
 * `verify_inits`: the sequence of `VerifyInit` messages constructed in the
   previous step. Here `verify_inits_length` is the length of the HTTP message
   content ({{!RFC9110, Section 6.4}}), minus the lengths in octets of the
-  encoded `agg_param` and `extensions` fields. That is, the remainder
-  of the HTTP message consists of `verify_inits`.
+  `verification_key_id` field and the encoded `agg_param` and `extensions`
+  fields. That is, the remainder of the HTTP message consists of `verify_inits`.
 
 {:aside}
 > IMPORTANT: this does not change the security requirements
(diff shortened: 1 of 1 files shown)


draft-ietf-ppm-dap.md
```diff
@@ -2469,6 +2476,10 @@ conditions:
 * Whether the `AggregationJobInitReq` is malformed. If so, the the Helper MUST
   fail the job with error `invalidMessage`.
 
+* Whether `AggregationJobInitReq.verification_key_id` identifies a VDAF
+  verification key configured for the task. If not, the Helper MUST fail the job
+  with error `invalidMessage`.
```
tgeoghegan: I'm not sure if we should use `invalidMessage` here. A message containing an unrecognized verification key ID is well-formed and parseable, just its contents are wrong. The precedent here is handling of unrecognized HPKE configurations, in which case aggregators are required to use the distinct error type `hpke_decrypt_error`. So I think we ought to define a new `ReportError` variant `unknown_verification_key_id` and use that here.

(in passing I notice that we define but don't use `ReportError.hpke_decrypt_error`. We should either ues it or delete it. But let's not do that in this PR: #785

draft-ietf-ppm-dap.md
```diff
@@ -4511,13 +4525,11 @@ verification key must be kept secret from Clients.
 
 Furthermore, for a given report, it may be possible to craft a verification key
 which leaks information about that report's measurement during verification.
-Therefore, the verification key for a task SHOULD be chosen before any reports
-are generated. To achieve this, the current design and analysis assume that
-the verification key is fixed for the lifetime of the task.
```
tgeoghegan, line 4516: I'd like @martinthomson and @cjpatton to comment here: when we added verification key ID, did that change our expectations around the key's arity?

tgeoghegan (commented):
Thank you for improving these nuanced lines of text. I think the changes here are good, but there's some details we need to hash out here.

draft-ietf-ppm-dap.md
```diff
@@ -1514,7 +1519,7 @@ Finally, the Collector is configured with the HPKE secret key corresponding to
 
 A task's parameters are immutable for the lifetime of that task. The only way to
 change parameters or to rotate secret values like Collector HPKE configuration
-or the VDAF verification key is to configure a new task.
+is to configure a new task.
```
cjpatton: Why this change?

draft-ietf-ppm-dap.md
cjpatton: I think a distinguished error would be useful for this case.

draft-ietf-ppm-dap.md
cjpatton: What do you mean by arity? Are you asking whether the same key can be used across tasks or whether multiple keys can be used within a task?

draft-ietf-ppm-dap.md
StanFromIreland: I understood `verification_key_id` as allowing multiple keys within a task, I'll revert it.

draft-ietf-ppm-dap.md
StanFromIreland: I added `unknown_verification_key_id`

draft-ietf-ppm-dap.md
tgeoghegan: Both! Are tasks : verification keys 1:1, 1:many, many:1 or many:many? Does the change to the text here change our assumptions at all?

draft-ietf-ppm-dap.md
```diff
@@ -2422,8 +2430,8 @@ The Leader proceeds as follows with each report:
 1. Else if the `VerifyResp` has type "reject", then the Leader rejects the
    report and removes it from the candidate set. The Leader MUST NOT include
    the report in a subsequent aggregation job, unless the report error is
-   `report_too_early`, in which case the Leader MAY include the report in a
-   subsequent aggregation job.
+   `report_too_early` or `     `, in which case the
```
tgeoghegan: Something missing in this blank space? Was this meant to be `unknown_verification_key_id`? If so, I think we might need more guidance here. Should the Leader retry after some out of band interaction to "activate" the new key ID, or retry with a known-valid key ID?

StanFromIreland:
```suggestion
   `report_too_early` or `unknown_verification_key_id`, in which case the
```
Ah oops, I must have cut.

StanFromIreland: Hmm, a retry makes sense, WDYT about:

> The Leader MAY retry the report after selecting a verification key ID known to the Helper

tgeoghegan: I think that's good enough, since the goal of the verification key ID stuff was not to spell out the whole lifecycle of the things, but just to add enough of an affordance in the protocol for something robust to be done by implementations or in later documents extending DAP.

draft-ietf-ppm-dap.md
cjpatton: I don't think we should change this right now. The point of the key ID is to allow the key to be rotated independently of the task lifetime, but as the text here suggests, this is only safe to do under specific conditions. In the absence of a concrete mechanism for managing key rotation, I think we should be conservative.

In other words, @StanFromIreland I suggest reverting this change.

cjpatton (approved):

tgeoghegan (approved):
This looks good to me. @StanFromIreland: our sincere thanks for improving the draft.

tgeoghegan:
I will fix the merge conflicts, rebase into a single commit, and merge to `main` later this morning.

StanFromIreland:
> I will fix the merge conflicts

Hmm, I'd be happy to fix them but I don't see any?

> rebase into a single commit

I suggest ["Squash merging"](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/incorporating-changes-from-a-pull-request/about-pull-request-merges#squash-and-merge-your-commits), it's far easier IMO, I use it for example in python/cpython.

tgeoghegan:
> Hmm, I'd be happy to fix them but I don't see any?

Just a minor issue with the error types table. I cleaned it up and GitHub is now happy. I also took the opportunity to add this PR # to the changelog, since the new error type is wire breaking and thus worth calling out.
 
> I suggest ["Squash merging"](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/incorporating-changes-from-a-pull-request/about-pull-request-merges#squash-and-merge-your-commits), it's far easier IMO, I use it for example in python/cpython.

I know, I do the same on the repos I control. We just never got around to doing it here.

PR #786, opened 2026-05-28
Audit `ReportError` variants and make sure they're used properly

### LINKED ISSUE #785: Audit `ReportError` variants and make sure they're used properly
We define:

```
enum {
  reserved(0),
  batch_collected(1),
  report_replayed(2),
  report_dropped(3),
  hpke_unknown_config_id(4),
  hpke_decrypt_error(5),
  vdaf_verify_error(6),
  task_expired(7),
  invalid_message(8),
  report_too_early(9),
  task_not_started(10),
  outdated_config(11),
  (255)
} ReportError;
```

We should audit these error codes and make sure that they're used appropriately in the text.

For example, `hpke_unknown_config_id` is not currently mentioned anywhere but the enum definition and the error registry, but I think it would be appropriate for `{{input-share-decryption}}` to require using it when an aggregator receives an HPKE ciphertext encrypted to an unknown config ID.

tgeoghegan:
See individual commits for detailed discussion. The headline changes:

- Delete `ReportError::task_expired, task_not_started` and use `report_dropped` instead
- Delete `outdated_config` and use `hpke_unknown_config_id` instead

What's most likely to be controversional here is whether we still want specific `ReportError` values for when the task interval extension is in use and the report timestamp is outside of it.

My position is: no, `report_dropped` suffices. My argument is: if the Task Interval extension is configured, then the Client has to have agreed to its parameters because it gets included in the `TaskConfiguration` and then HPKE AADs. That means that before it ever uploads, the Client can assure itself that its report timestamp is inside the task's time interval, and thus we don't need a recovery mechanism that would nudge the Client into waiting until the task starts to upload.

Closes #785

### DIFF
--- a/draft-ietf-ppm-dap.md
+++ b/draft-ietf-ppm-dap.md
@@ -1343,7 +1343,6 @@ enum {
   invalid_message(8),
   report_too_early(9),
   task_not_started(10),
-  outdated_config(11),
   unknown_verification_key_id(12),
   (255)
 } ReportError;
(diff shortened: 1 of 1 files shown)


draft-ietf-ppm-dap.md
```diff
@@ -182,12 +182,15 @@ aggregator.
 
 (RFC EDITOR: Remove this section.)
 
+19:
 
-18:
+- Delete unused or redundant `ReportError` variants. (\*) (#786)
 
 - Add verification key ID to aggregation jobs to enable but not require
   verification key management. (\*) (#766, #784)
```
divergentdave: The changes represented by this item are partly before the draft-18 tag, and partly after.

divergentdave (approved):

draft-ietf-ppm-dap.md
tgeoghegan: Yep, I was getting too clever there.

cjpatton (approved):
Looks good. Consider compacting the codepoints once more.

tgeoghegan:
> Consider compacting the codepoints once more.

Do they need more compacting?

cjpatton:
> > Consider compacting the codepoints once more.
> 
> Do they need more compacting?

Nope! I was looking at the commits individually and didn't notice they're now in order :) Disregard

PR #789, opened 2026-06-26
Fix report extension validation error handling

### LINKED ISSUE #788: Error handling for bad reports mixes ReportError and problem documents
Section 4.4.2.2 discusses how the leader processes an incoming upload request. Since upload requests may contain multiple reports, errors related to just one report are transmitted back to the client via an `UploadErrors` response. Errors with the whole request are transmitted via a problem document response. However, certain parts of this section still discuss problem details responses and problem type URIs for some report validation steps.

The report extension validation paragraph mentions aborting with `unsupportedExtension` and providing additional information in an extension member of the problem document. I think we should be fine with not replicating the additional information in the `UploadErrors` response, it should be apparent to operators which report extensions their software supports, and thus easy to identify which extension caused the issue.

The next paragraph, covering duplicate report extensions, says to abort with `invalidMessage`, when it should just mark the report with `invalid_message`.

divergentdave:
This fixes #788 by adding an `unsupported_extensions variant to `ReportError` and making it clear that issues with report extensions are treated as per-report errors only.

### DIFF
--- a/draft-ietf-ppm-dap.md
+++ b/draft-ietf-ppm-dap.md
@@ -1347,6 +1347,7 @@ enum {
   invalid_message(7),
   report_too_early(8),
   unknown_verification_key_id(9),
+  unsupported_extension(10),
   (255)
 } ReportError;
 ~~~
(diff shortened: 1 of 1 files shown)


divergentdave:
I spotted one more old error type in the Operational Considerations section, and pushed a commit to fix that as well.

jcjones (approved):
Good catch! Thanks for the forethought!

tgeoghegan:
Let's take a beat to unpack this statement from the issue, which informs the part of this PR where the aggregators no longer indicate which extensions are unsupported:

> I think we should be fine with not replicating the additional information in the UploadErrors response, it should be apparent to operators which report extensions their software supports, and thus easy to identify which extension caused the issue.

I believe I agree with this, because now that `TaskConfiguration` gets mixed into the AAD, the client can't even upload reports unless it agrees with the aggregators about which extensions are in use and how they're configured. So it makes sense to me that we don't need an in-band mechanism for aggregators to indicate supported extensions: DAP assumes that this has been negotiated before the start of a task. Put another way, we are not interested in enabling a flow like this:

- client constructs reports with extensions a, b, c
- leader rejects the upload indicating extension b is unsupported
- client tries again with just a, c
- leader now accepts the report

Some other dynamic task provisioning protocol could decide to incorporate extension negotiation, but DAP is putting it out of scope. I think this is correct and sensible (it rules out extension downgrade attacks) but want to highlight this implication for other reviewers.

Otherwise I think it's sensible to narrow these errors to per-report instead of them causing the rejection of an entire `UploadReq` or `AggregationJobInitReq`. The nuance here is that because individual reports or report shares in those messages are not encoded opaquely, the failure to decode a single element can have implications for decoding everything else. For example, if a length field is wrong in the n-th report, then we can't really continue trying to parse the n+1-th and beyond reports because we're left in an ambiguous state.

But the specific cases here are well-defined encodings of invalid data (out of order extensions, duplicate extension type). So I think it's reasonable to continue trying to parse other elements in these arrays. It makes me wish we'd somehow made duplicate or out of order extensions impossible to represent, but I don't think that's possible in TLS presentation/encoding language (maybe CBOR? but we're not doing CBOR in DAP).

tgeoghegan (approved):

cjpatton (approved):

Issue #791, opened 2026-07-27
Batch selector, query, aggregate share ambiguities


tgeoghegan:
While implementing AAD binding in Janus (https://github.com/divviup/janus/pull/4714), we ran into some ambiguities in the specification that should be cleared up.

## Aggregate share and collection job IDs

In the latest DAP versions, we have asserted that HTTP servers always choose resource identifiers. But [this paragraph](https://datatracker.ietf.org/doc/html/draft-ietf-ppm-dap-19#section-4.6.4-6) in 4.6.4 is ambiguous at best if not contradictory:

> A Helper aggregate share is identified by a server-selected identifier, unique within the scope of the task, assigned during resource creation ([Section 3.2](https://datatracker.ietf.org/doc/html/draft-ietf-ppm-dap-19#resource-creation)). Since this resource corresponds to exactly one collection job, the Leader might use the collection job ID as the aggregate share ID.

Sentence 1 correctly asserts that the Helper chooses the aggregate share ID, but sentence 2 implies that the Leader might chooses it to match the collection job ID, which isn't right. The simple fix is to replace "Leader" with "Helper" in sentence 2, and since this is just a "might", not even a "MAY", I think we can avoid laboriously spelling out all other possible scenarios.

## Consistency of time interval collection job query and batch selector

A _query_ is constructed by the Collector and send to the Leader, which then constructs a corresponding _batch selector_ that satisfies the query, which will be relayed to the Helper. The batch selector must be _consistent_ with the query according to per-batch-mode semantics.

In the time interval batch mode, both the query and the batch selector are `Interval`s. [5.1.2](https://datatracker.ietf.org/doc/html/draft-ietf-ppm-dap-19#section-5.1.2) makes it clear that they must be the same:

> The payload of BatchSelector.config is
>
> struct {
>   Interval batch_interval;
> } TimeIntervalBatchSelectorConfig;
>
> where batch_interval is the batch interval requested by the Collector.

But the very following paragraph disagrees:

> A TimeIntervalBatchSelectorConfig.config is consistent with Query.config for this batch mode if the batch selector's interval is within the queried time interval.

[...]

PR #793, opened 2026-07-29
Batch selector, query, aggregate share ambiguities

tgeoghegan:
See #791 for discussion of these problems and individual commits for details discussion of the fixes.

### DIFF
--- a/draft-ietf-ppm-dap.md
+++ b/draft-ietf-ppm-dap.md
@@ -3549,8 +3549,6 @@ The Leader then combines the values inside the batch bucket as follows:
 
 A Helper aggregate share is identified by a server-selected identifier, unique within
 the scope of the task, assigned during resource creation ({{resource-creation}}).
-Since this resource corresponds to exactly one collection job, the Leader might use
-the collection job ID as the aggregate share ID.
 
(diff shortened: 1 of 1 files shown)


draft-ietf-ppm-dap.md
```diff
@@ -1411,6 +1411,11 @@ is said to be after the interval. A time that is either before or after an
 interval is said to be outside the interval. A time that is neither before nor
 after an interval is said to be inside or fall within the interval.
 
+An interval `a` is said to be inside or fall within another interval `b` if:
+
+* `a.start` is not before `b` and
+* the the end of `a` (`a.start + a.duration`) is not after `b`.
```
cjpatton, line 1417: I find this way more readable. It allows you to visualize the intervals on a number line.

```suggestion
An interval `a` is said to be inside or fall within another interval `b` if `b.start >= a.start` and `a.start+a.duration <= b.start+b.duration`.
```

tgeoghegan: `a.start` and `a.start + a.duration` are `Time`s, and in this section we clearly define what it means for a time to be before or after another, but now that it means to do "<=" or ">=". I want to build on definitions this document provides.

PR #794, opened 2026-07-29
Clarify nature of resource IDs

tgeoghegan:
In the past, we concretely defined aggregation job, aggregate share and collection job IDs as strings of 16 bytes, and accordingly specified that they get rendered into URL path components via Base64 URL encoding.

More recently, we relaxed the specification these IDs, choosing to leave their size and structure up to implementations.

This commit revisits language in section 3 HTTP Usage around these IDs. Rather than stating that whatever their form is should be Base64 URL encoded, we make it clear that the actual value of IDs is itself a valid URL path component, meaning no special transformation is needed to render them into URLs.

An implementation is free to meet this requirement by generating 16 random bytes and base64url encoding it, which is what a DAP-17 implementation would do.


draft-ietf-ppm-dap.md
```diff
@@ -1159,7 +1160,8 @@ Several DAP interactions involve creating new resources on an HTTP server:
 aggregation jobs ({{agg-init}}), collection jobs ({{collect-init}}), and
 aggregate shares ({{collect-aggregate}}). In each case, the HTTP client sends a
 POST request to a creation URL and the HTTP server assigns the new resource a
-unique identifier.
+unique identifier. These identifiers MUST be URL path components conforming to
+the definition in {{!RFC3986, Section 3.3}}.
```
tgeoghegan: This is nuanced: I don't want draft-ietf-ppm-dap to spend any text describing what a valid path component is, hence the reference to RFC 3986. But maybe I'm missing some subtlety in that document's definition of a path that would burn us.

martinthomson (approved):

draft-ietf-ppm-dap.md
```diff
@@ -1159,7 +1160,8 @@ Several DAP interactions involve creating new resources on an HTTP server:
 aggregation jobs ({{agg-init}}), collection jobs ({{collect-init}}), and
 aggregate shares ({{collect-aggregate}}). In each case, the HTTP client sends a
 POST request to a creation URL and the HTTP server assigns the new resource a
-unique identifier.
+unique identifier. These identifiers MUST be URL path components conforming to
```
```suggestion
unique identifier. These identifiers MUST be URL path segments conforming to
```

draft-ietf-ppm-dap.md
martinthomson: So easy to miss commentary...

I think that you want to use the word "segment" rather than "component".  RFC 3986 uses the word "component" to refer to the entire path.

tgeoghegan: I considered that. It seems to me there's no harm in allowing multiple path segments. If some helper implementation wants to serve aggregation jobs at `https://example.com/tasks/{task_id}/aggregation_jobs/eu-west/1/somethingelse/another-segment/unique-identifier` then I don't see a reason to forbid it. I don't think it introduces any ambiguities in handling requests. But then again maybe it's a good idea to restrict what implementations can do.

martinthomson: I used "segments", plural, in my suggestion because I share your view :)  And reasoning.  The path "component" is a complete path, which only approximately makes sense when pluralized and not a lot of sense when tacked on to another URL.  That said, what you described probably syntactically-compatible, in the sense that concatenating multiple path components produces another path component.

There is also the possibility of this including a query component.  That seems eminently reasonable also.

That's what you get when you start to prescribe (or proscribe): complication.

Maybe you can say this instead:

> When these identifiers are formed into a URL, the result MUST conform to URL syntax {{!RFC3986}}.

tgeoghegan: Query gets complicated, because DAP itself uses a query param in at least one place, and I think it could get hairy if we try to spell out how to reconcile those. But I like your notion of carefully using the plural "segments"! And it sounds like we agree about the desired outcome.

tgeoghegan: Last nuance here, I think: on its own, the 3986 definition of "segment" forbids "/" and doesn't tell you how to join segments together. So I chose to point to the `path-rootless` rule, which I think has exactly the semantics we want.

